PR #64, opened 2015-01-22
removed redundant condition in ScoreDirectorFactoryConfig

snurkabill:
IDE (and I when writing tests) had found this piece of code redundant.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
@@ -247,7 +247,7 @@ public class ScoreDirectorFactoryConfig {
                     + ") or bendableSoftLevelsSize (" + bendableSoftLevelsSize + ").");
         }
         if (scoreDefinitionClass != null) {
-            if (scoreDefinitionType != null || bendableHardLevelsSize != null || bendableSoftLevelsSize != null) {
+            if (scoreDefinitionType != null) {
                 throw new IllegalStateException("With scoreDefinitionClass (" + scoreDefinitionClass
                         + ") there must be no scoreDefinitionType (" + scoreDefinitionType
                         + ") or bendableHardLevelsSize (" + bendableHardLevelsSize


optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -247,7 +247,7 @@ public ScoreDefinition buildScoreDefinition() {
                     + ") or bendableSoftLevelsSize (" + bendableSoftLevelsSize + ").");
         }
         if (scoreDefinitionClass != null) {
-            if (scoreDefinitionType != null || bendableHardLevelsSize != null || bendableSoftLevelsSize != null) {
```
ge0ffrey: I don't understand why this is a redundant condition (at least with the code on master, I am seeing this on line 262, not on line 250, please "git pull --rebase" master to confirm.

Here's a config I believe should trigger this fail fast error and wouldn't without the change:

    <scoreDirectorFactory>
        <scoreDefinitionClass>MyCustomDef</>
        <bendableHardLevelsSize>4</>
    </>

Ps: if it's redundant, then the IllegalStateException should probably be adjusted to not mention bendableHardLevelsSize.

snurkabill:
======= reproducer ======= 

ScoreDirectorFactoryConfig factory = new ScoreDirectorFactoryConfig();
            factory.setScoreDefinitionClass(SimpleScoreDefinition.class);
            // 1.
            //factory.setScoreDefinitionType(ScoreDefinitionType.SIMPLE);
            // 2.
            //factory.setBendableHardLevelsSize(2);
            factory.buildScoreDefinition();

try to run reproducer with 1. and then with 2. line. with 1. line you get IllegalStateException(), with 2. line you get IllegalArgumentException().

It is not dependent on ScoreDefinitionClass at all. Point is, that  "bendableHardLevelsSize != null || bendableSoftLevelsSize != null" is alwyas false, because it was already checked in previous conditions. 

tested with latest master branch.

ge0ffrey:
Agreed, thanks for clarification :)
The bendableHardLevelsSize and bendableSoftLevelsSize nullcheck is redundant. Removing it as proposed and it's reference in the error message itself too.
Also improving error message of the first check to make more sense.
Also changing the IllegalStateException into a IllegalArgumentException, as that's more appropriate.
Adding a unit test to check the good path (no checks yet on config errors).

I won't merge this PR, but I 'll give you credit for that commit :)

ge0ffrey:
Manually merged: 
  http://github.com/droolsjbpm/optaplanner/commit/e43e55ff6

snurkabill:
np, I am glad I could help :)

PR #68, opened 2015-02-27
Redundant condition

snurkabill:
there are redundant ternary operators in NearbySelectionConfig


manstis:
I'm not an OptaPlanner developer so will wait for Geoffrey de Smet.. but the data-type is an Integer and there's no check for it being null prior to the lines you're changing so why is the null check a redundant term?

snurkabill:
https://github.com/snurkabill/optaplanner/blob/redundancy2/optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/common/nearby/NearbySelectionConfig.java#L197

on this line, we determine if linear distribution is enabled. and then we have condition: 

 } else if (linearDistributionEnabled) {
           int sizeMaximum = linearDistributionSizeMaximum == null ? Integer.MAX_VALUE : linearDistributionSizeMaximum;

where "linearDistributionSizeMaximum == null" is redundant, because the information is already stored in linearDistributionEnabled boolean value.

ge0ffrey:
Agreed. Good found (there might be more valid ones in findbugs or IntelliJ inspection, keep any you think are valid coming).
But I need to add support for nearbySelectionType (which I will do now on master so we're not stuck in a limbo state), which will make that null check required. Once that's pushed, I 'll reference that commit here and I 'll close this PR without merging.

ge0ffrey:
Fixed alternatively, by adding NearbySelectionDistributionType:
  https://github.com/droolsjbpm/optaplanner/commit/cb288f817f048f90ad8037f3d93503873605c758

PR #71, opened 2015-03-09
PLANNER-275 Add Drools score rules for example CheapTime Scheduling

tkobayas:
ACC = 670 with sample01.xml

Some points:
- I insert CheapTimeSolution as a fact because only it knows globalPeriodRangeFrom/globalPeriodRangeTo
- I insert Period as a fact to create MachinePeriodPart conveniently. Also it is used in rule "taskPowerCost".
- Rule "machinePeriodPart" seems to be a major bottle-neck but I have no other idea at this moment.
- Note that CheapTimeIncrementalScoreCalculator is used by default. Please uncomment scoreDrl when you test

I'm open to change anything if you suggest in review. Thanks!

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/CheapTimeSolution.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/CheapTimeSolution.java
@@ -154,6 +154,7 @@ public class CheapTimeSolution extends AbstractPersistable implements Solution<H
         facts.addAll(taskList);
         facts.addAll(taskRequirementList);
         facts.addAll(periodPowerPriceList);
+        facts.add(this);
         // Do not add the planning entity's (taskAssignmentList) because that will be done automatically
         return facts;
     }
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/MachinePeriodPart.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/MachinePeriodPart.java
@@ -1,0 +1,120 @@
+package org.optaplanner.examples.cheaptime.solver.drools;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.lang.builder.CompareToBuilder;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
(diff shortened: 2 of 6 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/MachinePeriodPart.java
```diff
@@ -0,0 +1,120 @@
...
+    private void addTaskAssignment(TaskAssignment taskAssignment) {
+        active = true;
+        Task task = taskAssignment.getTask();
+        for (int i = 0; i < resourceAvailableList.length; i++) {
+            TaskRequirement taskRequirement = task.getTaskRequirementList().get(i);
+            resourceAvailableList[i] -=  taskRequirement.getResourceUsage();
+        }
+    }
+
+	public boolean isActive() {
```
ge0ffrey: Code style (indentation). I 'll reformat this after merging.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/CheapTimeSolution.java
```diff
@@ -154,6 +154,7 @@ public void setScore(HardMediumSoftLongScore score) {
         facts.addAll(taskList);
         facts.addAll(taskRequirementList);
         facts.addAll(periodPowerPriceList);
+        facts.add(this);
```
ge0ffrey: I am personally not a fan of adding the solution instance in the problem facts, but there's nothing wrong with that: OptaPlanner should support it completely.
Merging anyway, of course.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/Period.java
```diff
@@ -0,0 +1,52 @@
+package org.optaplanner.examples.cheaptime.solver.drools;
+
+import org.apache.commons.lang.builder.CompareToBuilder;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+public class Period {
```
ge0ffrey: The class name Period creates confusion: every where else vars names period are of the type int.
I am going to rename this class to PeriodWrapper to avoid ambiguity.

ge0ffrey:
Thanks for the PR! :)

PR #72, opened 2015-03-12
BetaDistribution: parameters alpha and beta need to be positive. 

snurkabill:
see https://en.wikipedia.org/wiki/Beta_distribution

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/BetaDistributionNearbyRandom.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/BetaDistributionNearbyRandom.java
@@ -25,13 +25,13 @@ public class BetaDistributionNearbyRandom implements NearbyRandom {
     protected final BetaDistribution betaDistribution;
 
     public BetaDistributionNearbyRandom(double betaDistributionAlpha, double betaDistributionBeta) {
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/BetaDistributionNearbyRandom.java
```diff
@@ -25,13 +25,13 @@
     protected final BetaDistribution betaDistribution;
 
     public BetaDistributionNearbyRandom(double betaDistributionAlpha, double betaDistributionBeta) {
-        if (betaDistributionAlpha < 0) {
+        if (betaDistributionAlpha <= 0) {
             throw new IllegalArgumentException("The betaDistributionAlpha (" + betaDistributionAlpha
-                    + ") must be at least 0.");
+                    + ") must be positive.");
```
ge0ffrey: 0 is positive, so the error message is not accurate.
Note: 0 is also negative. 0 is not strictly positive. 0 is also not strictly negative. https://proofwiki.org/wiki/Definition:Strictly_Positive

Changing into "must be greater than 0", to avoid confusion.

snurkabill: nice point, ok

PR #73, opened 2015-03-12
DistributioTests: added corner cases, more tested values

snurkabill:
You may not going to like the cycle over all values which we want to pick.

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/BetaDistributionNearbyRandomTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/BetaDistributionNearbyRandomTest.java
@@ -40,14 +40,12 @@ public class BetaDistributionNearbyRandomTest {
         Random random = mock(Random.class);
         NearbyRandom nearbyRandom = new BetaDistributionNearbyRandom(1.0, 1.0);
 
-        when(random.nextDouble()).thenReturn(0.0);
-        assertEquals(0, nearbyRandom.nextInt(random, 500));
-        when(random.nextDouble()).thenReturn(1.0 / 500.0);
-        assertEquals(1, nearbyRandom.nextInt(random, 500));
-        when(random.nextDouble()).thenReturn(2.0 / 500.0);
-        assertEquals(2, nearbyRandom.nextInt(random, 500));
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/BetaDistributionNearbyRandomTest.java
```diff
@@ -40,14 +40,12 @@ public void nextIntUniform() {
...
-        when(random.nextDouble()).thenReturn(0.0);
-        assertEquals(0, nearbyRandom.nextInt(random, 500));
-        when(random.nextDouble()).thenReturn(1.0 / 500.0);
-        assertEquals(1, nearbyRandom.nextInt(random, 500));
-        when(random.nextDouble()).thenReturn(2.0 / 500.0);
-        assertEquals(2, nearbyRandom.nextInt(random, 500));
-        when(random.nextDouble()).thenReturn(3.0 / 500.0);
-        assertEquals(3, nearbyRandom.nextInt(random, 500));
+        double roundingConstant = 0.000001; // normalizes calculation against over-rounding error
+        for (int i = 0; i < 500; i++) {
```
ge0ffrey: I prefer hard coded numbers in unit tests, instead of dynamic code to validate all numbers. Because there's a high risk that that dynamic test code is written under the same logic as the production code, victim to the same bugs. Hard coded numbers on the other hand, predicted in advance on paper (so written without using the production code or even a bit of code to calculate the result), are a better approach in my opinion.

So I am going to revert this class to the original code.

ge0ffrey: DRY (Do not repeat yourself), extract methods, etc is a must do in production code, but I don't think it's a must do in unit test code, at least not to calculate the expected result of an assert.

ge0ffrey:
Thanks jiri, merged, great work :)
I did make some changes on top of it, see comments above for explenation: http://github.com/droolsjbpm/optaplanner/commit/6fef9a86b

snurkabill:
I see your point. I think that I will make some tests for distributions in RH test suite based on some macro-polo methods, just to be sure. anyway, you are welcome as always :)

PR #74, opened 2015-03-18
BlockDistribution: minSize should be at least 1

snurkabill:


snurkabill:
closed, the test is no longer needed

PR #75, opened 2015-03-18
BlockDistribution: minSize should be at least 1

snurkabill:


ge0ffrey:
In NearbySelectionConfig it currently defaults to 0:
  int sizeMinimum = blockDistributionSizeMinimum == null ? 0 : blockDistributionSizeMinimum;

That will not end up well.

ge0ffrey:
Merging and adjusting default

ge0ffrey:
Thanks for the PR jiri. Would you mind prefixing the bugzilla id (for example "BZ-123456" ) in the commitId next time? Productization etc require that for all commits landing on the release branch that affect production code.

snurkabill:
I didn't know that, sure I can! :) no problem

ge0ffrey:
thanks :)

PR #76, opened 2015-04-11
PLANNER-273 Benchmark report should mention logging level (because that can effect performance)

mcimbora:
* As SLF4J doesn't expose log level, manual resolving needs to be done
* Level resolved for the most general category (org.optaplanner), as sub-categories can define their own logging levels.


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
```diff
@@ -394,6 +420,8 @@ protected static PlannerBenchmarkResult createMergeSingleton(List<SingleBenchmar
                             newResult.availableProcessors, oldResult.availableProcessors);
                     newResult.maxMemory = ConfigUtils.mergeProperty(
                             newResult.maxMemory, oldResult.maxMemory);
+                    newResult.loggingLevel = ConfigUtils.mergeProperty(
+                            newResult.loggingLevel, oldResult.loggingLevel);
```
ge0ffrey: In all other cases, loggingLevel is handled before maxMemory, so I 'll move it above that.

Great PR in general :)

ge0ffrey:
Merged, thanks! :)

PR #79, opened 2015-04-27
Fix for mvn exec:java

bibryam:


optaplanner-examples/pom.xml
```diff
@@ -59,15 +59,11 @@
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>exec-maven-plugin</artifactId>
           <configuration>
-            <executable>java</executable>
```
ge0ffrey: By deleting this line doesn't the configuration apply to "mvn exec:exec" instead of "mvn exec:java"?
However, it seems that we need to be using "mvn exec:exec" indeed, according to the jira below, so this change looks right.
  https://issues.jboss.org/browse/PLANNER-287
We need to be sure that what is use portable to windlows' fragile command line length limitation too.
I'll need to check if the docs never reference the wrong mvn form.

ge0ffrey:
Merged, you're right that it should be like that.

Thanks for the PR :)

ge0ffrey:
With these changes I am running into

    [ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.4.0:exec (default-cli) on project optaplanner-examples: The parameters 'executable' for goal org.codehaus.mojo:exec-maven-plugin:1.4.0:exec are missing or invalid -> [Help 1]

Investigating what's wrong.

ge0ffrey:
Ok, I figured it out:
*It is impossible to write a configuration that is compatible with both exec:java and exec:exec.*

Bilgin and Richard were using exec:java. I (and the docs) were using exec:exec. Because "exec:exec" depends on java being in the PATH (which is not always the case), I 've switched everything (including docs) to "exec:java" (as Bilgin and Richard were using). Downside: Now anyone that uses "exec:exec" gets a confusing error message (otherwise it's vica versa).

PR #81, opened 2015-04-30
Update DefaultSolver.java

dimiii:
Hello, for some reasons in our project is required to reset value of 'solving' if some exceptions occur during planning.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -149,16 +149,20 @@ public class DefaultSolver implements Solver {
             throw new IllegalArgumentException("The planningProblem (" + planningProblem
                     + ") must not be null.");
         }
-        solverScope.setBestSolution(planningProblem);
-        outerSolvingStarted(solverScope);
-        boolean restartSolver = true;
-        while (restartSolver) {
-            solvingStarted(solverScope);
-            runPhases();
-            solvingEnded(solverScope);
-            restartSolver = checkProblemFactChanges();
+        try {
+            solverScope.setBestSolution(planningProblem);
+            outerSolvingStarted(solverScope);
+            boolean restartSolver = true;
+            while (restartSolver) {
+                solvingStarted(solverScope);
+                runPhases();
+                solvingEnded(solverScope);
+                restartSolver = checkProblemFactChanges();
+            }
+            outerSolvingEnded(solverScope);
+        } finally {
+            solving.set(false);
         }
-        outerSolvingEnded(solverScope);
     }
 
     public void outerSolvingStarted(DefaultSolverScope solverScope) {


snurkabill:
@dimiii , could you please send us some reproducer? or at least which exception is thrown?

ge0ffrey:
@snurkabill It could be a user exception, for example in the score DRL, or the difficulty comparator or in a customer move.

@dimiii Why do you need the solving set to false? I need to understand the problem better.
If the Solver throws an exception (and you want to recover from that), I 'd recommend to throw the Solver instance away and build a new one with `SolverFactory.buildSolver()`.
However, if you rely on `Solver.isSolving()` from another thread, your code change might be required indeed. In any case, I need to understand the problem first.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -149,16 +149,20 @@ public final void solve(Solution planningProblem) {
...
+            boolean restartSolver = true;
+            while (restartSolver) {
+                solvingStarted(solverScope);
+                runPhases();
+                solvingEnded(solverScope);
+                restartSolver = checkProblemFactChanges();
+            }
+            outerSolvingEnded(solverScope);
+        } finally {
+            solving.set(false);
```
ge0ffrey: Because we always call solving.set(false) here, we 'll to remove it from the outerSolvingEnded() method.

dimiii:
@ge0ffrey  You are right - it's an user exception thrown in a custom implementation of `IncrementalScoreCalculator`. 

Actually there is one thread in which planning is executed. The planning is limited to, say 5 minutes, so we rely on `Solver.isSolving()` and use `Solver.iterminateEarly()` from other thread to get then the best solution and reuse the `Solver` for the next task. 

In our case the `Solver` is kind of a service which instantiated and managed by Spring framework - so I particularly did not consider the possibility of restarting of the `Solver` - it seems be bulky  work.  And I also thought it be useful in common case - to guaranteed change the state of `Solver` if something goes wrong.

Anyway I'll ask more experienced coworker about restarting `Solver`, thanks for advise.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
dimiii: @ge0ffrey You are right - it's an user exception thrown in a custom implementation of IncrementalScoreCalculator.

Actually there is one thread in which planning is executed. The planning is limited to, say 5 minutes, so we rely on Solver.isSolving() and use Solver.iterminateEarly() from other thread to get then the best solution and reuse the Solver for the next task.

In our case the Solver is kind of a service which instantiated and managed by Spring framework - so I particularly did not consider the possibility of restarting of the Solver - it seems be bulky work. And I also thought it be useful in common case - to guaranteed change the state of Solver if something goes wrong.

Anyway I'll ask more experienced coworker about restarting Solver, thanks for advise.

dimiii: About `outerSolvingEnded()` - I have tried to reuse it in a finally block,  but the method handles some variables which at the time of exception are not completely initialized and my attempt ended NPE. I have little knowledge about opta-planner guts, and for the first glance simple `solving.set(false)` in a finally block was an easier modification.

snurkabill: @dimiii , I know nothing about your integration of planner, but how about catching exception in scoreFunction and returning Double.NEGATIVE_INFINITY (in right level of your score)? 

It might be completely bad, but on the other hand, you could avoid stopping planner on the first place and if using prepared heuristic for it, it shouldn't broke your solution.

dimiii:
@snurkabill Thank you - it's sounds like a simple and reasonable workaround of my issue at the application level. I think to close request.

snurkabill:
@dimiii glad I could help, but still, this issue worth discussing

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
ge0ffrey: @snurkabill @dimiii Catching the exception and returning Double.NEGATIVE_INFINITY will effectively eat your exception. That's very bad.
Instead, surround the Solver.solve() method's call with try-catch (in your code) and report it there properly (for example call an BrokenSolverEvent, similar to the BestSolutionChangedEvent (note that that one is called in the solver thread too).

I do agree, that in OptaPlanner's future SolverManager (= a high-level wrapper around the low-level Solver that manages threads too, does load balancing, multitenancy, etc) will need to cover exception handling and notify exceptions correctly.

ge0ffrey: https://issues.jboss.org/browse/PLANNER-338

snurkabill: @ge0ffrey I think that eating exception is not that bad in this case, as @dimiii explained. But I agree we need some modul to catch exceptions from planner.

PR #82, opened 2015-04-30
Added feature that allows loading Drools rules from a KJAR via kie-ci (i...

DuncanDoyle:
....e. Maven GAV).

Geoffrey, as discussed, an initial version of the feature that allows one to load score rules from a Drools KJAR via Maven (kie-ci and Maven GAV). Including unit-tests ;-)

To use it in, for example, the cloud-balancing example, you would need to externalize the score rules in a KJAR and reference the KJAR from the solver config like this:

```xml
<!-- Score configuration -->
<scoreDirectorFactory>
    <scoreDefinitionType>HARD_SOFT</scoreDefinitionType>
    <scoreKJarReleaseId>
    	<groupId>org.optaplanner.example.cloudbalancing</groupId>
    	<artifactId>CloudBalancingRules</artifactId>
    	<version>1.0.0</version>
    </scoreKJarReleaseId>	
</scoreDirectorFactory>
```

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/config/score/director/ScoreKJarReleaseIdConfigTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/config/score/director/ScoreKJarReleaseIdConfigTest.java
@@ -25,7 +25,7 @@ public class ScoreKJarReleaseIdConfigTest extends CommonTestMethodBase {
 	private static KieServices kieServices;
 	
 	//Simple rule that doesn't really do anything. Just for test.
-		private static String drlOne = "package org.optaplanner.test;\n " +
+	private static String drlOne = "package org.optaplanner.test;\n " +
 		
 				"import org.optaplanner.core.api.score.buildin.simple.SimpleScoreHolder;\n" +
 				"import java.lang.String;\n" +
@@ -53,7 +53,6 @@ public class ScoreKJarReleaseIdConfigTest extends CommonTestMethodBase {
 		createAndDeployJar(kieServices, KMODULE_CONTENT, releaseId, drlOneResource);
 	}
 	
-	
 	@Test
 	public void testKJarAvailable() {
 		
@@ -76,7 +75,7 @@ public class ScoreKJarReleaseIdConfigTest extends CommonTestMethodBase {
 		assertEquals("Expected rule with this specific name in KieBase.", "SimpleConstraintRule", ruleNames.get(0));
 	}
 	
-	@Test
+	@Test(expected=RuntimeException.class)
 	public void testKJarNotAvailable() {
 		boolean caughtException = false;
 		ScoreKJarReleaseIdConfig config = new ScoreKJarReleaseIdConfig();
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/config/score/director/ScoreKJarReleaseIdConfigTest.java
```diff
@@ -0,0 +1,96 @@
...
+		config.setGroupId(groupId);
+		config.setArtifactId(artifactId);
+		//Use a non-ex
+		config.setVersion("1.0.1");
+		try {
+			DroolsScoreDirectorFactory scoreDirectorFactory = config.build();
+		} catch (RuntimeException re) {
+			caughtException = true;
+		}
+		assertEquals("Expected exception when trying to load a KJAR which doesn't exist.", true, caughtException);
```
triceo: ```@Test``` annotation has an attribute accepting the expected ```Exception``` type, which does this exception handling for you.

optaplanner-core/src/test/java/org/optaplanner/core/config/score/director/ScoreKJarReleaseIdConfigTest.java
```diff
@@ -0,0 +1,90 @@
...
+				ruleNames.add(nextRule.getName());
+			}
+		}
+		assertEquals("Expected only 1 rule in KieBase.", 1, ruleNames.size());
+		assertEquals("Expected rule with this specific name in KieBase.", "SimpleConstraintRule", ruleNames.get(0));
+	}
+	
+	@Test(expected=RuntimeException.class)
+	public void testKJarNotAvailable() {
+		boolean caughtException = false;
```
triceo: This bit is no longer necessary.

DuncanDoyle: Only line 80, right? I still want to test whether the correct KJAR is loaded, hence the asserts on the number of rules and the rule-name in the KieBase.

triceo: Yes - the variable declared on line 80 is not used anywhere in the method.

ge0ffrey:
Related jira: https://issues.jboss.org/browse/PLANNER-253
Will process next week. (holiday today :)
Great work!

ge0ffrey:
I won't get around to merging this this week, but it's on the top of my list after next week's training. Sorry for the delay

ge0ffrey:
Bad news, @DuncanDoyle : perfect code, but me and Lukas have just come to the conclusion that our feature requirement is faulty. Let me explain:

- To author rules the domain model most be available. So to other a "CPU capacity hard constraint" rule, the "Process" and "Computer" classes must be in the classpath. This means the domain classes are included in the kjar (or a dependency of the kjar).
- If the domain classes are included in the kjar, OptaPlanner either needs to load them from the kjar (which it currently doesn't do) or optaplanner's solver xml should be part of the kjar itself. 
- In OptaPlanner Workbench (which will be part of Business Central), it will be possible to design the domain model and add OptaPlanner annotations. It will also be possible to author a solver xml.

This means we will end up with 2 different ways of doing the same thing:
A) OptaPlanner solver xml (not in a kjar) that can use score rules (and domain classes??) from a kjar. 
B) Kjar that contains domain classes, score rules AND solver xml.

One of the design philosophies is not to expose the user to 2 different ways to do the same thing if 1 way is consistently superior to the other way. Basically KISS principle. I believe B) (once implemented...) will be superior to A) at which point having to keep A) for backwards compatibility support is undesired...

Nevertheless, this PR isn't a waste:
- I 'll reuse the testing code (it's great!) for testing B)
- This PR has made us realize the problem of our feature requirement (again, not in your code). During research, failure is also a good result.

ge0ffrey:
Closing this PR. Changed made on master:
- The <scoreDirectorFactory> now supports <ksessionName> which maps to a ksession name in kmodule.xml
- New method SolverFactory.createFromKieContainerXmlResource(kieContainer, ".../solverConfig.xml")
- I've reused the testcase from @DuncanDoyle 's PR to tes the methods above. Thanks, Duncan!

PR #84, opened 2015-05-17
docs: better manipulation with core library flag

tomasdavidorg:


optaplanner-docs/src/main/docbook/en-US/Chapter-Integration/Chapter-Integration.xml
```diff
@@ -221,12 +221,20 @@
...
-          <para>Add <literal>--core-library</literal> flag to the <literal>dx</literal> file in
-          <literal>build-tools</literal> folder in your Android SDK, because otherwise it will not accept classes in the
-          <literal>java.*</literal> package of the <literal>javabeans-1.0.jar</literal>. The last line of that file
-          should look something like this:</para>
+          <para>Add lines to <literal>android</literal> area in the <literal>build.gradle</literal> file in your Android
+          project to allow classes of <literal>java.*</literal> package inside the <literal>javabeans-1.0.jar</literal>
+          file, because otherwise file will not be accepted:</para>
 
-          <programlisting>exec java $javaOpts -jar "$jarpath" --core-library "$@"</programlisting>
+          <programlisting language="gradle">android {
```
ge0ffrey: Isn't the language used in the gradle build files called "groovy" instead of "gradle"?

ge0ffrey: Fixed :) http://github.com/droolsjbpm/optaplanner/commit/5f584a888 Turns out there were 2 instances of this, so I missed the other one originally too. That should mean we have syntax coloring for it now in the html:
  https://github.com/isagalaev/highlight.js/tree/master/src/languages

ge0ffrey: Reverted my fix. Ignore, my I am not awake yet: highlight.js does support "gradle" so there must be a reason to classify it as a separate language.
  http://github.com/droolsjbpm/optaplanner/commit/f30890daf

PR #85, opened 2015-05-21
Integration tests

snurkabill:
Proposal of integration tests.


optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/scenario/simplequeens/TestdataQueenEntity.java
```diff
@@ -0,0 +1,41 @@
+package org.optaplanner.core.impl.testdata.domain.scenario.simplequeens;
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
+import org.optaplanner.core.impl.testdata.domain.scenario.simplequeens.solution.TestdataQueenDifficultyWeightFactory;
+import org.optaplanner.core.impl.testdata.domain.scenario.simplequeens.solution.TestdataRowStrengthWeightFactory;
+
+@PlanningEntity(difficultyWeightFactoryClass = TestdataQueenDifficultyWeightFactory.class)
+public class TestdataQueenEntity extends TestdataEntity {
```
ge0ffrey: I am not confortable with putting another copy of the queen class in our repository. That's too much maintenance.
Can we either move these tests into the optaplanner-examples test, so they can reuse the real Queen class. Or can we simply reuse any of the existing testdata classes?

ge0ffrey:
This has become a big bang PR. That's no good, as it becomes a "take or leave" situation. Let's see if we can partition in it and first merge the non-debatable parts.

snurkabill:
Don't worry, I am couting with rejecting PR, I just wanted to show how It could look.

snurkabill:
Since there are more tests we can have wider discussion

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/scenario/simplequeens/TestdataQueenEntity.java
snurkabill: I understand. On the other hand, there shouldn't be many opportunities for maintenance. Tests needn't use up-to-date queen scenario

snurkabill: I used nqeens because it is very easy to tests CH (it is very simple to see the error if it occurs, also it's very simple write those tests)

ge0ffrey:
Closing this PR because Jiri's newer PR's have salvaged the parts we want from this PR.

PR #86, opened 2015-05-24
Additional unit test + helper method.

DieterDePaepe:
Created a test to understand the randomSelection flag in ChangeMoveSelector. Frankly, I still don't see the purpose of this flag, as it seems to assume that the valueSelector will produce random values (otherwise, not all possible moves can be created). You can create the same behaviour by simply supplying a random entity/value selector to a non-random ChangeMoveSelector?

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/SelectorTestUtils.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/SelectorTestUtils.java
@@ -16,11 +16,11 @@
 
 package org.optaplanner.core.impl.heuristic.selector;
 
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.ListIterator;
+import java.util.*;
 
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.ListMultimap;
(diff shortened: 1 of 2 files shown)


ge0ffrey:
A changeMoveSelector in complete original order needs to select A1, A2, A3, ..., B1, B2, B3, ...
A changeMoveSelector in complete random order needs to select B5, C2, A7, ...

A changeMoveSelector in original order with an original order entity selector and a random order value selector will select A5, A2, A7, ...
A changeMoveSelector in random order with an original order entity selector and a random order value selector will select A5, B2, C7, ...

ge0ffrey:
Merged - Thanks for contributing :)

DieterDePaepe:
I too would have expected the order that you describe. But this isn't the order I see happening. If you look at the test, the random order is selected, but the moves are A1, B1, C1, ..., A1, B1, ... The value never changes.

I don't see how the benefit the randomSelection flag.
If I wanted the order A1, A2, A3, ..., B1, B2, B3... I could use a ChangeMoveSelector with non-random EntitySelector and non-random ValueSelector.
If I wanted the order A5, B2, C7, ... I could use a ChangeMoveSelector with non-random EntitySelector and random ValueSelector. (No need to use the randomSelection flag?)
If I wanted the order B5, C2, A7, ... I could use a ChangeMoveSelector with random EntitySelector and random ValueSelector. (No need to use the randomSelection flag?)

PR #90, opened 2015-06-01
Clarification in docs of getMediumScore()

wmeyers:
Added "hard" to score to clarify when the medium score value is irrelevant.


ge0ffrey:
Thanks for the fix :)

PR #92, opened 2015-06-02
Allow the Benchmarking results page to display a schedule with up to …

code4dc:
…15 different score levels. (Limited for visualization reasons)

Discussion can be viewed [here.](https://groups.google.com/forum/#!topic/optaplanner-dev/Lw2GZCC2-Oc)


ge0ffrey:
Merged, thanks for the PR. Not sure if it made the 6.3.0.Beta1 tag (if not, it will be in 6.3.0.Beta2).

Agreed that this is an improvement over the current situation, but I am still a bit worried about how 15 levels will actually look like in the report (is there enough tab room to show them etc). Have you tried the report with actually 15 levels (not just 10 and took 5 extra just to be sure)?

PR #94, opened 2015-06-05
Integration ch tests

snurkabill:


optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/integration/util/QueenCoordinates.java
```diff
@@ -0,0 +1,25 @@
+package org.optaplanner.examples.nqueens.integration.util;
+
+public class QueenCoordinates  {
+
+    public static QueenCoordinates createQueenCoordinates(int x, int y) {
+        return new QueenCoordinates(x, y);
+    }
+
+    private final int x;
```
ge0ffrey: In NQueens, it's not called x and y, but rowIndex and columnIndex.
For consistency, can we use the same naming?

Note that (x,y) starts from the top left of a screen - so using a different name might imply that there is some translation calculation when going from (x, y) to (rowIndex, columnIndex), which there is not I presume.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/integration/constructionheuristic/ConstructionHeuristicTest.java
```diff
@@ -0,0 +1,201 @@
...
+        config.setPhaseConfigList(Collections.<PhaseConfig>singletonList(chConfig));
+
+        NQueens solution = new NQueens();
+        List<Queen> entities = new ArrayList<Queen>();
+        int dimension = 8;
+        List<Row> rowList = new ArrayList<Row>();
+        List<Column> columnList = new ArrayList<Column>();
+        for (int i = 0; i < dimension; i++) {
+            Column column = new Column();
+            column.setIndex(i);
```
ge0ffrey: Id's aren't set on Column, Row or Queen. The id's are used in the rules. That's dangerous.

This code should be replaced by a call to NQueensGenerator.createNQueens().

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/integration/constructionheuristic/ConstructionHeuristicTest.java
```diff
@@ -0,0 +1,201 @@
...
+                                     List<QueenCoordinates> expectedCoordinates) {
+        this.constructionHeuristicType = constructionHeuristicType;
+        this.entitySorterManner = entitySorterManner;
+        this.valueSorterManner = valueSorterManner;
+        this.expectedCoordinates = expectedCoordinates;
+    }
+
+    @Test
+    public void testConstructionHeuristics() {
+        SolverConfig config = new SolverConfig();
```
ge0ffrey: I'd prefer loading the nqueensSolverConfig.xml and then removing all solver phases, instead of having to declare score calculation etc here too. See ConstructionHeuristicTest#buildSolverFactory() for inspiration.

ge0ffrey:
Good stuff. Please fix the feed-back and I'll merge :)

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/integration/constructionheuristic/ConstructionHeuristicTest.java
snurkabill: Personally I don't like dependency on some config file. Consider you will change (and you will) queen config. what then? the test may be broken after change

ge0ffrey: yes, but the inverse happens more, in my experience: if I refactor nqueens, then the test is broken if it's refactored along if it doesn't reuse the xml config for the stuff it doesn't care about (such as the score director configuration and entity class config etc).

ge0ffrey: In both cases, I 'll rerun the test of course.

PR #95, opened 2015-06-15
Style changes and minor editing to 4.1 to 4.3.3 of Planning configura…

gsheldon:
…tion chapter

Started making minor docs style changes.

### DIFF
--- a/optaplanner-docs/src/main/docbook/en-US/Chapter-Planner_configuration/Chapter-Planner_configuration.xml
+++ b/optaplanner-docs/src/main/docbook/en-US/Chapter-Planner_configuration/Chapter-Planner_configuration.xml
@@ -97,7 +97,7 @@
   &lt;/localSearch&gt;
 &lt;/solver&gt;</programlisting>
 
-      <para>Notice the 3 parts in it:</para>
+      <para>Notice the three parts in it:</para>
 
       <itemizedlist>
         <listitem>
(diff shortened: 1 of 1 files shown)


optaplanner-docs/src/main/docbook/en-US/Chapter-Planner_configuration/Chapter-Planner_configuration.xml
```diff
@@ -253,8 +253,8 @@
       </note>
 
       <para><emphasis role="bold">In Planner all problems facts and planning entities are plain old JavaBeans
-      (POJO's).</emphasis> You can load them from a database, an XML file, a data repository, a noSQL cloud, ... (see
-      <link linkend="integration">Integration</link>): Planner doesn't care.</para>
+      (POJOs).</emphasis> You can load them from a database, an XML file, a data repository, a noSQL cloud, ... (see
+      <link linkend="integration">Integration</link>).</para>
```
ge0ffrey: About "Planner doesn't care": The way it's stated is probably too negatively/unproffesionally, but it is an important checkbox in people's head: the fact that optaplanner doesn't care if those POJO's come from XML or from a relational database or whatsoever. It means it doesn't matter which what it is combined, there are no pitfalls when integrating with any technology as long as that technology is able to deliver POJO's.

I'll merge this PR (so delete the original statement), but if you see anyway to describe the above sentiment in a good way, it's welcome :)

optaplanner-docs/src/main/docbook/en-US/Chapter-Planner_configuration/Chapter-Planner_configuration.xml
```diff
@@ -436,6 +436,7 @@ public class Queen {
         anyway. For example, a <literal>Queen</literal>'s <literal>row</literal> variable should not be used.</para>
       </section>
     </section>
+<!-- gemma: edited to here -->
```
ge0ffrey: Hey Gemma,

Would you mind not adding these comments in the document itself, but track it externally?
When I change something on the docs with my docbook WYSIWYG editor, this line changes too automatically, so that would create a merge conflict with your changes I fear. 
Also note that we can release a Beta of master at any time: it should always be in a Beta releasable state.

I'll delete this line (if you deleted it locally meanwhile, that's the same change so it probably won't give a merge conflict). If you do get a merge conflict: that can happen, resolve it :)

PR #96, opened 2015-06-24
Fix compilation error: HardSoftScore -> HardSoftLongScore

psiroky:


ge0ffrey:
Thanks Petr :) I didn't run the webexamples (only the normal examples) so my IDE didn't compile them. I should really build an arquillian test some day for the webexamples and run that too.

PR #99, opened 2015-07-07
Local search integration test

snurkabill:


snurkabill:
I am not exactly sure the difference between Late Acceptance and Step Counting Hill Climbing

PR #101, opened 2015-07-16
PLANNER-341: Disabled LogarithmicAxis' check for non-positive values

oskopek:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -514,7 +515,9 @@ private XYPlot createScalabilityPlot(List<XYSeries> seriesList,
             String yAxisLabel, NumberFormat yAxisNumberFormat) {
         NumberAxis xAxis = new NumberAxis(xAxisLabel);
         if (useLogarithmicProblemScale(seriesList)) {
-            xAxis = new LogarithmicAxis(xAxis.getLabel() + " (logarithmic)");
+            LogarithmicAxis xLogAxis = new LogarithmicAxis(xAxis.getLabel() + " (logarithmic)");
+            xLogAxis.setAllowNegativesFlag(true);
+            xAxis = xLogAxis;
```
ge0ffrey: No need to use a separate xLogAxis variable (especially as it's a local variable, so there is no multi-threading danger whatsoever). Also change the original code to avoid unnecessary instantiation:

NumberAxis xAxis;
if (!...) {
  xAxis = new NumberAxis(xAxisLabel);
} else {
  xAxis = new LogarithmicAxis(...)
  ...
}

ge0ffrey:
Otherwise looks good, I 'll merge when the changes in the comment above are applied

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: the `setAllowNegativesFlag()` is a method of LogarithmAxis, that's why I found this way better than doing `((LogarithmicAxis)xAxis).setAllowNegativesFlag(true);`

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -512,9 +513,12 @@ private BarRenderer createBarChartRenderer(NumberFormat numberFormat) {
     private XYPlot createScalabilityPlot(List<XYSeries> seriesList,
             String xAxisLabel, NumberFormat xAxisNumberFormat,
             String yAxisLabel, NumberFormat yAxisNumberFormat) {
-        NumberAxis xAxis = new NumberAxis(xAxisLabel);
+        NumberAxis xAxis;
         if (useLogarithmicProblemScale(seriesList)) {
-            xAxis = new LogarithmicAxis(xAxis.getLabel() + " (logarithmic)");
+            xAxis = new LogarithmicAxis(xAxisLabel + " (logarithmic)");
+            ((LogarithmicAxis)xAxis).setAllowNegativesFlag(true);
```
ge0ffrey: I agree with your comment in your mail, having a separate variable is indeed better than casting. I didn't realize it was a subclass specific method.
Call the separate variable name logarithmicAxis, that way it's clear it's a LogarithmicAxis specific method. There's no need to prefix that variable with x, because it's scoped within that if method, so it's very local.

Anyway, nitpicking here :)

PR #106, opened 2015-07-20
PLANNER-334: Added tooltips to explain uninitialized/infeasible/failed benchmark badges in report

oskopek:
**NOTE: Merge only after #105**

* Tested on Firefox and Chrome
* Works even without javascript enabled (tooltips then look like standard html tooltips)

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
@@ -55,6 +55,7 @@ public class SingleBenchmarkResult {
     private Long usedMemoryAfterInputSolution = null;
 
     private Boolean succeeded = null;
+    private Boolean initialized = null;
     private Score score = null;
     private long timeMillisSpent = -1L;
     private long calculateCount = -1L;
@@ -134,6 +135,14 @@ public class SingleBenchmarkResult {
         this.succeeded = succeeded;
     }
 
+    public Boolean getInitialized() {
+        return initialized;
+    }
+
+    public void setInitialized(Boolean initialized) {
+        this.initialized = initialized;
+    }
+
     public Score getScore() {
         return score;
     }
@@ -199,6 +208,10 @@ public class SingleBenchmarkResult {
 
     public boolean isSuccess() {
         return succeeded != null && succeeded.booleanValue();
+    }
+
+    public boolean isInitialized() {
+        return initialized != null && initialized.booleanValue();
     }
 
     public boolean isFailure() {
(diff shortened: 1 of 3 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
```diff
@@ -201,6 +210,10 @@ public boolean isSuccess() {
         return succeeded != null && succeeded.booleanValue();
     }
 
+    public boolean isInitialized() {
+        return initialized != null && initialized.booleanValue();
```
ge0ffrey: There's common-lang's BooleanUtils.isTrue(initialized) that does the same thing (null safe)

But I think this method  boolean isInitialized() is not desired:
I think it's better to use Boolean getInitialized() WITHOUT a null check (neither here nor in the calling code). Why? Because we can guarantee that setInitialized() has been called with a non-null value before getInitialized() is used. If that is not the case, that's a bug and I 'd rather see a NullPointerException than it silently presuming its not initialized, which might be incorrect (which might lead users to make big wrong decisions, like throw away an actually better solver configuration).

This is the principle of Fail Fast: if the state is corrupted, don't sweep it under the carpet, but throw any exception immediately, so the user knows there's something broken and he can report it to us. In the long term, this means we spot bugs sooner, which means less bugs and people thrust our software more. Thrust means not giving incorrect information (even if that means stating that something went wrong and it's unable to deliver the information).

ge0ffrey: Hmm, but I do it myself in isSuccess() and isFailure(), because the solver config can throw an exception at which point it never gets to the success/failure/initialized population code, but it does report it in the report. This is because the benchmarker is the only thing in all of planner that is special in the sense that it doesn't fail fast: if a single benchmark report fails, the benchmarker takes note of that and continues with the other benchmarks: it does write the report, but it does also throw an exception, at the very last moment. This is as special case, "an exception that confirms the rule".

So my previous comment doesn't apply here :) Ignore it.

optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
```diff
@@ -26,16 +26,17 @@
...
-        <#else>
-        <span class="badge">${singleBenchmarkResult.ranking}</span>
-        </#if>
-        <#if !singleBenchmarkResult.scoreFeasible>
-        <span class="badge badge-warning">!</span>
-        </#if>
+    <span class="badge">${singleBenchmarkResult.ranking}</span>
+    </#if>
+
+    <#if !singleBenchmarkResult.initialized>
```
ge0ffrey: Shouldn't this also be an elseif?
If the "Failed" badge is shown, there is no point in showing the "Initialized" or "Infeasible" badge.

I am already merging this PR, as it's a big step forward, but do take a look at this issue.

PR #109, opened 2015-07-22
PLANNER-302-fix: Added beans.xml to enable CDI

oskopek:
Quick fix for PLANNER-302 on JBoss EAP

### DIFF
--- a/optaplanner-webexamples/src/main/webapp/vehiclerouting/WEB-INF/beans.xml
+++ b/optaplanner-webexamples/src/main/webapp/vehiclerouting/WEB-INF/beans.xml
@@ -1,0 +1,23 @@
+<!--
+  ~ Copyright 2015 JBoss Inc
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Agreed, and this file needs to be deleted (will do so):
https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-webexamples/src/main/webapp/cloudbalancing/WEB-INF/beans.xml

ge0ffrey:
Must have been a refactor that went wrong

oskopek:
We could also add one beans.xml into `webapp/WEB-INF/beans.xml` and wouldn't have to add each one individually per webexample (depends on what the requirements are - if it needs to be easier to manage or more modular)

ge0ffrey:
the beans.xml only works if it's in `mywarfile.war!WEB-INF/beans.xml`, so only if it's in webapp/WEB-INF/beans.xml`, because WEB-INF (direclty under the root) is a magic directory. However, `cloudbalancing/WEB-INF` is not magic, so the beans.xml in there does nothing: it's just dead weight. Removing now.

ge0ffrey:
Hmm, the fact that you added `vehicleouring/WEB-INF/beans.xml` and that made it work, means that `cloudbalancing/WEB-INF/beans.xml` is indeed being picked up too...

PR #110, opened 2015-07-22
PLANNER-48: Multi-threaded benchmark warm-up

oskopek:
At this point I'm not sure if this PR satisfies all requirements or is completely correct, but a quick peer review might be a good idea now.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
@@ -122,25 +122,26 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
     }
 
     private void warmUp() {
-        if (plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit() > 0L) {
-            logger.info("================================================================================");
-            logger.info("Warm up started");
-            logger.info("================================================================================");
-            long startingTimeMillis = System.currentTimeMillis();
-            long timeLeft = plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit();
-            List<ProblemBenchmarkResult> unifiedProblemBenchmarkResultList = plannerBenchmarkResult.getUnifiedProblemBenchmarkResultList();
-            Iterator<ProblemBenchmarkResult> it = unifiedProblemBenchmarkResultList.iterator();
-            while (timeLeft > 0L) {
-                if (!it.hasNext()) {
-                    it = unifiedProblemBenchmarkResultList.iterator();
-                }
-                ProblemBenchmarkResult problemBenchmarkResult = it.next();
-                timeLeft = warmUp(problemBenchmarkResult, startingTimeMillis, plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit(), timeLeft);
-            }
-            logger.info("================================================================================");
-            logger.info("Warm up ended");
-            logger.info("================================================================================");
+        if (plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit() <= 0L) {
+            return;
         }
+        logger.info("================================================================================");
+        logger.info("Warm up started");
+        logger.info("================================================================================");
+        long startingTimeMillis = System.currentTimeMillis();
+        long timeLeft = plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit();
+        List<ProblemBenchmarkResult> unifiedProblemBenchmarkResultList = plannerBenchmarkResult.getUnifiedProblemBenchmarkResultList();
+        Iterator<ProblemBenchmarkResult> it = unifiedProblemBenchmarkResultList.iterator();
+        while (timeLeft > 0L) {
+            if (!it.hasNext()) {
+                it = unifiedProblemBenchmarkResultList.iterator();
+            }
+            ProblemBenchmarkResult problemBenchmarkResult = it.next();
+            timeLeft = warmUp(problemBenchmarkResult, startingTimeMillis, plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit(), timeLeft);
+        }
+        logger.info("================================================================================");
+        logger.info("Warm up ended");
+        logger.info("================================================================================");
     }
 
     protected long warmUp(ProblemBenchmarkResult problemBenchmarkResult,
@@ -182,7 +183,7 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
                 futureMap.put(singleBenchmarkRunner, future);
             }
         }
-        // wait for the benchmarks to complete
+        // Wait for the benchmarks to complete
         for (Map.Entry<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureEntry : futureMap.entrySet()) {
             SingleBenchmarkRunner singleBenchmarkRunner = futureEntry.getKey();
             Future<SingleBenchmarkRunner> future = futureEntry.getValue();
@@ -195,16 +196,18 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
                     throw new IllegalStateException("Score is null. TODO fix PLANNER-46.");
                 }
             } catch (InterruptedException e) {
-                logger.error("The singleBenchmarkRunner (" + singleBenchmarkRunner.getName() + ") was interrupted.", e);
+                singleBenchmarkRunnerExceptionLogger.error("The singleBenchmarkRunner ({}) was interrupted.",
+                        singleBenchmarkRunner, e);
                 failureThrowable = e;
             } catch (ExecutionException e) {
                 Throwable cause = e.getCause();
-                singleBenchmarkRunnerExceptionLogger.warn("The singleBenchmarkRunner ("
-                        + singleBenchmarkRunner.getName() + ") failed.", cause);
+                singleBenchmarkRunnerExceptionLogger.warn("The singleBenchmarkRunner ({}) failed.",
+                        singleBenchmarkRunner, cause);
                 failureThrowable = cause;
             } catch (IllegalStateException e) {
                 // TODO WORKAROUND Remove when PLANNER-46 is fixed.
-                singleBenchmarkRunnerExceptionLogger.warn("The singleBenchmarkRunner (" + singleBenchmarkRunner.getName() + ") failed.", e);
+                singleBenchmarkRunnerExceptionLogger.warn("The singleBenchmarkRunner ({}) failed.",
+                        singleBenchmarkRunner, e);
                 failureThrowable = e;
             }
             if (failureThrowable == null) {
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
@@ -25,8 +25,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 import org.optaplanner.benchmark.api.PlannerBenchmark;
 import org.optaplanner.benchmark.api.PlannerBenchmarkException;
@@ -131,22 +129,29 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
         logger.info("Warm up started");
         logger.info("================================================================================");
         long timeLeft = plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit();
+        int solverBenchmarkCount = plannerBenchmarkResult.getSolverBenchmarkResultList().size();
+        long minimumCyclesCount
+                = Math.round(Math.ceil(solverBenchmarkCount / (double) plannerBenchmarkResult.getParallelBenchmarkCount()));
+        long timeLeftPerSolverBenchmark;
         long timeEnd = System.currentTimeMillis() + timeLeft;
         Map<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureMap
                 = new HashMap<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>>();
         Map<SolverBenchmarkResult, TerminationConfig> originalTerminationConfigMap
-                = new HashMap<SolverBenchmarkResult, TerminationConfig>();
+                = new HashMap<SolverBenchmarkResult, TerminationConfig>(solverBenchmarkCount);
+        Map<SolverBenchmarkResult, Integer> singleBenchmarkResultIndexMap
+                = new HashMap<SolverBenchmarkResult, Integer>(solverBenchmarkCount);
         while (timeLeft > 0L) {
-            warmUpStarted(futureMap, originalTerminationConfigMap, timeLeft);
-            warmUp(futureMap, timeEnd);
+            timeLeftPerSolverBenchmark = Math.round(Math.floor((timeLeft / (double) minimumCyclesCount)));
+            warmUpStarted(futureMap, originalTerminationConfigMap, singleBenchmarkResultIndexMap, timeLeftPerSolverBenchmark);
+            warmUp(futureMap);
             timeLeft = timeEnd - System.currentTimeMillis();
             futureMap.clear();
         }
         warmUpEnded(originalTerminationConfigMap);
-        warmUpExecutorService.shutdownNow(); // Terminate warm up solvers early
-        List<Runnable> notFinishedWarmUpList = warmUpExecutorService.shutdownNow(); // Make sure they're terminated
+        List<Runnable> notFinishedWarmUpList = warmUpExecutorService.shutdownNow();
         if (!notFinishedWarmUpList.isEmpty()) {
-            throw new IllegalStateException("Impossible state: notFinishedWarmUpList size (" + notFinishedWarmUpList + ").");
+            throw new IllegalStateException("Impossible state: notFinishedWarmUpList is non-empty ("
+                    + notFinishedWarmUpList + ").");
         }
         logger.info("================================================================================");
         logger.info("Warm up ended");
@@ -154,33 +159,45 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
     }
 
     private void warmUpStarted(Map<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureMap,
-                               Map<SolverBenchmarkResult, TerminationConfig> originalTerminationConfigMap, long timeLeft) {
+                               Map<SolverBenchmarkResult, TerminationConfig> originalTerminationConfigMap,
+                               Map<SolverBenchmarkResult, Integer> singleBenchmarkResultIndexMap,
+                               long timeLeftPerSolverConfig) {
 
         for (SolverBenchmarkResult solverBenchmarkResult : plannerBenchmarkResult.getSolverBenchmarkResultList()) {
             TerminationConfig originalTerminationConfig = solverBenchmarkResult.getSolverConfig().getTerminationConfig();
             TerminationConfig tmpTerminationConfig = originalTerminationConfig == null
                     ? new TerminationConfig() : originalTerminationConfig.clone();
-            tmpTerminationConfig.shortenTimeMillisSpentLimit(timeLeft);
+            tmpTerminationConfig.shortenTimeMillisSpentLimit(timeLeftPerSolverConfig);
             solverBenchmarkResult.getSolverConfig().setTerminationConfig(tmpTerminationConfig);
-            originalTerminationConfigMap.put(solverBenchmarkResult, originalTerminationConfig);
-            for (SingleBenchmarkResult singleBenchmarkResult : solverBenchmarkResult.getSingleBenchmarkResultList()) {
+            if (!originalTerminationConfigMap.containsKey(solverBenchmarkResult)) {
+                originalTerminationConfigMap.put(solverBenchmarkResult, originalTerminationConfig);
+            }
+
+            Integer singleBenchmarkResultIndex = singleBenchmarkResultIndexMap.get(solverBenchmarkResult);
+            if (singleBenchmarkResultIndex == null ||
+                singleBenchmarkResultIndex >= solverBenchmarkResult.getSingleBenchmarkResultList().size()) {
+                singleBenchmarkResultIndex = 0;
+            }
+            SingleBenchmarkResult singleBenchmarkResult
+                    = solverBenchmarkResult.getSingleBenchmarkResultList().get(singleBenchmarkResultIndex);
+            if (singleBenchmarkResult != null) {
                 SingleBenchmarkRunner singleBenchmarkRunner = new SingleBenchmarkRunner(singleBenchmarkResult);
                 Future<SingleBenchmarkRunner> future = warmUpExecutorService.submit(singleBenchmarkRunner);
                 futureMap.put(singleBenchmarkRunner, future);
+                singleBenchmarkResultIndexMap.put(solverBenchmarkResult, singleBenchmarkResultIndex + 1);
             }
         }
     }
 
-    private void warmUp(Map<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureMap, long timeEnd) {
-        // Wait for the warm up benchmarks to complete or time-out
+    private void warmUp(Map<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureMap) {
+        // Wait for the warm up benchmarks to complete
         for (Map.Entry<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureEntry : futureMap.entrySet()) {
             SingleBenchmarkRunner singleBenchmarkRunner = futureEntry.getKey();
             Future<SingleBenchmarkRunner> future = futureEntry.getValue();
             Throwable failureThrowable = null;
             try {
                 // Explicitly returning it in the Callable guarantees memory visibility
-                long timeLeft = timeEnd - System.currentTimeMillis();
-                singleBenchmarkRunner = future.get(timeLeft, TimeUnit.MILLISECONDS);
+                singleBenchmarkRunner = future.get();
             } catch (InterruptedException e) {
                 singleBenchmarkRunnerExceptionLogger.error("The warm up singleBenchmarkRunner ({}) was interrupted.",
                         singleBenchmarkRunner, e);
@@ -190,12 +207,6 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
                 singleBenchmarkRunnerExceptionLogger.warn("The warm up singleBenchmarkRunner ({}) failed.",
                         singleBenchmarkRunner.getName(), cause);
                 failureThrowable = cause;
-            } catch (TimeoutException e) {
-                // Do not update failureThrowable on TimeoutException as it is expected to occur
-                singleBenchmarkRunnerExceptionLogger.debug("The warm up singleBenchmarkRunner ({}) timed out.",
-                        singleBenchmarkRunner);
-                singleBenchmarkRunner.terminateSolverEarly();
-                future.cancel(true);
             }
 
             if (failureThrowable != null) {
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
@@ -19,13 +19,14 @@ package org.optaplanner.benchmark.impl;
 import java.io.File;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.optaplanner.benchmark.api.PlannerBenchmark;
 import org.optaplanner.benchmark.api.PlannerBenchmarkException;
@@ -35,7 +36,6 @@ import org.optaplanner.benchmark.impl.result.PlannerBenchmarkResult;
 import org.optaplanner.benchmark.impl.result.ProblemBenchmarkResult;
 import org.optaplanner.benchmark.impl.result.SingleBenchmarkResult;
 import org.optaplanner.benchmark.impl.result.SolverBenchmarkResult;
-import org.optaplanner.core.api.solver.Solver;
 import org.optaplanner.core.config.solver.termination.TerminationConfig;
 import org.optaplanner.core.config.util.ConfigUtils;
 import org.slf4j.Logger;
@@ -53,6 +53,7 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
     private BenchmarkReport benchmarkReport = null;
 
     private ExecutorService executorService;
+    private ExecutorService warmUpExecutorService;
     private BenchmarkResultIO benchmarkResultIO;
 
     private long startingSystemTimeMillis = -1L;
@@ -107,6 +108,7 @@ public class PlannerBenchmarkRunner implements PlannerBenchmark {
         initBenchmarkDirectoryAndSubdirs();
         plannerBenchmarkResult.initSystemProperties();
         executorService = Executors.newFixedThreadPool(plannerBenchmarkResult.getParallelBenchmarkCount());
+        warmUpExecutorService = Executors.newFixedThreadPool(plannerBenchmarkResult.getParallelBenchmarkCount());
         benchmarkResultIO = new BenchmarkResultIO();
         logger.info("Benchmarking started: solverBenchmarkResultList size ({}), parallelBenchmarkCount ({}).",
                 solverBenchmarkResultList.size(), plannerBenchmarkResult.getParallelBenchmarkCount());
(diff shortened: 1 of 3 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +124,94 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+                = new HashMap<SolverBenchmarkResult, TerminationConfig>();
+        while (timeLeft > 0L) {
+            warmUpStarted(futureMap, originalTerminationConfigMap, timeLeft);
+            warmUp(futureMap, timeEnd);
+            timeLeft = timeEnd - System.currentTimeMillis();
+            futureMap.clear();
+        }
+        warmUpEnded(originalTerminationConfigMap);
+        warmUpExecutorService.shutdownNow(); // Terminate warm up solvers early
+        List<Runnable> notFinishedWarmUpList = warmUpExecutorService.shutdownNow(); // Make sure they're terminated
```
ge0ffrey: Why 2 calls to warmUpExecutorService.shutdownNow()?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +124,94 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+        if (!notFinishedWarmUpList.isEmpty()) {
+            throw new IllegalStateException("Impossible state: notFinishedWarmUpList size (" + notFinishedWarmUpList + ").");
```
ge0ffrey: The string says it's a size, but the variable is a list. I'd drop the word "size" in the string.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +124,94 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
             try {
-                Solver solver = solverBenchmarkResult.getSolverConfig().buildSolver();
-                solver.solve(problemBenchmarkResult.readPlanningProblem());
-            } catch (RuntimeException e) {
-                singleBenchmarkRunnerExceptionLogger.warn("The warmUp of singleBenchmark ("
-                        + singleBenchmarkResult.getName() + ") failed.", e);
-                // TODO update firstFailureSingleBenchmarkRunner (when warmups happen in a Runner too in parallel)
+                // Explicitly returning it in the Callable guarantees memory visibility
+                long timeLeft = timeEnd - System.currentTimeMillis();
+                singleBenchmarkRunner = future.get(timeLeft, TimeUnit.MILLISECONDS);
```
ge0ffrey: Suppose for 1 warmup single benchmark we have 10 seconds. If that 10 seconds is set in the terminationConfig (and optaplanner respects that!), why do we need to use that 10 seconds as a timeout too?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +124,94 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+        logger.info("================================================================================");
+        logger.info("Warm up started");
+        logger.info("================================================================================");
+        long timeLeft = plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit();
+        long timeEnd = System.currentTimeMillis() + timeLeft;
+        Map<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>> futureMap
+                = new HashMap<SingleBenchmarkRunner, Future<SingleBenchmarkRunner>>();
+        Map<SolverBenchmarkResult, TerminationConfig> originalTerminationConfigMap
+                = new HashMap<SolverBenchmarkResult, TerminationConfig>();
+        while (timeLeft > 0L) {
```
ge0ffrey: Suppose we have a warmup of 400 seconds. Suppose we have a parallel benchmark count of 4. Suppose we have 10 solver configurations. How long does each warmup single benchmark get?

The philosophy is that each solver config needs to be warmed up (but not every problem dataset), but that the user only specifies the entire budget (usually 30 seconds).
In the old code, so with parallel count 1, if we have 400 seconds with 10 solver configurations, each solver config would get 40 seconds. If the first one only took 13 seconds (instead of 40), then the other 9 would have gotten (400-13)/9 = 43 seconds. Basically, any time not soaked up by previous solver configuration runs, is donated to later runs.

ge0ffrey: So with 400 seconds and a parallel count of 4 and 10 solver configs, I would expect that we run 3 cycles of up to 4 solvers of 133 seconds each. Why? ceil(10/4) is 3. And 400/3 is 133 seconds per cycle.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: They were needed to 1. kill the threads 2. verify their dead. Not needed anymore.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Fixed

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: For some reason, I chose not to use the TerminationConfig then. I've rewritten this part to be more clean.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: This is a good idea! I didn't know about the requirement to run each solver config at least once, fixed that too.

oskopek:
Just one problem: Not all threads are being used at all times during the warm up. Imagine that in a 10 second warm up I run 5 solvers in 5 threads for a maximum time of 5s (because I need to run a few more later to cover all the solver configs). Now lets say 4 of those solvers end after 1s, and the fifth uses up all of its 5 seconds. The 4 threads aren't used until the last solver ends and we populate the ExecutionService with new single benchmarks.

In practice, this isn't as big of a problem as it seems, but I'd like to discuss it anyway.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -53,6 +54,8 @@
     private BenchmarkReport benchmarkReport = null;
 
     private ExecutorService executorService;
+    private ExecutorService warmUpExecutorService;
+    private ExecutorCompletionService<SingleBenchmarkRunner> warmUpExecutorCompletionService;
```
ge0ffrey: trivial: I'd order executorService after the 2 warmup ones, because it used later (and has the same volatility).

Elaborate philosophy:
I tend to organize fields, first on volatility (static final, static, final instance field, rarely-changing instance field, consistently-changing) and then on usage order (= some sort of logical order).
I'd even say there is logical order between things that are equal in importance. For example: in DefaultSolver, the Termination is logically before the BestSolutionRecaller field. This means that phaseStarted(), stepEnded() and similar events will be called in that order: first on Termination, then on BestSolutionRecaller. So if in the future, BestSolutionRecaller need something from Termination, it can safely presume it's already initialized & up to date. If on the other hand Termination needs something from BestSolutionRecaller, they need to be reserved in field & method call order (which implying checking if there's no dependency in the other direction).

Having a consistent ordering can actually prevents bugs in rare occasions, such as deadlocks. For example:Thread 1 takes A then takes B and thread 2 takes B then takes A: if they both start with A this is impossible.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+        for (int i = 0; i < cyclesCount; i++) {
+            long timeCycleEnd = System.currentTimeMillis() + timeLeftPerCycle;
+            for (int j = 0; j < parallelBenchmarkCount; j++) {
+                solverBenchmarkResultCycleArray[j] = plannerBenchmarkResult.
+                        getSolverBenchmarkResultList().get(solverBenchmarkResultIndex % solverBenchmarkResultCount);
+                solverBenchmarkResultIndex++;
+            }
+            warmUpPopulate(futureMap, singleBenchmarkResultIndexMap, solverBenchmarkResultCycleArray, timeLeftPerCycle);
+            warmUp(futureMap, singleBenchmarkResultIndexMap, timeCycleEnd);
+            futureMap.clear();
```
ge0ffrey: Why reuse the futureMap? Instead I 'd limit the scope of the futureMap to this inside this for loop, so do it's local variable declaration inside the for loop.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+        long cyclesCount = Math.round(Math.ceil(solverBenchmarkResultCount / (double) parallelBenchmarkCount));
+        long timeLeftPerCycle = Math.round(Math.floor((timeLeftTotal / (double) cyclesCount)));
+        ConcurrentMap<Future<SingleBenchmarkRunner>, SingleBenchmarkRunner> futureMap
+                = new ConcurrentHashMap<Future<SingleBenchmarkRunner>, SingleBenchmarkRunner>(parallelBenchmarkCount);
+        Map<SolverBenchmarkResult, TerminationConfig> originalTerminationConfigMap
+                = new HashMap<SolverBenchmarkResult, TerminationConfig>(solverBenchmarkResultCount);
+        ConcurrentMap<SolverBenchmarkResult, Integer> singleBenchmarkResultIndexMap
+                = new ConcurrentHashMap<SolverBenchmarkResult, Integer>(solverBenchmarkResultCount);
+
+        warmUpStarted(originalTerminationConfigMap);
```
ge0ffrey: maybe warmUpStarted(originalTerminationConfigMap) and warmUpEnded(originalTerminationConfigMap) should be renamed to backupTerminationConfigs() and restoreTerminationConfigs(originalTerminationConfigMap).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+        SolverBenchmarkResult[] solverBenchmarkResultCycleArray = new SolverBenchmarkResult[parallelBenchmarkCount];
```
ge0ffrey: Rename to solverBenchmarkResultCycles: for arrays and Collections we use plural (suffix with "s"), for List and Set and Map we suffix with "List" and "Set" and "Map"

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+            }
+            warmUpPopulate(futureMap, singleBenchmarkResultIndexMap, solverBenchmarkResultCycleArray, timeLeftPerCycle);
+            warmUp(futureMap, singleBenchmarkResultIndexMap, timeCycleEnd);
+            futureMap.clear();
+        }
+        warmUpEnded(originalTerminationConfigMap);
+        List<Runnable> notFinishedWarmUpList = warmUpExecutorService.shutdownNow();
+        if (!notFinishedWarmUpList.isEmpty()) {
+            throw new IllegalStateException("Impossible state: notFinishedWarmUpList is non-empty ("
+                    + notFinishedWarmUpList + ").");
```
ge0ffrey: Rest of optaplanner code does the () immediately after the var itself, so:
"... notFinishedWarmUpList (" + notFinishedWarmUpList + ") is not empty."

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+    private void warmUpPopulate(Map<Future<SingleBenchmarkRunner>, SingleBenchmarkRunner> futureMap,
+                                ConcurrentMap<SolverBenchmarkResult, Integer> singleBenchmarkResultIndexMap,
+                                SolverBenchmarkResult[] solverBenchmarkResultArray,
+                                long timeLeftPerSolverConfig) {
+        for (SolverBenchmarkResult solverBenchmarkResult : solverBenchmarkResultArray) {
             TerminationConfig originalTerminationConfig = solverBenchmarkResult.getSolverConfig().getTerminationConfig();
             TerminationConfig tmpTerminationConfig = originalTerminationConfig == null
                     ? new TerminationConfig() : originalTerminationConfig.clone();
-            tmpTerminationConfig.shortenTimeMillisSpentLimit(timeLeft);
+            tmpTerminationConfig.overwriteTimeMillisSpentLimit(timeLeftPerSolverConfig);
```
ge0ffrey: Why overwrite instead of shorten? If we have a warmup budget of 1000 seconds in this cycle, but we have a solver config that has a termination set to 50 seconds, I think we should just run it for 50 seconds on dataset 1, then 50 seconds on dataset 2, etc.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
             solverBenchmarkResult.getSolverConfig().setTerminationConfig(tmpTerminationConfig);
 
+            Integer singleBenchmarkResultIndex = singleBenchmarkResultIndexMap.get(solverBenchmarkResult);
+            if (singleBenchmarkResultIndex == null ||
+                    singleBenchmarkResultIndex >= solverBenchmarkResult.getSingleBenchmarkResultList().size()) {
+                singleBenchmarkResultIndex = 0;
+            }
+            SingleBenchmarkResult singleBenchmarkResult
+                    = solverBenchmarkResult.getSingleBenchmarkResultList().get(singleBenchmarkResultIndex);
+            if (singleBenchmarkResult != null) {
```
ge0ffrey: When is singleBenchmarkResult null? If it is null, shouldn't we cycle back to the first problem?
So do a % of the singleBenchmarkResultIndex?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+                if (firstFailureSingleBenchmarkRunner == null) {
+                    firstFailureSingleBenchmarkRunner = singleBenchmarkRunner;
+                }
+            }
+
+            SolverBenchmarkResult solverBenchmarkResult = singleBenchmarkRunner.getSingleBenchmarkResult().getSolverBenchmarkResult();
+            long timeLeftInCycle = timePhaseEnd - System.currentTimeMillis();
+            if (timeLeftInCycle > 0L) {
+                warmUpPopulate(futureMap, singleBenchmarkResultIndexMap, new SolverBenchmarkResult[]{solverBenchmarkResult}, timeLeftInCycle);
+                tasksCount++;
```
ge0ffrey: Isn't it possible that the warmUpPopulate adds more than just 1 Future into the futureMap? so tasksCounts++ might not be accurate.

I think the code would be more readable if the for(i to taskCounts) loop would be replaced by a while (futureMap isn't empty, pop the first one)
Note: pop() is a method on Queue. It can be mimiced on Map with values().iterator().next(), but a Queue is probably more natural here?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+                continue;
             }
 
-            solverBenchmarkResult.getSolverConfig().setTerminationConfig(originalTerminationConfig);
-            long timeSpent = System.currentTimeMillis() - startingTimeMillis;
-            timeLeft = warmUpTimeMillisSpentLimit - timeSpent;
-            if (timeLeft <= 0L) {
-                return timeLeft;
+            Throwable failureThrowable = null;
+            SingleBenchmarkRunner singleBenchmarkRunner = futureMap.get(future);
```
ge0ffrey: Why
  singleBenchmarkRunner = futureMap.get(future);
if you then do this immediately 2 lines below:
  singleBenchmarkRunner = future.get();


I believe/hope the futureMap is obsolete

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Fixed and noted :)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Wouldn't it be more memory friendly if it is just reused and not allocated/deallocated each time?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Fixed, that's indeed a better name.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Noted the convention, but in this case, `solverBenchmarkResultCycle` makes more sense, as its contents are replaced at the beginning of each single cycle.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Oops, fixed, of course.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: The point of overwrite was to act as a protection against illogical uses, such that we wouldn't spent almost all of the time with start up/shut down of individual single benchmarks.

It is basically just a corner case: in most practical uses, this will shorten the time anyway. The decision is up to you.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
triceo: Maybe a little bit. But your performance will suffer - ```clear()```ing a map is much slower than just leaving it for the GC to remove. Generally unless you've really got a mem-hungry algorithm, I'd go for local variables.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: singleBenchmarkResult is never null in this context, the singleBenchmarkResult**Index** is null when the singleBenchmarkResultIndexMap is uninitialized (when `.get()` returns null), therefore we initialize it with 0.

It basically acts as an index through the datasets of each solver config.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: @triceo: Didn't know that, thanks.

 This map has `< number of threads` elements, I'm moving the map into the loop.

ge0ffrey: +1 for triceo's comment. Objecting pooling was more memory friendly in Java 1.1, but not in a modern Java version.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: That's a good point of view too, but I 'd argue that those users would actually want the startup/shutdown more warmup up than anyway, as it becomes relatively more important as the termination time decreases.
I think the advantage of doing more different datasets (especially if dataset 1 never triggers certain rules such as Penalty based rules, so that code doesn't get jitted during warmup), outweights the other advantage.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: So what does line 200 "if (singleBenchmarkResult != null) {" check?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: It isn't, because the third parameter (`new SolverBenchmarkResult[]{solverBenchmarkResult}`) is a an array of size 1.

As for readability, this might be a bit more readable:
```java
if (timeLeftInCycle > 0L) {
    SolverBenchmarkResult[] solverBenchmarkResultSingleton = {solverBenchmarkResult};
    warmUpPopulate(futureMap, singleBenchmarkResultIndexMap, solverBenchmarkResultSingleton, timeLeftInCycle);
    tasksCount++;
}
```

The problem is, we're not going through the map in any given order. The map is passed into this function just as a reference for looking up the returned `SingleBenchmarkRunners` from the `warmUpExecutorCompletionService` and so that we can pass it to `warmUpPopulate`.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Your argument makes sense now that I think about it. Fixing.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: I really have no clue how that got in there. Fixing.

Maybe I should move the initialization of the `singleBenchmarkResultIndexMap` into a different method and convert this `if` on line 194 into
```java
singleBenchmarkResultIndex = singleBenchmarkResultIndex % solverBenchmarkResult.getSingleBenchmarkResultList().size();
```

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Because the first is the copy of `singleBenchmarkRunner `we put into the `warmUpExecutorCompletionService` so it's non-null (needed in the `if (failureThrowable != null)` block a bit down).

The second one is what we got back from the `warmUpExecutorCompletionService`.

I suspect it won't be obsolete, because `warmUpExecutorCompletionService.take()` doesn't throw information like `future.get()` does in case the singleBenchmark failed (that's the reason why there are two initializations currently).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: Both approaches are fine. You could even do:

    singleBenchmarkResultIndex = (singleBenchmarkResultIndex == null) ? 0
            : singleBenchmarkResultIndex % solverBenchmarkResult.getSingleBenchmarkResultList().size();

It's all good.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: Agreed, for the failure path we need to keep track of the futureMap. That's a pity but I agree there's no other way.

For readability, I might move the "singleBenchmarkRunner = futureMap.get(future);" in the 2 catches, so the line above can just be "SingleBenchmarkRunner singleBenchmarkRunner;" (declaration but no initialization).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
-                Solver solver = solverBenchmarkResult.getSolverConfig().buildSolver();
-                solver.solve(problemBenchmarkResult.readPlanningProblem());
-            } catch (RuntimeException e) {
-                singleBenchmarkRunnerExceptionLogger.warn("The warmUp of singleBenchmark ("
-                        + singleBenchmarkResult.getName() + ") failed.", e);
-                // TODO update firstFailureSingleBenchmarkRunner (when warmups happen in a Runner too in parallel)
+                future = warmUpExecutorCompletionService.take();
+            } catch (InterruptedException e) {
+                singleBenchmarkRunnerExceptionLogger.error("Waiting for a warm up singleBenchmarkRunner was interrupted.", e);
+                continue;
```
ge0ffrey: When will this happen? Can we test it locally realistically? Will it ever happen?

Don't continue, don't log an error, just fail fast and throw an exception. When in doubt, leave it out: when in doubt, don't support the functionality, break and fix it when users starts complaining about it. This way we don't have to support misbehavior for ever due to backwards compatibility.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,134 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
+        for (int i = 0; i < tasksCount; i++) {
+            Future<SingleBenchmarkRunner> future;
             try {
-                Solver solver = solverBenchmarkResult.getSolverConfig().buildSolver();
-                solver.solve(problemBenchmarkResult.readPlanningProblem());
-            } catch (RuntimeException e) {
-                singleBenchmarkRunnerExceptionLogger.warn("The warmUp of singleBenchmark ("
-                        + singleBenchmarkResult.getName() + ") failed.", e);
-                // TODO update firstFailureSingleBenchmarkRunner (when warmups happen in a Runner too in parallel)
+                future = warmUpExecutorCompletionService.take();
```
ge0ffrey: What if we use poll() instead of take() here and use a `while (future != null)` loop?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: Aggreed on tasksCounts++.

I still dislike the trick of using the tasksCounts to predict when the queue will be empty and the loop should stop. See my comment on poll() and take() above.
BTW: the reason there is no isEmpty() method in that service is for multi-threaded use cases (even though we only have 1 consumer thread): the poll() method does the isEmpty() check inherently in a lock safe manner.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: I'm not sure when this happens, hasn't happened once locally. AFAIK it shouldn't happened: I see no reason why the waiting thread should be interrupted (at least not when everything is working correctly).

Substituted the catch block with:
 ```java
catch (InterruptedException e) {
    throw new RuntimeException("Waiting for a warm up singleBenchmarkRunner was interrupted.", e);
}
```

File: optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java (hunk 15 above)
Comment (oskopek) [inline optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java]: I dislike the idea of actively waiting for the result of a concurrent operation (that takes potentially seconds/minutes/hours) in this way. The main thread will just use up CPU cycles pointlessly.

File: optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java (hunk 13 above)
Comment (oskopek) [inline optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java]: Fixed the readability issue. Since we have to keep the map, does anything change on the issue above (Queue vs. Map iterator vs. `int tasksCount`)?

File: optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java (hunk 14 above)
Comment (oskopek) [inline optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java]: Options/opinions on how to handle this correctly:

* A: Logging and continuing
```java
logger.error("Waiting for a warm up singleBenchmarkRunner was interrupted.", e);
continue;
```

* B: Throwing the InterruptedException to a higher level
  * Widely recommended, but does it make sense in our case?
  * What is it that we want to tell the caller?

* C: Setting the interrupted status and returning from the method
  * Widely recommended, but does it make sense in our case?
  * Do we actually tell the caller anything? What does returning too soon mean to it?

```java
logger.error("Waiting for a warm up singleBenchmarkRunner was interrupted.", e);
Thread.currentThread().interrupt();
return;
```

* D: Throwing a RuntimeException and failing (too?) fast

```java
throw new RuntimeException("Waiting for a warm up singleBenchmarkRunner was interrupted.", e);
```

* E: Actively waiting (doesn't throw an InterruptedException)
  * Uses the CPU extensively for an uninteresting and not needed task

```java
while (future == null) {
    future = warmUpExecutorCompletionService.poll();
}
```

ge0ffrey: A will - in most cases - effectively eat the exception (= sometimes goes wrong, but the developer/user doesn't see the notification). Eating the exception is about the worse thing than can happen, because it doesn't force the developer/user to fix the problem, it silently creates explainable problems which reduce long term thrust in our software. The benchmarker does uses the logger.error() itself once too, but that's only to delay the fail fast, instead of eating it: it still ends up throwing an exception instead of returning to the user. -1 on A (even with a delayed throw actually)

B is impossible/undesirable because InterruptedException is a checked exception and we don't want to make our benchmark users jumpt through some hoops of try-catch loops that they don't know what to do in the catch anyway. Big -1 on B

C in that form eats the exception, see reasoning of A. There might be a variation of this that is acceptable.

D should be a subclass of RuntimeException, such as IllegalStateException or IllegalArgumentException. I'd recommend IllegalStateException. +1 on D with an IllegalStateException and the InterruptedException on the cause

E will cause an infinite loop. Big -1 on E

oskopek: We decided to go with D for the moment. See [PLANNER-402](https://issues.jboss.org/browse/PLANNER-402) for more information.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -122,54 +127,132 @@ private void initBenchmarkDirectoryAndSubdirs() {
...
-            logger.info("================================================================================");
         }
     }
 
-    protected long warmUp(ProblemBenchmarkResult problemBenchmarkResult,
-            long startingTimeMillis, long warmUpTimeMillisSpentLimit, long timeLeft) {
-        for (SingleBenchmarkResult singleBenchmarkResult : problemBenchmarkResult.getSingleBenchmarkResultList()) {
-            SolverBenchmarkResult solverBenchmarkResult = singleBenchmarkResult.getSolverBenchmarkResult();
+    private void warmUpPopulate(Map<Future<SingleBenchmarkRunner>, SingleBenchmarkRunner> futureMap,
+                                ConcurrentMap<SolverBenchmarkResult, Integer> singleBenchmarkResultIndexMap,
```
ge0ffrey: Code style: method parameters are 2 idents in (so 8 spaces), instead of being aligned to param

PR #111, opened 2015-07-23
PLANNER-334: Fixed hiding of other badges when a benchmark fails

oskopek:
This should be more understandable and logical.


optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
```diff
@@ -27,17 +27,18 @@
 <#macro addSingleRankingBadge singleBenchmarkResult>
     <#if !singleBenchmarkResult.ranking??>
     <span class="badge badge-important" data-toggle="tooltip" title="Failed benchmark">F</span>
-    <#elseif singleBenchmarkResult.winner>
+    <#else>
+    <#if singleBenchmarkResult.winner>
```
ge0ffrey: Would you mind indenting this nested if, so it's clear its nested?

oskopek:
I decided to indent all the if blocks. It is more readable in the template and as far as I can tell doesn't affect rendering in HTML in any way.

ge0ffrey:
Agreed, they should have been indented in the first place (IIRC it was my intention even, must have missed some), as a clear control flow is important because code readability is very important.

PR #113, opened 2015-07-27
PLANNER-396: define classpath for examples using wildcards

psiroky:
* tested on Linux and Windows Server 2012 R2

 * works even with long base directory
   (tried with one that had more than 160 chars)

### DIFF
--- a/optaplanner-distribution/src/main/assembly/scripts/runExamples.bat
+++ b/optaplanner-distribution/src/main/assembly/scripts/runExamples.bat
@@ -4,8 +4,7 @@ setLocal enableExtensions enableDelayedExpansion
 rem Most examples run (potentially slower) with max heap of 128 MB (so -Xmx128m), but 1 example's dataset requires 1.5 GB
 set "jvmOptions=-Xms256m -Xmx1536m -Dorg.optaplanner.examples.dataDir=sources/data/"
 set "mainClass=org.optaplanner.examples.app.OptaPlannerExamplesApp"
-set "mainClasspath="
-for %%i in (binaries\*.jar) do (set "mainClasspath=!mainClasspath!;%%i")
+set "mainClasspath=binaries/*;../binaries/*"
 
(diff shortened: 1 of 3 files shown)


optaplanner-distribution/src/main/assembly/scripts/runExamples.bat
```diff
@@ -4,8 +4,7 @@ setLocal enableExtensions enableDelayedExpansion
 rem Most examples run (potentially slower) with max heap of 128 MB (so -Xmx128m), but 1 example's dataset requires 1.5 GB
 set "jvmOptions=-Xms256m -Xmx1536m -Dorg.optaplanner.examples.dataDir=sources/data/"
 set "mainClass=org.optaplanner.examples.app.OptaPlannerExamplesApp"
-set "mainClasspath="
-for %%i in (binaries\*.jar) do (set "mainClasspath=!mainClasspath!;%%i")
+set "mainClasspath=binaries/*;../binaries/*"
```
ge0ffrey: I am pretty sure I tried this too and it didn't work on some platform. I suggest to merge and to ask QA to test more platforms...
Any reason why it's `*` and not `*.jar`

ge0ffrey: Did it run in powershell on Windows Server 2012 R2?

psiroky: The reason for `*` instead of `*.jar` is http://stackoverflow.com/a/1275467 . Looks like the `*.jar` does not work on Windows for whatever reason.

I've tried both powershell and the standard windows `cmd`. However, It would be definitely good to try on Win 7 or Win 8 as well.

psiroky: Just tried on Windows 7 (32bit) and works as well (with long paths).

PR #116, opened 2015-07-30
Added modules to the maven-checkstyle-plugin

oskopek:
Feature proposal:
* Added 25 (+3 out-commented) new modules to the `maven-checkstyle-plugin` (only module currently being run is `FileTabCharacter` which checks for tabs in files)
* All of the modules except for one are run in the warning severity level
  * `FileTabCharacter` is run in severity "error" as in the parent pom
* All errors/warnings/info are reported to the console *(divergent behavior from the parent pom)*
* All modules that fail and have a severity of "error" **fail the build** *(divergent behavior from the parent pom)*
  * Note: currently only one module, `FileTabCharacter`
* Adding/removing modules can be done incrementally as new issues emerge
  * The severity of modules can also be increased/decreased on the go (on a per module basis)
  * **Unfortunately, I have failed at finding modules that correspond with JDK8's javadoc doclint checks**, the only option I see is writing our own, which isn't really satisfactory.
* Does it slow the build? Yes, but not really by much.
  * On an average of 5 runs on my local machine, the complete build (`mvn clean install -Dfull`) runs about 15 seconds slower (which given the ~280s build is about 6%).
  * **Note**: If you don't `clean` between the builds, there is virtually no delay (because of caching)
    * See the logs without `clean` in their name.
  * For details on the benchmark, see [my benchmark script and log](https://gist.github.com/oskopek/ab01b23ee951d3362bcc)
* Currently, checkstyle reports ~100 warnings in the whole project. They're all fixed as part of this PR.
* Also, there's a nice [plugin for IntelliJ](https://github.com/jshiell/checkstyle-idea)

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSingleStatistic.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSingleStatistic.java
@@ -30,7 +30,7 @@ import org.optaplanner.benchmark.impl.statistic.single.pickedmovetypebestscore.P
 import org.optaplanner.benchmark.impl.statistic.single.pickedmovetypestepscore.PickedMoveTypeStepScoreDiffSingleStatistic;
 
 /**
- * 1 statistic of {@link SingleBenchmarkResult}
+ * 1 statistic of {@link SingleBenchmarkResult}.
  */
 @XStreamInclude({
         ConstraintMatchTotalBestScoreSingleStatistic.class,
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/ReflectionHelper.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/ReflectionHelper.java
@@ -114,7 +114,7 @@ public final class ReflectionHelper {
                 + (propertyName.isEmpty() ? "" : propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1));
         try {
             return containingClass.getMethod(getterName);
-        } catch (NoSuchMethodException e) {}
+        } catch (NoSuchMethodException e) { }
         String isserName = PROPERTY_ACCESSOR_PREFIX_IS
                 + (propertyName.isEmpty() ? "" : propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1));
         try {
@@ -122,7 +122,7 @@ public final class ReflectionHelper {
             if (method.getReturnType() == boolean.class) {
                 return method;
             }
-        } catch (NoSuchMethodException e) {}
+        } catch (NoSuchMethodException e) { }
         return null;
     }
 
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
@@ -386,7 +386,7 @@ public class BenchmarkAggregatorFrame extends JFrame {
             final Object benchmarkResult = mixedCheckBox.getBenchmarkResult();
 
             JPanel mainPanel = new JPanel(new BorderLayout());
-            String benchmarkResultTextFieldText = null; 
+            String benchmarkResultTextFieldText = null;
             if (benchmarkResult instanceof SolverBenchmarkResult) {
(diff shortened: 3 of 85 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSingleStatistic.java
```diff
@@ -30,7 +30,7 @@
 import org.optaplanner.benchmark.impl.statistic.single.pickedmovetypestepscore.PickedMoveTypeStepScoreDiffSingleStatistic;
 
 /**
- * 1 statistic of {@link SingleBenchmarkResult}
+ * 1 statistic of {@link SingleBenchmarkResult}.
```
ge0ffrey: All sentences should end with a dot (or question mark etc), but this is not a sentence because it has no verb, so we shouldn't add a dot here

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/ReflectionHelper.java
```diff
@@ -114,7 +114,7 @@ public static Method getGetterMethod(Class containingClass, String propertyName)
                 + (propertyName.isEmpty() ? "" : propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1));
         try {
             return containingClass.getMethod(getterName);
-        } catch (NoSuchMethodException e) {}
+        } catch (NoSuchMethodException e) { }
```
ge0ffrey: Why is the extra space needed?

ge0ffrey:
Sorry that I let this one rot... I though it was just the checkstyle config, not the actual fixes.
I definitely want these fixes in (not sure if the checkstyle plugin needs to be active already, might just comment it out for now). Would you mind rebasing it and deal with the conflicts (and the 2 comments I made)? I 'll merge it quickly afterwards to avoid letting it rot again.

oskopek:
I'll rebase, fix the conflicts, and submit different PRs for the config and the fixes. Closing. Will open new PRs in less than 3 hours, going on a short break now.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/ReflectionHelper.java
oskopek: Because the WhitespaceAround (quite a useful check) can't be disabled to allow empty blocks. Fixed to
```
...
} catch (NoSuchMethodException e) {
    // intentionally empty
}
...
```

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSingleStatistic.java
oskopek: I'd argue (from a language syntactical standpoint) that there actually is a verb implicitly in the sentence (therefore, it is a sentence):
```
(This is) 1 statistic of (a) {@link SingleBenchmarkResult}.
```
I may be wrong on that account thought. Anyway, the `JavadocStyle` check expects the first sentence to be a summary of what the method/class/field/etc does/is, and it makes sense for it to expect some sort of punctuation at the end. This can be configured (either disabling the first sentence check or supplying your own regex) and I will do it if you want to, but I'm in favor of just adding the dot here.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/ReflectionHelper.java
ge0ffrey: Meh. False negatives are very annoying. From practical experience, I believe that any rule which has false negatives needs to be disabled entirely - or they endanger developers from disliking CheckStyle entirely. I've seen this in practice with the maven-enforcer-plugin and Jenkins too. The messenger of bad news (= Jenkins, CheckStyle, enforcer) cannot afford to make mistakes because a developer team working against a deadlines under stress is more than willing to shoot it (= to disable it or ignore it entirely).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSingleStatistic.java
ge0ffrey: I've just seen a javadocs guidelines page that also uses dots there in an example, so looks like you're right :)
  http://www.oracle.com/technetwork/articles/java/index-137868.html
My memory says the opposite though - we might want to consult Gemma about this - product docs probably has a guideline about this sort of thing.

PR #117, opened 2015-07-31
examples: run tests in parallel

psiroky:
* just a suggestion. Tested locally without any test failures. 

* 4 seems like a good number given current processors have 2-4 cores. We could even use something like 2C (two threads per core). But this gets tricky when using e.g. Hyper Threading, because the number of cores visible to surefire is doubled, so in the end we have 4 threads per one "real" core.

 * adding it only to `-examples` as this module takes by far the most time and contains most of the tests

 * locally, the speedup is about 33% (from 3 min to 2 min)

### DIFF
--- a/optaplanner-examples/pom.xml
+++ b/optaplanner-examples/pom.xml
@@ -43,6 +43,13 @@
     <pluginManagement>
       <plugins>
         <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <configuration>
+            <forkCount>0.5C</forkCount>
+          </configuration>
+        </plugin>
+        <plugin>
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>exec-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Interesting idea. The gain is indeed very attractive, but I am scared of platform depended builds:
1) What if this gets run on a jenkins slave or contributor with only 1 or 2 cores?
Ideally, we want to use the number of cores divided by 2. This way the OS has a core too, and performance of the integration tests that solve to attain a specific score won't suffer much, meaning that they are less likely to time out.
2) What about memory consumption? Do the tests share the same JVM? If so, we need a test heap with potentially 4 times as much memory if we run 4 in parallel?

psiroky:
ad 1) The ideal number of forks is very dependent on the actual machine the tests are running on. I agree that 2 would be safer for code that in mostly CPU bound (like the optaplanner is?)

ad 2) by default the tests use different JVM (and they reuse these JVMs), so basically the memory consumption is ~<nr-of-forks> times higher.

ge0ffrey:
Merged, based on discussion on optaplanner-dev IRC. On first sign of trouble, we will revert and figure out trouble locally.

PR #120, opened 2015-08-05
PLANNER-391: Fixed ranking of solver configs with uninitialized solutions

oskopek:
* **NOTE**: Merge only after PLANNER-390 PR #119

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
@@ -34,6 +34,7 @@ import org.optaplanner.benchmark.impl.statistic.StatisticType;
 import org.optaplanner.core.api.score.FeasibilityScore;
 import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.solver.Solver;
+import org.optaplanner.core.impl.score.ScoreUtils;
 
 /**
  * Represents 1 benchmark for 1 {@link Solver} configuration for 1 problem instance (data set).
@@ -55,8 +56,8 @@ public class SingleBenchmarkResult {
     private Long usedMemoryAfterInputSolution = null;
 
     private Boolean succeeded = null;
-    private Boolean initialized = null;
     private Score score = null;
+    private Integer uninitializedVariableCount = null;
     private long timeMillisSpent = -1L;
     private long calculateCount = -1L;
 
@@ -135,20 +136,20 @@ public class SingleBenchmarkResult {
         this.succeeded = succeeded;
     }
 
-    public Boolean getInitialized() {
-        return initialized;
-    }
-
-    public void setInitialized(Boolean initialized) {
-        this.initialized = initialized;
-    }
-
     public Score getScore() {
         return score;
     }
 
     public void setScore(Score score) {
         this.score = score;
+    }
+
+    public Integer getUninitializedVariableCount() {
+        return uninitializedVariableCount;
+    }
+
+    public void setUninitializedVariableCount(Integer uninitializedVariableCount) {
+        this.uninitializedVariableCount = uninitializedVariableCount;
     }
 
     public long getTimeMillisSpent() {
@@ -211,7 +212,7 @@ public class SingleBenchmarkResult {
     }
 
     public boolean isInitialized() {
-        return initialized != null && initialized.booleanValue();
+        return uninitializedVariableCount != null && uninitializedVariableCount == 0;
     }
 
     public boolean isFailure() {
@@ -241,6 +242,10 @@ public class SingleBenchmarkResult {
 
     public SingleStatistic getSingleStatistic(StatisticType statisticType) {
         return effectiveSingleStatisticMap.get(statisticType);
+    }
+
+    public String getScoreWithUninitializedPrefix() {
+        return ScoreUtils.getScoreWithUninitializedPrefix(uninitializedVariableCount, score);
     }
 
     // ************************************************************************
@@ -285,8 +290,8 @@ public class SingleBenchmarkResult {
         // Skip oldResult.reportDirectory
         // Skip oldResult.usedMemoryAfterInputSolution
         newResult.succeeded = oldResult.succeeded;
-        newResult.initialized = oldResult.initialized;
         newResult.score = oldResult.score;
+        newResult.uninitializedVariableCount = oldResult.uninitializedVariableCount;
         newResult.timeMillisSpent = oldResult.timeMillisSpent;
         newResult.calculateCount = oldResult.calculateCount;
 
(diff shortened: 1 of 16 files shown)


ge0ffrey:
OptaPlanner comes with 3 out-of-the-box solver ranking implementations (should be in docs iirc - also adjust/add a sentence in docs to mention that a higher uninitialized count (and failures count if it doesn't do so already) get ranked later):
TotalRankSolverRankingWeightFactory, TotalScoreSolverRankingComparator, WorstScoreSolverRankingComparator.

I would have expected the existing failure count check and this new uninitialized count check to be in those (in an abstract superclass of those even), so that users can overwrite it if they implement their own ranking. I think the old (and new code) needs to refactor to that. What do you think?

ge0ffrey:
Note: A *WeightFactory has the same purpose as a *Comparator, but has a different API because it allows calculating the weight once in advance for efficiency during sorting.

ge0ffrey:
I can imagine a scenario were a user implements a custom ranking to prefer 1 solver config that has a good solution and uninitialized solution over another solver config that has a bad solution and an infeasible solution. (not that we'd want to support that out-of-the-box.)

oskopek:
Makes sense to let the user override the ranking and make the minimum (no?) hard-coded choices on behalf of them. I'll create a pair of abstract superclasses on top of the existing ones and keep optaplanner's actual behavior.

oskopek:
Use case: We have N solvers, which all solve M datasets. Each of the N solvers has an uninitialized solution on one (the same one) dataset. One of the solvers has worse scores on all of the datasets, but has a better (lets say 0/0) score on the uninitialized dataset. Therefore, it wins on TotalScore comparing (but it shouldn't, I think). Suggestions:

A. add an accumulated variable to SolverBenchmarkResult called "totalScoreWithoutUninitialized"/"averageScoreWithoutUninitialized" and use that in the comparisions
B. Change the existing "totalScore"/"averageScore" to ignore uninitialized single benchmarks

Either way, we should refer to this in the docs.

ge0ffrey:
Good question. I think total score should do this in order:
- Less failing configs wins: if config A has less failures than B, then A wins.
- Less uninitialized configs wins: if config A has less uninitialized solutions than B, then A wins.
- Less uninitialized variables wins: if config A has less sum(unitialized var count) than B, then A wins
- Less hard constraints broken wins. Well actually, higher hard score wins.
- Higher soft score wins.

Most business users will say that anything that fails, is uninitialized or infeasible is useless - and they don't care about their inner ranking - but they way to get feasible, is first to fix all failures, then fix all unitialized vars and then fix all infeasibles, so we do care about their inner ranking.

oskopek:
Follow up from a call.

Ranking (most important to least):
* Failing single benchmarks - unranked.
* Most uninitialized configs
* Most uninitialized variables - check if API allows this right now (add a JIRA about refactoring Score to support scoring an uninitialized solution and link it to the 7.0 release task)
* RankingComparator/RankingWeightFactory specific ranking

ToDo:
* Show uninitialized score in the report (prefixing "Nuninitialized/..." to the score, where N is the number of uninitialized variables) -- display it in the same way as in the examples swing app
* Finish refactoring of the ranking code

ge0ffrey:
These 2 should be seen as 1:
- Most uninitialized variables - check if API allows this right now (add a JIRA about refactoring Score to support scoring an uninitialized solution and link it to the 7.0 release task)
- RankingComparator/RankingWeightFactory specific ranking

Only for TotalScore, it should look at Most uninitialized variables total. For Worst Score, it should look at Most uninitialized variables of worst singleB. For TotalRank, the singleB ranking should already take the uninitialized var count into account.

oskopek:
Should Worst Score compare the uninitialized variable count before or after comparing the scores? Which score is worse: A (0uninitialized/0hard/-1soft) or B (1uninitialized/0hard/0soft)?

I'd say B.

ge0ffrey:
B is worst. Consider uninitialized as a super_hard level (a level before the hard level).
Or stated in big words (TM): The "/" in the string pattern are the token separators for lexicographic comparison.

oskopek:
* I'm unsure about the naming or the class javadoc comment of `UniversalBenchmarkComparatorTool`
* The build fails between the first two commits. If this is a problem, will fix.
  * Note: will make smaller commits (especially the first two) next time...

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
```diff
@@ -55,8 +56,8 @@
     private Long usedMemoryAfterInputSolution = null;
 
     private Boolean succeeded = null;
-    private Boolean initialized = null;
     private Score score = null;
+    private Integer uninitializedVariableCount = null;
```
ge0ffrey: uninitializedVariableCount should be before the field score, because it's logically earlier that. I am going to merge anyway already - if you could send in a separate PR later

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverBenchmarkResult.java
```diff
@@ -58,6 +58,7 @@
 
     private Integer failureCount = null;
     private Integer uninitializedSolutionCount = null;
+    private Integer uninitializedVariableCountSum = null;
```
ge0ffrey: Consistency: for score we call it totalScore, not scoreSum.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/ScoreUtils.java
```diff
@@ -79,6 +79,14 @@ public static double calculateTimeGradient(Number[] totalDiffNumbers, Number[] s
         return timeGradient;
     }
 
+    // TODO remove me (and all occurences) once https://issues.jboss.org/browse/PLANNER-405 is fixed
+    public static String getScoreWithUninitializedPrefix(int uninitializedVariableCount, Score score) {
```
ge0ffrey: We 'll want the Swing UI to use this method too, as it does this code there somewhere already

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
```diff
@@ -438,7 +438,12 @@
         </listitem>
       </itemizedlist>
 
-      <para>You can also use a custom ranking, by implementing a <literal>Comparator</literal>:</para>
+      <para>All of the supported rankings do not rank solvers which have at least one failed single benchmark.</para>
+
+      <para>If <literal>Solver</literal>s have single benchmarks with uninitialized solutions,
+      they are by default ranked last - the bigger the sum of uninitialized variables in single benchmarks
+      for a given <literal>Solver</literal> is, the worse its ranking will be.
```
ge0ffrey: "the bigger the sum of uninitialized variables... the worse its ranking " Is that still accurate in our new viewpoint? I might just state it more generally to avoid going into implementation details - I doubt anyone would care about them as the report raw data is clear for them to reverse engineer the ranking manually. In this particular case, I'd keep it short and focus on intent instead of specifics:
"Solvers with at least one failed benchmark don't get a ranking." [1] "Solvers with not fully initialized solutions are ranked worse."
[1] (because IIRC that's not even overridable - as it's a clear user error there is no need to make it overridable)

ge0ffrey:
Merged, thanks - great work. If you want to handle those last 4 comments, send in a separate PR please.

PR #121, opened 2015-08-05
PLANNER-391-refactor: Changed rank number if solvers are to be equally ranked

oskopek:
* Instead of numbering equal solver with the highest possible rank, number them with the lowest possible rank.
  * Example: if two solver configs are to be ranked after three better solver configs, their rank numbers would be: 0, 1, 2, 4, 4. Originally: 0, 1, 2, 3, 3.
* This is just a cosmetic touch and not important, also depends on personal preference. I personally found the numbering to be more logical this way. Open to discussion/closing the PR.
* **NOTE**: Merge after PLANNER-390 PR #119 **and** PLANNER-391 PR #120

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
@@ -320,10 +320,11 @@ public class PlannerBenchmarkResult {
                 benchmarkReport, rankableSolverBenchmarkResultList);
         int ranking = 0;
         for (List<SolverBenchmarkResult> sameRankingList : sameRankingListList) {
+            ranking += sameRankingList.size() - 1;
             for (SolverBenchmarkResult solverBenchmarkResult : sameRankingList) {
                 solverBenchmarkResult.setRanking(ranking);
             }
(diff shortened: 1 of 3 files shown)


ge0ffrey:
In English, when they say "They are tied for 2th place" - unless I am mistaken - it means there's someone on 1th place and no one on 3th place.
We should always follow scientific standards, then en_US locale standards (or even better adjust dynamically based on the benchmarkReport.locale) and then our personal preference.

So -1 on this PR because of the "they are tied for 2th place" saying which favors "0, 1, 2, 3, 3". I am not aware of any scientific standards for the other way.
Note: we could argue that solver rankings should start at 1 (check with Lukas and JiriV if they think this should be changed - I don't prefer either way).

oskopek:
Not sure if it's exactly a scientific standard, but I found this [paragraph on Wikipedia](https://en.wikipedia.org/wiki/Ranking#Standard_competition_ranking_.28.221224.22_ranking.29) about the "Standard competition ranking", i.e. the "1224" ranking, which agrees with your notion. I agree that the way it is right now is the right way to do it.

As for starting at 1, I'm not really sure; people generally count from 1, but software engineers (most of benchmarker's users?) don't :) Will talk with Jirka and possibly Lukas and let you know what they think.

ge0ffrey:
Agreed :) I 'll close this PR.

PR #126, opened 2015-08-19
PLANNER-391: Show uninitialized score in swingui

oskopek:
Didn't find any place where the Swing UI would use an "uninitialized" score prefix, so I added it (not very cleanly with all the casts, I must admit). The colors were changed to correspond with the usage in the benchmarker.


ge0ffrey:
This is the best way to stopgap it for now. But the lacking of uninitializedCount in AbstractScore is starting to create a lot of technical debt in Swing UI, in the benchmarker, in BestSolutionChangedEvent, in DefaultSolverScope, ...
We should add a comment "// TODO Clean when fixing jiraLink" in all these places and consider fixing that jira for 6.3. (I believe there's a jira already but I might be wrong).
Wdyt?

oskopek:
I agree, had an uneasy feeling when I wrote this part of the code. Will find/create the JIRA and add TODO comments.

ge0ffrey:
Thanks! :)

PR #127, opened 2015-08-19
Add solver ranking class and environment mode to benchmarker report

oskopek:
Addendum to #120.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
@@ -38,6 +38,7 @@ import org.optaplanner.benchmark.impl.report.BenchmarkReport;
 import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.solver.Solver;
 import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.config.solver.EnvironmentMode;
 import org.optaplanner.core.config.util.ConfigUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -65,6 +66,7 @@ public class PlannerBenchmarkResult {
 
     private Integer parallelBenchmarkCount = null;
     private Long warmUpTimeMillisSpentLimit = null;
+    private EnvironmentMode environmentMode = null;
 
     @XStreamImplicit(itemFieldName = "solverBenchmarkResult")
     private List<SolverBenchmarkResult> solverBenchmarkResultList = null;
@@ -149,6 +151,10 @@ public class PlannerBenchmarkResult {
 
     public void setWarmUpTimeMillisSpentLimit(Long warmUpTimeMillisSpentLimit) {
         this.warmUpTimeMillisSpentLimit = warmUpTimeMillisSpentLimit;
+    }
+
+    public EnvironmentMode getEnvironmentMode() {
+        return environmentMode;
     }
 
     public List<SolverBenchmarkResult> getSolverBenchmarkResultList() {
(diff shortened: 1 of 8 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
```diff
@@ -60,6 +61,7 @@
     private Score score = null;
     private long timeMillisSpent = -1L;
     private long calculateCount = -1L;
+    private EnvironmentMode environmentMode = null;
```
ge0ffrey: The EnvironmentMode is set in the solver config, so 7 single benchmarks belonging to the same solver benchmark will always have the same environmentMode, so this field is denormalized.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
```diff
@@ -277,6 +283,7 @@ public void accumulateResults(BenchmarkReport benchmarkReport) {
 
     private void determineTotalsAndAverages() {
         failureCount = 0;
+        boolean environmentModeSet = false;
```
ge0ffrey: Any variable ending with Set is usually expected to be a java.util.Set. For clarity, we might want to rename it.

Or just do a boolean first:

boolean first = true;
for (...) {
   if (first) {
      ...
      first = false;
   } else {
     ...
   }
   ...
}

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
oskopek: The problem is, if the solver config has no environment mode set, the default doesn't get reflected in the config, therefore `solverBenchmarkResult.getSolverConfig().getEnvironmentMode()` is always null, f.e. in the GeneralOptaPlannerBenchmark.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
oskopek: Fixed and noted.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SingleBenchmarkResult.java
ge0ffrey: See this line in SolverConfig:
  EnvironmentMode environmentMode_ = defaultIfNull(environmentMode, EnvironmentMode.REPRODUCIBLE);

I would recommend to extract a method called determineEnvironmentMode() of that line and reuse that from the benchmark report.

In any case, every singleBResult for the same solverBResult will always have the same environmentMode, so solverBResult, right? Even in a aggregation this still holds true. This means that the singleBResult should not have a field environmentMode (due to rule of normalization).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverBenchmarkResult.java
```diff
@@ -359,6 +361,10 @@ protected void determineStandardDeviation() {
         }
     }
 
+    protected EnvironmentMode determineEnvironmentMode() {
+        return ObjectUtils.defaultIfNull(solverConfig.getEnvironmentMode(), EnvironmentMode.REPRODUCIBLE);
```
ge0ffrey: This duplicates SolverConfig line 172: if the default ever changes there (to PRODUCTION for example), then this one would have to change too and the developer might miss that.
I'd rather see the  determineEnvironmentMode() method on solverconfig and make line 172 use that as well as this method (which I would name getEnvironmentMode(), even though it delgates to determineEnvironmentMode() because on Config objects the getters/setters must be simple (for XStream, JAXB IIRC), but on Result objects they don't need to be simple)

PR #128, opened 2015-08-19
PLANNER-284: Ability to report multiple scores from a single rule

oskopek:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -235,11 +235,21 @@ private BigDecimalConstraintMatchTotal findBigDecimalConstraintMatchTotal(RuleCo
         return matchTotal;
     }
 
-    private AgendaItem prepareAgendaItemForUnMatchListener(RuleContext kcontext) {
+    private AgendaItem prepareAgendaItemForUnMatchListener(RuleContext kcontext, int scoreLevel) {
         AgendaItem agendaItem = (AgendaItem) kcontext.getMatch();
-        if (agendaItem.getActivationUnMatchListener() != null) {
-            // Both parameters null because they are not used by the ActivationUnMatchListener created anyway
-            agendaItem.getActivationUnMatchListener().unMatch(null, null);
+        ConstraintMatchTotal matchTotal = null;
```
ge0ffrey: Why is this declared outside of the if?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -235,11 +235,21 @@ private BigDecimalConstraintMatchTotal findBigDecimalConstraintMatchTotal(RuleCo
...
+            matchTotal = constraintMatchTotalMap.get(Arrays.asList(rule.getPackageName(), rule.getName(), scoreLevel));
+            if (agendaItem.getActivationUnMatchListener() != null && matchTotal != null) {
+                // Both parameters null because they are not used by the ActivationUnMatchListener created anyway
+                agendaItem.getActivationUnMatchListener().unMatch(null, null);
+            }
+        } else {
+            if (agendaItem.getActivationUnMatchListener() != null) {
+                // Both parameters null because they are not used by the ActivationUnMatchListener created anyway
+                agendaItem.getActivationUnMatchListener().unMatch(null, null);
+            }
```
ge0ffrey: I don't see how this fixes the bug, especially if constraintMatch is not enabled?

HardSoftScoreHolder.addHardConstraintMatch(...) calls registerIntConstraintMatch() which calls:
  agendaItem.setActivationUnMatchListener(constraintUndoListener);
It's important to note the "set" in that drools method "setActivationUnMatchListener". It means any existing ActivationUnMatchListener is removed.

So if a user calls addHardConstraintMatch(...) and addSoftconstraintMatch(), they will still uising the same agendaItem, so last-one-wins: only the last one's ActivationUnMatchListener actually works.

ge0ffrey: I don't think we can actually make that AgendaItem unique for 2 addConstraintMatch calls for the same score rule. Which means we probably need to always register is a multiplex ActivationUnMatchListener to the AgendaItem and add our specific ActivationUnMatchListener in that multiplexer... The mutiplexer can have at most 1 ActivationUnMatchListener per score level.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
oskopek: Refactoring fail, fixed.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
oskopek: It doesn't, as I mentioned in the PR description above.

The problem is exactly like you described it in the first comment. Proposal B's solution of a complex ActivationUnMatchListener is probably a better way to go than creating a "unMatchMap" in the score holder, as this would essentially either duplicate or (in the case of constraint matching disabled) imitate the constraintMatchMap. I'll make a quick implementation of B and lets discuss it from there.

oskopek:
Please check the tests behave like they should (especially the `assertEquals(N, scoreHolder.getConstraintMatchTotals().size())` assertion)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendablebigdecimal/BendableBigDecimalScoreHolder.java
```diff
@@ -80,7 +80,7 @@ public void undo() {
 
     public void addSoftConstraintMatch(RuleContext kcontext, final int softLevel, final BigDecimal weight) {
         softScores[softLevel] = softScores[softLevel].add(weight);
-        registerBigDecimalConstraintMatch(kcontext, softLevel, weight, new BigDecimalConstraintUndoListener() {
+        registerBigDecimalConstraintMatch(kcontext, getHardLevelsSize() + softLevel, weight, new BigDecimalConstraintUndoListener() {
```
ge0ffrey: Nice fix - it might even deserve it's own jira as it means that BENDABLE_BIG_DECIMAL score was unreliable, apparently since JUN-2014 according to git log.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -67,16 +68,44 @@ public boolean isConstraintMatchEnabled() {
     // Worker methods
     // ************************************************************************
 
+    protected class MultiLevelConstraintUndoListener implements ActivationUnMatchListener {
+
+        private Map<Integer, ActivationUnMatchListener> activationUnMatchListenerMap;
+
+        public MultiLevelConstraintUndoListener(int scoreLevel, ActivationUnMatchListener activationUnMatchListener) {
+            this.activationUnMatchListenerMap = new HashMap<Integer, ActivationUnMatchListener>();
```
ge0ffrey: I wondering about the memory impact of creating all these HashMaps.
We have a listener instance for each rule for each match. The HashMap.DEFAULT_INITIAL_CAPACITY is 16.

oskopek: Few ideas:
* A: Initialize capacity statically to 1 or 2 (most score types have two levels, and 99% of rules will only need 1 level (backwards compatibility))
* B: Initialize capacity to `scoreLevel` (might be good in the average case, although is a bit "pulled out of a hat")
* C: Somehow set it according to the score type to an exact (maximal) size for the given type

ge0ffrey: D: if there's only 1 undo listener, put it in directly, only use the MultiLevelConstraintUndoListener if there are 2 score levels. But that won't work as we'll need to know the undo listener's score level

A: +1 Good idea. HashMap's automatically expand anyway 
B: -1 If we have a bendable with 1000 levels, then we have a serious scalability problem: even with 10k matches each using just 1 level, we reserve 10m memory. Also, the AbstractScoreHolder needs to add an abstract method to get the score level size from it's children.
C: +0
D: +0 painful

ge0ffrey: Doing some experiments with visualvm. I 've got 2 checkouts and 2 intellij windows of optaplanner (one with these changes added and one which I call old-optaplanner because it's without those changes) - running them next to each other to see difference.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -67,16 +68,44 @@ public boolean isConstraintMatchEnabled() {
...
+            for (Integer scoreLevel : activationUnMatchListenerMap.keySet()) {
+                unMatch(scoreLevel);
+            }
+        }
+
+        public void unMatch(int scoreLevel) {
+            if (activationUnMatchListenerMap.containsKey(scoreLevel)) {
+                // Both parameters null because they are not used by the ActivationUnMatchListener created anyway
+                activationUnMatchListenerMap.get(scoreLevel).unMatch(null, null);
+                activationUnMatchListenerMap.remove(scoreLevel);
```
ge0ffrey: containsKey, get and remove each do a lookup of the scoreLevel, so the lookup happens 3 times per scoreLevel. Replacing with entrySet() and clear() usage.

oskopek: Agreed, performance > re-usability in this case.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -67,16 +68,44 @@ public boolean isConstraintMatchEnabled() {
     // Worker methods
     // ************************************************************************
 
+    protected class MultiLevelConstraintUndoListener implements ActivationUnMatchListener {
```
ge0ffrey: inner class can be static

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -67,16 +68,44 @@ public boolean isConstraintMatchEnabled() {
...
+        public final void unMatch(RuleRuntime ruleRuntime, Match match) {
+            for (Integer scoreLevel : activationUnMatchListenerMap.keySet()) {
+                unMatch(scoreLevel);
+            }
+        }
+
+        public void unMatch(int scoreLevel) {
+            if (activationUnMatchListenerMap.containsKey(scoreLevel)) {
+                // Both parameters null because they are not used by the ActivationUnMatchListener created anyway
+                activationUnMatchListenerMap.get(scoreLevel).unMatch(null, null);
```
ge0ffrey: We have a composition pattern now, but we could actually use a real ConstraintUndoListener interface, instead of reusing the ActivationUnMatchListener and presuming this null parameters validity. The problem is that will affect backwards compatiblity for the AbstractScoreHolder user definied subclasses... but there's a workaround for that too.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolderTest.java
```diff
@@ -46,12 +46,19 @@ public void addConstraintMatch(boolean constraintMatchEnabled) {
...
         scoreHolder.addSoftConstraintMatch(ruleContext3, 0, -3); // Rule match modified
+        scoreHolder.addHardConstraintMatch(ruleContext3, 0, -300); // Rule of different level added
+        scoreHolder.addHardConstraintMatch(ruleContext3, 0, -400); // Rule of different level modified
 
         scoreHolder.addSoftConstraintMatch(createRuleContext("scoreRule4"), 1, -4); // Rule match added
 
-        assertEquals(BendableScore.valueOf(new int[]{-1000}, new int[]{-3, -4}), scoreHolder.extractScore());
+        RuleContext ruleContext5 = createRuleContext("scoreRule5");
+        scoreHolder.addHardConstraintMatch(ruleContext5, 0, -1);
+        scoreHolder.addSoftConstraintMatch(ruleContext5, 0, -1);
```
ge0ffrey: One trick I use is to avoid using the same decimal order twice in a number used in a unit test.
It's far easier to sum 1000, 200, 30 and 4 than the sum of 1, 2, 3 and 4, so therefor it's far easier to verify the validity of a unit test by reading it :)

ge0ffrey:
I 've done some general benchmark comparisons and it seems there's be a performance degradation - I am going to take my time to do some experiments to see if there's a way around it, if it doesn't make CR2, so be it (I 'll adjust the jira accordingly).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
oskopek: I'd go with A, specifically initial size 1.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
oskopek: +1

optaplanner-core/src/test/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolderTest.java
oskopek: Nice trick, will keep that in mind :)

oskopek:
@ge0ffrey ok, also, don't forget to adjust the release notes.

ge0ffrey:
Merged manually, thanks! :)

PR #129, opened 2015-08-20
Added overheating warning for parallelBenchmarkCount > 1 to benchmarker docs

oskopek:
Not sure if we want to have 3 indented blocks in a row.


optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
```diff
@@ -952,6 +952,16 @@
           If it's higher, it will be automatically decreased.</para>
         </note>
 
+        <warning>
```
ge0ffrey: I'd make it a note. I still presume your laptop type is the exception to the rule.

oskopek: Fixed.

PR #130, opened 2015-08-21
PLANNER-409: Name benchmark directories uniquely + test

oskopek:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
```diff
@@ -226,13 +226,23 @@ public void initBenchmarkReportDirectory(File benchmarkDirectory) {
         if (StringUtils.isEmpty(name)) {
             name = timestamp;
         }
-        benchmarkReportDirectory = new File(benchmarkDirectory,
-                BooleanUtils.isFalse(aggregation) ? timestamp : timestamp + "_aggregation");
-        boolean benchmarkReportDirectoryAdded = benchmarkReportDirectory.mkdirs();
-        if (!benchmarkReportDirectoryAdded) {
-            throw new IllegalArgumentException("The benchmarkReportDirectory (" + benchmarkReportDirectory
-                    + ") creation failed. It probably already exists.");
+        if (!benchmarkDirectory.mkdirs()) {
```
ge0ffrey: The if and while condition aren't just read methods, they also have the side-effect of making the directory. That can hurt readability / code expectations, although it's very debatable too.

For consistency with the rest of optaplanner code, would you mind creating a boolean variable as in the original code?

ge0ffrey: Nevermind this. It's too debatable indeed.

PR #131, opened 2015-08-21
Constraint matches dialog redraw fix

oskopek:
SwingUI Constraint matches dialog now redraws correctly when viewed multiple times in a single solver run

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/common/swingui/ConstraintMatchesDialog.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/common/swingui/ConstraintMatchesDialog.java
@@ -56,7 +56,7 @@ public class ConstraintMatchesDialog extends JDialog {
         this.solutionBusiness = solutionBusiness;
     }
 
-    public void resetContentPanel() {
+    public void showDialog() {
         JPanel buttonPanel = new JPanel(new FlowLayout());
         Action okAction = new AbstractAction("OK") {
             @Override
@@ -114,6 +114,7 @@ public class ConstraintMatchesDialog extends JDialog {
         }
         pack();
         setLocationRelativeTo(getParent());
+        setVisible(true);
     }
 
     public String buildConstraintMatchSetText(ConstraintMatchTotal constraintMatchTotal) {
@@ -134,7 +135,7 @@ public class ConstraintMatchesDialog extends JDialog {
         }
 
         public int getRowCount() {
-            return constraintMatchTotalList.size();
+            return constraintMatchTotalList == null ? 0 : constraintMatchTotalList.size();
         }
 
         public int getColumnCount() {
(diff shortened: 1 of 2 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/common/swingui/ConstraintMatchesDialog.java
```diff
@@ -56,6 +56,14 @@ public ConstraintMatchesDialog(SolverAndPersistenceFrame solverAndPersistenceFra
         this.solutionBusiness = solutionBusiness;
     }
 
+    @Override
+    public void setVisible(boolean visible) {
+        if (!visible) {
+            setContentPane(new JPanel()); // force redraw on visible(true)
```
ge0ffrey: setVisible() shouldn't mess with the contentPane.
I suspect this works because it calls Container.validate() indirectly.

IIRC, there are 2 methods in Swing to fix these kinds of bugs:
 repaint(); // typically with a custom paint() method this needs to be called, otherwise it's overkill
 validate(); (or invalidate()?) // typically changing something invalidates the layout, and Swing wants to wait until all your changes are done before calculating the new layout. validate() is used to signal to Swing that you're done. IIRC...

oskopek: You're right, of course. I wasn't aware of `invalidate()`, only of `validate()`, and that didn't work. Fixed, will push ASAP.

oskopek: Seems not. Will investigate further.

oskopek:
Decided to refactor a bit (we now create a dialog each time and let the GC take care of it afterwards). I tried just about every combination of `(re|in|^)validate` and `repaint`, none of which worked reliably. Also, as this is a modal dialog window, it really makes sense to create it and throw away afterwards.

psiroky:
dear Jenkins, retest this please (just for checking the phrase is accepted and test results are correctly reported)

ge0ffrey:
I can't reproduce this issue (haven't applied the code changes yet). Might you remember a reproduce recipe?

ge0ffrey:
@oskopek This is good to merge, if we can prove that there is indeed a redraw issue currently (I can't seem to reproduce it).

oskopek:
Great! I'll take a look at this on Thu.

On Jan 26, 2016 2:24 PM, "Geoffrey De Smet" <notifications@github.com>
wrote:
>
> @oskopek This is good to merge, if we can prove that there is indeed a
redraw issue currently (I can't seem to reproduce it).

oskopek:
Cannot reproduce either. Pushed a correctly rebased commit just for fun.

Some other change must have fixed it, or my updated system/JRE did. Closing.

PR #132, opened 2015-08-25
PLANNER-259: Persist SingleStatistic point lists to save memory

oskopek:
* On longer benchmark runs with a large count of single statistics, the memory savings can be quite large: about 20MB per statistic per single benchmark (the number is not accurate at all). See JIRA attached screenshots.
* The toll for the memory optimization comes at the end, when generating the report: There is a ~11% slow down (about 17 seconds) in a very big benchmark (61 single benchmarks, 10min each).

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
@@ -228,9 +228,10 @@ public class BenchmarkReport {
                     for (SingleStatistic singleStatistic : problemStatistic.getSingleStatisticList()) {
                         try {
                             singleStatistic.unhibernatePointList();
-                        } catch (IllegalStateException ise) {
+                        } catch (IllegalStateException e) {
                             if (!plannerBenchmarkResult.getAggregation()) {
-                                throw ise;
+                                throw new IllegalStateException("Failed to unhibernate point list of SingleStatistic ( "
+                                        + singleStatistic + " ) of ProblemStatistic ( " + problemStatistic + " ).", e);
                             }
                             logger.trace("This is expected, aggregator doesn't copy CSV files. Could not read CSV file "
                                     + "( {} ) of single statistic ( {} ).", singleStatistic.getCsvFile().getAbsolutePath(), singleStatistic);
@@ -250,9 +251,10 @@ public class BenchmarkReport {
                         for (PureSingleStatistic pureSingleStatistic : singleBenchmarkResult.getPureSingleStatisticList()) {
                             try {
                                 pureSingleStatistic.unhibernatePointList();
-                            } catch (IllegalStateException ise) {
+                            } catch (IllegalStateException e) {
                                 if (!plannerBenchmarkResult.getAggregation()) {
-                                    throw ise;
+                                    throw new IllegalStateException("Failed to unhibernate point list of "
+                                            + "PureSingleStatistic ( " + pureSingleStatistic + " ).", e);
                                 }
                                 logger.trace("This is expected, aggregator doesn't copy CSV files. Could not read CSV file "
                                         + "( {} ) of pure single statistic ( {} ).", pureSingleStatistic.getCsvFile().getAbsolutePath(), pureSingleStatistic);
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
@@ -67,8 +67,12 @@ import org.optaplanner.benchmark.impl.statistic.PureSingleStatistic;
 import org.optaplanner.benchmark.impl.statistic.SingleStatistic;
 import org.optaplanner.benchmark.impl.statistic.common.MillisecondsSpentNumberFormat;
 import org.optaplanner.core.impl.score.ScoreUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class BenchmarkReport {
+
+    protected final transient Logger logger = LoggerFactory.getLogger(getClass());
 
     public static final int CHARTED_SCORE_LEVEL_SIZE = 15;
     public static final int LOG_SCALE_MIN_DATASETS_COUNT = 5;
@@ -222,14 +226,30 @@ public class BenchmarkReport {
             if (problemBenchmarkResult.hasAnySuccess()) {
                 for (ProblemStatistic problemStatistic : problemBenchmarkResult.getProblemStatisticList()) {
                     for (SingleStatistic singleStatistic : problemStatistic.getSingleStatisticList()) {
-                        singleStatistic.unhibernatePointList();
+                        try {
+                            singleStatistic.unhibernatePointList();
+                        } catch (IllegalStateException ise) {
+                            if (!plannerBenchmarkResult.getAggregation()) {
+                                throw ise;
+                            }
+                            logger.trace("This is expected, aggregator doesn't copy CSV files. Could not read CSV file "
+                                    + "( {} ) of single statistic ( {} ).", singleStatistic.getCsvFile().getAbsolutePath(), singleStatistic);
+                        }
                     }
                     problemStatistic.writeGraphFiles(this);
                 }
                 for (SingleBenchmarkResult singleBenchmarkResult : problemBenchmarkResult.getSingleBenchmarkResultList()) {
                     if (singleBenchmarkResult.isSuccess()) {
                         for (PureSingleStatistic pureSingleStatistic : singleBenchmarkResult.getPureSingleStatisticList()) {
-                            pureSingleStatistic.unhibernatePointList();
+                            try {
+                                pureSingleStatistic.unhibernatePointList();
+                            } catch (IllegalStateException ise) {
+                                if (!plannerBenchmarkResult.getAggregation()) {
+                                    throw ise;
+                                }
+                                logger.trace("This is expected, aggregator doesn't copy CSV files. Could not read CSV file "
+                                        + "( {} ) of pure single statistic ( {} ).", pureSingleStatistic.getCsvFile().getAbsolutePath(), pureSingleStatistic);
+                            }
                             pureSingleStatistic.writeGraphFiles(this);
                         }
                     }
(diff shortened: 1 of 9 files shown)


ge0ffrey:
Has the Aggregator been manually tested too to assure it doesn't cause any trouble for it?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
```diff
@@ -178,6 +178,18 @@ public void readCsvStatisticFile() {
...
     }
 
+    public void retrievePointList() {
+        if (getCsvFile().exists() && pointList == null) {
+            initPointList();
+            readCsvStatisticFile();
+        }
+    }
+
+    public void persistPointList() {
```
ge0ffrey: the name "persist" in not entirely correct. It does more than persist them, it also removes them from memory.

What if we call these 2 methods "hibernatePointList" and "unhibernatePointList"? Although "hibernate" isn't perfect as it might make devs think of Hibernate ORM (the JPA impl). But I didn't find a good synonym.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -220,11 +221,15 @@ public void writeReport() {
...
+                    for (SingleStatistic singleStatistic : problemStatistic.getSingleStatisticList()) {
+                        singleStatistic.retrievePointList();
+                    }
                     problemStatistic.writeGraphFiles(this);
                 }
                 for (SingleBenchmarkResult singleBenchmarkResult : problemBenchmarkResult.getSingleBenchmarkResultList()) {
                     if (singleBenchmarkResult.isSuccess()) {
                         for (PureSingleStatistic pureSingleStatistic : singleBenchmarkResult.getPureSingleStatisticList()) {
+                            pureSingleStatistic.retrievePointList();
                             pureSingleStatistic.writeGraphFiles(this);
```
ge0ffrey: Should the pointList not be hibernated again after the graph is generated (or a bit later due to the problem statistic stuff)? Otherwise it remains in memory and that means that at some point all pointLists are in memory (which could still be too big for the memory even though the solver isn't running so that's not taking up memory).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
```diff
@@ -178,6 +178,18 @@ public void readCsvStatisticFile() {
         }
     }
 
+    public void retrievePointList() {
+        if (getCsvFile().exists() && pointList == null) {
```
ge0ffrey: Definitely fail fast if the csv file doesn't exist. Suppose I remove the CSV file manually during benchmark and this method doesn't crash - then I get some obscure NPE later, instead of a nice IllegalStateException telling me about the missing CSV file.

I would fail fast if the pointList isn't null. In theory this method is never called when pointList isn't null, and the rest of our coding presumes that. So because we haven't actively dealt with the situation of that happening, we should fail fast so it's sticks out like sore thumb when it does happen (which I presume will be never).

Suppose the CVS doesn't exist -

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: Good question. This could actually occur, but is it realistic? Even on large benchmarks (61 singles, 10min each, 5+ single statistics each), all the pointLists have about 400MB on disk (=> not much more in memory AFAIK). On 5 threads and large datasets, the memory usage is almost 4GB at times.

Of course, in a single threaded benchmark, with small datasets that run for a long time and with a low-memory environment, this could become an issue. But the time it takes to write/read the pointLists to/from disk repeatedly might be quite a time killer in all the other cases.

Will try to run a few benchmarks to see how this turns out.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
oskopek: I didn't either, so I used "persist". As for "hibernate":
* -1 exactly because of the issue you raised with the Hibernate project, it even confused me for a second
* +1 as it does effectively describe removing the pointList from memory

I'll keep on the lookout for anything better and fall back to "hibernate" if I don't find anything.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
oskopek: Fixed and noted.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
ge0ffrey: 2 votes on the same proposal is cheating :) I guess you wanted to use "pro" and "con" and you're +0 on it (which means you're hoping there's a better alternative):

According to apache voting guidelines:
-1 implies a (soft?) *veto* for as long as the comment after the "-1" cannot be countered (adding a comment is mandatory). The new situation is worse than the old.
+1 implies I think this is a good idea. The new situation is better than the old.
+0 (or -0) implies I don't care or this change can slip through or I prefer another alternative even though this has it's merits. The new situation isn't worse than the old.

The philopshy is that one +1 vote can get a change in, as long as there are none -1's. And also that -1's need to be explicitly dealt with (by turning them into +0 for example).

oskopek: I'm sorry O:) Yes, I meant to enumerate pros and cons. Just in case anyone reads this in the future: http://www.apache.org/foundation/voting.html

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: Benchmarked this. The slowdown is hardly noticeable (~3%, although, do note that this is on a very new machine with a fast SSD), and the memory efficiency is a lot better (we basically keep only the pointlists that are being used right now in memory). It could still be a problem on old hard drives though.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
ge0ffrey: Is the fix in this pull request? (waiting on this to merge)

oskopek: This PR is kind-of blocked by https://issues.jboss.org/browse/PLANNER-427 and #137. I don't want to merge this if I'm not sure it works, that's why I didn't push the fixes yet.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -220,12 +225,44 @@ public void writeReport() {
         for (ProblemBenchmarkResult problemBenchmarkResult : plannerBenchmarkResult.getUnifiedProblemBenchmarkResultList()) {
             if (problemBenchmarkResult.hasAnySuccess()) {
                 for (ProblemStatistic problemStatistic : problemBenchmarkResult.getProblemStatisticList()) {
+                    for (SingleStatistic singleStatistic : problemStatistic.getSingleStatisticList()) {
+                        try {
+                            singleStatistic.unhibernatePointList();
+                        } catch (IllegalStateException ise) {
+                            if (!plannerBenchmarkResult.getAggregation()) {
+                                throw ise;
```
ge0ffrey: Never throw an exception that is already thrown, as that eats the original stacktrace.
Just chain it:
throw new IllegalStateException("...", ise);

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -220,12 +225,44 @@ public void writeReport() {
         for (ProblemBenchmarkResult problemBenchmarkResult : plannerBenchmarkResult.getUnifiedProblemBenchmarkResultList()) {
             if (problemBenchmarkResult.hasAnySuccess()) {
                 for (ProblemStatistic problemStatistic : problemBenchmarkResult.getProblemStatisticList()) {
+                    for (SingleStatistic singleStatistic : problemStatistic.getSingleStatisticList()) {
+                        try {
+                            singleStatistic.unhibernatePointList();
+                        } catch (IllegalStateException ise) {
```
ge0ffrey: The default variable name for an exception is e.
There a few these very short default names for which it's widely accepted not to give a good meaningful name:
for (int i = 0;...)
catch (...Exception e)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/ProblemBasedSingleStatistic.java
```diff
@@ -34,4 +34,9 @@ public ProblemStatisticType getStatisticType() {
         return problemStatisticType;
     }
 
+    @Override
+    public String toString() {
+        return singleBenchmarkResult.toString() + "_" + problemStatisticType.toString();
```
ge0ffrey: when concatening strings with +, it's not needed to do .toString().
This gives the same result: 

    return singleBenchmarkResult + "_" + problemStatisticType;

Furthermore, it even avoids a nullpointerexception if for example problemStatisticType is null (which it never is in this particular case)

oskopek: Both are good comments. Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: Good point. Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: I didn't actually know that, thank you. Fixed.

PR #133, opened 2015-08-25
PLANNER-284: Code style + performance

oskopek:
* Creating a new Map instead of clearing it is debatable, feel free to leave that commit out.


ge0ffrey:
I 'll revert the clear() line indeed: if they do heavily use 20 different score levels for 1 match, then the Map capacity will have expanded to 20 already and remain that way after the clear(), improving later performance

PR #136, opened 2015-08-27
PLANNER-420: Fix

oskopek:


ge0ffrey:
thanks for fixing it quickly :)

oskopek:
No problem :) Btw, just confirmed that this really doesn't affect 6.3.x (wasn't #126 isn't backported).

PR #137, opened 2015-08-31
Refactoring for PLANNER-259

oskopek:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/ProblemStatistic.java
```diff
@@ -120,4 +121,19 @@ protected File writeChartToImageFile(JFreeChart chart, String fileNameBase) {
...
+        if (graphFileList == null || graphFileList.isEmpty()) {
+            return null;
+        } else if (graphFileList.size() > 1) {
+            throw new IllegalStateException("Cannot get graph file for the ProblemStatistic ( " + this
+                    + " ) because it has more than 1 graph file. See method getGraphList() and "
+                    + "ProblemStatisticType.hasScoreLevels()");
+        } else {
+            return graphFileList.get(0);
+        }
+    };
```
ge0ffrey: the ";" is probably obsolete

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
```diff
@@ -150,6 +150,9 @@ public void readCsvStatisticFile() {
             }
             Map<String, String> stringDuplicationRemovalMap = new HashMap<String, String>(1024);
             for (line = reader.readLine(); line != null && !line.isEmpty(); line = reader.readLine()) {
+                if (line.equals("Failed")) { // Reading statistics of a failed benchmark
+                    break;
```
ge0ffrey: I would expect it to do either of these 2 things when it finds "Failed":
- set a flag to mark this single statistic as failed (I presume that it already knows that through the benchmark xml loaded earlier)
- Check if that flag was already set (see above) and fail-fast if it didn't.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -78,6 +78,18 @@
         return ObjectUtils.equals(a, b) ? a : null;
     }
 
+    public static <T> T meldProperty(T a, T b) {
```
ge0ffrey: The difference between meld and merge isn't inherently clear. Please add a javadoc to explain the difference or rename both methods more self-explanatory.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -453,9 +453,9 @@ public int hashCode() {
                             it.remove();
                         }
                     }
-                    newResult.entityCount = ConfigUtils.mergeProperty(oldResult.entityCount, newResult.entityCount);
-                    newResult.variableCount = ConfigUtils.mergeProperty(oldResult.variableCount, newResult.variableCount);
-                    newResult.problemScale = ConfigUtils.mergeProperty(oldResult.problemScale, newResult.problemScale);
+                    newResult.entityCount = ConfigUtils.meldProperty(oldResult.entityCount, newResult.entityCount);
```
ge0ffrey: When would a meld (that is not a merge) happen? So when would a.entityCount be null and b.entityCount not be null? I would argue that can only happen when loading older benchmarks, in which the migration code should take care of that (or do nothing and just leave it to become null as merged), and not the agregation code here?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/ProblemStatistic.java
oskopek: How did that ever get in there.. Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/ProblemStatistic.java
```diff
@@ -120,4 +121,19 @@ protected File writeChartToImageFile(JFreeChart chart, String fileNameBase) {
...
     }
 
+    public File getGraphFile() {
+        List<File> graphFileList = getGraphFileList();
+        if (graphFileList == null || graphFileList.isEmpty()) {
+            return null;
+        } else if (graphFileList.size() > 1) {
+            throw new IllegalStateException("Cannot get graph file for the ProblemStatistic ( " + this
+                    + " ) because it has more than 1 graph file. See method getGraphList() and "
+                    + "ProblemStatisticType.hasScoreLevels()");
```
oskopek: Nice trick. Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SingleStatistic.java
oskopek: Fixed with the second approach.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
oskopek: It was a bug when merging a problem benchmark result with only failed singles. In SingleBenchmarkRunner, `registerScale()` never gets called (for example, the others too). It's version agnostic. So you fix the error, rerun benchmarker, and want to merge those reports. That's when it happens.

Yes, we can leave it to become null, but we know the problem scale from one of them, why not use it? It hasn't changed, the dataset is the same (see `fileToNewResultMap`.. of course, assuming the user didn't change the dataset between runs).

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
oskopek: I couldn't find a more suitable name. Will add the javadoc.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
ge0ffrey: Makes sense. Would you mind adding a comment to indicate that meld is used to deal with failed singles? I 'll already merge - just add it to another PR.

oskopek: Fixed in #140.

PR #138, opened 2015-09-01
Fixes uninitialized variable count averaging

oskopek:
Round up, more is worse.


ge0ffrey:
Nice one indeed that it needs to be ceil. In AbstractScore we should probably make uninitializedCount negative so it can follow the consistent logic of "higher is better".

oskopek:
Agreed. One more thing: I went running, and found an even better fix for this. Will open a new PR.

PR #139, opened 2015-09-01
Fixes uninitialized variable count averaging without floating point arithmetic

oskopek:
Round up, more is worse.

Sorry for creating 2 PRs, but I had this idea later and found it way better than the first one. Added a test as compensation.


ge0ffrey:
no problem for me :) Better a PR too many than too little

PR #141, opened 2015-09-02
PLANNER-427: Fix

oskopek:
*Ready for review.*


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -169,13 +172,41 @@ private void warmUp() {
...
+                            + " ) already contained key ( " + singleBenchmarkResult + " ) with value ( "
+                            + singleBenchmarkStatisticPutResult + " ).");
+                }
+                singleBenchmarkResult.setPureSingleStatisticList(Collections.<PureSingleStatistic>emptyList());
+
+                ProblemBenchmarkResult problemBenchmarkResult = singleBenchmarkResult.getProblemBenchmarkResult();
+                List<ProblemStatistic> originalProblemStatisticList = problemBenchmarkResult.getProblemStatisticList();
+                if (!originalProblemStatisticMap.containsKey(problemBenchmarkResult)) { // TODO: After Java 8, do Map#putIfAbsent
+                    List<ProblemStatistic> problemStatisticPutResult = originalProblemStatisticMap.put(problemBenchmarkResult, originalProblemStatisticList);
+                    if (problemStatisticPutResult != null && !originalProblemStatisticList.isEmpty()) {
```
ge0ffrey: the !originalProblemStatisticList.isEmpty() check is obsolete because the put() on that map will always return null if it's empty.

oskopek: Well, yes, in this example, it should be obsolete, but in theory, it isn't: What if the actual value is null for that given key?

Of course, all of this gets obsoleted by the `containsKey` check a few lines above. Fixing.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -338,4 +378,23 @@ public long calculateTimeMillisSpent() {
         return now - startingSystemTimeMillis;
     }
 
+    private static final class ConfigBackup {
```
ge0ffrey: ConfigBackup is too general - it doesn't state it's purpose.

I would include the word "warmUp" or even "duringWarmUp" to indicate it's purpose/reason of existence

Some proposals:
- WarmUpConfigBackup
- BenchmarkConfigBackupDuringWarmUp
- OriginalConfigPartsDuringWarmUp
- ...

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: It's a put(), not a get(). Whether the actual value is null doesn't matter. See put()'s javadoc return value :)

Furthermore, for a get() that returns null but the key does exist, an isEmpty() check wouldn't suffice as it would be be broken if the map contains other keys too.

oskopek: Citings put's javadoc: "**returns**    the previous value associated with key, or null if there was no mapping for key. (A null return can also indicate that the map previously associated null with key, if the implementation supports null values.)"

As far as I can tell, put behaves exactly like get regarding their return value (put additionally overrides the value for the key in the map, but that doesn't affect the return value).

Of course, you're right that the `!isEmpty()` check doesn't really make sense here: it will always be false, because we put a value in the line above, regardless of the putted value being null/non-null.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: A: +1, short, simple, concise 
B: -1, too long
C: -1, originally, I had something like BackupSolverConfigParts, but I found it to be quite unreadable, just as (C)

D1: OriginalConfigBackup
D2: OriginalBenchmarkConfigBackup

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -169,13 +172,41 @@ private void warmUp() {
...
+                    }
+                }
+                singleBenchmarkResult.getProblemBenchmarkResult().setProblemStatisticList(Collections.<ProblemStatistic>emptyList());
+                singleBenchmarkResult.initSingleStatisticMap();
+            }
+            ConfigBackup configBackupPutResult = configBackupMap.put(solverBenchmarkResult, configBackup);
+            if (configBackupPutResult != null) {
+                throw new IllegalStateException("ConfigBackupMap already contained key ( " + solverBenchmarkResult
+                        + " ) with value ( " + configBackupPutResult + " ).");
             }
```
ge0ffrey: Can we move the whole backup and restoring logic into a method in that new inner class (currently called BackupConfig)?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: Ah yes - I stand corrected on (the put() javadoc) - was thrusting my memory too much :)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
oskopek: Good idea. Will fix.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
ge0ffrey: +1 for A
+0 for others

PR #142, opened 2015-09-02
Small fixes

oskopek:
(Mostly) 1 fix per commit. See commit messages.

**Merge only after** #141.

**Ready to merge after build passes.**

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/config/util/ConfigUtilsTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/config/util/ConfigUtilsTest.java
@@ -24,6 +24,34 @@ import static org.optaplanner.core.config.util.ConfigUtils.*;
 
 public class ConfigUtilsTest {
 
+    @Test
+    public void testMergeProperty() {
+        Integer a = null;
+        Integer b = null;
+        assertEquals(null, mergeProperty(a, b));
+        a = Integer.valueOf(1);
+        assertEquals(null, mergeProperty(a, b));
+        b = Integer.valueOf(10);
+        assertEquals(null, mergeProperty(a, b));
+        b = Integer.valueOf(1);
+        assertEquals(Integer.valueOf(1), mergeProperty(a, b));
+        a = null;
+        assertEquals(null, mergeProperty(a, b));
+    }
+
+    @Test
+    public void testMeldProperty() {
+        Integer a = null;
+        Integer b = null;
+        assertEquals(null, meldProperty(a, b));
+        a = Integer.valueOf(1);
+        assertEquals(Integer.valueOf(1), meldProperty(a, b));
+        b = Integer.valueOf(10);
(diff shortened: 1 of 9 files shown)


ge0ffrey:
Ondrej, did you automatically get a mail because the jenkins build failed?

oskopek:
@ge0ffrey no, I didn't. Did you?

ge0ffrey:
No, me neither. Checking with Petr - it would be nice that the submitter of a PR gets a mail if it fails on jenkins.

cc @psiroky

psiroky:
I agree it would be nice. However, with the current setup it is not as easy as it may seem. I'll need to check the Jenkins plugins doing the PR builds if this is easily possible.

psiroky:
retest this please

(testing if the e-mailing for failed build works)

psiroky:
Looks like the Jenkins is refusing to re-run the build, because there were no changes since the last build. That technically makes sense, but still I thought it would simply run the build if told so. To test the e-mailing there would need to be change in the PR branch (new commit or force push that changes the SHAs)

oskopek:
@psiroky will open a new PR where we can test this ASAP

ge0ffrey:
Looks all good, but it seems to include PR 141.
Shall I wait with merging until PR 141 is fixed and will this PR be updated accordingly? Or is there another way to avoid havoc?

oskopek:
@ge0ffrey That is on purpose, else the new tests would fail. Will rebase after #141 is merged and ping you when that's done.

optaplanner-core/src/test/java/org/optaplanner/core/config/util/ConfigUtilsTest.java
```diff
@@ -0,0 +1,88 @@
...
+        a = Integer.valueOf(1);
+        assertEquals(Integer.valueOf(1), meldProperty(a, b));
+        b = Integer.valueOf(10);
+        assertEquals(mergeProperty(Integer.valueOf(1), Integer.valueOf(10)), meldProperty(a, b));
+        a = null;
+        assertEquals(Integer.valueOf(10), meldProperty(a, b));
+    }
+
+    private static void checkCeilDivision(int dividend, int divisor) {
+        long result = Math.round(Math.ceil(dividend / (double) divisor));
```
ge0ffrey: This is too much logic for unit tests.

I 'd just write a test like this:
assertEquals(4, ceilDivide(20, 5));
assertEquals(4, ceilDivide(19, 5));
assertEquals(5, ceilDivide(21, 5));
assertEquals(-4, ceilDivide(-20, 5));
assertEquals(-3, ceilDivide(-19, 5));
assertEquals(-5, ceilDivide(-21, 5));

oskopek: Although I disagree (mainly because I wrote the tests first and then checked the implementation against the tests), I see your point of view on this. The coverage is exactly the same as I had with this sort of test implementation (apart from the test logic, ofc).

This is your call, do I change it back or keep what we have?

PR #146, opened 2015-09-09
FIX: SolverBenchmarkResult Correct failed singles handling in average score with uninitialized prefix

oskopek:
Very sorry for breaking the build with #145. This will fix it.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverBenchmarkResult.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverBenchmarkResult.java
@@ -269,6 +269,9 @@ public class SolverBenchmarkResult {
     }
 
     public String getAverageScoreWithUninitializedPrefix() {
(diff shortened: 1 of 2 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverBenchmarkResult.java
```diff
@@ -269,6 +269,9 @@ public String getTotalScoreWithUninitializedPrefix() {
     }
 
     public String getAverageScoreWithUninitializedPrefix() {
+        if (getSuccessCount() == 0) {
+            return null; // return null, not String "null" so we can use a default value
```
ge0ffrey: We should never use the string "null" anywhere - there really is never a valid case for it - so I wouldn't even bother adding this comment.

Merging already.

oskopek: Will fix both comments.

PR #149, opened 2015-09-16
Replace ARQ with Cargo for integration testing

psiroky:
* tests renamed to *IntegrationTest to be automatically
   picked up by failsafe instead of surefire
 * mvn clean install will start WFLY, run the tests,
   stop WFLY

There are two main disadvantages when using Cargo (but the advantages outweigh these in my opinion):
 * you can't create and run in-container tests (which you are not trying to do in this case it seems)
 * you can't debug the tests by single IDE click since the container is managed by maven plugin (not by JUnit runner like with ARQ). What I usually do is to start+deploy with `mvn cargo:run` and then debug the tests from IDE (just the debugging for test code, for remote debugging there is additional config needed)


psiroky:
The test will fail, but it is not related to the Cargo itself.

oskopek:
Just a personal note: I know about the dependency issue that arquillian is causing us from the upstream bom, nevertheless I personally do prefer directly arquillian over other libraries/tools (including cargo). + it's a JBoss community project

ge0ffrey:
@oskopek Upgrading to ShrinkWrap Resolver  2.2.0-beta-2 might solve our problem too (as that does depend on a newer aether version). But this is an interesting alternative none the less. The version hell of arquillian (this was just the last issue of many) is painful.

ge0ffrey:
I'll try both approaches tomorrow (or day after, need to finish the hibernate issue first)

optaplanner-webexamples/src/test/java/org/optaplanner/webexamples/common/rest/service/AbstractClientArquillianTest.java
```diff
@@ -1,68 +0,0 @@
...
-public abstract class AbstractClientArquillianTest {
-
-    private static final String POM_DIRECTORY_NAME = "optaplanner-webexamples";
-
-    @Deployment(testable = false)
-    public static WebArchive createDeployment() {
-        File file = findPomFile();
-        return ShrinkWrap.create(MavenImporter.class)
-                .loadPomFromFile(file)
-                .importBuildOutput()
```
ge0ffrey: The fact that we can get rid of this is very nice, as this has limitations in case we'd do GWT compilation etc.

optaplanner-webexamples/pom.xml
```diff
@@ -34,41 +35,81 @@
...
+            </deployable>
+          </deployables>
+
+        </configuration>
+        <executions>
+          <execution>
+            <id>start-container</id>
+            <phase>pre-integration-test</phase>
+            <goals>
+              <goal>start</goal>
```
ge0ffrey: I presume this means we can't just click on the test in IntelliJ and run it? We first need to start a wildfly for it to run on (and preferably not the same wildfly installation as the one we're using for our manual tests)?

psiroky: Yes, as mentioned in the PR description, you can't run/debug the tests using single click in IDE (at least now right now, it would require some plugin for the IDE).

ge0ffrey:
I am rejecting this PR (even though it was very educational and interesting) because we got Arquillian to work and mainly because Arquillian allows in IDE testing and client-side debugging (which is very handy during development).
Sorry Petr.

PR #150, opened 2015-09-21
PLANNER-401: Statistical benchmarking

oskopek:
* There are probably a lot of code style errors - I will try to fix them continuously as we find them
  * Hopefully fixed
* We should discuss the **naming and the usage of the `SolverProblemBenchmarkResult` interface**
  * Fixed.
* Please see the TODO in SubSingleBenchmarkResult: for some reason, **the `@XStreamImplicit` tag doesn't work there** - I'm not sure why.
  * Fixed (automagically, through removing stats from SingleBR)
* There are a lot of **incompatibilities with older versions** of `plannerBenchmarkResult`s -> this is a problem that's observed while running the aggregator. They are directly related to me changing the SingleBenchmarkResult API. We should discuss this.
  * Wont-fix (we ignore it for now).
* Unfortunately, the commits aren't exactly what I'd call tidy. That means, some commits may break the build, but it will get fixed in a later commit, or the messages don't describe the changes correctly (I really need to learn to make logical, independent commits early on). Best thing to do would probably be to review this all at once. - If you want me to change a commit message or restructure a few commits, it's no problem.
  * Wont-fix, but the newer commits are better structured.


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/SolverBenchmarkConfig.java
```diff
@@ -88,6 +100,10 @@ protected void validate() {
             throw new IllegalStateException("The solverBenchmark name (" + name
                     + ") is invalid because it starts or ends with whitespace.");
         }
+        if (subSingleCount == null || subSingleCount < 1) {
```
ge0ffrey: If it's null it's invalid? If it's null, shouldn't it be valid (and default to 1 basically)?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/SolverBenchmarkConfig.java
```diff
@@ -101,6 +117,9 @@ public void inherit(SolverBenchmarkConfig inheritedConfig) {
         } else if (inheritedConfig.getProblemBenchmarksConfig() != null) {
             problemBenchmarksConfig.inherit(inheritedConfig.getProblemBenchmarksConfig());
         }
+        if (subSingleCount == null) {
+            subSingleCount = ObjectUtils.defaultIfNull(inheritedConfig.getSubSingleCount(), 1);
```
ge0ffrey: Reading an Config object and then writing it again should not alter the XML. This code will do that.

This is better:
if (subSingleCount == null) {
            subSingleCount = inheritedConfig.getSubSingleCount();
}
But there's a ConfigUtils.method to do that for you.

Don't do defaulting in the inherit() method, do it in locally in the build methods (so it doesn't change the global fields of a config class.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/ranking/SingleBenchmarkRankingComparator.java
```diff
@@ -31,8 +31,8 @@
     public int compare(SingleBenchmarkResult a, SingleBenchmarkResult b) {
         return new CompareToBuilder()
                 .append(b.isFailure(), a.isFailure()) // Reverse, less is better (redundant: failed benchmarks don't get ranked at all)
-                .append(b.getUninitializedVariableCount(), a.getUninitializedVariableCount()) // Reverse, less is better
-                .append(a.getScore(), b.getScore(), resilientScoreComparator)
+                .append(b.getAverageUninitializedVariableCount(), a.getAverageUninitializedVariableCount()) // Reverse, less is better
+                .append(a.getAverageScore(), b.getAverageScore(), resilientScoreComparator)
```
ge0ffrey: The upgrade recipe will need to make note of this (as [MINOR]), as some users might have a custom Ranking implementation.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -205,11 +214,21 @@ public String getSolverRankingClassFullName() {
...
+    public List<File> getSubSingleBenchmarkAggregationChartFileMapEntry(ProblemBenchmarkResult problemBenchmarkResult) {
+        return subSingleBenchmarkAggregationChartFileMap.get(problemBenchmarkResult);
+    }
+
     // ************************************************************************
     // Write methods
     // ************************************************************************
 
     public void writeReport() {
+        logger.info("Generating benchmarking report...");
```
ge0ffrey: Good idea! This feedback to the user has been missing.

Replace "benchmarking report" by "benchmark report", as most of the code/manual calls is the later term.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -529,6 +549,64 @@ private void writeBestScorePerTimeSpentSummaryChart() {
         }
     }
 
+    private void writeSubSingleBenchmarkScoreCharts() {
```
ge0ffrey: Thinking aloud: Does this method belong on this BenchmarkReport?
It only had summary charts. These subsingle score charts are per problem I guess? But it doesn't hurt to generate it all the time (iff subSingleCount > 1 of course) as it doesn't impact the solver speed. So it is kinda like a summary chart and it's not... Let's keep it here for now.

Will these charts be made obsolete by the summary version of this chart:
  https://issues.jboss.org/browse/PLANNER-443
Or does that summary version has less detailed information? Let's worry about that later (we can easily delete this one if needed).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -79,6 +78,9 @@
     private Long variableCount = null;
     private Long problemScale = null;
 
+    @XStreamOmitField // loaded lazily from singleBenchmarkResults
```
ge0ffrey: Comments start with a capital

If this XStream ommitted, shouldn't we XStream omit other fields potentially too?
- Maybe ProblemBenchmarkResult.problemScale
- Definitely SolverBenchmarkResult.averageAverageCalculateCountPerSecond
Please make a separate jira to capture that potential improvement, but let's ignore it for now to avoid going out of scope.

All these fields get refreshed when reading from xml (such as in an Aggregation) through PlannerBenchmarkResult#accumulateResults(...). I presume this field hasNonDefaultSubSingleCount is refreshed too when accumulateResults is called?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -79,6 +78,9 @@
...
+    private Boolean hasNonDefaultSubSingleCount = null;
```
ge0ffrey: Naming. For what is this boolean used?
If it is used to determine whether or not something should be shown because showing just 1 subsingle is pointless,
then it should not be named "...NonDefault..." because the default value doesn't matter then (as it's a coincidence that the default is 1 too), instead a name like "subSingleCountIsMultiple" or "subSingleCountAtLeastTwo" or maybe track an int such as "int subSingleCount" or "int maximumSubSingleCount" (in case there are multiple singles per problem and they could have a different subSingleCount such as during an Aggregator).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -220,6 +222,20 @@ public boolean hasProblemStatisticType(ProblemStatisticType problemStatisticType
         return false;
     }
 
+    public boolean hasSubSingleCountBiggerThanOne() {
+        if (hasNonDefaultSubSingleCount != null) {
+            return hasNonDefaultSubSingleCount;
+        }
+        for (SingleBenchmarkResult singleBenchmarkResult : singleBenchmarkResultList) {
+            if (singleBenchmarkResult.getSolverBenchmarkResult().getSubSingleCount() > 1) {
+                hasNonDefaultSubSingleCount = true;
```
ge0ffrey: This calculation should happen indirectly through PlannerBenchmarkResult#accumulateResults(...), so somewhere during accumulateResults(). It should not happen in the lazy getter.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -220,6 +222,20 @@ public boolean hasProblemStatisticType(ProblemStatisticType problemStatisticType
         return false;
     }
 
+    public boolean hasSubSingleCountBiggerThanOne() {
+        if (hasNonDefaultSubSingleCount != null) {
+            return hasNonDefaultSubSingleCount;
```
ge0ffrey: The name of this getter and the boolean field should be the same (as long as they are the same thing which they currently are)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
```diff
@@ -0,0 +1,334 @@
...
+
+    protected static final transient Logger logger = LoggerFactory.getLogger(SubSingleBenchmarkResult.class);
+
+    @XStreamOmitField
+    private SingleBenchmarkResult singleBenchmarkResult;
+
+    private int subSingleBenchmarkIndex = -1;
+
+    // @XStreamImplicit() // TODO FIXME
+    private List<PureSingleStatistic> subPureSingleStatisticList = null;
```
ge0ffrey: A SubSingleResult has a List of PureSingleStatistics? That seems wrong to me.
Either it has a List of PureSubSingleStatistics or it parent Single has a list of PureSingleStatistics.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
```diff
@@ -0,0 +1,334 @@
...
+/**
+ * Represents 1 benchmark run for 1 Single Benchmark configuration for 1 {@link Solver} configuration for 1 problem
+ * instance (data set).
+ */
+@XStreamAlias("subSingleBenchmarkResult")
+public class SubSingleBenchmarkResult implements SolverProblemBenchmarkResult {
+
+    protected static final transient Logger logger = LoggerFactory.getLogger(SubSingleBenchmarkResult.class);
+
+    @XStreamOmitField
```
ge0ffrey: Add the comment "// Bi-directional relationship restored through BenchmarkResultIO"

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
```diff
@@ -0,0 +1,334 @@
...
+                    oldSingleStatistic.getStatisticType().buildPureSingleStatistic(newResult));
+        }
+
+        newResult.initSubSingleStatisticMap();
+        for (SingleStatistic singleStatistic : newResult.effectiveSingleStatisticMap.values()) {
+            SingleStatistic oldSingleStatistic = oldResult.getSingleStatistic(singleStatistic.getStatisticType());
+            if (!oldSingleStatistic.getCsvFile().exists()) {
+                if (oldResult.isFailure()) {
+                    singleStatistic.initPointList();
+                    logger.debug("Old result ( {} ) is a failure, skipping merge of it's single statistic ( {} ).",
```
ge0ffrey: "( {} )" => "({})" the other logs don't add spaces between the braces.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatsUtil.java
```diff
@@ -0,0 +1,87 @@
...
+
+package org.optaplanner.benchmark.impl.statistic;
+
+import java.util.List;
+
+import org.optaplanner.benchmark.impl.result.SolverProblemBenchmarkResult;
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.score.ScoreUtils;
+
+public class StatsUtil {
```
ge0ffrey: Mind renaming this to StatisticUtils?
- All other Util(s) are plural: Utils not Util
- Don't abbreviate if it's not too verbose.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/single/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreSingleStatistic.java
```diff
@@ -205,7 +205,7 @@ public void writeGraphFiles(BenchmarkReport benchmarkReport) {
             }
             plot.setDataset(seriesCollection);
             JFreeChart chart = new JFreeChart(
-                    singleBenchmarkResult.getName() + " constraint match total best score diff level " + scoreLevelIndex + " statistic",
+                    solverProblemBenchmarkResult + " constraint match total best score diff level " + scoreLevelIndex + " statistic",
```
ge0ffrey: The lacking of the .getName() call means it will use to toString() method instead.
As this shows up in the UI (the benchmark report), it should still use the getName() method. toString() is primary for debugging and logs. In many cases this is hair splitting indeed, as they are often the same.

optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
```diff
@@ -351,7 +379,35 @@
...
+                                                            <#--<li class="dropdown-header"><strong>Standard Deviation</strong></li>-->
+                                                            <#--<li>${singleBenchmarkResult.standardDeviationStringAverageCalculateCountPerSecond!""}/s</li>-->
+                                                            <#--<li class="dropdown-header"><strong>Best</strong></li>-->
+                                                            <#--<li>${singleBenchmarkResult.bestAverageCalculateCountPerSecond!""}/s</li>-->
+                                                            <#--<li class="dropdown-header"><strong>Worst</strong></li>-->
+                                                            <#--<li>${singleBenchmarkResult.worstAverageCalculateCountPerSecond!""}/s</li>-->
+                                                            <#--<li class="dropdown-header"><strong>Median</strong></li>-->
+                                                            <#--<li>${singleBenchmarkResult.medianAverageCalculateCountPerSecond!""}/s</li>-->
```
ge0ffrey: When do these comments expire?

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
```diff
@@ -974,6 +974,49 @@
...
+  ...
+  &lt;solverBenchmark&gt;
+    &lt;name&gt;...&lt;/name&gt;
+    &lt;solver&gt;
+      ...
+    &lt;/solver&gt;
+    &lt;subSingleCount&gt;10&lt;subSingleCount&gt;
+  &lt;/solverBenchmark&gt;
+  ...
+&lt;/plannerBenchmark&gt;</programlisting>
```
ge0ffrey: I would show an example with the inheritedSolverBenchmark instead of the solverBenchmark because 99% of the time people will only use it there. So we should explain the most common use case first. See also writing of problemBenchmarks: even though that works on a solverBenchmark, the docs first teach you that it works on inheritedSolverBenchmark (as that's where you'll use it) and only then they briefly mention IIRC that it also works on a solverBenchmark level (without an example as that's just wasting paper space).

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
```diff
@@ -974,6 +974,49 @@
...
+
+      <para>You can also configure <literal>subSingleCount</literal> in the
+        <link linkend="inheritedSolverBenchmark"><literal>&lt;inheritedSolverBenchmark&gt;</literal></link> element
+        and possibly overwrite it in the individual <literal>&lt;solverBenchmark&gt;</literal> elements.</para>
+
+      <note>
+        <para><literal>subSingleCount</literal> is by default set to 1. Any positive integer is a
+          valid value.</para>
+      </note>
```
ge0ffrey: Might be noting the obvious? We probably already have too many note and warning elements in the docs already. Maybe just add "(defaults to 1)" somewhere in the paragraphs above - and I might not even mention it's a positive integer (as the example code makes that obvious - is a count ever negative? - if a user is distracted enough to put a negative number there, he'll get a clear error message anyway).

PS: It's a strictly positive integer, not a positive integer because 0 isn't allowed :)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatsUtil.java
oskopek: Fixed & noted.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverProblemBenchmarkResult.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.optaplanner.benchmark.impl.result;
+
+import java.io.File;
+
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.config.solver.SolverConfig;
+
+public interface SolverProblemBenchmarkResult {
```
ge0ffrey: This SolverProblemBenchmarkResult should be renamed back to SingleBenchmarkResult (I presume it had this temporary name due to refactoring),
as well as all the fields/getters that were renamed (excluding those that needed to be renamed to SubSingle* of course).

Oh wait. This is an interface. Why has this interface been extracted? If we extract it here, why don't we extract it for the others too? If we want to keep the interface, then this interface should be named SingleBenchmarkResult and it's implemtation DefaultSingleBenchmarkResult.
But do we want to keep this interface?

ge0ffrey:
Great work! Looking forward to merging this and trying it out :)
This is a big step forward for OptaPlanner Benchmark.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/single/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreSingleStatistic.java
oskopek: Fixed & noted.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
oskopek: Fixed along with all other occurrences.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: Had a similar thought process when I wrote this, and decided to leave it here.

Not sure about this chart being obsoleted by PLANNER-443, as that animal is a bit different from my point of view.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
oskopek: Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
ge0ffrey: Agreed to leave it here (I don't see a better place either).
As for PLANNER-443, we 'll cross that bridge when we get there.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/SolverBenchmarkConfig.java
oskopek: Fixed and added tests.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/SolverBenchmarkConfig.java
oskopek: Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
oskopek: Fixed.

optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
oskopek: As a professor of mine used to say: "Copy-pasting is a common source of errors." Fixed (removed).

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
oskopek: Fixed. Wasn't aware of the terminology: we use the term positive for [1,infinity) and non-negative for [0,infinity).

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
oskopek: Makes sense. Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
oskopek: Fixed.

Makes sense. Created JIRA: https://issues.jboss.org/browse/PLANNER-453

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
oskopek: Fixed. `maximumSubSingleCount` is the best name here, as it gives us more potential information for the future at no cost. Kept the `boolean ...()` method though, just renamed it to `isMaximumSubSingleCountMultiple()`. Moved the calculation into `accumulateResults`, specifically `determineTotalsAndAveragesAndRanking` to only take the maximum from successful benchmarks.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
oskopek: Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
oskopek: Sort-of-fixed: reworked to:
```java
public boolean isMaximumSubSingleCountMultiple() {
    return maximumSubSingleCount != null ? maximumSubSingleCount > 1 : false;
}
```

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
ge0ffrey: +1

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
oskopek: Doesn't really make sense to create additional Statistics, as they are the same as before. If you meant only renaming them, well, that doesn't sound like an improvement either, as that would make a mess in SingleBenchmarkResult.

+1 that we could somehow decide to store the statistics differently: right now, they're copied from the parent Single when the SubSingles are created, and upon accumulating, the median's statistics are copied back to the parent. See [this JIRA comment](https://issues.jboss.org/browse/PLANNER-401?focusedCommentId=13105187&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13105187) proposal D2c. (Proposal: move D4 into a new JIRA - we could add those charts Jirka is talking about, but separately - maybe)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/ranking/SingleBenchmarkRankingComparator.java
oskopek: The question here is, if we want to do this change (readability vs. backwards compatibility). We could keep using `.getScore()` and `.getUninitializedVariableCount()` and just internally flip the switch to mirror `.getAverageScore()` and `.getAvg...Count()`. This would mean reverting the changes I made all across the benchmarker code base.

See `SingleBenchmarkResult#getScore()`.  WDYT?

ge0ffrey: That was also in mind, but I think we need to make this change anyway.
That way the API is clear that it's an average.

I think we can count the number of users that have a custom ranking can be counted on 1 hand.
But even if that wasn't the case: I think impl changes (so non public api changes!) that impact user code should not influence our decision on whether or not to make them significantly. Long-term it's our API that matters and at some point that impl api will become public api so it should be as perfect as possible by then. But we need to clearly document every potential user impacting change in the Upgrade recipe, to at least minimize the pain inflicted on our users.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
ge0ffrey: Agreed that the renaming isn't the right solution here, which means there's a normalization problem instead:
Does this work instead of making a field for them: getSingleBenchmarkResult().getPureSingleStatisticList()?

oskopek: Concurrency would fail, as SubSingles of the same Single could be run at the same time. That's why we're technically building new instances during copying from Single->SubSingle.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/ranking/SingleBenchmarkRankingComparator.java
oskopek: +1, agreed. Will add a note to the upgrade recipe.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverProblemBenchmarkResult.java
oskopek: This interface was created due to Single Statistics, at first. Then it became apparent, that we need something representing a "SolverProblemBenchmark" (anything that can be described as a cross section of a Problem and Solver Benchmark - Singles and Subsingles): when calculating the standard deviation of scores, in the report when adding badges.

`SolverProblemBenchmarkResult` was the best thing I came up with, as it fits into the model quite nicely. I'm not sure if we want to have more of them, as there's no necessary need so far. As for naming, I agree that is debatable, although I'm not sure that SubSingleBenchmark implementing SingleBenchmark and DefaultSingleBenchmark implementing SingleBenchmark is the way to go, but it's not necessarily bad.

oskopek: Fixed. Renamed to `BenchmarkResult` as agreed in a meeting.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SubSingleBenchmarkResult.java
oskopek: Fixed, removed stats from SingleBR completely.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/SolverBenchmarkConfig.java
```diff
@@ -20,16 +20,20 @@
...
 import org.optaplanner.benchmark.impl.result.SolverBenchmarkResult;
 import org.optaplanner.core.config.solver.SolverConfig;
+import org.optaplanner.core.config.util.ConfigUtils;
 
 @XStreamAlias("solverBenchmark")
 public class SolverBenchmarkConfig {
 
     private String name = null;
 
+    private Integer subSingleCount = null;
```
ge0ffrey: IIRC, the order should be solverConfig, problemBenchmarksConfig, subSingleCount because users first think about the others before deciding how many subsingles to run

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/ProblemBenchmarksConfig.java
```diff
@@ -178,22 +179,35 @@ private ProblemBenchmarkResult buildProblemBenchmark(PlannerBenchmarkResult plan
...
             }
         }
-        singleBenchmarkResult.setPureSingleStatisticList(pureSingleStatisticList);
-        singleBenchmarkResult.initSingleStatisticMap();
+        singleBenchmarkResult.initSubSingleStatisticMaps();
         solverBenchmarkResult.getSingleBenchmarkResultList().add(singleBenchmarkResult);
         problemBenchmarkResult.getSingleBenchmarkResultList().add(singleBenchmarkResult);
     }
 
+    private void buildSubSingleBenchmarks(SingleBenchmarkResult parent, int count) {
```
ge0ffrey: nitpicking: I might name the count param subSingleCount to avoid any confusion (and to make it clear it indirectly comes from the <subSingleCount> element in the configs)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/statistic/SubSingleStatisticType.java
```diff
@@ -17,33 +17,33 @@
...
-import org.optaplanner.benchmark.impl.statistic.single.constraintmatchtotalstepscore.ConstraintMatchTotalStepScoreSingleStatistic;
-import org.optaplanner.benchmark.impl.statistic.single.pickedmovetypebestscore.PickedMoveTypeBestScoreDiffSingleStatistic;
-import org.optaplanner.benchmark.impl.statistic.single.pickedmovetypestepscore.PickedMoveTypeStepScoreDiffSingleStatistic;
+import org.optaplanner.benchmark.impl.statistic.subsingle.constraintmatchtotalbestscore.ConstraintMatchTotalBestScoreSubSingleStatistic;
+import org.optaplanner.benchmark.impl.statistic.subsingle.constraintmatchtotalstepscore.ConstraintMatchTotalStepScoreSubSingleStatistic;
+import org.optaplanner.benchmark.impl.statistic.subsingle.pickedmovetypebestscore.PickedMoveTypeBestScoreDiffSubSingleStatistic;
+import org.optaplanner.benchmark.impl.statistic.subsingle.pickedmovetypestepscore.PickedMoveTypeStepScoreDiffSubSingleStatistic;
 
-public enum SingleStatisticType implements StatisticType {
+public enum SubSingleStatisticType implements StatisticType {
```
ge0ffrey: SingleStatisticType must not be renamed to SubSingleStatisticType:
They are configured on a single, not on a subsingle.
This is similar to how a ProblemStatisticType is configured on a problem, not a single or a subsingle.
Both ProblemStatisticType and SingleStatisticType result into PureSubSingleStatistic objects being made.
So the build() method changes below are good.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResult.java
```diff
@@ -0,0 +1,51 @@
...
+     * @return true if there is a failed child benchmark and the variable is initialized
+     */
+    public boolean hasAnyFailure();
+
+    /**
+     * @return true if all child benchmarks were a success and the variable is initialized
+     */
+    public boolean hasAllSuccess();
+
+    public String getName();
```
ge0ffrey: Nitpicking: the getName() is consistently ordered first IIRC (at least before hasAnyFailure())

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SubSingleStatistic.java
```diff
@@ -187,10 +200,10 @@ private void readCsvStatisticFile() {
     public void unhibernatePointList() {
         if (!getCsvFile().exists()) {
             throw new IllegalStateException("The csvFile ( " + getCsvFile() + " ) of the statistic ( " + getStatisticType()
-                    + " ) of the single benchmark ( " + singleBenchmarkResult + " ) doesn't exist.");
+                    + " ) of the single benchmark ( " + subSingleBenchmarkResult + " ) doesn't exist.");
```
ge0ffrey: Consistency nitpicking: no spaces inside the braces:

    + " ) of the single benchmark ( " + subSingleBenchmarkResult + " ) doesn't exist.");

becomes

    + ") of the single benchmark (" + subSingleBenchmarkResult + ") doesn't exist.");

optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
```diff
@@ -182,10 +182,38 @@
...
+                                                                <li class="dropdown-header"><strong>Individual runs</strong></li>
+                                                                <li role="separator" class="divider"></li>
+                                                                <#list singleBenchmarkResult.subSingleBenchmarkResultList as subSingleBenchmarkResult>
+                                                                    <li class="dropdown-header"><strong>Run #${subSingleBenchmarkResult.getSubSingleBenchmarkIndex()}</strong></li>
+                                                                    <li>${subSingleBenchmarkResult.scoreWithUninitializedPrefix!""}&nbsp;<@addSolverProblemBenchmarkResultBadges solverProblemBenchmarkResult=subSingleBenchmarkResult/></li>
+                                                                </#list>
+                                                                <li role="separator" class="divider"></li>
+                                                                <li class="dropdown-header"><strong>Average</strong></li>
```
ge0ffrey: We should probably look into replacing these strong usages with a CSS class, but let's worry about that in a different jira, as this probably the tip of the iceberg.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/util/PlannerTestUtils.java
```diff
@@ -85,6 +89,21 @@ public Score calculateScore(Solution solution) {
 
     }
 
+    public static String readResourceToString(Class clazz, String resource) throws IOException {
```
ge0ffrey: Use IOUtils instead, this method is obsolete.
We have 3 libraries supplying such common stuff for us: commons-lang3, commons-io and guava. One option is to use common-io's IOUtil.readLines(getClass().getResourceAsStream("..."), "UTF-8") but that returns a List<String>. IIRC, of the other util classes has a methods to return just a String.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/util/PlannerTestUtils.java
```diff
@@ -85,6 +89,21 @@ public Score calculateScore(Solution solution) {
...
+        InputStream resourceInputStream = clazz.getResourceAsStream(resource);
+        String resourceContent = "";
+        String buffer;
+        BufferedReader br = new BufferedReader(new InputStreamReader(resourceInputStream));
```
ge0ffrey: Never create a reader without specifying that the encoding is UTF-8, or we get platform dependent code (= bugs on windows that default to the encoding "windows-1252"). findbugs finds these bugs for us.

optaplanner-distribution/src/main/assembly/filtered-resources/UpgradeFromPreviousVersionRecipe.txt
```diff
@@ -2791,3 +2791,22 @@ After in *.java:
...
+[MINOR] If you're using SingleBenchmarkResult: the functionality of the method getScore() and getUninitializedVariableCount() has moved to
+getAverageScore() and getAverageUninitializedVariableCount() because of the addition of SubSingleBenchmarkResult (Statistical benchmarking).
+Before in *.java:
+    singleBenchmarkResult.getScore();
+    singleBenchmarkResult.getUninitializedVariableCount();
+After in *.java:
+    singleBenchmarkResult.getAverageScore();
+    singleBenchmarkResult.getAverageUninitializedVariableCount();
+
+[MINOR] If you're using SingleStatistic or any of it's subclasses: all of them were renamed to include "Sub" as a part of
```
ge0ffrey: I'd change this to an [IMPL_DETAIL]. Avoid any confusion that the <singleStatisticType> elements in benchmark configs need to be renamed too, because they don't. (see earlier comment)

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
```diff
@@ -114,7 +114,7 @@
         not work on Linux and Mac.</para>
       </note>
 
-      <para>To lower verbosity, the common parts of multiple <literal>&lt;solverBenchmark&gt;</literal> elements are
+      <para xml:id="inheritedSolverBenchmark">To lower verbosity, the common parts of multiple <literal>&lt;solverBenchmark&gt;</literal> elements are
```
ge0ffrey: We don't id's to para's because it's hard select those in the browser when copy pasting into a stackoverflow answer.

It seems as inheritedSolverBenchmark warrants it's own section, so it can be linked to easily?

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
```diff
@@ -779,14 +779,14 @@
     &lt;problemBenchmarks&gt;
       ...
       &lt;problemStatisticType&gt;...&lt;/problemStatisticType&gt;
-      &lt;singleStatisticType&gt;PICKED_MOVE_TYPE_BEST_SCORE_DIFF&lt;/singleStatisticType&gt;
+      &lt;subSingleStatisticType&gt;PICKED_MOVE_TYPE_BEST_SCORE_DIFF&lt;/subSingleStatisticType&gt;
```
ge0ffrey: See comment earlier, this won't change, as the type is configured on a single.

Furthermore (regardless that it makes more sens to configure on single than on subsingle), even if we wanted to change this: we can't because the xml is considered backwards compatible... (If and only if using the element in question does not require you to use classes from the impl package. For example the element `<terminationClass>` isn't guaranteed backwards compatible because no one can use it without also using the impl class Termination.)

optaplanner-benchmark/src/test/resources/org/optaplanner/benchmark/impl/testdataInputSolutionFile.xml
```diff
@@ -0,0 +1,8 @@
+<TestdataSolution id="1">
```
ge0ffrey: Where is the XStream alias that makes sure this is "TestdataSolution" instead of "org.optaplanner.core.impl.testdata.domain.TestdataSolution"?

I am not sure about these test xml... it could create a maintenance burden. Normally I use the example datasets xml for integration tests and for unit tests just generate the domain model at runtime.

Of course, having a test is more important than perfecting it.

ge0ffrey: Leave these tests in, let me worry about it when I pull it locally.

oskopek: IIRC I did add aliases to the domain classes in -core.

ge0ffrey: Firefox's ctrl-f is telling me differently? Ignore if it's just my FF.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SubSingleStatistic.java
oskopek: Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResult.java
oskopek: Fixed

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/ProblemBenchmarksConfig.java
oskopek: +1, fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/SolverBenchmarkConfig.java
oskopek: Agreed, fixed.

optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
oskopek: I'll leave the creation of this JIRA up to you, as I don't know the reasons for this change (not saying there aren't any).

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/util/PlannerTestUtils.java
oskopek: Couldn't find the right utils class. Thanks. Fixed to `IOUtils.toString(getClass().getResourceAsStream(solverConfigResource), "UTF-8");`

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/util/PlannerTestUtils.java
oskopek: Thanks, will take a note. Won't fix, as I removed the method.

optaplanner-benchmark/src/test/resources/org/optaplanner/benchmark/impl/testdataInputSolutionFile.xml
oskopek: Wow, that's weird. I'm basically sure I added them in. Adding them in again.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/statistic/SubSingleStatisticType.java
oskopek: Just to confirm I understand what you're saying: *SingleStatistic should be renamed (as is), SingleStatisticType should not be renamed.

optaplanner-distribution/src/main/assembly/filtered-resources/UpgradeFromPreviousVersionRecipe.txt
oskopek: Fixed.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/statistic/SubSingleStatisticType.java
oskopek: Fixed.

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
oskopek: Added a new section. Fixed & noted.

optaplanner-docs/src/main/docbook/en-US/Chapter-Benchmarking_and_tweaking/Chapter-Benchmarking_and_tweaking.xml
oskopek: Fixed (from the above changes) & noted.

ge0ffrey:
Merged. Very nice!

oskopek:
Woo! :)

PR #152, opened 2015-09-26
Added formatted names to StatisticTypes and benchmark report

oskopek:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticType.java
```diff
@@ -23,4 +23,9 @@
      */
     public String name();
 
+    /**
+     * @return never null
+     */
+    public String formattedName();
```
ge0ffrey: Good idea to pretty up the report, just some thoughts on naming, to be consistent with the rest of the code:

Proposal A) Either we use "getLabel()" instead of "getFormattedName()" (and rename name() to getName())
This is in sync with the example domain objects which often have a toString() for logging and getLabel() for the UI (but no getName();

Proposal B) "getFormattedName()" and "name()" both becomes "getName()". A StatisticType gets a field name and a constructor that sets it. This is similar to JobType.
The filename for the csv stats is the getName(), just like the solver benchmark name and the problem benchmark name are used for the directory names.
Advantages:
- name can differ from enum key. For example: "Semi-special stat" and SEMI_SPECIAL_STAT.
- better localization opportunity

oskopek: Good proposals. I'm not really sure which one we should choose.

A) I agree, but we should consider keeping `name()` - as our enum overrides it by itself IIRC: http://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#name%28%29

Else `getName()` would basically just `return name()`

B) This proposal is intriguing, but I don't really like the idea that we would localize something we use for folder/file names. I know we already support it for benchmark names, but doesn't feel right either.

ge0ffrey: Good point on B). For benchmarks names we have no choice, as Japanese users even use japanese characters (and that works) to name them.

Let's call that proposal C) to keep name() and rename getFormattedName() to getLabel(). That class was originally an enum, which explains.

+1 on all 3 proposals A), B) and C) - which is probably very helpful :) - your call which to go for.

oskopek: I don't really have any preference, but I'm going with C as that has the least impact for now. Fixed.

PR #155, opened 2015-09-28
Fixed large number handling of getStandardDeviationString and added tests

oskopek:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
```diff
@@ -76,10 +76,24 @@ public static String getStandardDeviationString(double[] standardDeviationDouble
             // Abbreviate to 2 decimals
             // We don't use DecimalFormat to abbreviate because it's written locale insensitive (like java literals)
             int dotIndex = abbreviated.lastIndexOf('.');
-            if (dotIndex >= 0 && dotIndex + 3 < abbreviated.length()) {
-                abbreviated = abbreviated.substring(0, dotIndex + 3);
+            StringBuilder result = new StringBuilder().append(abbreviated.substring(0, dotIndex < 0 ? abbreviated.length() : dotIndex));
```
ge0ffrey: Nitpick: StringBuilder has this method:
  public StringBuilder append(CharSequence s, int start, int end)

This avoids the substring()'s string creation and is trivially, completely unnoticeable faster.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
```diff
@@ -76,10 +76,24 @@ public static String getStandardDeviationString(double[] standardDeviationDouble
...
+                        break;
+                    }
+                    char currentChar = abbreviatedChars[dotIndex + i];
+                    if (Character.isDigit(currentChar) || currentChar == '.') {
+                        result = result.append(currentChar);
+                    }
+                }
+                int expIndex = abbreviated.lastIndexOf('E');
+                if (expIndex >= 0) {
+                    result = result.append(abbreviated.substring(expIndex));
```
ge0ffrey: no need to put the append return value back into the result. it's always the same instance with StringBuilder (unlike String and BigDecimal). The reason they return this is simply for method chaining: result.append("i = ").append(i);

optaplanner-benchmark/src/test/java/org/optaplanner/benchmark/impl/statistic/StatisticUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Test
+    public void testGetStandardDeviationString() throws Exception {
+        assertEquals(null, StatisticUtils.getStandardDeviationString(null));
+        assertEquals("2.0", StatisticUtils.getStandardDeviationString(new double[]{2d}));
+        assertEquals("1.41", StatisticUtils.getStandardDeviationString(new double[]{Math.sqrt(2)}));
+        assertEquals("1.60E8", StatisticUtils.getStandardDeviationString(new double[]{160338212.294d}));
+        assertEquals("2.0E9", StatisticUtils.getStandardDeviationString(new double[]{2000000000d}));
+        assertEquals("2.0E10", StatisticUtils.getStandardDeviationString(new double[]{20000000000d}));
```
ge0ffrey: We should use DecimalFormat for this kind of stuff, not format it ourselves
  http://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormat.html

DecimalFormat is locale sensitive, so pass it the reportLocale.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
oskopek: Oh, I assumed otherwise automatically. Fixed.

optaplanner-benchmark/src/test/java/org/optaplanner/benchmark/impl/statistic/StatisticUtilsTest.java
ge0ffrey: DecimalFormat is used elsewhere in the code too.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
oskopek: Fixed and hopefully remembered.

optaplanner-benchmark/src/test/java/org/optaplanner/benchmark/impl/statistic/StatisticUtilsTest.java
oskopek: @ge0ffrey What about this comment in `getStandardDeviationString`?
```
// Abbreviate to 2 decimals
// We don't use DecimalFormat to abbreviate because it's written locale insensitive (like java literals)
```

Of course, I can refactor the code to use DecimalFormat and keep the original behavior.

ge0ffrey: Good question why that comment is there. No idea why we wouldn't want to write a number locale sensitive in the benchmark report that clearly notes it's locale at the bottom.

ge0ffrey: Because the comment doesn't motivate why it should be locale insensitive, git annotate doesn't refresh my memory and I currently believe the opposite is true, I think we should remove it now and print a locale sensitive value. Wdyt?

oskopek: I thought it was to imitate the printing of the other scores -> i.e. to print the StdDev as if it were a score

ge0ffrey: That makes sense. That's it indeed. The other scores are always printed as java literals (so always "." as decimal marker for example, even though nl_BE uses ","). And in the long term we might want everything to be localized to the report locale, but then it should happen for standard deviation and the scores at the same time. We might not want to.

In any case, for now use a new DecimalFormat() instance, fix the locale on Locale.en_US  and supply a pattern so it only prints the "." and the "E" if needed, but no "'" etc. Adjust the comment to indicate that it doesn't use the report locale because the other scores don't use it yet either.

oskopek: Fixed, please confirm that the output is formatted as should be (see test).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
```diff
@@ -72,12 +77,12 @@ public static String getStandardDeviationString(double[] standardDeviationDouble
...
             }
-            String abbreviated = Double.toString(standardDeviationDouble);
-            // Abbreviate to 2 decimals
-            // We don't use DecimalFormat to abbreviate because it's written locale insensitive (like java literals)
-            int dotIndex = abbreviated.lastIndexOf('.');
-            if (dotIndex >= 0 && dotIndex + 3 < abbreviated.length()) {
-                abbreviated = abbreviated.substring(0, dotIndex + 3);
+            // See http://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#toString%28double%29
+            String abbreviated;
+            if (0.001d <= standardDeviationDouble && standardDeviationDouble <= 10000000d) {
```
ge0ffrey: About the "d" suffix for the numbers: the rest of the code uses capital L for longs (because lower case it's hard to see the difference with a 1), capital F for floats, but no suffix for ints and a ".0" for doubles (but no "D").

optaplanner-benchmark/src/test/java/org/optaplanner/benchmark/impl/statistic/StatisticUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.junit.Assert.*;
+
+import static org.mockito.Mockito.*;
+
+public class StatisticUtilsTest {
+
+    private static final double DELTA = 0.001d;
+
+    @Test
+    public void testSingleDetermineStandardDeviationDoubles() throws Exception {
```
ge0ffrey: The rest of the code doesn't start test methods with the prefix "test" (as that's a JUnit 3 relic, now the @Test annotation makes it clear it's a test). This makes the test results in the intellij windows easier to read.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
ge0ffrey: PS: That lowercase L problem is a famous java puzzler (I forgot which year):
  https://www.parleys.com/tutorial/the-java-puzzlers
  https://www.parleys.com/tutorial/java-puzzlers-scraping-bottom-barrel

oskopek: Fixed & bookmarked.

optaplanner-benchmark/src/test/java/org/optaplanner/benchmark/impl/statistic/StatisticUtilsTest.java
oskopek: The test prefix was automatically generated by IntelliJ (even for JUnit 4) and I forgot to remove them. Fixed.

PR #159, opened 2015-09-30
PLANNER-431: Migrate upgrade recipes to AsciiDoc

oskopek:


optaplanner-core/src/main/java/org/optaplanner/core/api/package-info.java
```diff
@@ -18,7 +18,8 @@
  * The public API of OptaPlanner.
  * <p>
  * All classes in this namespace are backwards compatible in future releases (especially minor and hotfix releases).
- * If a major version number changes, a few specific classes might not be backwards compatible,
- * in which case the upgrade recipe will clearly document those cases.
+ * If a major version number changes, a few specific classes might not be backwards compatible, in which case
+ * <a href="http://www.optaplanner.org/download/upgradeRecipe/upgradeRecipeArchive.html">the upgrade recipe</a>
```
ge0ffrey: Maybe we should rename upgradeRecipeArchive.html to index.html (same for releaseNotesArchive.html), so then we can use a shorter URL here: href="http://www.optaplanner.org/download/upgradeRecipe/"
and it's likely that some users will craft that url themselfes to find a list of all upgrade recipe's (when looking at 1 upgrade recipe) - in which case they'd currently see "forbidden".

ge0ffrey:
The Upgrade recipe file should be removed too <rant>(important because never move anything without deleting the original - because then the rest of your time will still edit the original file and you'll have effectively forked it while bring a lot of pain when the mess is discovered years later. This can get so bad that often one is taken as canonical and the other's file changes are simply lost. Instead, delete the original and let git do it's job as a backup)</rant>

 Actually let's not remove it, but empty it out completely and just put a link to the new URL location. This way it's still there in the download zip assemblies, so long-time users can still find it.

optaplanner-core/src/main/java/org/optaplanner/core/api/package-info.java
oskopek: +1 will fix.

oskopek:
And I did just that already :) See e5f508d

PR #160, opened 2015-10-01
PLANNER-440: Unfinished

oskopek:
* PLANNER-440: ClassLoader refactoring in benchmark
  * *rebased, waiting to be finished*
  * TODOs (still left):
    * Why can't scoreDrl in classloader...Config.xml be prefixed with `diverted.../` prefix?
    * Move these tests to `-examples` too?
    * Add the Freemarker test


ge0ffrey:
Merged and fixed - will add freemarker test soon

PR #161, opened 2015-10-14
replaced deprecated ObjectUtils by java.util.Objects

Burhan1991:
minor changes, that will be need for all examples..


ge0ffrey:
Thanks for the PR, but unfortunately I can't merge it: currently we're still supporting Java 6 (libraries != end-user app) and Objects is since Java 7.

ge0ffrey:
ok to test

PR #164, opened 2015-10-21
Update exception messages

marksamman:
getDeclaringClass() already includes "class" in the name, so it currently becomes "class class ClassName".

PlanningEntityCollection should be PlanningEntityCollectionProperty


ge0ffrey:
Thanks for contributing :)

PR #167, opened 2015-12-14
Added GeoJSON support to VRP webexample

Darkade:
Added support for [GeoJSON](http://geojson.org/) alongside with plain JSON to webexamples.
Modified the leaflet.jsp module to render the map via GeoJSON.

GeoJSON is an open standard format designed for representing simple geographical features, along with their non-spatial attributes. This makes the VRP standards compliant and the JSON it returns more portable;  http://geojson.io and GitHub will render this format out of the box.

[Google maps](https://developers.google.com/maps/documentation/javascript/datalayer) also supports this format but the example wasn't modified at this stage.

Greetings

### DIFF
--- a/optaplanner-webexamples/src/main/java/org/optaplanner/webexamples/vehiclerouting/rest/service/DefaultVehicleRoutingRestService.java
+++ b/optaplanner-webexamples/src/main/java/org/optaplanner/webexamples/vehiclerouting/rest/service/DefaultVehicleRoutingRestService.java
@@ -41,6 +41,13 @@ import org.optaplanner.webexamples.vehiclerouting.rest.domain.JsonMessage;
 import org.optaplanner.webexamples.vehiclerouting.rest.domain.JsonVehicleRoute;
 import org.optaplanner.webexamples.vehiclerouting.rest.domain.JsonVehicleRoutingSolution;
 
+import ws.arete.arms.engine.domain.geojson.GeoJsonCustomer;
+import ws.arete.arms.engine.domain.geojson.GeoJsonCustomerProperties;
+import ws.arete.arms.engine.domain.geojson.GeoJsonGeometry;
+import ws.arete.arms.engine.domain.geojson.GeoJsonRoute;
+import ws.arete.arms.engine.domain.geojson.GeoJsonRouteProperties;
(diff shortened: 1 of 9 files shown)


Darkade:
I didn't changed the name of the package added. Not sure if I should, let me know.

ge0ffrey:
ok to test

ge0ffrey:
Very interesting stuff. Does geojson support adding images (such as a vehicle svg) on the map too?

Darkade:
Not in the geojson itself but every _feature_ has a _properties_ member. And there you can put any value you may use. Then, you can instruct leaflet to do a certain action whenever that property is encountered. So you could add something like `{"properties": {"vechile": "truck", "color":"#ffffff"}}` and have leaflet load the required file.
I'll try to give this a spin later today.

Darkade:
We've added a minor change that allows to make custom icons for locations. Every customization option available to leaflet.js seems to be compatible with the native geojson integration.

Greetings

ge0ffrey:
Great stuff, sorry I haven't had time to merge yet. Hopefully between christmas and new year.

Darkade:
That would be great. I'll check to make the appropriate changes to the documentation then.

ge0ffrey:
Sorry for the delay on this, I 'll review and potentially merge it after
the holidays.

2015-12-28 16:53 GMT+01:00 Ivan Reyes <notifications@github.com>:

> That would be great. I'll check to make the appropriate changes to the
> documentation then.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/droolsjbpm/optaplanner/pull/167#issuecomment-167592227>
> .
>



-- 
With kind regards / Met vriendelijke groeten,
Geoffrey De Smet

ge0ffrey:
@Darkade Sorry for the kie-ci spam and not being able address this PR in reasonable time.

I am still on the fence about this one though (undecided) - it's a big influence - breaking some rest approach that we don't want to delete, but OTH adding the geo standard api which we also want - so I don't just want to merge it nor just remove it. I intend to revisit this issue if and when optashift-vehicle-routing is build.

ge0ffrey:
Created a jira to track this:
  https://issues.jboss.org/browse/PLANNER-803

Darkade:
@ge0ffrey it's no problem. I do agree it's big. I'd be happy to see this implemented in some fashion in opta

PR #170, opened 2015-12-16
Added OK button to Constraint match dialog

oskopek:
... when constraint matches are disabled (GNOME3 doesn't render a close button). There's no way to close the dialog when that happens, except Alt+F4, which is a bad UX.

See: http://stackoverflow.com/questions/26162178/jdialog-not-showing-minimize-close-button

screenshot from 2015-12-16 11-46-49


ge0ffrey:
Thanks Ondrej! :)

PR #172, opened 2016-01-13
First step to run persistance tests with database

snurkabill:
not finished yet, we can track and discuss changes

### DIFF
--- a/optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
+++ b/optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
@@ -71,6 +71,7 @@
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-core</artifactId>
       <!-- Used at compile time by *ScoreHibernateType -->
+      <!--<scope>test</scope>-->
       <optional>true</optional>
     </dependency>
     <!-- Test -->
@@ -79,6 +80,8 @@
       <artifactId>hibernate-entitymanager</artifactId>
       <scope>test</scope>
     </dependency>
+
+
     <dependency>
       <groupId>com.h2database</groupId>
       <artifactId>h2</artifactId>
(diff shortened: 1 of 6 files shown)


optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
```diff
@@ -84,11 +84,85 @@
...
+      <dependencies>
+        <!--  Add the downloaded jdbc driver jar for the database to the classpath.
+              !: <groupId>, <artifactId> and <version> are just placeholders and DO NOT MATTER! -->
+        <dependency>
+          <groupId>org.jbpm</groupId>
+          <artifactId>downloaded-jdbc-driver-jar</artifactId>
+          <scope>system</scope>
+          <systemPath>${maven.jdbc.driver.jar}</systemPath>
+        </dependency>
+      </dependencies>
```
psiroky: Just a thought: Surefire can be directly configured to use additional classpath elements. See http://maven.apache.org/surefire/maven-surefire-plugin/examples/configuring-classpath.html. This may be a better alternative to `<scope>system</scope>` dependency (which will likely be removed in Maven 4).

ge0ffrey: @psiroky This uses the same trick as the drools-persistence-jpa tests  and it's a bit special: the artifactId downloaded-jdbc-driver-jar doesn't actually exist: it gets replaced on the fly base don the jenkins job configuration's database type it's running against. There are many database types (oracle, mysql, ...).

Would switching to that surefire approach still allow that? If so, we need to do that both here and on the drools-persistence-jpa pom.

psiroky: Yes, I believe that is exactly the use case for this additional classpath config. I go agree though that we should be consistent with other -jpa modules, so I guess we can stick the current solution and later on possibly fix this at all places (assuming it will be worth the trouble).

PR #174, opened 2016-01-21
Fixed getStandardDeviationString for localized systems

oskopek:
`DecimalFormat` replaces the dot with a comma

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
@@ -17,6 +17,7 @@
 package org.optaplanner.benchmark.impl.statistic;
 
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.util.List;
 
 import org.optaplanner.benchmark.impl.result.BenchmarkResult;
@@ -68,8 +69,12 @@ public class StatisticUtils {
         StringBuilder standardDeviationString = new StringBuilder(standardDeviationDoubles.length * 9);
         // Abbreviate to 2 decimals
         // We don't use a local sensitive DecimalFormat, because other Scores don't use it either (see PLANNER-169)
+        DecimalFormatSymbols decimalFormatSymbols = DecimalFormatSymbols.getInstance();
+        decimalFormatSymbols.setDecimalSeparator('.');
         DecimalFormat exponentialFormat = new DecimalFormat("0.0#E0");
+        exponentialFormat.setDecimalFormatSymbols(decimalFormatSymbols);
         DecimalFormat decimalFormat = new DecimalFormat("0.0#");
+        decimalFormat.setDecimalFormatSymbols(decimalFormatSymbols);
         boolean first = true;
         for (double standardDeviationDouble : standardDeviationDoubles) {
             if (first) {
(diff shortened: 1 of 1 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticUtils.java
```diff
@@ -68,8 +69,12 @@ public static String getStandardDeviationString(double[] standardDeviationDouble
         StringBuilder standardDeviationString = new StringBuilder(standardDeviationDoubles.length * 9);
         // Abbreviate to 2 decimals
         // We don't use a local sensitive DecimalFormat, because other Scores don't use it either (see PLANNER-169)
+        DecimalFormatSymbols decimalFormatSymbols = DecimalFormatSymbols.getInstance();
```
ge0ffrey: Changing the DecimalFormatSymbols.getInstance() of the entire JVM has lasting side effects.

1) Presume we have a benchmark server that generates the reports on the fly (based on previous benchmark runs - much like the aggregator). In parallel, it can get asked to generated 2 or more requests, with a different locale (the local is determined by the browser of the client, and is part of the HTTP request - the server's default local isn't relevant in this case). If that happens, there could be a race condition that the 1th request gets its number formatted in the locale of the 2th request.

2) Presume that the JVM app that embeds our benchmarker runs something else after it that uses DecimalFormatSymbols and doesn't explicitly set the default locale's setting for it (because it's the default even though we changed it).

ge0ffrey: Would something like this work?

DecimalFormatSymbols symbols = new DecimalFormatSymbols(EN_US);
DecimalFormat exponentialFormat = new DecimalFormat("0.0#E0", symbols);

oskopek: Definitely, thanks for catching that.

I'd prefer the following, as it doesn't depend on any *specific* locale being present (uses the default one):
```
DecimalFormatSymbols decimalFormatSymbols = new DecimalFormatSymbols();
decimalFormatSymbols.setDecimalSeparator('.');
DecimalFormat exponentialFormat = new DecimalFormat("0.0#E0", decimalFormatSymbols);
DecimalFormat decimalFormat = new DecimalFormat("0.0#", decimalFormatSymbols);
```

WDYT?

ge0ffrey: As long as PLANNER-169 isn't implemented, we want those numbers to be formatted locale-indepent, like java literals (such a "double d = 1.5;"). By default, the Score.toString() also follow the java literal format (similar to Integer.toString() and BigDecimal.toString()).

Java literals use en_US. Notice how we don't write "double d = 1,5;". So I think it needs to be:

    new DecimalFormatSymbols(Locale.US)

oskopek: +1, will fix ASAP. Also, according to the docs, Locale.US will always be present (or a compatible substitute)

ge0ffrey: thanks :)

oskopek:
Please, do not merge yet.

ge0ffrey:
Thanks Ondrej!

PR #175, opened 2016-01-21
PLANNER-502 + refactored NQueensXStreamXmlPlannerBenchmarkFactoryTest

oskopek:


ge0ffrey:
Thanks, merged, that FileConverter is indeed a good idea!

PR #179, opened 2016-02-29
Small optimization in the SolutionDescriptor

triceo:
While doing research for PLANNER-356, i found out this one TODO that was easy enough to solve.

This PR removes fixes the TODO and also moves the list sorting out of the loop, since the list is never modified other than by removing the first item.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -296,21 +296,6 @@ public class SolutionDescriptor {
 
     private void determineGlobalShadowOrder() {
         // Topological sorting with Kahn's algorithm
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -331,8 +316,8 @@ public int compare(Pair<ShadowVariableDescriptor, Integer> a, Pair<ShadowVariabl
             }
         }
         int globalShadowOrder = 0;
+        Collections.sort(pairList, (a, b) -> Integer.compare(a.getValue(), b.getValue()));
         while (!pairList.isEmpty()) {
-            Collections.sort(pairList, comparator);
```
ge0ffrey: The sort needs to happen in the loop for Kahn's algorithm to work :)

ge0ffrey: That's actually the reason a while loop, instead of a normal for loop.

triceo:
I didn't even run the tests, that's how overconfident I was. Well, this will teach me.

PR #180, opened 2016-02-29
PLANNER-356: No more getProblemFacts()

triceo:
This is a proof of concept, DO NOT MERGE YET.

What I did:

- The method ```getProblemFacts()``` is gone. It is ignored if present on a ```Solution```.
- Instead, we have two new annotations in the public API, ```PlanningFactProperty``` and ```PlanningFactCollectionProperty```.
- These annotations have the exact same behavior as the equivalent entity property annotations. This is because they share most of the code...
- ... and in order to achieve that, I have done a little refactoring of ```SolutionDescriptor```. It now uses some Java 8 in places, specifically streams and lambdas, and reuses as many code paths as possible.
- The refactor did not touch places it didn't need to. As a result, the class is now mixing some imperative-style methods with some more functional-style ones.

What I need to do:

- Write some tests for the new behavior.
- Fix anything that comes out of code review.
- Add a note to migration guide.
- Squash all of this into one commit, since my "iterative" development doesn't look good here. :-)

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactCollectionProperty.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactCollectionProperty.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2011 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.api.domain.solution;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import java.util.Collection;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Specifies that a property (or a field) on a {@link Solution} class is a {@link Collection} of planning facts.
+ */
+@Target({METHOD, FIELD})
+@Retention(RUNTIME)
+public @interface PlanningFactCollectionProperty {
+
+}
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactCollectionProperty.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactCollectionProperty.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2011 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactProperty.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactProperty.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2011 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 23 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactCollectionProperty.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2011 Red Hat, Inc. and/or its affiliates.
```
triceo: 2016

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactProperty.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2011 Red Hat, Inc. and/or its affiliates.
```
triceo: Ditto.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/CheapTimeSolution.java
```diff
@@ -31,6 +28,9 @@
 import org.optaplanner.examples.common.domain.AbstractPersistable;
 import org.optaplanner.persistence.xstream.impl.score.XStreamScoreConverter;
 
+import java.util.List;
+
+// FIXME originally had itself as a fact
```
triceo: We need to fix this somehow.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -617,33 +639,38 @@ public int countReinitializableVariables(ScoreDirector scoreDirector, Solution s
...
-            Collection<Object> entityCollection = extractEntityCollection(entityCollectionMemberAccessor, solution);
+            Collection<Object> entityCollection = extractCollection(entityCollectionMemberAccessor, solution);
             iteratorList.add(entityCollection.iterator());
         }
         return Iterators.concat(iteratorList.iterator());
     }
 
-    private Object extractEntity(MemberAccessor entityMemberAccessor, Solution solution) {
-        return entityMemberAccessor.executeGetter(solution);
+    private Object extract(MemberAccessor memberAccessor, Solution solution) {
```
triceo: This method should probably move to ```MemberAccessor``` itself. It has no business being on ```SolutionDescriptor``` and may be made static.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -617,33 +639,38 @@ public int countReinitializableVariables(ScoreDirector scoreDirector, Solution s
...
+        return memberAccessor.executeGetter(solution);
     }
 
-    private Collection<Object> extractEntityCollection(
-            MemberAccessor entityCollectionMemberAccessor, Solution solution) {
-        Collection<Object> entityCollection = (Collection<Object>) entityCollectionMemberAccessor.executeGetter(solution);
+    private Collection<Object> extractCollection(MemberAccessor collectionMemberAccessor, Solution solution,
+                                                 boolean isFact) {
```
triceo: Likewise.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -63,6 +42,12 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.*;
+
 public class SolutionDescriptor {
```
triceo: A whole lot of methods on this class could be made static.

ge0ffrey:
Discussion summary:

- We need this. We need to get rid of the Solution interface (in green field projects). The pro's outweigh the cons.
- All corner cases (such as a solution that is a problem fact, people that want to cache the problem facts list for whatever reason, calculated at runtime problem facts) are handled by adding a @ProblemFacts(Collection)Property on a method.
- AbstractSolution can use that trick too.
- The solution will have a @PlanningScore annotation on for example the method getScore(). This means every solution will still have a score. There is debate on whether this is desirable - but that's my decision for now - we can reopen the debate after we got rid of the solution interface if desired.
- The debate of the score definition duplication is separate - I do agree we need to get rid of it in the xml - but that debate will start once we got rid of the solution interface.
- For backwards compatibility, we want to keep the Solution interface in the next few weeks - if Drools ends up doing big backwards compatibility DRL changes in 7.0, then we 'll remove that Solution interface from the public api. If they don't, then we'll debate at that time (but I'd currently favor to keeping it.)

ge0ffrey:
Lukas, if you rebase to master: there's a new example now taskassigning. I do think that if you just adjust TaskAssigningSolution too, like you've adjusted the rest of the examples, you 'll be fine.
(presuming there is no merge conflict during the rebase - might be worth branch your branch first as a backup ;)

triceo:
I rebased the branch to get those changes.

triceo:
Closing this as it's been replaced by #182.

PR #181, opened 2016-03-07
6.3.0.final ip patch

ohlle:
This is a workaround for a NPE that I received. The NPE seems to occur when the planner runs out of valid moves before the bailout occurs. I managed to reproduce it when planning on a really small dataset with a Selector that has very hard restrictions. I'm guessing the error is really in LocalSearchDecider.decideNextStep since the erroneous SwapMove was created there. Here is the full stacktrace. This fix probably only cures the sympton but doesn't fix the cause.

```
java.lang.NullPointerException
        at sun.reflect.GeneratedMethodAccessor107.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at org.optaplanner.core.impl.domain.common.accessor.BeanPropertyMemberAccessor.executeGetter(BeanPropertyMemberAccessor.java:67)
        at org.optaplanner.core.impl.domain.variable.descriptor.VariableDescriptor.getValue(VariableDescriptor.java:89)
        at org.optaplanner.core.impl.heuristic.selector.move.generic.SwapMove.isMoveDoable(SwapMove.java:63)
        at org.optaplanner.core.impl.localsearch.decider.LocalSearchDecider.decideNextStep(LocalSearchDecider.java:118)
        at org.optaplanner.core.impl.localsearch.DefaultLocalSearchPhase.solve(DefaultLocalSearchPhase.java:72)
        at org.optaplanner.core.impl.solver.DefaultSolver.runPhases(DefaultSolver.java:214)
        at org.optaplanner.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:176)
```


ge0ffrey:
The PR's mentions over 150 changed files and a few dozen commits. I presume the only commit that matters to be merged is this one?
  https://github.com/IntraPhone/optaplanner/commit/51f98d7c069e8a522ff81baa6c5596ed58a3a1d6

ohlle:
@ge0ffrey yes, I think merged to/from wrong branch. It is only that commit, should I create an other PR?

ge0ffrey:
I'll merge your commit directly from your fork, no problem (I might add another commit on top of it with some peer review changes). I 'll close this PR once that's done (probably not today).

ge0ffrey:
Thanks for submitting PR btw :)

snurkabill:
Is PillarSwapMove also affected?

ohlle:
No problem, glad I can help out a bit with this awesome framework.

snurkabill:
Just reminder: check also pillarSwapMove

ge0ffrey:
Is there a jira or a forum or dev list thread related to this PR? I recall something like that and I want to understand the use case better.
Reason: if the selector is wrong to select null moves than the move shouldn't have to work around it.

ohlle:
I never created a Jira issue or something like that. I asked for help regarding this on IRC a couple of weeks ago. What information do you want?

ge0ffrey:
What's the use case?

(Is your PlanningVariable nullable=true? If not, then the Local Search can only start from an initialized solution, which would guarantee that `new TestdataEntityProvidingEntity("a", Arrays.asList(v1, v2, v3), null)` in the test is impossible. Not that I think that this matters.)

Do you use a custom MoveFactory? If not and you've just a swapMoveSelector, then it's a bug that it selected a rightEntity that is null, so `new SwapMove(..., a, null)` is impossible. I'd like to see the solver config in that case. If you do have a custom MoveFactory then I think that's to blame, but SwapMove can be improved to fail fast during construction with a nice error message.

ohlle:
It is an extension of the vehicle routing problem with more restrictions on who can go where and how long time they stay etc.

My planning variables aren't nullable.
I'm not using a custom move factory but I do use a movableEntitySelectionFilter on one of my PlanningEntities which have quite hard restrictions because of built in hard restrictions. When I remove that one I don't get the NPE any more, and also if I change it to reject everything it doesn't produce the NPE either.

Here is my complete solverconfig.xml
```
<solver>
  <solutionClass>net.intraphone.route.domain.RouteSolution</solutionClass>
  <entityClass>net.intraphone.route.domain.Standstill</entityClass>
  <entityClass>net.intraphone.route.domain.Task</entityClass>

  <randomType>MERSENNE_TWISTER</randomType>

  <scoreDirectorFactory>
    <scoreDefinitionType>HARD_MEDIUM_SOFT_LONG</scoreDefinitionType>

    <easyScoreCalculatorClass>net.intraphone.route.RoutingScoreCalculator</easyScoreCalculatorClass>
  </scoreDirectorFactory>

  <constructionHeuristic>
    <constructionHeuristicType>FIRST_FIT_DECREASING</constructionHeuristicType>
  </constructionHeuristic>

  <localSearch>
    <unionMoveSelector>
      <changeMoveSelector/>
      <swapMoveSelector/>
      <subChainChangeMoveSelector>
        <selectReversingMoveToo>true</selectReversingMoveToo>
      </subChainChangeMoveSelector>
      <subChainSwapMoveSelector>
        <selectReversingMoveToo>true</selectReversingMoveToo>
      </subChainSwapMoveSelector>
    </unionMoveSelector>
    <acceptor>
      <lateAcceptanceSize>200</lateAcceptanceSize>
    </acceptor>
    <forager>
      <acceptedCountLimit>1</acceptedCountLimit>
    </forager>
    <termination>
      <secondsSpentLimit>60</secondsSpentLimit>
      <unimprovedSecondsSpentLimit>10</unimprovedSecondsSpentLimit>
    </termination>
  </localSearch>
</solver>
```

ge0ffrey:
Intresting. This looks like a bug in the SwapMoveSelector. Thanks for reporting.
Fixing it in the SwapMove would be fixing the symptom, not the cause, so let's not do that. I 've got your commit locally but I 'll stash that because it fixes the symptom.

I've created a jira to track this bug:
  https://issues.jboss.org/browse/PLANNER-550
I'd like to able to reproduce this locally, so I can diagnose it. Would it be possible to attach a reproducer to that jira issue? 

PS: Are you using a normal solution ValueRange or an entity depended value range?

ohlle:
I will try to make a reproducer. It is quite coupled with our database right now but I think it is possible.

I'm not really sure if I understand your question. I'm using this in the PlanningEntity
```
    @PlanningVariable(valueRangeProviderRefs = {"staffRange", "taskRange"}, graphType = PlanningVariableGraphType.CHAINED)
    public Standstill getPreviousStandstill() {
        return previousStandstill;
    }

```
And this in the Solution class
```
    @PlanningEntityCollectionProperty
    @ValueRangeProvider(id = "staffRange")
    public List<Staff> getStaffList() {
        return staffList;
    }

    @PlanningEntityCollectionProperty
    @ValueRangeProvider(id = "taskRange")
    public List<Task> getTaskList() {
        return taskList;
    }
```

ge0ffrey:
That looks good. How does your movableEntityFilter look like?

ohlle:
```
@PlanningEntity(difficultyWeightFactoryClass = TaskDifficultyWeightFactory.class, movableEntitySelectionFilter = TaskSelectionFilter.class)
public class Task extends AbstractPersistable implements Standstill {
```

The TaskSelectionFilter is only task.accept(). In Task.accept() it checks required staff and also if the task has any required attributes that the staff isn't fulfilling. I'm working on fixing the reproducer but it was a lot harder to decouple the code from our codebase than I expected.

Some more information that can be worth mentioning. When I first ran into this problem I was working on automated testing of our code and I had a very minimal data set just to verify our rules. I then moved some of the constraints from the accept method to the score calculator and added a hard score. That worked out in the test but on the real data the performance hit was far to great and it took way to long time to optimize the solution. I'm using the EasyScoreCalculator if that makes any difference.

ohlle:
I've finally been able to make a small reproducer that I've attached to the jira issue. I've added some more explanation as a comment there as well.

snurkabill:
Planner finds possible issue when FULL_ASSERT environment mode is enabled, see JIRA

ge0ffrey:
Fixed through https://github.com/droolsjbpm/optaplanner/commit/24778c4f33eb329537931681bda823bb13f39df1 (see also the jira)

PR #182, opened 2016-03-12
PLANNER-436 + PLANNER-356: Solution interface is no more

triceo:
This pull request does a great many things, since unfortunately all of them need to be done at once.

- The interface **```Solution``` has been ```@Deprecated``` and all of its uses from everywhere in Planner have been completely removed**. (Except on ```Solver``` interface.) ```Solution``` is now only useful for  backwards compatibility with 6.x, but it is an unnecessary code complexity increase and I suggest it **be killed with fire**.
- **Certain classes needed to be generified** in order to remove ```Solution``` references. Planner core is now more type-safe than before, and it's by the actual type of the user's solution, not the ```Solution``` interface. This makes for a cleaner code. (I only added the generic type where it was absolutely necessary to directly replace ```Solution``` - if we wanted to make Planner completely generified, there is a lot more places where this would need to happen.)
- We have **two new annotations** in the public API, ```PlanningFactProperty``` and ```PlanningFactCollectionProperty```. These annotations have the exact same behavior as the equivalent entity property annotations. This is because they share most of the code.
- We have **one more new annotation** in the public API, ```PlanningScore```.
- ```SolutionDescriptor``` **now scans all solution superclasses** for fields and methods. This was necessary, otherwise ```AbstractSolution``` would no longer work. (Solutions need to extend that, and it contains annotations which otherwise wouldn't have been read.)
- As a good citizen, I also refactored places which I had to touch in order to leave the code better off than when I started. This includes some Java 7+8 stuff, such as collection streams, try-with-resources, diamond operators, ...

What I need to do:

- Write some tests for the new stuff.
- Fix anything that comes out of code review.
- Describe all of this in migration guide.
- ```optaplanner-wb``` may need a commit or two. I'll worry about that after this is code-reviewed.
- Squash all of this into one commit, since my "iterative" development doesn't look good here. :-)


triceo:
@ge0ffrey I would appreciate expediting the code review - it is a lot of code and I would hate if it got out of sync with master. 

I'm sure you'll have comments and I promise to fix them, but maybe we could do that in subsequent PR and not block this one.

snurkabill:
+1 for merge

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/PlannerBenchmarkRunner.java
```diff
@@ -273,7 +258,7 @@ protected void runSingleBenchmarks() {
             }
         }
         // Wait for the benchmarks to complete
-        for (Map.Entry<SubSingleBenchmarkRunner, Future<SubSingleBenchmarkRunner>> futureEntry : futureMap.entrySet()) {
+        futureMap.entrySet().forEach(futureEntry -> {
```
ge0ffrey: These kind of changes are bloating the PR a bit for me, making it hard to grok.
Although I agree we should start using the diamond operator, etc - and through our entire codebase - it rather deal with that in isolation (and preferably after 6.4.0.Final is out to avoid backporting merge conflicts).

Specifically, replacing a for(each) with a .forEach() is a debatable change as it's probably slower (use of streams and lamba's without need for parallel()).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -74,6 +50,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.imageio.ImageIO;
```
ge0ffrey: (Suggestion, not important)
About import ordering: I believe the default import ordering in our default code style is different. Our default code style file is here:
  https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml

Nevertheless, this is fine to merge, as sooner or later we do an "optimize imports" and it goes back to what it was :)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
```diff
@@ -108,9 +99,6 @@ protected PlannerBenchmarkResult readPlannerBenchmarkResult(File plannerBenchmar
             String benchmarkReportDirectoryName = plannerBenchmarkResultFile.getParentFile().getName();
             plannerBenchmarkResult = PlannerBenchmarkResult.createUnmarshallingFailedResult(
                     benchmarkReportDirectoryName);
-        } catch (XStreamException e) {
```
ge0ffrey: Important.
Deleting this runtime exception XStreamException will remove the actual filename from the stacktrace message when a user has a bug in his solver configuration. That will leave the user clueless (= not good).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSubSingleStatistic.java
```diff
@@ -39,7 +39,8 @@
         PickedMoveTypeBestScoreDiffSubSingleStatistic.class,
         PickedMoveTypeStepScoreDiffSubSingleStatistic.class
 })
-public abstract class PureSubSingleStatistic<P extends StatisticPoint> extends SubSingleStatistic<P> {
+public abstract class PureSubSingleStatistic<Solution_, P extends StatisticPoint>
```
ge0ffrey: Can we rename P to StatisticPoint_ ?
The idea is to get generic types that are self-explanatory by using the superclassname + "_"

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactProperty.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Specifies that a property (or a field) on a {@link Solution} class is a planning fact.
```
ge0ffrey: Does {@link solution} still work?

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningScore.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Specifies that a property (or a field) on a {@link Solution} class is a planning score.
```
ge0ffrey: suggestion: `{@link Score}`

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningFactProperty.java
triceo: It does, since I brought ```Solution``` back in the final commit. But I removed all references from javadoc, in order to make ```rm Solution.java``` very easy. :-)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/PureSubSingleStatistic.java
triceo: Will do. Although I'm sure you see that it's been like this before, I just didn't change it. :-)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
triceo: I'll make sure I have my IDE configured properly.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/AbstractSolution.java
```diff
@@ -16,53 +16,52 @@
...
+    protected Collection<?> getProblemFacts() {
         Class<? extends AbstractSolution> instanceClass = getClass();
-        addProblemFactsFromClass(factList, instanceClass);
-        return factList;
+        return getProblemFactsFromClass(instanceClass);
     }
 
-    private void addProblemFactsFromClass(List<Object> factList, Class<?> instanceClass) {
+    @PlanningFactCollectionProperty
+    private Collection<Object> getProblemFactsFromClass(Class<?> instanceClass) {
```
ge0ffrey: Doesn't the @PlanningFactCollectionProperty need to be on getProblemFacts()?

If the  @PlanningFactCollectionProperty is on a method that contains 1 or more arguments, OptaPlanner must fail-fast during buildSolver() with a clear error message. We have a test for AbstractSolution, so I presume that fail-fast might be missing. 

Some btw if @PlanningFactCollectionProperty is on a method that doesn't return a collection: that should also fail fast.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/ClassBrowser.java
```diff
@@ -0,0 +1,55 @@
+package org.optaplanner.core.impl.domain.solution.descriptor;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.*;
+
+class ClassBrowser {
```
ge0ffrey: All classes are public, to encourage reuse of classes with optaplanner itself, across packages.

Note: not all classes are considered "module public" - but until JDK 9 we can't enforce that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/ClassBrowser.java
```diff
@@ -0,0 +1,55 @@
...
+        while (superClass != Object.class) {
+            classes.add(superClass);
+            superClass = superClass.getSuperclass();
+        }
+        return classes;
+    }
+
+    /**
+     *
+     * @param root The class in which to start the search.
```
ge0ffrey: suggestion: add either "never null" or "sometimes null" for params and return javadocs.

For example: `@param root Never null, the class in which to start the search.`

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/easy/EasyScoreDirector.java
```diff
@@ -16,25 +16,26 @@
...
+public class EasyScoreDirector<Solution_>
+        extends AbstractScoreDirector<Solution_, EasyScoreDirectorFactory<Solution_>> {
 
-    private final EasyScoreCalculator easyScoreCalculator;
+    private final EasyScoreCalculator<Solution_> easyScoreCalculator;
 
-    public EasyScoreDirector(EasyScoreDirectorFactory scoreDirectorFactory, boolean constraintMatchEnabledPreference,
-            EasyScoreCalculator easyScoreCalculator) {
+    public EasyScoreDirector(EasyScoreDirectorFactory<Solution_> scoreDirectorFactory,
+                             boolean constraintMatchEnabledPreference,
```
ge0ffrey: Code style: Our code style uses 2 idents for the next line, instead of aligning parameter on the `(` because otherwise long method names create off-page content.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/AbstractSolution.java
triceo: This is interesting. You are right that it should be on a method without arguments.

However, the code checking this is *the exact same code* that does that for ```@PlanningEntityCollectionProperty``` - literally, it's the same method. So, if there is a bug, it is there as well.

This should answer the point about not returning a collection - if the other annotations do it, this one will too. It's the same code.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/ClassBrowser.java
triceo: Ok, will change.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/ClassBrowser.java
triceo: Will do. That said, we should be using ```Optional``` for this.

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/PlanningScore.java
triceo: Will do.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/recaller/BestSolutionRecallerTest.java
```diff
@@ -87,17 +88,22 @@ public void unimprovedInitializedProcessWorkingSolutionDuringStep() {
 
     @Test
     public void improvedUninitializedProcessWorkingSolutionDurintStep() {
+        DefaultSolverScope solverScope = new DefaultSolverScope();
+        InnerScoreDirector scoreDirector = mock(InnerScoreDirector.class);
+        SolutionDescriptor solutionDescriptor = mock(SolutionDescriptor.class);
+        when(scoreDirector.getSolutionDescriptor()).thenReturn(solutionDescriptor);
+        solverScope.setScoreDirector(scoreDirector);
```
ge0ffrey: This boilerplate code should be extracted into or simply using
  org.optaplanner.core.impl.testdata.util.PlannerTestUtils
For example, that already has PlannerTestUtils.mockScoreDirector()

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
triceo: IDE was reporting this as having no effect, since ```XStreamException extends IOException```. That said, if you want it there, please put it back when you're merging.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/reflect/field/TestdataFieldAnnotatedSolution.java
```diff
@@ -75,6 +74,7 @@ public void setScore(SimpleScore score) {
     // Complex methods
     // ************************************************************************
 
+    @PlanningFactCollectionProperty
```
ge0ffrey: For TestdataFieldAnnotatedSolution, I 'd annotated the @PlanningFactCollectionProperty on the fields.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/CheapTimeSolution.java
```diff
@@ -16,36 +16,39 @@
...
 @XStreamAlias("CtCheapTimeSolution")
-public class CheapTimeSolution extends AbstractPersistable implements Solution<HardMediumSoftLongScore> {
+public class CheapTimeSolution extends AbstractPersistable {
 
     private int timeResolutionInMinutes;
     private int globalPeriodRangeFrom; // Inclusive
     private int globalPeriodRangeTo; // Exclusive
 
+    @PlanningFactCollectionProperty
     private List<Resource> resourceList;
```
ge0ffrey: Consistency: Each example should either use ALL **field annotations** or ALL **method annotations **(usually getter annotations).
Most examples currently use method annotations, including the CheapTimeSolution. This is because JPA-hibernate also recommends method annotations over field annotations (but I recently started prefering reading field annotations). In any case, we should mix in 1 example.

The taskassigning examples uses field annotations for @PlanningVariable etc, so it should use field annotations for @PlanningFactCollectionProperty too. But for all other examples, its' the other way around (unless we decide to switch a number of - not all of them - to field annotations, but I 'd rather decide that in isolation from this issue again).

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/easy/EasyScoreDirector.java
triceo: If you want me to go over all the files to fix that, I will.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/recaller/BestSolutionRecallerTest.java
triceo: Will do.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/CheapTimeSolution.java
triceo: Ok, I'll review the examples and make sure we are consistent (all-field or all-method) in each of them.

The reason I did it like this was to have a "test case" ensuring that both approach work in real life, not just in mocked tests.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
```diff
@@ -217,24 +213,21 @@ public boolean isSolving() {
     }
 
     public void registerForBestSolutionChanges(final SolverAndPersistenceFrame solverAndPersistenceFrame) {
-        solver.addEventListener(new SolverEventListener<Solution_>() {
-            // Not called on the event thread
```
ge0ffrey: This comment "// Not called on the event thread" is important info

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/reflect/field/TestdataFieldAnnotatedSolution.java
triceo: We've just agreed on IRC that we're not changing this and sticking to method-based annotations, except for one example to demonstrate the capability.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
ge0ffrey: Motivation: race conditions are tricky, so in concurrent code, it's important that it's clear which thread is handling what.

triceo: Will put it back.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
ge0ffrey: I am not seeing that with my version of xstream:

    XStreamException extends BaseException
    BaseException extends RuntimeException

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/easy/EasyScoreDirector.java
ge0ffrey: It would save me the time :) Otherwise I will - I 'd rather not mix code styles as _once the gate is open, it's hard to close it again_.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/reflect/field/TestdataFieldAnnotatedSolution.java
ge0ffrey: Yes, but this class is called Testdata **Field Annotated** Solution :)

triceo: Crap, of course, you're right. :-)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
triceo: Interesting. I was assuming it does, because why else would IDEA warn me about it? Anyway, I'll put it back.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/bestscore/BestScoreProblemStatistic.java
```diff
@@ -73,14 +73,16 @@ public SubSingleStatistic createSubSingleStatistic(SubSingleBenchmarkResult subS
     public void writeGraphFiles(BenchmarkReport benchmarkReport) {
         List<XYPlot> plotList = new ArrayList<XYPlot>(BenchmarkReport.CHARTED_SCORE_LEVEL_SIZE);
         int seriesIndex = 0;
-        for (SingleBenchmarkResult singleBenchmarkResult : problemBenchmarkResult.getSingleBenchmarkResultList()) {
+        List<SingleBenchmarkResult> results = problemBenchmarkResult.getSingleBenchmarkResultList();
+        for (SingleBenchmarkResult singleBenchmarkResult : results) {
```
oskopek: Is there any advantage to writing for-each loops like this, instead of the original version?

Also, we usually name lists ...List, f.e. "pointList" (an array would be "points") IIRC.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
triceo: Done.

optaplanner-docs/src/main/docbook/en-US/images/Chapter-Planner_introduction/distributionZip.svg
```diff
@@ -7671,7 +7671,7 @@ RA4qFfXPnY0j10ZplDbpUWtjFU3TFEWR85SWlgKAKIoAYDKZym+VryY+za7Z111nBsnbNxYUAFIJ
 YoHsFI2t2tlAAVBaGiAVuzwsgNna3jB/Lrtmp3NJ20sIj/Vc4qCy5uDBdQ13XCfQ2qwHD9aZQfCc
 yX0upGnbwc6Voktad6/RAefxJAUAEBAAAE0BAOTPcLXc39pyERLuQQGMVbvbdPM1qZr2rO8+8PRu
 6wPZjlyU8UdGcktZC0VSyjYChBDHcTqdTq1WK7sQlIUNSZLf+6emyx+Gn3/GqrgFBEFQFFVSUqJW
-qxOJhCiKFEXJsizwwpbGLb+P/J5il9qLYGhr7LZn9+/oLHaNDiD6/kenB5RZDe85c817vMF1eEvl
+qxOJhCiKFEXJsizwwpbGLb+StatisticPoint_/J5il9qLYGhr7LZn9+/oLHaNDiD6/kenB5RZDe85c817vMF1eEvl
```
oskopek: This probably wasn't intentional.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/bestscore/BestScoreProblemStatistic.java
triceo: The only advantage is that if I keep it like the original version had it, it won't compile. :-) It has something to do with the new generics I introduced, but I was busy finishing this work and didn't really look into it.

optaplanner-docs/src/main/docbook/en-US/images/Chapter-Planner_introduction/distributionZip.svg
triceo: Nice catch!

PR #184, opened 2016-03-18
PLANNER-436, PLANNER-356: Solution interface no longer necessary

triceo:
This replaces pull request #182 and is ready to merge, following up on the comments on that PR. Final summary of what I did in this PR:

- The interface **```Solution``` has been ```@Deprecated``` and all of its uses from everywhere in Planner have been completely removed**. (Except on ```Solver``` interface.) ```Solution``` is now only useful for  backwards compatibility with 6.x, but it is an unnecessary code complexity increase and I suggest it **be killed with fire**.
- **Certain classes needed to be generified** in order to remove ```Solution``` references. Planner core is now more type-safe than before, and it's by the actual type of the user's solution, not the ```Solution``` interface. This makes for a cleaner code. (I only added the generic type where it was absolutely necessary to directly replace ```Solution``` - if we wanted to make Planner completely generified, there is a lot more places where this would need to happen.)
- We have **two new annotations** in the public API, ```PlanningFactProperty``` and ```PlanningFactCollectionProperty```. These annotations have the exact same behavior as the equivalent entity property annotations. This is because they share most of the code.
- We have **one more new annotation** in the public API, ```PlanningScore```.
- ```SolutionDescriptor``` **now scans all solution superclasses** for fields and methods. This was necessary, otherwise ```AbstractSolution``` would no longer work. (Solutions need to extend that, and it contains annotations which otherwise wouldn't have been read.)
- As a good citizen, I also refactored places which I had to touch in order to leave the code better off than when I started. This includes some Java 7+8 stuff, such as collection streams, try-with-resources, diamond operators, ...

What @ge0ffrey said he would do:

- Get rid of the forEach() constructs, since they're not necessary for this.
- Maybe fix some code formatting in places. (My IDEA just doesn't get it right.)
- Describe all of this in migration guide.

It is possible that, following up on this, optaplanner-wb will need a commit or two. I can work on that immediately after this is merged.


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/ClassBrowser.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param root Never null, the class in which to start the search.
+     * @return Never null; all fields in the class, and all non-private non-static fields in super-types.
+     */
+    public static Set<Field> getAllVisibleFields(Class<?> root) {
+        Set<Class<?>> classes = getAllClassesInHierarchy(root);
+        Set<Field> visibleFields = new HashSet<>();
+        classes.forEach(c -> Arrays.stream(c.getDeclaredFields()).forEach(f -> {
+            int mods = f.getModifiers();
+            if (!Modifier.isStatic(mods) && (c == root || !Modifier.isPrivate(mods))) {
```
triceo: @ge0ffrey You were mentioning that SolutionDescriptor should fail fast if it finds an annotated fact/entity getter with non-zero number of attributes. And you noticed that AbstractSolution does just that, yet Planner doesn't fail. I find out it is because of this code.

This code makes sure that, in parent classes of a class (solution, in this case), it never returns a field or method which is not visible. (```private``` field/method from a super-type will not be visible.) This was exactly the case with ```AbstractSolution``` - the code always worked, and it worked as it was designed.

Just to be sure, I've created and pushed ```SolutionDescriptorTest``` verifying that this situation works out as it should.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -464,24 +600,17 @@ public VariableDescriptor findVariableDescriptorOrFail(Object entity, String var
...
-            throw new IllegalStateException("The solution (" + solution
-                    + ")'s method getProblemFacts() should never return null.");
-        }
-        facts.addAll(problemFacts);
-        for (MemberAccessor entityMemberAccessor : entityPropertyAccessorMap.values()) {
-            Object entity = extractEntity(entityMemberAccessor, solution);
-            if (entity != null) {
-                facts.add(entity);
+    public Collection<Object> getAllFacts(Solution_ solution) {
+        Collection<Object> facts = new ArrayList<>();
```
triceo: @ge0ffrey Perhaps we should make this a ```Set```? Why repeat facts?

ge0ffrey:
Merged - I added a bunch of stuff (not all related to this PR but also a bunch of unit tests, a few fixes, docs and a bunch of consistency with rest of code).
  http://github.com/droolsjbpm/optaplanner/compare/c230e2e...2053205

Thanks Lukas! Great work :)
Doing upgrade recipe & release notes now.

PR #185, opened 2016-04-04
PLANNER-555: replace anonymous classes with lambdas

triceo:
I have applied various automated IDE code inspections:

- Replaced anonymous classes with lambdas.
- Removed unnecessary imports.
- Used diamond operators in places that didn't already use them.
- Removed needless boxing and unboxing.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
@@ -380,7 +380,7 @@ public class ProblemBenchmarkResult<Solution_> {
             long registeringProblemScale) {
         if (entityCount == null) {
             entityCount = registeringEntityCount;
-        } else if (entityCount.longValue() != registeringEntityCount) {
+        } else if (entityCount != registeringEntityCount) {
             logger.warn("The problemBenchmarkResult ({}) has different entityCount values ([{},{}]).",
                     getName(), entityCount, registeringEntityCount);
             // The entityCount is not unknown (null), but known to be ambiguous
@@ -388,7 +388,7 @@ public class ProblemBenchmarkResult<Solution_> {
         }
         if (variableCount == null) {
             variableCount = registeringVariableCount;
-        } else if (variableCount.longValue() != registeringVariableCount) {
+        } else if (variableCount != registeringVariableCount) {
             logger.warn("The problemBenchmarkResult ({}) has different variableCount values ([{},{}]).",
                     getName(), variableCount, registeringVariableCount);
             // The variableCount is not unknown (null), but known to be ambiguous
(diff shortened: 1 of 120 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -380,7 +380,7 @@ public void registerScale(long registeringEntityCount, long registeringVariableC
             long registeringProblemScale) {
         if (entityCount == null) {
             entityCount = registeringEntityCount;
-        } else if (entityCount.longValue() != registeringEntityCount) {
+        } else if (entityCount != registeringEntityCount) {
```
ge0ffrey: These changes are dangerous. The confuse wether the entityCount is being unboxed or the registeringEntityCount is being boxed (which is a different result when entityCount > 255!).
-1 on these changes. Clarity trumps verbosity for long term projects.

Because this is a big commit, with many changes (and a lot of code just different indented, so any changes in there are invisible to me), I would like to ask to have no other changes to be sneaked in than the lambda change.

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/value/chained/SubChainSelectorConfig.java
```diff
@@ -26,9 +26,6 @@
 import org.optaplanner.core.config.heuristic.selector.value.ValueSelectorConfig;
 import org.optaplanner.core.config.util.ConfigUtils;
 import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
-import org.optaplanner.core.impl.heuristic.selector.move.MoveSelector;
-import org.optaplanner.core.impl.heuristic.selector.move.generic.ChangeMoveSelector;
```
ge0ffrey: Weird that this is possible, as I did an optimize imports on optaplanner-core... checking

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/ProblemBenchmarkResult.java
triceo: Thy will shall be done.

triceo: I would just like to remind you that these are automated code changes - I sincerely doubt that IDEA would be recommending these changes if they had even the slightest potential of actually being harmful. (And, of course, no tests failed.)

ge0ffrey:
I need to take this into account before closing this PR:

<lpetrovi> ge0ffrey, when you're making the lambda changes, there's two classes you can optionally remove. (IntelliJ won't do that.)
<lpetrovi> ge0ffrey, one has to do with asserts and my PR removes that. (it is easily replaceable by Consumer<T>.)
<lpetrovi> ge0ffrey, the other... CodeAssertable... is easily replacable by Supplier<T>, but the code then looks IMO worse.

ge0ffrey:
Similar fixes done on master: it should include all mentioned here, except the boxing stuff (as that can cause bugs or ambiguity) and the CodeAssertable. Thanks for the patience!

PR #186, opened 2016-04-10
Ticket #549 TemporalValueRange added with tests

Blackjack92:
The `contains` method should be improved when possible.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
@@ -1,0 +1,134 @@
+package org.optaplanner.core.impl.domain.valuerange.buildin.temporal;
+
+import org.optaplanner.core.impl.domain.valuerange.AbstractCountableValueRange;
+import org.optaplanner.core.impl.domain.valuerange.util.ValueRangeIterator;
+import org.optaplanner.core.impl.solver.random.RandomUtils;
+
+import java.time.temporal.*;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Random;
+
+/**
+ * Created by kevin on 08.04.2016.
+ */
+public class TemporalValueRange extends AbstractCountableValueRange<Temporal> {
+
+    private final Temporal from;
+    private final Temporal to;
+    private final long incrementAmount;
+    private final TemporalUnit incrementUnit;
+
+    public TemporalValueRange(Temporal from, Temporal to, long incrementAmount, TemporalUnit incrementUnit) {
+        this.from = from;
+        this.to = to;
+        this.incrementAmount = incrementAmount;
+        this.incrementUnit = incrementUnit;
+
+        if (from == null || to == null) {
+            throw new IllegalArgumentException("Parameters should not be null.");
+        }
+
+        if (incrementAmount <= 0) {
+            throw new IllegalArgumentException("Increment amount must be greater 0");
+        }
+
+        if (!from.isSupported(incrementUnit) || !to.isSupported(incrementUnit)) {
+            throw new IllegalArgumentException("The increment unit " + incrementUnit + " is not supported.");
+        }
+
+        if (from.until(to, incrementUnit) < 0) {
+            throw new IllegalArgumentException("From is before to.");
+        }
+
+
+    }
+
+    private long getDuration() {
+        return from.until(to, incrementUnit);
+    }
+
+    @Override
+    public long getSize() {
+        return (getDuration() / incrementAmount);
+    }
+
+    @Override
+    public Temporal get(long index) {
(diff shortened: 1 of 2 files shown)


ge0ffrey:
ok to test

Blackjack92:
I first wanted to add a `TemporalAmount` instead of the `long incrementAmount, TemporalUnit incrementUnit` but because of the `until` method of the `Temporal` a `TemporalUnit` is needed.

ge0ffrey:
Merged, thanks for your contribution @Blackjack92 . I 'll add a line in the release notes "Contributed by Kevin Wallis", ok?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -0,0 +1,144 @@
...
+
+import java.time.temporal.Temporal;
+import java.time.temporal.TemporalUnit;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Random;
+
+/**
+ * Created by kevin on 08.04.2016.
+ */
```
ge0ffrey: Removing this, as we don't keep author tags in the files (it gets out of date, instead we use Git history which is 100% accurate + we mention contributors in release notes + in team page).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public boolean contains(Temporal value) {
+        if (value == null || !value.isSupported(incrementUnit)) {return false;}
+
+        // long delta = ...
+        // return from.until(value, incrementUnit) >= 0 && to.until(value, incrementUnit) < 0 && delta == 0;
+
+        for (long i = 0; i < getSize(); i++) {
```
ge0ffrey: There's a scaling issue here (BigO notation is bad). I 'll take a look at a better scaling implementation. But the current implementation is a good start of course :)

ge0ffrey:
Great PR, very nice quality. As usual, I did a few changes on top of it during the review (and found some issues in the other ranges meanwhile): http://github.com/droolsjbpm/optaplanner/compare/bda3f32...bb3626f

Thanks for contributing :)
Mentioned in the release notes here: http://www.optaplanner.org/download/releaseNotes/releaseNotes7.0.html
If you want your name listed as a contributor too on the team page, follow the instructions on the top of that page:
http://www.optaplanner.org/community/team.html

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        @Override
+        public Temporal next() {
+            if (upcoming.until(to, incrementUnit) < 0) {
+                throw new NoSuchElementException();
+            }
+
+            Temporal next = upcoming;
+            upcoming = upcoming.plus(incrementAmount, incrementUnit);
```
ge0ffrey: I found a very interesting, cool bug with this one

31-JAN + 1 month + 1 month returns 28-MAR
31-JAN + 2 months returns 31-MAR

So get() and createOriginalIterator() weren't in sync. Fixed

Blackjack92: Okay very good code changes :+1: , i have not recognized this bug while implementing :/
Why do you use the generic Temporal_?

ge0ffrey: To be able to have a CountableValueRange<LocalDate>, see ValueRangeFactory.createLocalDateRange() which is just a facade method. It works on all other Temporal subtypes too.

Blackjack92:
A short question because of the generic class:
`public class TemporalValueRange<Temporal_ extends Temporal> extends AbstractCountableValueRange<Temporal_>`

would it be possible to change this code to: 
`public class TemporalValueRange<Temporal_ extends Temporal & Comparable<? super Temporal_>> extends AbstractCountableValueRange<Temporal_>`?

so line 99: `Comparable<Temporal_> comparableValue = (Comparable<Temporal_>) value;` would not be needed, and i think it would be more typsafe and fail-safe.

ge0ffrey:
@Blackjack92 Great idea, Kevin. I 've adjusted the code accordingly :)

Blackjack92:
@ge0ffrey i think in the `class TemporalValueRange` the code from line 51 `Comparable<Temporal_> comparableFrom = (Comparable<Temporal_>) from;` can also be removed and the line 53 can be updated. Is this correct?

ge0ffrey:
yes, thanks @Blackjack92 :)

PR #189, opened 2016-04-22
[DROOLS-1137] better granularity for imported BOMs

psiroky:
* optaplanner now depends only on kie-third-party-bom,
   drools-bom and optaplanner-bom

 * before this change, the kie-parent-with-dependencies
   brought in also uberfire-boms, dashbuilder-bom, errai-bom,
   guvnor-bom and other BOMs which optaplanner does not need

Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/184

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.kie</groupId>
-    <artifactId>kie-parent-with-dependencies</artifactId>
+    <artifactId>kie-parent</artifactId>
     <version>7.0.0-SNAPSHOT</version>
     <!-- relativePath causes out-of-date problems on hudson slaves -->
     <!--<relativePath>../droolsjbpm-build-bootstrap/pom.xml</relativePath>-->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -62,6 +62,45 @@
...
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
```
psiroky: These dependencies should be I think moved directly into modules which use them. But that will require many more changes, so I did not want to mess this PR with those.

ge0ffrey: They are so basic :) Don't we already have junit in the kie-parent?

psiroky: `kie-parent` should not contain any dependencies I believe. I am fine with leaving these in the top-level POMs. There will be certain duplication, but I don't think that should be an issue since it will done once and likely won't change in a long time (maybe when upgrade to junit5).

PR #191, opened 2016-05-02
webexamples: run tests on WildFly 10.0.0.Final (instead of 9.0.1.Final)

psiroky:
* aligns with the overall KIE platform which uses (or will use)
   WildFly 10 by default


ge0ffrey:
Thanks Petr

PR #194, opened 2016-06-01
Add code coverage for phase lifecycle and fix a few bugs

yurloc:
I started learning to use [PIT](http://pitest.org/) for mutation testing. I picked the first line with missing code coverage and find a few bugs related to solver, phases and steps lifecycle. I have fixed the bugs and added code coverage.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
@@ -107,14 +107,12 @@ public abstract class AbstractPhase<Solution_> implements Phase<Solution_> {
     public void solvingStarted(DefaultSolverScope<Solution_> solverScope) {
         // bestSolutionRecaller.solvingStarted(...) is called by DefaultSolver
         termination.solvingStarted(solverScope);
-        phaseLifecycleSupport.fireSolvingStarted(solverScope);
     }
 
     @Override
     public void solvingEnded(DefaultSolverScope<Solution_> solverScope) {
         // bestSolutionRecaller.solvingEnded(...) is called by DefaultSolver
         termination.solvingEnded(solverScope);
-        phaseLifecycleSupport.fireSolvingEnded(solverScope);
     }
 
     @Override
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -312,9 +312,9 @@ public class DefaultSolver<Solution_> implements Solver<Solution_> {
     }
 
     public void removePhaseLifecycleListener(PhaseLifecycleListener<Solution_> phaseLifecycleListener) {
-        phaseLifecycleSupport.addEventListener(phaseLifecycleListener);
+        phaseLifecycleSupport.removeEventListener(phaseLifecycleListener);
         for (Phase phase : phaseList) {
-            phase.addPhaseLifecycleListener(phaseLifecycleListener);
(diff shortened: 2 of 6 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
```diff
@@ -107,14 +107,12 @@ public void setAssertShadowVariablesAreNotStaleAfterStep(boolean assertShadowVar
     public void solvingStarted(DefaultSolverScope<Solution_> solverScope) {
         // bestSolutionRecaller.solvingStarted(...) is called by DefaultSolver
         termination.solvingStarted(solverScope);
-        phaseLifecycleSupport.fireSolvingStarted(solverScope);
```
yurloc: Due to this, solvingStarted event was fired for each phase. I have moved this to DefaultSolver. As a result, solvingStarted event fires exactly once per Solver#solve(solution). Same for solvingEnded.

ge0ffrey: Very nice bug finding.

I 'll merge this PR after the big local refactor I am working on for initScore is merged (and I 'll deal with any merge conflicts).

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -301,14 +305,16 @@ public void removeEventListener(SolverEventListener<Solution_> eventListener) {
     }
 
     public void addPhaseLifecycleListener(PhaseLifecycleListener<Solution_> phaseLifecycleListener) {
+        phaseLifecycleSupport.addEventListener(phaseLifecycleListener);
```
ge0ffrey: There was and still is a design issue here:
This solver's phaseLifecycleSupport might be reused internally - as well as the phase's phaseLifecycleSupports. However currently the design presumes that any listener is in both (and that would mean that one created intended specifically for one phase will probably also get stepEnded events from other phases).

Because - by luck - neither the solver nor the phase phaseLifecycleSupport are reused beyond this method (which is used heavily by the statistics) - there are no bugs yet.

I'll refactor to separate them out and to have each phase also notify the "global" solver phaseLifecycleSupports upon phase/step events (but not on solver events to avoid duplication) by registering a bridge.

optaplanner-core/src/test/java/org/optaplanner/core/impl/phase/PhaseLifecyleTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
+import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+import org.optaplanner.core.impl.testdata.util.PlannerAssert;
+import org.optaplanner.core.impl.testdata.util.PlannerTestUtils;
+
+@RunWith(MockitoJUnitRunner.class)
+public class PhaseLifecyleTest {
+
+    @Mock
+    private PhaseLifecycleListener<TestdataSolution> listener;
```
ge0ffrey: The @Mock is interesting mockito feature, but I 'll remove this for now because:
- Code encapsulation: The listener can be a local variable. Furthermore, if more methods are added and verify() is used to check the number of mock calls, sharing the same mock between tests is error-prone.
- Consistency: If we start using this style, we should try to migrate existing code too - otherwise our code becomes an inconsistent mess - and if such partial introductions of technologies happen a few times, bugs start to appear.
- There can only be one @RunWith annotation: What if we want to write a unit test that runs on arquillian, is a parameterized test and uses a bit of mocking too? We'd need to configure 3 Runners somehow.

PR #198, opened 2016-06-18
Fixes after IP BOM 7.0.0.CR4 upgrade

psiroky:


ge0ffrey:
+1 Do it needs to be merged at the same time with others?

psiroky:
Yes, all PRs need to be merged together (will add a comment to each of them just to be sure).

ge0ffrey:
ok, can you merge in such occasions?

this is not a one time thing, such a multi-repo merge. We might want to standarize a tag or phrase, for example "multi-repo merge". Which implies that you need a +1 of each repo gatekeeper, but you 'll do the merge yourself - and the gatekeepers shouldn't merge because otherwise they'll break their repo (unless they merge all others too). Wydt?

psiroky:
Yeah, me or @csadilek are usually merging these PRs. We should definitely be putting a clear comment in the PR so that gatekeepers know all the PRs need to be merged at the same time. We could standardize a phrase or something, but not not sure if it is worth it. These kinds of big upgrades are not that common.

PR #200, opened 2016-06-22
[DROOLS-1216] exclude commons-logging and use jcl-over-slf4j instead

psiroky:
Part of an ensemble:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/226
https://github.com/droolsjbpm/drools/pull/815
https://github.com/droolsjbpm/optaplanner/pull/200
https://github.com/droolsjbpm/jbpm/pull/487
https://github.com/droolsjbpm/droolsjbpm-integration/pull/509
https://github.com/droolsjbpm/kie-uberfire-extensions/pull/29
https://github.com/droolsjbpm/guvnor/pull/328
https://github.com/droolsjbpm/drools-wb/pull/209
https://github.com/droolsjbpm/jbpm-designer/pull/302
https://github.com/droolsjbpm/jbpm-console-ng/pull/432
https://github.com/droolsjbpm/kie-wb-distributions/pull/303

### DIFF
--- a/optaplanner-webexamples/pom.xml
+++ b/optaplanner-webexamples/pom.xml
@@ -157,6 +157,12 @@
       <groupId>org.jboss.resteasy</groupId>
       <artifactId>resteasy-jaxrs</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-logging</groupId>
+          <artifactId>commons-logging</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.jboss.resteasy</groupId>


optaplanner-webexamples/pom.xml
```diff
@@ -157,6 +157,12 @@
       <groupId>org.jboss.resteasy</groupId>
       <artifactId>resteasy-jaxrs</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
```
ge0ffrey: 1) In the latest version, I don't see the commons-logging dep any more
  http://central.maven.org/maven2/org/jboss/resteasy/resteasy-jaxrs/3.0.17.Final/resteasy-jaxrs-3.0.17.Final.pom
As we're moving to EAP 7/ Wildfly 11, I presume we 'll be using that version?

2) Can we do these kind of exclusions on a higher level, such as ip-bom or kie-parent?

psiroky: 1) We are currently using RestEasy 3.0.16.Final (3.0.17.Final likely not fixes this). It comes transitively from `httpclient`:
```
+- org.jboss.resteasy:resteasy-jaxrs:jar:3.0.16.Final:test
|  +- org.jboss.spec.javax.annotation:jboss-annotations-api_1.2_spec:jar:1.0.0.Final:test
|  +- javax.activation:activation:jar:1.1.1:test
|  +- org.apache.httpcomponents:httpclient:jar:4.5:test
|  |  +- org.apache.httpcomponents:httpcore:jar:4.4.1:test
|  |  \- commons-logging:commons-logging:jar:1.1.1:test
|  \- net.jcip:jcip-annotations:jar:1.0:test
```

2) We could do the exclusions in e.g. kie-parent. However, the issue with this is that it works _only_ in case one uses our kie-parent's depMgmt. If users simply depend on our artifacts, without using our depMgmt, the commons-logging/log4j creeps back in (this is IMO a bug in Maven -- I believe we also talked about it briefly in the past). 

We could possibly differentiate the uses cases -- e.g. if it is for testing (so won't end-up in the transitive dep. closure for users using the artifacts), we don't need to explicitly exclude it. For direct compile/runtime dependencies we explicitly exclude it. I don't like this approach very much as it makes the exclusions inconsistent and one needs to know which approach to use in specific scenario. It also makes it harder for us to detect these use cases as everything will work for us (when using the kie-parent's depMgm), but could actually fail when used in standalone app/library.

ge0ffrey:
+1 to merge

mbiarnes:
+1

PR #202, opened 2016-06-28
PLANNER-582 Automatically generate Drools reproducer

yurloc:
Ready for review, not yet for merge.

Nice to have: more comments, more tests.

Need to decide: how to activate the reproducer? Currently it's done by setting `org.optaplanner.drools.reproducer` logger level to `INFO` or lower, which is probably not a good idea because it will also become enabled if just `org.optaplanner` logging level is low enough. New environment mode? Or just a simple system property?

Another serious issue to be resolved: how to include the scoring rules in the test so that it doesn't need to be inserted manually? Currently nurse rostering example score DRL is hard-coded. If path(s) to scoring DRL(s) were accessible from `DroolsScoreDirector` it would be OK. Another option, which I like more, would be to include the DRL source in the test, so that another manual step of copy&pasting could be left out when preparing a PR with the reproducer.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/reproducer/fact/ValueFact.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.optaplanner.core.impl.domain.common.ReflectionHelper;
+import org.optaplanner.core.impl.domain.common.accessor.BeanPropertyMemberAccessor;
+import org.slf4j.Logger;
+
+public class ValueFact implements Fact {
```
ge0ffrey: I'd really like to avoid that these reproducer classes would be confused for "real" classes, so I think we should prefix all classnames with for example "Reproducer"

triceo:
My 2 cents: this code should be its own Maven module, only included on the classpath optionally. Furthermore, the tool should be refactored so that it uses [standard Drools listeners](https://docs.jboss.org/drools/release/6.4.0.Final/drools-docs/html_single/#d0e15360).

This way, to enable the reproducing code is to include the module on the classpath. Also, it has the additional benefits of no extra code being called unless the listeners are actually registered. Which could also be done only in case the module is actually on the classpath, making the impact of this code on the non-reproducing run to be exactly zero. (As opposed to some negligible, but still present, overhead of the NoopReproducer.)

yurloc:
@triceo Using listeners it will be slightly more difficult (and error prone) to discern the list of initial inserts (which can be reduced only after reducing the updates - that's why it's kept separately). Otherwise it's a good idea and a 100% clean solution.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/reproducer/DroolsReproducer.java
```diff
@@ -0,0 +1,312 @@
...
+    private void printSetup() {
+        reproducerLog.info(
+                "    @Before\n" +
+                "    public void setUp() {\n" +
+                "        KieServices kieServices = KieServices.Factory.get();\n" +
+                "        KieModuleModel kieModuleModel = kieServices.newKieModuleModel();\n" +
+                "        KieFileSystem kfs = kieServices.newKieFileSystem();\n" +
+                "        kfs.writeKModuleXML(kieModuleModel.toXML());\n" +
+                // TODO don't hard-code score DRL
+                "        kfs.write(kieServices.getResources().newClassPathResource(\"org/optaplanner/examples/nurserostering/solver/nurseRosteringScoreRules.drl\").setResourceType(ResourceType.DRL));\n" +
```
yurloc: This needs to be resolved. The ideal solution would be to get all rules (not the source files) through KIE API, prune them automatically and finally inline source of the minimal set of rules that still reproduces the bug. I don't know if that's possible.

triceo:
@yurloc Unless "slightly more difficult" actually means "impossible," I'd still like us to implement this "clean" solution, provided that @ge0ffrey agrees.

(I'm sure that we could insert to the director a call to some reproducer singleton which would let the listeners and/or the post-processing logic know when the initial insertion phase is over.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/reproducer/DroolsReproducer.java
```diff
@@ -0,0 +1,312 @@
...
+            }
+        }
+        journal = new KieSessionJournalImpl(journal.getFacts(), journal.getInitialInserts(), m.getResult());
+        log.info("{} updates remaining.", journal.getMoveOperations().size());
+    }
+
+    private void pruneInserts() {
+        log.info("Pruning inserts...", journal.getInitialInserts().size());
+        RemoveRandomBlockMutator<KieSessionInsert> m = new RemoveRandomBlockMutator<KieSessionInsert>(journal.getInitialInserts());
+        while (m.canMutate()) {
```
baldimir: This could be extracted to separate method and called from both pruneUpdates() and pruneInserts().

ge0ffrey:
I haven't got the time to review today (or yesterday) - I will do once I
have the time.

This is a great contribution, but I too worry about mixing this in
optaplanner-core.
I really like Lukas's idea to somehow extract this code from
optaplanner-core into a separate module. For some time I 've been thinking
of creating optaplanner-test (to extract AbstractPerformanceTest for
example) and this could fit in there too.
But we 'd still need a way to register the "extended" DroolsScoreDirector -
I agree that drools listeners won't suffice. I think this registering
should somehow impact the ScoreDirectorConfig class, which builds the
DroolsScoreDirectorFactory which builds the director itself.

I fear it will probably take me until next week to discuss this further.

2016-06-29 3:03 GMT-07:00 Lukáš Petrovický <notifications@github.com>:

> @yurloc <https://github.com/yurloc> Unless "slightly more difficult"
> actually means "impossible," I'd still like us to implement this "clean"
> solution, provided that @ge0ffrey <https://github.com/ge0ffrey> agrees.
>
> (I'm sure that we could insert to the director a call to some reproducer
> singleton which would let the listeners and/or the post-processing logic
> know when the initial insertion phase is over.)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/droolsjbpm/optaplanner/pull/202#issuecomment-229313583>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAKy8I-K0XbzwAwYQt7bF8bYjkP4LvE_ks5qQkMGgaJpZM4JAln3>
> .
>



-- 
With kind regards / Met vriendelijke groeten,
Geoffrey De Smet

yurloc:
Just realized an important point: this only catches exceptions coming from `kieSession.fireAllRules()`. Corrupted score is handled on a higher level. So if we want this to reproduce score corruption too, we have one more requirement to take into account when redesigning the architecture.

yurloc:
During discussion with Geoffrey we have agreed on the following plan:

KIE API listeners are not going to be used for 2 reasons: a) there is no easy way to listen for the event of calling `fireAllRules()`, b) it would be much harder to determine which planning variable has changed without using OptaPlanner's `VariableDescriptor`. We can afford not using listeners because the KIE session recording is not going to be done by `DroolsScoreDirector` but by a new subclass. This way `DroolsScoreDirector` will stay untouched and so it will not contain any non-production code and there will be no performance impact to care about.

Classpath activation, either using `ServiceLoader` or `Class.forName()`, was rejected due to the downside of having to make the decision before starting JVM. The test generation code will rather stay in `optaplanner-core` and it will be possible to enable or disable it any time simply by changing solver configuration.

The above should also make it easier to use the code to generate reproducer for corrupted score. This will be done by overriding `AbstractScoreDirector#buildScoreCorruptionAnalysis()`.

The sub-package will be renamed to `testgen` and most of the classes will be prefixed with `TestGen` to avoid any confusion and unwanted IDE suggestions.

@triceo Any concerns?

triceo:
@yurloc Go ahead.

ge0ffrey:
+1 some other things discussed and extra idea's:

- DroolsScoreDirector and DroolsScoreDirectorFactory will be subclassed, so the original ones don't change - those are used in production code and should not suffer any perf penalty with this technique.
- The package is "testgen" (onder score.director.drools) and all classes are prefixed with "TestGen" so they aren't confused with production classes.
- ScoreDirectorFactoryConfig will use that TestGenDroolsScoreDirector(Factory) if a solver configuration property is specified. Some brainstorm idea's for that XMLelement: "generateDroolsTest", "generateDroolsUnitTest" "testGenEnabled", ... I believe it should contain the word drools as it's drools specific. It should be self-describing of course.
- ScoreDirectorFactoryConfig fails fast with a nice error message if easy or incremental JavaScoreDirector is combined with that generateDroolsTest element.
- Is that configuration property a Boolean (null or true basically) or is it a relative filename, for example null or "target/diagnostic/MyUnitTest.java" ? If it's a filename it should end with "File" or "OutputFile". The filename is relative to the working directory. If it's just a Boolean, the filename is hardcoded (which is fine too I think in this case). Wdyt?

optaplanner-examples/src/main/resources/logback.xml
```diff
@@ -31,4 +31,14 @@
...
+  <appender name="reproducerAppender" class="ch.qos.logback.core.FileAppender">
+    <file>DroolsReproducerTest.java</file>
+    <append>false</append>
+    <encoder>
+      <pattern>%m%n</pattern>
+    </encoder>
+  </appender>
+  <logger name="org.optaplanner.drools.reproducer" level="${logback.level.org.optaplanner.reproducer:-debug}" additivity="false">
+    <appender-ref ref="reproducerAppender"/>
+  </logger>
```
yurloc: TODO remove this when test printing via logger is replaced by direct File writing.

yurloc:
Done:
- [x] `ScoreDirectorFactoryConfig` based activation (`<generateDroolsTestOnError>true</>`)
- [x] fail fast if `generateDroolsTestOnError` is used with non-Drools score director
- [x] subclassed `DroolsScoreDirector` and its 2 factories to avoid any impact on production code
- [x] changed package name to `testgen`
- [x] added `TestGen` prefix to most of the classes that could be confused with a production code
- [x] implemented test generation for corrupted score

TODO:
- [ ] fix 2 failures in config related tests
- [ ] leave out the hard-coded nurse rostering score DRL from generated test
- [ ] add generated test file path in the thrown exception or at least log it
- [ ] replace logger based test printing with direct `File` writing to avoid having to define new `appender` (without it the test would be printed to stdout including timestamps, which is annoying)

Nice to have:
- [ ] add tests (there is almost zero code coverage)
- [ ] add some basic JavaDoc
- [ ] prune rules
- [ ] prune setup code
- [ ] use `scoreDrlList` or `scoreDrlFileList` from `ScoreDirectorFactoryConfig` when they are set
- [ ] print domain classes into the test to make it self-contained
- [ ] resolve all TODOs in the code
- [ ] change solver config to allow specifying generated test file path

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/TestGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+import org.optaplanner.core.impl.score.director.drools.testgen.mutation.HeadCuttingMutator;
+import org.optaplanner.core.impl.score.director.drools.testgen.mutation.RemoveRandomBlockMutator;
+import org.optaplanner.core.impl.score.director.drools.testgen.operation.TestGenKieSessionInsert;
+import org.optaplanner.core.impl.score.director.drools.testgen.operation.TestGenKieSessionOperation;
+import org.optaplanner.core.impl.score.director.drools.testgen.operation.TestGenKieSessionUpdate;
+import org.optaplanner.core.impl.score.director.drools.testgen.reproducer.OriginalProblemReproducer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class TestGenerator {
```
yurloc: Could be renamed to `TestGenJournalPruner`.

yurloc:
@ge0ffrey I consider this ready to merge, all requirements implemented. Small note to the `TestGen` prefix: I only prefixed public classes that could be confused with a production code. Let me know if you have any objections.

There's a lot things to improve but I think this can be postponed to the next time when the feature will be used (most probably by me).

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -243,6 +247,14 @@ public void setAssertionScoreDirectorFactory(ScoreDirectorFactoryConfig assertio
         this.assertionScoreDirectorFactory = assertionScoreDirectorFactory;
     }
 
+    public boolean isGenerateDroolsTestOnError() {
+        return generateDroolsTestOnError != null && generateDroolsTestOnError.booleanValue();
```
ge0ffrey: The getter should return null if it's null.
For all Config classes, it's important to distinguish between null (= not filled in) and the default value (the value it gets when it's null). This for example impacts the programmatic API after reading a solver config XML file. Also note that solver config files are sometimes written to XML (see benchmark report).

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -324,12 +336,20 @@ public ScoreDefinition buildDeprecatedScoreDefinition() {
                 throw new IllegalArgumentException("The scoreDirectorFactory cannot have "
                         + "both an easyScoreDirectorFactory and an droolsScoreDirectorFactory.");
             }
+            if (isGenerateDroolsTestOnError()) {
```
ge0ffrey: do not use getter, so it doesn't need to replace null by the default value.

I'd use BooleanUtils.isTrue(generateDroolsTestOnError)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/CorruptedScoreException.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.optaplanner.core.impl.score.director.drools.testgen;
+
+public class CorruptedScoreException extends RuntimeException {
```
ge0ffrey: If this extends IllegalStateException, we can reuse it for any score corruption in the normal ScoreDirectors too.

Should be it CorruptedScoreException or ScoreCorruptionException?
Or CorruptionException in case we use it for shadow variable corruption too?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/AbstractValueProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.optaplanner.core.impl.score.director.drools.testgen.fact;
+
+abstract class AbstractValueProvider implements ValueProvider {
```
ge0ffrey: Even though it's not public, it does show in ctrl-n in IntelliJ, so let's prefix all classes with TestGen?

Visibility doesn't really help there. They can be reverted public btw (but don't need to be, it's all good).

ge0ffrey:
Good to merge after:
- the TestGen prefix is on all classes (as even non-public classes show up on ctrl-n in intellij) an
- the Config class getters doen't return the default value if it's null
(see comments earlier)

Great work :) Looks very nice. Sorry that the Legacy bit forced you to duplicate that code. We'll deal with that later.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/CorruptedScoreException.java
yurloc: For me, CorruptedScoreException sounds more natural. I looked into Java docs and found the same pattern examples: UnsupportedOperationException, UndeclaredThrowableException.

Yeah, I think it is a good idea to use a custom exception for corrupted score and shadow variables but I'd rather make this change in a separate PR.

ge0ffrey: +1. Let's merge this as-is (when other comments are fixed) and then do a separate PR for that change.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/AbstractValueProvider.java
yurloc: +1 I knew this was the real reason for using the prefix :smile:

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/CorruptedScoreException.java
yurloc: Since the exception is currently only used internally by the TestGenCorruptedScoreReproducer I prefixed the exception with TestGen too.

yurloc:
@ge0ffrey Done.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -99,6 +102,8 @@
     @XStreamAlias("assertionScoreDirectorFactory")
     protected ScoreDirectorFactoryConfig assertionScoreDirectorFactory = null;
 
+    protected Boolean generateDroolsTestOnError = null;
```
ge0ffrey: Just noticed one more thing, that this property isn't inherited in the inherit() method of this Config class? It should be if it's not.

ge0ffrey:
Great work :) Merged (using the "squash and merge" button of the PR request).

PR #203, opened 2016-06-29
PLANNER-580: Make org.optaplanner.core.api.score.Score class hierarchy JavaScript-compile friendly

mcimbora:
This relates to reusing enums from `ConstructionHeuristicPhaseConfig` in the Workbench. As the enum methods introduced too many class dependencies, a split into enum & helper class is required.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/blueprint/SolverBenchmarkBluePrintType.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/blueprint/SolverBenchmarkBluePrintType.java
@@ -22,6 +22,7 @@ import java.util.List;
 import org.optaplanner.benchmark.config.SolverBenchmarkConfig;
 import org.optaplanner.core.config.constructionheuristic.ConstructionHeuristicPhaseConfig;
 import org.optaplanner.core.config.constructionheuristic.ConstructionHeuristicType;
+import org.optaplanner.core.config.constructionheuristic.ConstructionHeuristicTypeHelper;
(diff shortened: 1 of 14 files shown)


optaplanner-core/src/main/resources/org/optaplanner/core/OptaPlannerCoreAPI.gwt.xml
```diff
@@ -0,0 +1,26 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit 2.4.0//EN"
+    "http://google-web-toolkit.googlecode.com/svn/tags/2.5.0/distro-source/core/src/gwt-module.dtd">
+<module>
```
ge0ffrey: We won't make optaplanner-core a GWT module :) Let's look for another solution.

mcimbora: What we could do is to introduce new module (e.g. optaplanner-workbench-model), which would depend on optaplanner-core (and other modules, if required). Something like https://github.com/droolsjbpm/drools/tree/master/drools-workbench-models, however we wouldn't have to to re-implement the core classes, just adding them to module descriptor like this would be sufficient.

PR #205, opened 2016-07-13
persistence-jaxb: depend explicitly on JavaEE specs

psiroky:
* fix for JDK9 which does not by default includes these
 * still works on JDK8

@ge0ffrey please take a look and let me know if you have any concerns.

### DIFF
--- a/optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
+++ b/optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
@@ -54,12 +54,20 @@
     </dependency>
     <!-- XML and JSON -->
     <dependency>
+      <groupId>org.jboss.spec.javax.xml.bind</groupId>
+      <artifactId>jboss-jaxb-api_2.2_spec</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.sun.xml.bind</groupId>
       <artifactId>jaxb-core</artifactId>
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Thanks for having my back on JDK 9, Petr. Merged.

I wonder if they 'll get away with releasing JDK 9 GA like this too, as it kinda seems to break backwards compatibility, no? I would have expected the jars that have no module-info.java be able to reach out to those artifacts still.

psiroky:
Yeah, this definitely breaks backwards compatibility. Not sure if something changed in the past month or so, but before that it seemed like that this is the way to go for JDK9 compatible apps. One can basically use `-addmods java.se.ee` to add all the "EE" packages available in JDK9. However, that does not work in all cases as it can cause package conflicts. Case in point: module `java.sql` exports package `javax.transaction.xa`. The same package is also part of `transaction-api-spec-jar`, but this one contains more classes. So what happens in the runtime is that the the `javax.transaction.xa` package contains only limited number of classes (as the package from JDK is used) and this may result in ClassNotFoundExceptions. It's pretty confusing and it took me a while to figure this one out. More info in this jigsaw-dev thread: http://jigsaw-dev.1059479.n5.nabble.com/The-root-modules-when-compiling-code-in-the-unnamed-module-or-where-the-initial-class-is-loaded-fromh-tt5711268.html

PR #207, opened 2016-08-04
PLANNER-622 Prune nodes with optimistic bound == pessimistic bound

yurloc:


optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/scope/ExhaustiveSearchPhaseScope.java
```diff
@@ -92,7 +92,7 @@ public void registerPessimisticBound(Score pessimisticBound) {
             for (Iterator<ExhaustiveSearchNode> iterator = expandableNodeQueue.iterator(); iterator.hasNext(); ) {
                 // Prune it
                 ExhaustiveSearchNode node = iterator.next();
-                if (node.getOptimisticBound().compareTo(bestPessimisticBound) < 0) {
+                if (node.getOptimisticBound().compareTo(bestPessimisticBound) <= 0) {
```
ge0ffrey: Very nice bug finding.

This is the opposite end of ExhaustiveSearchDecider.java line 188 (which is fine):

                if (optimisticBound.compareTo(phaseScope.getBestPessimisticBound()) > 0) {
                    // It's still worth investigating this node further (no need to prune it)

Clearly, the opposite of > 0 is <= 0, not < 0.

It's these kinds of bugs that can waste a lot performance and yet never be discovered by production use alone.

PR #209, opened 2016-08-05
Review score directors, fix getContraintMatchTotals()

yurloc:
Contains a few minor JavaDoc fixes and makes all score directors throw
IllegalStateException from getConstraintMatchTotals() when
constraintMatchEnabled is false. Tests included.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
@@ -31,7 +31,6 @@ import org.kie.api.runtime.rule.Match;
 import org.kie.api.runtime.rule.RuleContext;
 import org.kie.api.runtime.rule.RuleRuntime;
 import org.kie.internal.event.rule.ActivationUnMatchListener;
-import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.score.constraint.ConstraintMatchTotal;
 import org.optaplanner.core.api.score.constraint.bigdecimal.BigDecimalConstraintMatch;
 import org.optaplanner.core.api.score.constraint.bigdecimal.BigDecimalConstraintMatchTotal;
@@ -41,7 +40,6 @@ import org.optaplanner.core.api.score.constraint.primint.IntConstraintMatch;
 import org.optaplanner.core.api.score.constraint.primint.IntConstraintMatchTotal;
 import org.optaplanner.core.api.score.constraint.primlong.LongConstraintMatch;
 import org.optaplanner.core.api.score.constraint.primlong.LongConstraintMatchTotal;
-import org.optaplanner.core.impl.score.director.ScoreDirector;
 
 /**
  * Abstract superclass for {@link ScoreHolder}.
@@ -64,6 +62,10 @@ public abstract class AbstractScoreHolder implements ScoreHolder, Serializable {
 
     @Override
     public Collection<ConstraintMatchTotal> getConstraintMatchTotals() {
+        if (!isConstraintMatchEnabled()) {
+            throw new IllegalStateException("When constraintMatchEnabled (" + isConstraintMatchEnabled()
+                    + ") is disabled, this method should not be called.");
+        }
         return constraintMatchTotalMap.values();
     }
 
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/score/holder/ScoreHolderTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/score/holder/ScoreHolderTest.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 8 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirector.java
```diff
@@ -42,6 +42,12 @@ public IncrementalScoreDirector(IncrementalScoreDirectorFactory scoreDirectorFac
                                     boolean constraintMatchEnabledPreference,
                                     IncrementalScoreCalculator<Solution_> incrementalScoreCalculator) {
         super(scoreDirectorFactory, constraintMatchEnabledPreference);
+        if (constraintMatchEnabledPreference
+                && !(incrementalScoreCalculator instanceof ConstraintMatchAwareIncrementalScoreCalculator)) {
+            throw new IllegalStateException("The score calculator " + incrementalScoreCalculator + " must implement "
+                    + ConstraintMatchAwareIncrementalScoreCalculator.class.getSimpleName()
+                    + " when constraintMatchEnabledPreference is true.");
+        }
```
yurloc: This causes the single test failure in optaplanner-examples. My intention was to fail fast if constraint matches are enabled (requested?) e.g. by one of the asserted modes. I am not sure if this is desired or not. Need to find out...

ge0ffrey: There's a reason why constraintMatchEnabledPreference is called "preference" (we might want to add a comment why that is). It's like a wish, a request: We prefer to have constraints match enabled (and are ready to pay the perf impact for that - it's for the UI explaining the score for example or a demanded supply of 1 of the algo's - most algo's don't demand it), but if the score director impl doesn't support it, then we don't get it anyway.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -64,6 +62,10 @@ public boolean isConstraintMatchEnabled() {
 
     @Override
     public Collection<ConstraintMatchTotal> getConstraintMatchTotals() {
+        if (!isConstraintMatchEnabled()) {
+            throw new IllegalStateException("When constraintMatchEnabled (" + isConstraintMatchEnabled()
+                    + ") is disabled, this method should not be called.");
+        }
```
ge0ffrey: +1 much better than an NPE indeed

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirector.java
ge0ffrey: I fear we won't be able to add this fail fast (although I always like any proposal to add more fail fasts :)). The other fail fasts added in the PR are fine.

ge0ffrey:
Jiri, see my previous comment - if that fail fast is deleted, I think I can merge this PR.

yurloc:
Done.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/holder/ScoreHolderTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.optaplanner.core.api.score.Score;
+
+import static org.junit.Assert.assertNotNull;
+
+public class ScoreHolderTest {
+
+    @Rule
+    public final ExpectedException thrown = ExpectedException.none();
```
ge0ffrey: Any reason why we're not using `@Test(expected = IllegalStateException.class)` on each test method instead? This approach is a bit hard to read and has less vertical density. OTH, I presume its to test the existence of "constraintMatchEnabled" in the message.

ge0ffrey:
Jenkins retest this

optaplanner-core/src/test/java/org/optaplanner/core/api/score/holder/ScoreHolderTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void illegalStateExceptionThrownWhenConstraintMatchNotEnabled() {
+        ScoreHolder scoreHolder = buildScoreHolder(false);
```
yurloc: For example, if I made a bug that would cause the `buildScoreHolder` method throw IllegalStateException, the test would still pass with `@Test(expected = IllegalStateException.class)`. That's why [JUnit documentation](https://github.com/junit-team/junit4/wiki/Exception-testing) warns to use the `expected` param with care.

ge0ffrey: agreed :)

Shall we rename the field "thrown" to "expectedException" so the link with expected is more clear?

yurloc: I intentionally keep the same name that is used in JUnit examples. Examples (of any project) form a de facto convention and if you follow this convention, your code might be easier to read for newcomers. Just to explain my reasoning.

If you like "expectedException" better, I will happily change it and, of course, keep using that in the future for the sake of consistency.

ge0ffrey: Yes, please refactor rename those to "expectedException". It's makes it easier to see that it's an alternative to the annotation parameter. It's also easier to deduce the type (ExceptedException). I think the JUnit examples were written in a jiffy.

It's a pity that ExpectedException ends with Exception but doesn't extend Exception... that's what really confused me. But "thrown" also confused me because I though it was what had been thrown, now what is expected to be thrown. Oh well, I guess "expectedExceptionRule" is confusing with Drools rules and "assertException" again hides the type... Let's stick with "expectedException"?

+1 for consistency in all usage of this approach. That's why I am think it's worth thinking twice on the name: it's a pattern for the future too.

yurloc:
Done.

ge0ffrey:
Thanks :) merged

PR #210, opened 2016-08-05
Fix for difficulty comparator

bibryam:
ShiftDate and ShiftType comparisons are done by name and id. It make more sence to order shift assigments by RequiredEmployeeSize. It gives better results too.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/solver/ShiftAssignmentDifficultyComparator.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/solver/ShiftAssignmentDifficultyComparator.java
@@ -30,10 +30,12 @@ public class ShiftAssignmentDifficultyComparator implements Comparator<ShiftAssi
         Shift aShift = a.getShift();
         Shift bShift = b.getShift();
         return new CompareToBuilder()
-                    .append(bShift.getShiftDate(), aShift.getShiftDate()) // Descending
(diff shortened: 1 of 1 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/solver/ShiftAssignmentDifficultyComparator.java
```diff
@@ -30,10 +30,12 @@ public int compare(ShiftAssignment a, ShiftAssignment b) {
         Shift aShift = a.getShift();
         Shift bShift = b.getShift();
         return new CompareToBuilder()
-                    .append(bShift.getShiftDate(), aShift.getShiftDate()) // Descending
-                    .append(bShift.getShiftType(), aShift.getShiftType()) // Descending
                     // For construction heuristics, scheduling the shifts in sequence is better
```
ge0ffrey: If it's indeed better, we should remove this comment.

In 2010, I ran with the date/type first way and your way, and on the datasets of this example, the benchmarker said that by date/type first way was better...

ge0ffrey:
ok to test

bibryam:
I think date/type is not true difficulty comparator as it is based on the date/type description/id.

ge0ffrey:
Closing this PR because it does not improve the solution quality, at least in my benchmarks. If you have benchmarks that show the opposite, please share and let's discuss :) Feel free to reopen if you disagree.

I did improve the comment for better clarity.
http://github.com/droolsjbpm/optaplanner/commit/9595a4877

PR #211, opened 2016-08-05
PLANNER-624 Ignore initScore in ES node comparators to avoid depth exploration

yurloc:


optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/node/comparator/ScoreFirstNodeComparator.java
```diff
@@ -35,8 +35,8 @@ public ScoreFirstNodeComparator(boolean scoreBounderEnabled) {
 
     @Override
     public int compare(ExhaustiveSearchNode a, ExhaustiveSearchNode b) {
-        // Investigate better score first
-        int scoreComparison = a.getScore().compareTo(b.getScore());
+        // Investigate better score first (ignore initScore to avoid depth first ordering)
```
ge0ffrey: Let's add suffix on all "// Investigate better score first" comments

ge0ffrey: Done (it was a small fix, so I've done it quickly meself), see http://github.com/droolsjbpm/optaplanner/commit/215351423

ge0ffrey:
thanks, good regression found, merged :)

PR #213, opened 2016-08-19
Benchmark report consistency fix

yurloc:
Score "calculation" count summary is used in the navigation button and in documentation.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
@@ -498,8 +498,8 @@ public class BenchmarkReport {
         }
         XYPlot plot = createScalabilityPlot(seriesList,
                 "Problem scale", NumberFormat.getInstance(locale),
-                "Average calculate count per second", NumberFormat.getInstance(locale));
-        JFreeChart chart = new JFreeChart("Average calculate count summary (higher is better)",
+                "Average calculation count per second", NumberFormat.getInstance(locale));
+        JFreeChart chart = new JFreeChart("Average calculation count summary (higher is better)",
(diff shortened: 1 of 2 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -498,8 +498,8 @@ private void writeAverageCalculateCountPerSecondSummaryChart() {
         }
         XYPlot plot = createScalabilityPlot(seriesList,
                 "Problem scale", NumberFormat.getInstance(locale),
-                "Average calculate count per second", NumberFormat.getInstance(locale));
-        JFreeChart chart = new JFreeChart("Average calculate count summary (higher is better)",
+                "Average calculation count per second", NumberFormat.getInstance(locale));
+        JFreeChart chart = new JFreeChart("Average calculation count summary (higher is better)",
```
ge0ffrey: Some time ago, I renamed "average calculation count per second (999)" to "score calculation speed (999/sec)". Looks like I missed these - we should rename them also.

yurloc: Ahh, sorry, I forgot to include an important notice: this is for 6.x stream, the fix doesn't apply to master.

yurloc:
There are other occurrences of the "calculate count" phrase but I only chose to fix this one because documentation uses "calculation count" when describing this part of the benchmark report (see 14.4.6. Average Calculation Count Summary (Graph and Table)).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
ge0ffrey: ah, I hadn't notice that, merging, thanks :)

PR #214, opened 2016-08-23
PLANNER-580 Make org.optaplanner.core.api.score.Score class hierarchy JavaScript-compile friendly

mcimbora:
The PR relates to OptaPlanner Workbench. As `org.optaplanner.core.api.score.AbstractScore` uses `Class.isInstance()` method which is not supported by gwt compiler, a refactoring is required to be able to reuse the Score classes on the client-side of the WB.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractScore.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractScore.java
@@ -19,7 +19,6 @@ package org.optaplanner.core.api.score;
 import java.io.Serializable;
 import java.math.BigDecimal;
 
-import org.optaplanner.core.api.domain.solution.PlanningSolution;
 import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
 
 /**
@@ -176,11 +175,6 @@ public abstract class AbstractScore<S extends Score> implements Score<S>, Serial
     @Override
     public boolean isSolutionInitialized() {
         return initScore >= 0;
-    }
-
-    @Override
-    public boolean isCompatibleArithmeticArgument(Score otherScore) {
-        return getClass().isInstance(otherScore);
     }
 
     protected String getInitPrefix() {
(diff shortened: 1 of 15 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScore.java
```diff
@@ -149,18 +149,6 @@ public int getLevelsSize() {
         return hardScores.length + softScores.length;
     }
 
-    /**
-     * @param level {@code 0 <= level <} {@link #getLevelsSize()}
-     * @return higher is better
-     */
-    public int getHardOrSoftScore(int level) {
```
ge0ffrey: Why is this method removed? It can be quite useful to users (although it might be not used by our examples or ourselves yet).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractScore.java
```diff
@@ -178,11 +177,6 @@ public boolean isSolutionInitialized() {
         return initScore >= 0;
     }
 
-    @Override
-    public boolean isCompatibleArithmeticArgument(Score otherScore) {
```
ge0ffrey: I 'll add something in upgrade recipe from 6.x to 7.x about this, in the exotic case that a user has defined a custom score.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScore.java
mcimbora: I didn't see any usages in the project (true, there can be clients who actually use it). I'll revert the change.

ge0ffrey: Thanks.

We should add a simple unit test for it, in BendableScoreTest etc.

mcimbora:
Thanks, I reverted the change & added tests for Bendable*ScoreHolder classes.

ge0ffrey:
Thanks, merged :)

ge0ffrey:
upgrade recipe adjusted:
http://github.com/droolsjbpm/optaplanner-website/commit/ede3aaebb

PR #215, opened 2016-08-23
Add home page URL to all POMs

yurloc:
There is a lot of artifacts under the `optaplanner.org` group ID but only `optaplanner` (the parent module) contains URL of www.optaplanner.org. When people search for OptaPlanner JARs they will probably look for optaplanner-core, optaplanner-benchmark etc. on mvnrepository.com where the URL is shown as a HomePage just after the artifact description. So why not providing a one-click access to www.optaplanner.org for people that possibly don't know about it? Is it a good idea?

Result:
maven repository org optaplanner optaplanner 7 0 0 beta1 - mozilla firefox_007

### DIFF
--- a/optaplanner-benchmark/pom.xml
+++ b/optaplanner-benchmark/pom.xml
@@ -20,6 +20,7 @@
 
     This module contains the benchmarker toolkit.
   </description>
+  <url>http://www.optaplanner.org</url>
 
   <properties>
     <osgi.Bundle-SymbolicName>org.optaplanner.benchmark</osgi.Bundle-SymbolicName>
--- a/optaplanner-core/pom.xml
+++ b/optaplanner-core/pom.xml
@@ -20,6 +20,7 @@
 
     This module contains the planning engine itself.
   </description>
+  <url>http://www.optaplanner.org</url>
 
   <properties>
     <osgi.Bundle-SymbolicName>org.optaplanner.core</osgi.Bundle-SymbolicName>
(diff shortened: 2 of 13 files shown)


ge0ffrey:
Agreed that we need to have this show up on mvnrepository.com, to pull the community together. Thanks for reporting and fixing :)

Normally, the url from the parent should be inherited, so is this 1) duplication as a workaround for a tmp problem in mvnrepository.com (so we need a comment above each line) or 2) the url isn't inherited (correctly).

mvnrepository.com only shows it if it's not inherited (so now only on the parent pom):
http://mvnrepository.com/artifact/org.optaplanner/optaplanner/7.0.0.Beta1
http://mvnrepository.com/artifact/org.optaplanner/optaplanner-core/7.0.0.Beta1

search.maven.org never shows the homepage, not even for the parent pom:
http://search.maven.org/#artifactdetails|org.optaplanner|optaplanner|7.0.0.Beta1|pom
http://search.maven.org/#artifactdetails%7Corg.optaplanner%7Coptaplanner-core%7C7.0.0.Beta1%7Cbundle

But here's the maybe surprising bit:
```
$ cd optaplanner-core
$ mvn help:effective-pom | grep -n5 url
...
<url>http://www.optaplanner.org/optaplanner-core</url>
...
```
So the url is not just inherited by maven, it's appended with the artifactId, because it presumes we're generating our homepage with mvn:site (hell no, it's no longer 2005). So it could be that mvnrepository.com is ignoring them because they are broken links. In any case, 2) applies: not inherited correctly, so it's not a workaround, so no comments needed and merging as is :)

Thanks

yurloc:
Wow, thanks for thorough investigation and a complete explanation. :+1:

PR #217, opened 2016-09-01
PLANNER-647 Workbench: Include ErraiApp.properties and score marshalling mappings in workbench-upstream-model module

mcimbora:
As the scores are value classes, custom errai mapping needs to be defined for them if they are ever to be used in Errai services. The problem with Bendable*Score types is that they don't expose hardScores & softScores variables, therefore it's not possible to marshall such value.
This PR adds getters for aforementioned variables. @ge0ffrey Are you ok with such change?

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScore.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScore.java
@@ -105,7 +105,15 @@ public final class BendableScore extends AbstractBendableScore<BendableScore>
         this.softScores = softScores;
     }
 
-    // Intentionally no getters for the hardScores or softScores int arrays to guarantee that this class is immutable
+    public int[] getHardScores() {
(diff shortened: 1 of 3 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScore.java
```diff
@@ -105,6 +105,14 @@ protected BendableScore(int initScore, int[] hardScores, int[] softScores) {
...
 
+    public int[] getHardScores() {
+        return Arrays.copyOf(hardScores, hardScores.length);
+    }
+
+    public int[] getSoftScores() {
+        return Arrays.copyOf(softScores, softScores.length);
+    }
+
     // Intentionally no getters for the hardScores or softScores int arrays to guarantee that this class is immutable
```
ge0ffrey: Can we remove this comment (actually put it above the Arrays.copyOf() and change it to explain why it's a copyOf).

PR #220, opened 2016-09-14
PLANNER-536 Workbench solver editor: bendable score combobox selection should allow filling in hard and soft level size

mcimbora:
The changes are required to allow score getter/setter overriding in `AbstractSolution` subclass. This is used by the Workbench when working with bendable score types to define `@PlanningScore` with respective `bendable(Hard|Soft)LevelsSize` annotation values.

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/extended/abstractsolution/TestdataExtendedAbstractSolutionOverridenScoreAccessors.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/extended/abstractsolution/TestdataExtendedAbstractSolutionOverridenScoreAccessors.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *       http://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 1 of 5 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/extended/abstractsolution/TestdataExtendedAbstractSolutionOverridenScoreAccessors.java
```diff
@@ -0,0 +1,88 @@
...
+        return SolutionDescriptor.buildSolutionDescriptor(TestdataExtendedAbstractSolutionOverridenScoreAccessors.class, TestdataEntity.class);
+    }
+
+    private List<TestdataValue> valueList;
+    private Object extraObject;
+
+    private List<TestdataEntity> entityList;
+
+    public TestdataExtendedAbstractSolutionOverridenScoreAccessors(String code) {
+        super();
```
ge0ffrey: the code parameter is lost, call "super(code);" instead

mcimbora: I removed the constructor completely. `code` parameter only makes sense when extending `org.optaplanner.core.impl.testdata.domain.TestdataObject`

PR #225, opened 2016-10-04
PLANNER-649 Migrate test asserts to AssertJ

yurloc:
* Replaces all usages of `org.junit.Assert` with `org.assertj.core.api.Assertions`.
* Using [PIT](http://pitest.org/) I have verified that changes are equivalent (no change in mutation coverage with respect to the default set of mutators). All tests pass.
* Added checkstyle rule that detects imports of `org.junit.Assert`.

@ge0ffrey I know you are working on the Partitioned Search, which is going to cause conflicts. Feel free to consider this PR dormant until you merge your work. I will then resolve the conflicts myself (and migrate newly introduced tests too).


optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/forager/finalist/StrategicOscillationByLevelFinalistPodiumTest.java
```diff
@@ -212,4 +213,8 @@ protected LocalSearchMoveScope buildMoveScope(LocalSearchStepScope stepScope,
         return moveScope;
     }
 
+    private static <E> void verifyListContainsExactlyOneInstance(List<E> list, E instance) {
```
jhrcek: Just for brevity: `assertThat(list).assertContainsExactly(instance);` serves the same purpose.

yurloc: @jhrcek Yes, you are almost correct. It serves the same purpose but there is a slight difference. The `verifyListContainsExactlyOneInstance` method makes sure that the single item in the list is exactly the expected _instance_. `containsExactly()`, on the other hand, only verifies the list items are _equal_ to the expected ones.

I was very careful to make _equivalent_ changes during the migration, otherwise I would use `containsExactly()`.

yurloc:
Rebased and ready to be merged. Squash recommended (I can do that if there's no button for it).

ge0ffrey:
I've discussed this with Jiri L and I am rejecting this incarnation of the PR because we 're favoring JUnit 4 (and later JUnit 5) over AssertJ. So as specced in PLANNER-649, we only want to use AssertJ to replace the PlannerAssert methods and the ExpectedException's of JUnit 4 (and some of those will be replaced once we start using JUnit 5).

PR #226, opened 2016-10-17
[PREVIEW] PLANNER-667 Introduce annotation-driven difficulty comparator definition

mcimbora:
Initial proposal of the mechanism to define difficulty comparator (and perhaps other comparators in the future) using annotations. This is especially useful in the Workbench app - we need to capture the fields used for the comparison of Planning entities and generate comparator code based on this.

https://github.com/droolsjbpm/optaplanner-wb/pull/106 currently uses a slight modification of the annotation hierarchy suggested here. Based on the discussion with @ge0ffrey it might be useful to move the generating code into the core module to provide an alternative way to define the comparator.

As described in the JIRA, this consists of 3-level hierarchy.
`ComparatorField` - Represents single problem fact field, e.g. Course -> lectureSize : int
`ComparatorFieldChain` - Chain of variables with defined order, e.g. Course -> teacher : Teacher -> age : String, false
`ComparatorDefinition` - Encapsulation of all problem fact fields used for planning entity comparison.

This is how the definition might look like with Cloud balancing example.
`@PlanningEntity(difficultyComparatorDefinition = 
@ComparatorDefinition(fieldChains = {
            @ComparatorFieldChain(fields = {@ComparatorField(name = "requiredCpuPower", type = int.class)}, ascending = true),
            @ComparatorFieldChain(fields = {@ComparatorField(name = "requiredMemory", type = int.class)}, ascending = true),
            @ComparatorFieldChain(fields = {@ComparatorField(name = "requiredNetworkBandwidth", type = int.class)}, ascending = true)})
)`

Things to agree on:
- naming
- location of the annotations - currently they reside in `org.optaplanner.core.api.domain.entity.comparator` package, but this may change if we want to reuse the comparator definition elsewhere
- implementation of the generating code - IMO this should be in first so that https://github.com/droolsjbpm/optaplanner-wb/pull/106 can be integrated. Implement the code generation in the second round.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/comparator/ComparatorDefinition.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/comparator/ComparatorDefinition.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


mcimbora:
@triceo @yurloc This might be interesting for you as well.

mcimbora:
Closing, implemented on the WB level.

PR #229, opened 2016-10-28
Partitioned Search: best solution event support

ge0ffrey:
@yurloc The PartitionQueue component is a going to be the concurrent place-to-be, so I can use a second pair of eyes on that code, to spot race conditions.


optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueue.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public PartitionQueue(int partCount) {
+        // TODO partCount * 100 is pulled from thin air
+        queue = new ArrayBlockingQueue<>(partCount * 100);
+        moveEventMap = new ConcurrentHashMap<>(partCount);
+        // HashMap because it's never modified after creation
+        nextEventIndexMap = new HashMap<>(partCount);
+        for (int i = 0; i < partCount; i++) {
+            nextEventIndexMap.put(i, new AtomicLong(0));
+        }
```
Zillode: This is merely a suggestion, but maybe you can enforce immutability of `nextEventIndexMap` using `Collections.unmodifiableMap()` if you don't want to rely on `ConcurrentHashMap`

ge0ffrey: Fixed, thanks for noticing :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/event/SolverEventSupport.java
```diff
@@ -32,19 +32,17 @@
...
     }
 
     public void fireBestSolutionChanged(DefaultSolverScope<Solution_> solverScope, Solution_ newBestSolution) {
         final Iterator<SolverEventListener<Solution_>> it = eventListenerSet.iterator();
-        long timeMillisSpent = solverScope.calculateTimeMillisSpentUpToNow();
-        Score newBestScore = solverScope.getSolutionDescriptor().getScore(newBestSolution);
         if (it.hasNext()) {
             final BestSolutionChangedEvent<Solution_> event = new BestSolutionChangedEvent<>(solver,
-                    timeMillisSpent, newBestSolution, newBestScore);
+                    solverScope.getBestSolutionTimeMillis(), newBestSolution, solverScope.getBestScore());
```
ge0ffrey: This definitely breaks stuff. spent (since start) != timeMillis (since epoch)... fixing.

ge0ffrey:
I intend to merge this later today, if no one objects.

ge0ffrey:
Jenkins retest this

ge0ffrey:
Jenkins retest this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -36,6 +38,13 @@
     protected EntityPlacer entityPlacer;
     protected ConstructionHeuristicDecider decider;
 
+    protected final boolean skipBestSolutionCloningInSteps = true;
```
yurloc: This is currently a constant. If it was the intention it should be renamed to upper case.

Since there is some decision logic based on this field, perhaps there should be a constructor parameter for it? Or is this planned for a future iteration?

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueue.java
```diff
@@ -0,0 +1,159 @@
...
+        long eventIndex = nextEventIndexMap.get(partIndex).getAndIncrement();
+        PartitionChangedEvent<Solution_> event = new PartitionChangedEvent<>(
+                partIndex, eventIndex, throwable);
+        queue.add(event);
+    }
+
+    @Override
+    public Iterator<PartitionChangeMove> iterator() {
+        // Currently doesn't be support to be called twice on the same instance
+        return new PartitionQueueIterator();
```
yurloc: Would it be worth throwing an exception on the second call of this method to avoid a possible bug? Depends on the probability of such situation and how long this limitation is going to be true...

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
ge0ffrey: Planned future interaction. It should be configurable, but I am not ready to expose it in the configuration just yet...I 'll add a todo.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueue.java
ge0ffrey: Currently it will never be true, because the phase uses it once and throws it away. You're right though, should be defensive. Not sure what to do yet... maybe just add a todo.

ge0ffrey: todo added

PR #230, opened 2016-11-05
VRP File With Multiple Depots

madhanpavan:
Vehicle Routing Example is assigning all vehicles to the first Depot in the Vehicle Routing Importer even when there are multiple Depots in the input file. This will force all vehicles to start at one Location. If we want the Vehicles to start from different locations, we can add one Depot for each Vehicle. If the number of Vehicles do not match with the number of Depots then just assign all of them to one Depot. I also modified example file and created belgium-road-time-n40-k10.vrp with 10 Depots.


optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/persistence/VehicleRoutingImporter.java
```diff
@@ -391,6 +391,12 @@ private void createVehicleList() {
                 vehicle.setId(id);
                 id++;
                 vehicle.setCapacity(capacity);
+                //Multiple Depot VRP assign one vehicle for each depot
+                if(depotList.size() == vehicleListSize){
+                    vehicle.setDepot(depotList.get(i));
+                } else {
+                    vehicle.setDepot(depotList.get(0));
+                }
                 vehicle.setDepot(depotList.get(0));
```
ge0ffrey: Shouldn't this line be removed? It puts the vehicle's depot back to depo 0.

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/persistence/VehicleRoutingImporter.java
```diff
@@ -391,6 +391,12 @@ private void createVehicleList() {
                 vehicle.setId(id);
                 id++;
                 vehicle.setCapacity(capacity);
+                //Multiple Depot VRP assign one vehicle for each depot
+                if(depotList.size() == vehicleListSize){
+                    vehicle.setDepot(depotList.get(i));
```
ge0ffrey: There might be more depots than vehicles or there might be less.
Probably something like depotList.get(i % depotList.size()) should be used.

optaplanner-webexamples/src/main/resources/org/optaplanner/webexamples/vehiclerouting/belgium-road-time-n40-k10.vrp
```diff
@@ -0,0 +1,174 @@
+NAME: belgium-road-time-n50-k10
```
ge0ffrey: filename and name here should somehow clearly note that it has multiple depots.

madhanpavan: I didn't want to change the existing behavior if the number of vehicles doesn't match with the depots. I thought this would be a quick fix with minimum changes to support multiple depots with one vehicle. If we decide upon the layout of the VRP file to support multiple vehicles for each depot then we can make changes accordingly.

madhanpavan: Do we have to make changes to DepotDistanceCustomerDifficultyWeightFactory and DepotAngleCustomerDifficultyWeightFactory? 
The weights are calculated based on the first Depot.

ge0ffrey:
Relates to https://issues.jboss.org/browse/PLANNER-689

madhanpavan:
Hi Geoffrey,

I made the suggested changes. Please review.

Thanks,
Eswar

ge0ffrey:
ok to test

ge0ffrey:
Thanks @madhanpavan ! Merged.

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/persistence/VehicleRoutingImporter.java
```diff
@@ -391,6 +391,8 @@ private void createVehicleList() {
                 vehicle.setId(id);
                 id++;
                 vehicle.setCapacity(capacity);
+                //Multiple Depot VRP assign one vehicle for each depot
+                depotList.get(i % depotList.size());
```
ge0ffrey: This doesn't actually do anything, fixing.

madhanpavan: This actually improves the solution for VRPs with multiple depots. Attached are the screenshots of the solution before and after the fix. I ran belgium-road-time-multi-depot-n40-k10.vrp. This file has 10 depots. 

after
before

yurloc: @madhanpavan The point is that the value returned from depotList.get() isn't used and since ArrayList.get() and ArrayList.size() methods have no side effects the line indeed does nothing (unless I'm missing something). Did you instead want to write:
```java
vehicle.setDepot(depotList.get(i % depotList.size()));
```
or something similar?

madhanpavan: Oh yeah. I messed up. It is supposed to be vehicle.setDepot(depotList.get(i%depotList.size()));
I apologize for the confusion.

ge0ffrey: I fixed it locally, but I am looking into generating a good set multidepot datasets.

ge0ffrey:
Just pushed a number of datasets so multi-depot VRP's can be benchmarked decently.

PR #231, opened 2016-11-18
Fix checkstyle warnings

yurloc:


optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueue.java
```diff
@@ -31,7 +31,7 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * Thread-safe
+ * Thread-safe.
```
ge0ffrey: According to javadoc guidelines, sentences should end with a dot and a sentance has at least 1 verb.
I'll change this into a sentence after merging :)

ge0ffrey: Looks like I had several other instances with a non-sentance that ended with a dot in javadocs. Fixed.

PR #233, opened 2016-11-23
TestGen fixes 2

yurloc:
Next round of fixes and improvements, this time based on the isolation work on DROOLS-1326 and adding reproducers for DROOLS-1174 and DROOLS-1187.

* if `<scoreDrl>` is used in solver config it is now written to the reproducer
* implemented fact setup pruning -- eliminates working fact setters that are not necessary for reproducing the issue


yurloc:
I think it's OK to squash the history.

PR #239, opened 2016-12-04
Fix various Javadoc issues

yurloc:
Most frequent issue was broken links due to usage of `Solution_`
parameter in method references. Javadoc tool cannot handle that but when
the method reference is used without parameters, the link is generated
correctly.

Another issue was with `@value` tag. It appears to require FQCN when
referencing a different class.

Also fixes missing period at ends of first sentences and typos.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/descriptor/EntityIndependentValueRangeDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/descriptor/EntityIndependentValueRangeDescriptor.java
@@ -25,10 +25,10 @@ import org.optaplanner.core.api.domain.valuerange.ValueRange;
 public interface EntityIndependentValueRangeDescriptor<Solution_> extends ValueRangeDescriptor<Solution_> {
 
     /**
-     * As specified by {@link #extractValueRange(Solution_, Object)}.
+     * As specified by {@link ValueRangeDescriptor#extractValueRange}.
      * @param solution never null
      * @return never null
-     * @see #extractValueRange(Solution_, Object)
+     * @see ValueRangeDescriptor#extractValueRange
      */
     ValueRange<?> extractValueRange(Solution_ solution);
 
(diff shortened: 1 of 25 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/descriptor/EntityIndependentValueRangeDescriptor.java
```diff
@@ -25,10 +25,10 @@
 public interface EntityIndependentValueRangeDescriptor<Solution_> extends ValueRangeDescriptor<Solution_> {
 
     /**
-     * As specified by {@link #extractValueRange(Solution_, Object)}.
+     * As specified by {@link ValueRangeDescriptor#extractValueRange}.
```
ge0ffrey: I don't understand: why is it ok to leave out the parameters here, but not in all the other cases with Solution_? What's the difference?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/domain/Meeting.java
```diff
@@ -26,7 +26,7 @@
 
     private String topic;
     /**
-     * Multiply by {@link TimeGrain#GRAIN_LENGTH_IN_MINUTES} to get duration in minutes
+     * Multiply by {@link TimeGrain#GRAIN_LENGTH_IN_MINUTES} to get duration in minutes.
```
ge0ffrey: Nitpick: This is technically not a sentence, because it has no verb (javadoc style says sentence parts should not end in dot IIRC). Not important, so I 'll just merge anyway.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/descriptor/EntityIndependentValueRangeDescriptor.java
ge0ffrey: Agreed, I must have confused the old for the new, ignore my original comment, merging. Thanks :)

ge0ffrey:
Thanks Jiri, great work. My apologies for letting this one drag on for so long, holidays :)

PR #240, opened 2016-12-07
Add revapi plugin

MarianMacik:
Hi, @ge0ffrey,

here is the revapi plugin for the optaplanner repo. Currently, the following is checked:

- optaplanner-benchmark - packages org.optaplanner.benchmark.api and org.optaplanner.benchmark.config
- optaplanner-core - packages org.optaplanner.core.api and org.optaplanner.core.config

Thanks,

Marian

### DIFF
--- a/optaplanner-benchmark/src/build/revapi-config.json
+++ b/optaplanner-benchmark/src/build/revapi-config.json
@@ -1,0 +1,95 @@
+{
+  "revapi": {
+    "java": {
+      "filter": {
+        "packages": {
+          "comment": "We don't want to check transitive classes, since we already check them in their own module.",
+          "regex": true,
+          "include": [
+            "org\\.optaplanner\\.benchmark\\.api.*",
+            "org\\.optaplanner\\.benchmark\\.config.*"
+          ]
+        }
+      }
+    },
+    "_comment": "Changes between 6.5.0.Final and master. These changes are desired and thus ignored. They should be removed when 7.0.0.Final is available.",
+    "ignore": [
+      {
+        "code": "java.field.removedWithConstant",
+        "old": "field org.optaplanner.benchmark.config.PlannerBenchmarkConfig.AVAILABLE_PROCESSOR_COUNT",
+        "justification": "PLANNER-491 empty classes for Partitioned Search POC + extract CofnigUtils.resolveThreadPoolSizeScript()."
+      },
+      {
+        "code": "java.method.returnTypeChanged",
+        "old": "method long org.optaplanner.benchmark.config.PlannerBenchmarkConfig::calculateWarmUpTimeMillisSpentLimit()",
+        "new": "method java.lang.Long org.optaplanner.benchmark.config.PlannerBenchmarkConfig::calculateWarmUpTimeMillisSpentLimit()",
+        "justification": "PLANNER-712 Benchmark should warm up for 30 seconds by default"
+      },
+      {
+        "code": "java.method.numberOfParametersChanged",
+        "old": "method void org.optaplanner.benchmark.config.ProblemBenchmarksConfig::buildProblemBenchmarkList(org.optaplanner.benchmark.impl.result.SolverBenchmarkResult)",
+        "new": "method void org.optaplanner.benchmark.config.ProblemBenchmarksConfig::buildProblemBenchmarkList(org.optaplanner.core.config.SolverConfigContext, org.optaplanner.benchmark.impl.result.SolverBenchmarkResult)",
+        "justification": "PLANNER-448 Benchmark report should state 'Hard score' and 'Soft score' instead of 'Score level 0' and 'Score level 1' by using ScoreDefinition.getLevelLabels()"
+      },
+      {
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -189,6 +189,47 @@
             </execution>
           </executions>
         </plugin>
+        <plugin>
+          <groupId>org.revapi</groupId>
+          <artifactId>revapi-maven-plugin</artifactId>
```
ge0ffrey: Can't we move this config into kie-parent - presume it will be the same for kie-api etc.

optaplanner-benchmark/src/build/revapi-config.json
```diff
@@ -0,0 +1,118 @@
...
+        "code": "java.missing.newClass",
+        "old": "missing-class sun.java2d.pipe.hw.ExtendedBufferCapabilities",
+        "new": "missing-class sun.java2d.pipe.hw.ExtendedBufferCapabilities",
+        "justification": "Probably a bug in revapi, nothing serious"
+      },
+      {
+        "code": "java.missing.oldClass",
+        "old": "missing-class sun.util.locale.provider.LocaleServiceProviderPool.LocalizedObjectGetter",
+        "new": "missing-class sun.util.locale.provider.LocaleServiceProviderPool.LocalizedObjectGetter",
+        "justification": "Probably a bug in revapi, nothing serious"
```
ge0ffrey: Can we filter out these bogus messages (anything not in org.optaplanner basically) in the plugin configuration?

pom.xml
MarianMacik: Actually a good point, but this is not 100% final config. I will move it to the parent from all other repositories once it is a final config. So if you do not mind I would leave it that way for now.

MarianMacik: After consulting with psiroky, we decided to move it to kie-parent. I will change it tomorrow.

ge0ffrey: thanks :)

MarianMacik:
OK, so the configuration has been moved to the parent. PR for it is waiting: droolsjbpm/droolsjbpm-build-bootstrap#326. As I said it is already filtered for the optaplanner packages, these missing classes are false-negatives, it is already reported as an issue in revapi, let's hope it is fixed in the near future. Also there are quite a lot of changes so feel free to drop justifications so it can be tracked. One sentence is mostly more than enough.

ge0ffrey:
Jenkins retest this.

My last commit probably broke this PR as I reverted a public API change back to the 6.x form. More info in:
http://github.com/droolsjbpm/optaplanner-website/commit/a0b712769

ge0ffrey:
Ok, as expected, let me understand what needs to happen on this PR to fix it, that's a good learning exercise for revapi for me.

MarianMacik:
It is good that you are willing to learn it :+1: But I am now upgrading it to the newest version (hopefully last time for a long time) since it addresses a couple of bugs. So please consider revapi 0.8.1 and its java dependency 0.13.1 when fixing this PR :) . The reason is that new version might discover new changes etc. Thanks!

MarianMacik:
Hi, @ge0ffrey, now it is ready so you can provide justifications. Once they are there we can merge it. Thanks!

ge0ffrey:
jenkins test this

ge0ffrey:
Ideally, each of these "ADD YOUR EXPLANATION FOR THE NECESSITY OF THIS CHANGE" should be looked up in the upgrade recipe and that text should be replaced with the title of the item there:
  http://www.optaplanner.org/download/upgradeRecipe/upgradeRecipe7.0.html
so we can see if any are not documented in that upgrade recipe.

That's a lot of work, but if we don't do that, adding this revapi is just pro forma.

optaplanner-benchmark/src/build/revapi-config.json
```diff
@@ -0,0 +1,95 @@
+{
+  "revapi": {
+    "java": {
+      "filter": {
+        "packages": {
+          "comment": "We don't want to check transitive classes, since we already check them in their own module.",
+          "regex": true,
+          "include": [
+            "org\\.optaplanner\\.benchmark\\.api.*",
+            "org\\.optaplanner\\.benchmark\\.config.*"
```
ge0ffrey: For config classes, I think the regex should only look at the fields (names and types) (optionally including the getters and setters), but not at the build and other methods, as those aren't guaranteed to be backwards compatible.

ge0ffrey: If revapi doesn't support this, we should seriously consider not including the config packages for now.

MarianMacik:
Hi, @ge0ffrey, I have added justifications to both files. Please check if you are satisfied with them. For things I did not find in the upgrade recipe I used a commit message or a sentence with a short explanation. I will now try to figure out how to filter out methods which are not getters and setters. Thanks so far!

ge0ffrey:
@MarianMacik Good enough to merge already, so I am merging (to aviod merge conflicts later).
But we indeed need a solution for those config clases. If we can't limit it to "only fields" or only "fields + getters and setters" then we need to remove them from the revapi list I think.

PR #244, opened 2016-12-10
Compiler warnings - introduce Score_ parameter to exhaustive search

yurloc:
Using the score type parameter allows safe ES node comparison and calculation of optimistic and pessimistic bounds.

Compiler warnings reduced by 28.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/exhaustivesearch/NodeExplorationType.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/exhaustivesearch/NodeExplorationType.java
@@ -18,6 +18,7 @@ package org.optaplanner.core.config.exhaustivesearch;
 
 import java.util.Comparator;
 
+import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.impl.exhaustivesearch.node.ExhaustiveSearchNode;
 import org.optaplanner.core.impl.exhaustivesearch.node.comparator.BreadthFirstNodeComparator;
 import org.optaplanner.core.impl.exhaustivesearch.node.comparator.DepthFirstNodeComparator;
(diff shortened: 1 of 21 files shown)


ge0ffrey:
hmmm, we already got the generic parameter Solution_ all over the place (not enough those, see ChangeMove vs SwapMove). And that has good benefits for the user, as their CustomPhaseCommand etc don't need to cast any more.
But what's the advantage of this Score_ generic parameter? If it's only to please the compiler, then maybe the solution is worse than the problem...?

yurloc:
I think it's a good approach not to ignore warnings, because some of them may point at a real issue. However when there is such a huge amount of warnings it's very difficult to find the ones that need fixing. So we should either fix them all or turn off `-Xlint` completely (if no one's ever going to pay attention to the warnings). My choice would be to fix them all to have a clean build, which makes it easy and encourages to resolve any new warnings that might occur.

While generics and their proper usage improves types safety, I understand that they're most beneficial in public APIs as opposed to internal ones where we have a better knowledge and control over types. For example in ES node comparators, we basically know that each node has the same score type so we could just as well say
```java
@SuppressWarnings("unchecked") // all nodes have the same score type
int scoreComparison = a.getScore().compareTo(b.getScore());
```
In my opinion the better alternative is to declare the requirement (or knowledge) that the score types are same in method signature even at the "cost" of making the method or class generic:
```java
<S extends Score> int compare(ExhaustiveSearchNode<S> a, ExhaustiveSearchNode<S> b);
```

Another example is the `Solution_`. Leaving aside the fact that it improves the user facing API, it also helps in reading the internal parts of the code. Even though we don't care about the actual type of the solution and we could just say `Object` at many places, it is still a helpful piece of information that the type we're looking at represents the solution. I think this is important since there are also other domain entities whose type is unconstrained, i.e. the planning entities and variables. Without generics, they all would have to be referred to as `Object` which could be confusing sometimes.

To conclude, I think that the approach of using of generics, apart from better type safety, improves code self-documentation by expressing otherwise hidden knowledge about the types. I agree this doesn't bring any value to the end-user however I personally see the added value for those who work with the internal parts of the code. On the other hand there is the increased verbosity, which might be considered a drawback. I am though not sure how serious drawback this is and there might be some other reasons for not adopting generics that widely?

ge0ffrey:
Because the generic Score_ parameter is likely to leak through all the code, such as into DefaultSolverScope.getBestScore() and from there to everywhere, such as into the hello world:

     Solver<CloudBalance, HardSoftScore> solver

I believe the fix is worse than the problem and I am currently rejecting this PR.

PR #251, opened 2017-01-08
Build fix for Windows.

bradh:


optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenValueFactTest.java
```diff
@@ -40,7 +38,7 @@ public void initialization() {
 
         StringBuilder sb = new StringBuilder(100);
         fact.printInitialization(sb);
-        assertEquals("    private final TestdataValue testdataValue_321 = new TestdataValue();\n", sb.toString());
+        assertEquals("    private final TestdataValue testdataValue_321 = new TestdataValue();" + System.lineSeparator(), sb.toString());
```
ge0ffrey: @yurloc Any idea why it's not only a "\n" on Windows too? Ideally all .java files only have linux line endings, regardless on what platform they are generated (we already enforce this for our normal .java files in .gitattributes)

yurloc: It's because TestGen classes actually use `System.lineSeparator()` when printing code. My intention was to produce output in a format which is native to the runtime environment.

The restriction you are mentioning is there to avoid mess in the files that are version controlled. TestGen output (the reproducing test) is not meant to be version controlled so it doesn't matter much what line endings are used there.

So we just need to make a trade off between two extreme situations:
1. A Windows user opens the generated test in some ancient editor that doesn't support linux endings. The user is not happy. This won't happen with the current solution.
2. The Windows user is happy and attaches the file to a JIRA, a Drool developer copy-pastes part of that file and IF he uses an IDE that doesn't convert line endings, he may commit a test with Windows line endings. This can happen with the current solution.

I can hardly imagine a Windows user/developer using a text editor/IDE that doesn't support `\n`s nowadays so I think it might actually be more consistent to use `\n`s under all circumstances when printing the generated reproducer. What do you think?

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenValueFactTest.java
```diff
@@ -20,14 +20,12 @@
...
 import java.util.List;
-
+import static org.junit.Assert.*;
 import org.junit.Test;
 import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
 import org.optaplanner.core.impl.testdata.domain.TestdataValue;
 import org.optaplanner.core.impl.testdata.domain.collection.TestdataEntityCollectionPropertyEntity;
 
-import static org.junit.Assert.*;
-
```
yurloc: @bradh This is a code style change that is a) not related to the actual fix, and b) doesn't comply with the [code style rules](https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml#L19) set for this project. Please revert it if you are updating this PR. If you plan to contribute on a regular basis, please consider importing the code style rules.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenValueFactTest.java
ge0ffrey: I think we need to use \n under all circumstances (including windows) because we already enforce that on the drools/optaplanner repo's (see README in build-bootstrap) and through .gitattributes (not sure if that's in the drools repo yet, but it should be)

yurloc: In that case I would make this test more relaxed and not care about white spaces around the statement. `trim()` before asserting should do it.

The change from `System.lineSeparator()` to `\n` should be done in a new PR. I'd like to take care of it because I also intend to replace some usage of `StringBuilder.append()` with plain string concatenation (`+`) to lower verbosity and increase readability.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenValueFactTest.java
bradh: @yurloc: I understand that you are going to propose an alternative fix, so the PR is not required, so I won't fix this one.

I would import the rules if there were some for NetBeans, or they were defined in a non-IDE specific way.

ge0ffrey:
Fixed with Jiri's PR https://github.com/droolsjbpm/optaplanner/pull/252

Thanks for reporting this issue, Brad.
Thanks for fixing it, Jiri.

PR #253, opened 2017-01-14
PLANNER-539 Reduce usage of commons.lang3 in optaplanner-benchmark.

bradh:


ge0ffrey:
ok to test

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/PlannerBenchmarkConfig.java
```diff
@@ -209,7 +205,8 @@ public PlannerBenchmark buildPlannerBenchmark(SolverConfigContext solverConfigCo
         plannerBenchmarkResult.setAggregation(false);
         int parallelBenchmarkCount = resolveParallelBenchmarkCount();
         plannerBenchmarkResult.setParallelBenchmarkCount(parallelBenchmarkCount);
-        plannerBenchmarkResult.setWarmUpTimeMillisSpentLimit(defaultIfNull(calculateWarmUpTimeMillisSpentLimit(), 30L));
+        Long warmupTimeLimit = calculateWarmUpTimeMillisSpentLimit();
+        plannerBenchmarkResult.setWarmUpTimeMillisSpentLimit(Objects.isNull(warmupTimeLimit) ? 30L : warmupTimeLimit);
```
ge0ffrey: Let's use "warmupTimeLimit == null" instead of Objects.isNull().
That isNull() method only exists to be used as a predicate for streams etc (see javadoc).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
```diff
@@ -397,7 +396,7 @@ public RenameNodeDialog(final DefaultMutableTreeNode treeNode) {
                 @Override
                 public void actionPerformed(ActionEvent e) {
                     String newBenchmarkResultName = benchmarkResultNameTextField.getText();
-                    if (StringUtils.isEmpty(newBenchmarkResultName)) {
+                    if (newBenchmarkResultName.isEmpty()) {
```
ge0ffrey: this is not null safe

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/measurement/ScoreDifferencePercentage.java
```diff
@@ -121,7 +120,7 @@ public String toString() {
     }
 
     public String toString(String locale) {
-        return toString(LocaleUtils.toLocale(locale));
+        return toString(new Locale(locale));
```
ge0ffrey: this fails for "en_US"
See javadoc differences between methods

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/ReportHelper.java
```diff
@@ -47,7 +47,16 @@ public static String escapeHtmlId(String rawHtmlId) {
...
+        if (Objects.isNull(s)) {
+            return null;
+        }
+        if (s.isEmpty()) {
+            return s;
+        }
+        StringBuilder sb = new StringBuilder();
+        sb.append(Character.toTitleCase(s.charAt(0)));
+        sb.append(s.substring(1));
+        return sb.toString();
```
ge0ffrey: Although I want to get rid of the commons-lang3 dependency if JDK 8 can support all those methods, I don't want to get rid of it by writing the methods ourselves - I think it's better to keep the commons-lang3 dependency then.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/PlannerBenchmarkResult.java
```diff
@@ -267,7 +265,7 @@ public void initBenchmarkReportDirectory(File benchmarkDirectory) {
             String directoryName = timestamp + (duplicationIndex == 0 ? "" : "_" + duplicationIndex);
             duplicationIndex++;
             benchmarkReportDirectory = new File(benchmarkDirectory,
-                    BooleanUtils.isFalse(aggregation) ? directoryName : directoryName + "_aggregation");
+                    aggregation ? directoryName + "_aggregation" : directoryName);
```
ge0ffrey: not null safe

ge0ffrey:
Brad, would you mind if I run an "optimize imports" on optaplanner-benchmark and reject this PR?
It seems that the solution of getting rid of the commons-lang3 dependency is worse than the problem. I proposed the JIRA, but I think the jira needs to change to "replace as much as possible with Java 8 methods".

Commons-lang3 isn't an ok dependency - they're backwards compatible. I worry more about the Guava dependency (they break backwards compatibility every 18 months).

bradh:
OK. I had hoped to get closer to full removal, but there are a few places where that will be messy (like removing public API).

ge0ffrey:
Thanks - I'll do an optimize imports after merging your other PR (which is waiting for changes, although if you're out of time I can handle them on my side).

PR #259, opened 2017-01-19
Eliminate some warnings related to Move<Solution_>

yurloc:
I was reviewing https://issues.jboss.org/browse/PLANNER-731 and I found some rawtype Moves usages that could be easily fixed without side effects like changing method signatures and cascading to other parts of the code. Using the Solution_ parameter where possible also eliminated a number of unchecked warnings.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
@@ -35,7 +35,6 @@ import org.optaplanner.core.impl.heuristic.move.Move;
 import org.optaplanner.core.impl.heuristic.selector.entity.EntitySelector;
 import org.optaplanner.core.impl.phase.AbstractPhase;
 import org.optaplanner.core.impl.score.director.InnerScoreDirector;
-import org.optaplanner.core.impl.score.director.ScoreDirector;
 import org.optaplanner.core.impl.solver.recaller.BestSolutionRecaller;
 import org.optaplanner.core.impl.solver.scope.DefaultSolverScope;
 import org.optaplanner.core.impl.solver.termination.Termination;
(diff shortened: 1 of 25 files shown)


yurloc:
Hi @ge0ffrey, this is also related to the discussion that we had about generics. What do you think about this sort of fixes?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/SwapMove.java
```diff
@@ -151,7 +151,7 @@ public boolean equals(Object o) {
         if (this == o) {
             return true;
         } else if (o instanceof SwapMove) {
-            SwapMove other = (SwapMove) o;
+            SwapMove<?> other = (SwapMove) o;
```
ge0ffrey: This PR is a good idea.
For this line, can't we avoid the ? and simply do:

    SwapMove<Solution_> other = (SwapMove) o;

yurloc: We could do that but that would be an unchecked conversion which I wanted to avoid and the exact type of the move is not needed for the equality check.

ge0ffrey: Agreed, you're right.

optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
```diff
@@ -210,14 +209,11 @@ protected void restoreWorkingSolution(ExhaustiveSearchStepScope<Solution_> stepS
...
-        List<Move> restoreMoveList = new ArrayList<>(oldMoveList.size() + newMoveList.size());
+        List<Move<Solution_>> restoreMoveList = new ArrayList<>(oldMoveList.size() + newMoveList.size());
         restoreMoveList.addAll(oldMoveList);
         Collections.reverse(newMoveList);
         restoreMoveList.addAll(newMoveList);
-        ScoreDirector scoreDirector = phaseScope.getScoreDirector();
-        for (Move restoreMove : restoreMoveList) {
-            restoreMove.doMove(scoreDirector);
-        }
+        restoreMoveList.forEach(restoreMove -> restoreMove.doMove(phaseScope.getScoreDirector()));
```
ge0ffrey: Nitpick: this actually does the getScoreDiretor() call n times, instead of just once. Not that matters much because it's a simple getter that is probably inlined, but Cliff Click (the JVM hotspot compiler guy) argues that inlining is not always guaranteed.

I 'll merge as is anyway.

yurloc: Good catch! I didn't realize that.

PR #261, opened 2017-02-01
Add jboss-transaction-api_1.2_spec as replacement for excluded geroni…

psiroky:
…mo-jta_1.1_spec

Should be merged together with https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/378


mbiarnes:
+1

PR #264, opened 2017-02-14
Add direct link to starter issues

yurloc:


ge0ffrey:
Good idea, makes it easier for contributors indeed :)

PR #265, opened 2017-02-14
Make TestGen compatible with Coach Shuttle Gathering example

yurloc:
CSG uses shadow variables (custom and inverse relationship), which lead to updates with values that are not working facts (e.g. transportTimeToHub is an Integer, transferShuttleList is an ArrayList). TestGen now handles these dynamic non-fact values using the `TestGenInlineValue`.

Another anomaly to cope with was the `Coach.destination` property whose type is `BusHub` but the getter is a supertype (`StopOrHub`).

A new "reproducer" (reproducibility decider) for corrupted variable listeners was added. It is not fully compatible with the current pruning strategy but it works when only dropping the oldest updates. FIXME comment included.

TestGen score director's logging was enhanced so that each Move updates are now visible in full detail:
```
// do move
Updating variable H1.transferShuttleList[List]: [S1, S2, S3, S4, S5, S6] → [S1, S2, S3, S4, S5]
Updating variable S6.destination[StopOrHub]: H1 → B2
Updating variable B2.transferShuttleList[List]: [] → [S6]
// undo move
Updating variable B2.transferShuttleList[List]: [S6] → []
Updating variable S6.destination[StopOrHub]: B2 → H1
Updating variable H1.transferShuttleList[List]: [S1, S2, S3, S4, S5] → [S1, S2, S3, S4, S5, S6]
Move index (2), score (0hard/-412040soft), accepted (true), move (S6 {H1 -> B2}).
```

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/TestGenDroolsScoreDirector.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/TestGenDroolsScoreDirector.java
@@ -93,6 +93,27 @@ public class TestGenDroolsScoreDirector<Solution_> extends DroolsScoreDirector<S
     }
 
     @Override
+    public void assertShadowVariablesAreNotStale(Score expectedWorkingScore, Object completedAction) {
+        try {
+            super.assertShadowVariablesAreNotStale(expectedWorkingScore, completedAction);
+        } catch (IllegalStateException e) {
+            // catch stale shadow variable exception and create a minimal reproducing test
+            // TODO check it's really stale shadow variable?
+//            TestGenCorruptedScoreReproducer reproducer = new TestGenCorruptedScoreReproducer(e.getMessage(), this);
+//            TestGenKieSessionJournal minJournal = TestGenerator.minimize(journal, reproducer);
+            try {
+//                minJournal.replay(createKieSession());
+//                throw new IllegalStateException();
+            } catch (TestGenCorruptedScoreException tgcse) {
+                writer.setCorruptedScoreException(tgcse);
+            }
+//            writer.print(minJournal, testFile);
+            writer.print(journal, testFile);
+            throw wrapOriginalException(e);
+        }
+    }
+
+    @Override
     public void assertWorkingScoreFromScratch(Score workingScore, Object completedAction) {
         try {
             super.assertWorkingScoreFromScratch(workingScore, completedAction);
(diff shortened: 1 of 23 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/location/RoadLocationArc.java
```diff
@@ -67,4 +67,12 @@ public void setShuttleDuration(int shuttleDuration) {
     // Complex methods
     // ************************************************************************
 
+    @Override
+    public String toString() {
+        return "Road arc [coach: "
```
ge0ffrey: Can we change this to be in the same style as the rest of the domain classes toString() methods?

yurloc: For example? I checked some domain classes and they usually return name but RoadLocationArc has no name. I couldn't find a pattern in the rest of domain classes.

ge0ffrey: Looking at Period.toString() and Machine.toString() I notice there is no consistency really currently (there should be though). I am not a fan of toStrings() that mention all their properties - they are the main cause of unreadble (too verbose) logging. But I agree in this case we don't have much choice.

I am ok with merging this as is once jenkins is green. I 'll go through all toStrings() at some point.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/TestGenDroolsScoreDirector.java
```diff
@@ -93,6 +97,37 @@ public Score calculateScore() {
...
+            // catch corrupted VariableListener exception and create a minimal reproducing test
+            if (e.getMessage().startsWith("Impossible")) {
+                TestGenCorruptedVariableListenerReproducer reproducer
+                        = new TestGenCorruptedVariableListenerReproducer(e.getMessage(), this);
+                // FIXME this is currently broken. The pruning needs to be smarter and not remove genuine variable
+                // updates that directly affect shadow variables in the last (corrupted) variable listeners update.
+                // If the genuine update is removed the shadow update obviously becomes inconsistent, which leads
+                // to a false positive and the journal no longer reproduces the original issue. This is the current
+                // state.
+                TestGenKieSessionJournal minJournal = TestGenerator.minimize(journal, reproducer);
```
baldimir: Please Jirka, create a JIRA for this if it doesn't exist, so we can plan fixing this sometimes in the future.

PR #268, opened 2017-02-16
PLANNER-695 Test coverage for Locator and Location Strategies

yurloc:
## Summary
I think all requirements are met except for testing `ImmutableLocationStrategy` with each `LocationStrategyType`. Instead of that it's tested with each out-of-the-box immutable type. Let me know if you think this needs to be extended.

I have implemented the check for `equals` and `hashCode` override for the Equality location strategy.

As part of the API review I have removed `<Solution_>` type parameter from `Locator` because it isn't used. This is a proposal and I am open to discussion.

## Notes on the test structure

Each location strategy except for Immutable is tested through a Locator instance. This approach also provides coverage for the strategy resolver.

IdOrNone, IdOrFail and None location strategy tests all have the same test methods, which I think makes it easy to compare how behavior differs between respective `LocationStrategyType`s on the `@PlanningSolution`.

Equality test is a bit different because it doesn't focus on `@PlanningId` but on `equals`. So the method names are different and some methods test completely different things like the `equals` implementation.

Immutability test is completely different. It focuses on all possible immutable types and since the expected behavior is always the same it could be easily parameterized.

On the other hand, IdOrNone, IdOrFail and None couldn't be parameterized easily because the behavior for each location strategy type is different.

## Questions
I've already made a comment on PLANNER-695. To conclude it, it's not clear if adding the same instance twice should pass or fail. More questions inline.


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/LocationStrategyResolver.java
```diff
@@ -88,6 +89,21 @@ public LocationStrategy determineLocationStrategy(Object object) {
...
+                        if (equals.getDeclaringClass().equals(Object.class)) {
+                            throw new IllegalArgumentException("The class (" + object.getClass().getSimpleName()
+                                    + ") doesn't override equals method, neither does any superclass.");
+                        }
+                        Method hashCode = object.getClass().getMethod("hashCode");
+                        if (hashCode.getDeclaringClass().equals(Object.class)) {
+                            throw new IllegalArgumentException("The class (" + object.getClass().getSimpleName()
+                                    + ") overrides equals but neither it nor any superclass overrides hashCode method."
+                            );
+                        }
```
yurloc: Do we also make sure that `hashCode` and `equals` are overridden in the same class? In that case, fail or only log a warning? For example, class overriding equals and relying on hashCode from superclass is weird but it could work because even if the subclass's equals is more specific, equal objects still would have the same hashCode. The other way around (hashCode in subclass) would likely result in equal objects with different hashCode, which breaks hashCode's contract.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/PlanningIdLocationStrategy.java
```diff
@@ -62,7 +62,7 @@ protected Object extractPlanningId(Object externalObject) {
         Object planningId = planningIdMemberAccessor.executeGetter(externalObject);
         if (planningId == null) {
             throw new IllegalArgumentException("The planningId (" + planningId
-                    + ") of the member (" + planningIdMemberAccessor + ") of the class (" + externalObject.getClass()
+                    + ") of the member (" + planningIdMemberAccessor
```
yurloc: `planningIdMemberAccessor.toString()` already contains the object's FQCN.

baldimir:
Test case from JIRA probably missing: Adding 2 different instances of the same class with the same PlanningId should fail fast.

yurloc:
I've added the test case for 2 different instances with same PlanningId. I also added this test case to the Equality strategy test, which now fails and raises the question if the same check should be made with Equality strategy. I understand the importance of checking that when same PlanningId is added twice the instances **must not be different**. What still isn't clear to me is whether it is allowed to add the same instance twice as well as two different instances which are equal. Should this be allowed or fail fast because there is no reason to add an instance twice?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/LocationStrategyResolver.java
```diff
@@ -104,8 +120,8 @@ public LocationStrategy determineLocationStrategy(Object object) {
             return null;
         }
         if (memberList.size() > 1) {
-            throw new IllegalStateException("The class (" + clazz
-                    + ") has " +  memberList.size() + " members (" + memberList + ") with a "
+            throw new IllegalArgumentException("The class (" + clazz
+                    + ") has " + memberList.size() + " members (" + memberList + ") with a "
```
yurloc: An object with two or more PlanningIds is an invalid argument regardless of the state (contents of the map).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/Locator.java
```diff
@@ -21,22 +21,20 @@
...
 import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
 import org.optaplanner.core.impl.score.director.ScoreDirector;
 
 /**
- * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  * @see PlanningId
  * @see ScoreDirector#locateWorkingObject(Object)
  */
-public class Locator<Solution_> {
+public class Locator {
```
yurloc: Please pay attention to this proposal and confirm if it is OK.

yurloc:
@ge0ffrey Ready for review. There are some unanswered questions in my previous comment. Also please pay attention to my inline comments.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/Locator.java
```diff
@@ -21,22 +21,20 @@
...
 
     private final LocationStrategyResolver locationStrategyResolver;
 
     private Map<Object, Object> idToWorkingObjectMap;
 
-    public Locator(SolutionDescriptor<Solution_> solutionDescriptor) {
+    public Locator(SolutionDescriptor<?> solutionDescriptor) {
```
ge0ffrey: Let's change this constructor signature to 
   Locator(LocationStrategyResolver locationStrategyResolver)
That way, we avoid the <?>, which is really better of as <Solution_>

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/Locator.java
ge0ffrey: thanks for pointing this out.

I am ok with it if we change the constructor to Locator(LocationStrategyResolver locationStrategyResolver) (see below)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/Locator.java
yurloc: On the other hand <?> is better than a raw type, don't you think?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/LocationStrategyResolver.java
ge0ffrey: Good question, but I think there might be valid cases indeed for which they are not overridden in the same class, as the example you gave. We can't afford any "false negatives", but we can afford a little as possible "false positives". So I 'd leave it as is. It might not fail fast on all negatives, but it never fails fast when it shouldn't.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/PlanningIdLocationStrategy.java
ge0ffrey: I don't think it does if the member is from a superclass.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/Locator.java
ge0ffrey: yes, but I dislike both:
- raw type gives intellij warning and can hide good compilations errors
- ? also doesn't give any guarantees that the generic type is consistent across all classes - and it can cause ackwards casts that cast Foo<?> to Foo<Solution_> if the Foo<?> leaks into other classes.

yurloc: The point is that this piece of code (Locator) doesn't work with Solution_ at all so the type parameter is redundant there. Also the solution descriptor is only needed to retrieve the location strategy resolver so there would be no point in casting SolutionDescriptor<?> to something else.

In fact we could replace the constructor parameter type from SolutionDescirptor to LocationStrategyResolver:

```java


    public Locator(LocationStrategyResolver locationStrategyResolver) {
        this.locationStrategyResolver = locationStrategyResolver;
    }
```

Which I think might actually be a good idea because the tests wouldn't need to construct the solution descriptor to set up the Locator.

yurloc: I apologize, I didn't read your proposal of the constructor change carefully enough and thought you wanted to change the parameter to `SolutionDescriptor` (rawtype). So my previous comment doesn't make sense because it proposes exactly what you intended. Please ignore it :-)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/Locator.java
yurloc: Done.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/locator/EqualsLocationStrategy.java
```diff
@@ -22,7 +22,11 @@
 
     @Override
     public void addWorkingObject(Map<Object, Object> idToWorkingObjectMap, Object workingObject) {
-        idToWorkingObjectMap.put(workingObject, workingObject);
+        Object oldAddedObject = idToWorkingObjectMap.put(workingObject, workingObject);
+        if (oldAddedObject != null) {
+            throw new IllegalStateException("An object that is equal to workingObject (" + workingObject
+                    + ") has already been added.");
+        }
```
yurloc: So I added this check so the methods behaves similar to the one of PlanningIdLocationStrategy. If it turns out this check is to strict it can be removed anytime. If there are no objections, the PR can be merged as is.

ge0ffrey: +1, but the error message should include the toString() of both the old and the new one. Adjusting that now.

yurloc: I think it is highly improbable there would be a difference in the two equal object that would affect toString result. But yes, an exception is an exceptional event and some extra information can't do any harm.

ge0ffrey:
Merged manually:
  http://github.com/droolsjbpm/optaplanner/compare/e9a676c...cd9d2a4
Made a few cosmetic changes, nothing important;

PR #271, opened 2017-02-20
PLANNER-632 ConstraintMatch and ConstraintMatchTotal should be able to return a Score instead of an int scoreLevel and weight

ge0ffrey:


ge0ffrey:
This PR has the side affect of changing a BigDecimalScore to start with ZERO instead of null.

PR #279, opened 2017-03-17
Add DefaultPartitionedSearchPhase test coverage

yurloc:
- Fixed a few typos in Javadoc.
- Fixed implementation of `PartitionChangeMove.createUndoMove()`. It now returns null instead of throwing `UnsupportedOperationException`.
- Added thread pool shutdown when PS phase ends.
- Improved `PartitionedSearchPhaseConfig` API. Now it's possible to set custom `SolutionPartitioner`.
- Added `@PlanningId` on `TestdataObject`.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
@@ -54,7 +54,7 @@ public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchP
     // Warning: all fields are null (and not defaulted) because they can be inherited
     // and also because the input config file should match the output config file
 
-    protected Class<SolutionPartitioner> solutionPartitionerClass = null;
+    protected Class<? extends SolutionPartitioner<?>> solutionPartitionerClass = null;
     @XStreamConverter(KeyAsElementMapConverter.class)
     protected Map<String, String> solutionPartitionerCustomProperties = null;
 
@@ -68,11 +68,11 @@ public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchP
     // Constructors and simple getters/setters
     // ************************************************************************
 
-    public Class<SolutionPartitioner> getSolutionPartitionerClass() {
+    public Class<? extends SolutionPartitioner<?>> getSolutionPartitionerClass() {
         return solutionPartitionerClass;
     }
 
-    public void setSolutionPartitionerClass(Class<SolutionPartitioner> solutionPartitionerClass) {
+    public void setSolutionPartitionerClass(Class<? extends SolutionPartitioner<?>> solutionPartitionerClass) {
         this.solutionPartitionerClass = solutionPartitionerClass;
     }
 
@@ -163,7 +163,7 @@ public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchP
 
     private SolutionPartitioner buildSolutionPartitioner() {
         if (solutionPartitionerClass != null) {
-            SolutionPartitioner solutionPartitioner = ConfigUtils.newInstance(this,
+            SolutionPartitioner<?> solutionPartitioner = ConfigUtils.newInstance(this,
                     "solutionPartitionerClass", solutionPartitionerClass);
             ConfigUtils.applyCustomProperties(solutionPartitioner, "solutionPartitionerClass",
                     solutionPartitionerCustomProperties);
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
@@ -19,10 +19,7 @@ package org.optaplanner.core.config.partitionedsearch;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import com.thoughtworks.xstream.annotations.XStreamConverter;
@@ -143,7 +140,7 @@ public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchP
                 phaseIndex, solverConfigPolicy.getLogIndentation(), bestSolutionRecaller,
                 buildPhaseTermination(phaseConfigPolicy, solverTermination));
         phase.setSolutionPartitioner(buildSolutionPartitioner());
-        phase.setThreadPoolExecutor(buildThreadPoolExecutor());
+        phase.setThreadFactory(buildThreadFactory());
         phase.setRunnablePartThreadLimit(resolvedActiveThreadCount());
         List<PhaseConfig> phaseConfigList_ = phaseConfigList;
         if (ConfigUtils.isEmptyCollection(phaseConfigList_)) {
@@ -182,18 +179,12 @@ public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchP
         }
     }
 
-    private ThreadPoolExecutor buildThreadPoolExecutor() {
-        ThreadFactory threadFactory;
+    private ThreadFactory buildThreadFactory() {
         if (threadFactoryClass != null) {
-            threadFactory = ConfigUtils.newInstance(this, "threadFactoryClass", threadFactoryClass);
+            return ConfigUtils.newInstance(this, "threadFactoryClass", threadFactoryClass);
         } else {
-            threadFactory = new DefaultSolverThreadFactory("PartThread");
+            return new DefaultSolverThreadFactory("PartThread");
         }
-        // Based on Executors.newCachedThreadPool(...)
-        return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
-                60L, TimeUnit.SECONDS,
-                new SynchronousQueue<>(),
-                threadFactory);
     }
 
     private Integer resolvedActiveThreadCount() {
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
@@ -102,7 +102,7 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
         int partCount = partList.size();
         phaseScope.setPartCount(partCount);
         phaseStarted(phaseScope);
-        ThreadPoolExecutor threadPoolExecutor = newThreadPoolExecutor();
+        ThreadPoolExecutor threadPoolExecutor = createThreadPoolExecutor();
         if (threadPoolExecutor.getMaximumPoolSize() < partCount) {
             throw new IllegalStateException(
                     "The threadPoolExecutor's maximumPoolSize (" + threadPoolExecutor.getMaximumPoolSize()
@@ -133,7 +133,8 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
                         partitionSolver.solve(part);
                         partitionQueue.addFinish(partIndex);
                     } catch (Throwable throwable) {
-                        // TODO is it a good idea to catch Throwables including OOME for example?
+                        // Any Exception or even Error that happens here (on a partition thread) must be stored
+                        // in the partitionQueue in order to be propagated to the solver thread
                         partitionQueue.addExceptionThrown(partIndex, throwable);
                     }
                 });
@@ -149,27 +150,15 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
                 stepEnded(stepScope);
                 phaseScope.setLastCompletedStepScope(stepScope);
             }
-        } catch (RuntimeException e) {
+        } finally {
             // If a partition thread throws an Exception, it is propagated here
             // but the other partition thread won't finish any time soon, so we need to terminate them
             childThreadPlumbingTermination.terminateChildren();
-            throw e;
-        } finally {
-            List<Runnable> partitionSolvers = threadPoolExecutor.shutdownNow();
-            logger.trace("{}         Awaiting termination of {} partition solvers.",
-                    logIndentation, partitionSolvers.size());
+            threadPoolExecutor.shutdownNow();
             try {
-                long start = 0;
-                if (logger.isTraceEnabled()) {
-                    start = System.nanoTime();
-                }
-                if (threadPoolExecutor.awaitTermination(1, TimeUnit.MILLISECONDS)) {
-                    if (logger.isTraceEnabled()) {
-                        logger.trace("{}         Thread pool shutdown took {} nanos.",
-                                logIndentation, System.nanoTime() - start);
-                    }
-                } else {
-                    logger.warn("{}Thread pool didn't terminate within the timeout.", logIndentation);
+                if (!threadPoolExecutor.awaitTermination(1, TimeUnit.SECONDS)) {
+                    logger.warn("{}Partitioned Search threadPoolExecutor didn't terminate within timeout (1 second).",
+                            logIndentation);
                 }
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
@@ -179,7 +168,7 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
         phaseEnded(phaseScope);
     }
 
-    private ThreadPoolExecutor newThreadPoolExecutor() {
+    private ThreadPoolExecutor createThreadPoolExecutor() {
         // Based on Executors.newCachedThreadPool(...)
         return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                 60L, TimeUnit.SECONDS,
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
@@ -20,7 +20,10 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import org.optaplanner.core.api.domain.solution.PlanningSolution;
 import org.optaplanner.core.config.heuristic.policy.HeuristicConfigPolicy;
@@ -51,7 +54,7 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
         implements PartitionedSearchPhase<Solution_>, PartitionedSearchPhaseLifecycleListener<Solution_> {
 
     protected SolutionPartitioner<Solution_> solutionPartitioner;
-    protected ThreadPoolExecutor threadPoolExecutor;
+    protected ThreadFactory threadFactory;
     protected Integer runnablePartThreadLimit;
 
     protected List<PhaseConfig> phaseConfigList;
@@ -66,8 +69,8 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
         this.solutionPartitioner = solutionPartitioner;
     }
 
-    public void setThreadPoolExecutor(ThreadPoolExecutor threadPoolExecutor) {
-        this.threadPoolExecutor = threadPoolExecutor;
+    public void setThreadFactory(ThreadFactory threadFactory) {
+        this.threadFactory = threadFactory;
     }
 
     public void setRunnablePartThreadLimit(Integer runnablePartThreadLimit) {
@@ -99,6 +102,7 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
         int partCount = partList.size();
         phaseScope.setPartCount(partCount);
         phaseStarted(phaseScope);
+        ThreadPoolExecutor threadPoolExecutor = newThreadPoolExecutor();
         if (threadPoolExecutor.getMaximumPoolSize() < partCount) {
             throw new IllegalStateException(
                     "The threadPoolExecutor's maximumPoolSize (" + threadPoolExecutor.getMaximumPoolSize()
@@ -109,7 +113,8 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
         }
         ChildThreadPlumbingTermination childThreadPlumbingTermination = new ChildThreadPlumbingTermination();
         PartitionQueue<Solution_> partitionQueue = new PartitionQueue<>(partCount);
-        Semaphore runnablePartThreadSemaphore = runnablePartThreadLimit == null ? null : new Semaphore(runnablePartThreadLimit);
+        Semaphore runnablePartThreadSemaphore
+                = runnablePartThreadLimit == null ? null : new Semaphore(runnablePartThreadLimit);
         try {
             for (ListIterator<Solution_> it = partList.listIterator(); it.hasNext();) {
                 int partIndex = it.nextIndex();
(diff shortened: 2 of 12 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -94,11 +93,11 @@ public String getPhaseTypeString() {
     @Override
     public void solve(DefaultSolverScope<Solution_> solverScope) {
         PartitionedSearchPhaseScope<Solution_> phaseScope = new PartitionedSearchPhaseScope<>(solverScope);
-        phaseStarted(phaseScope);
         List<Solution_> partList = solutionPartitioner.splitWorkingSolution(
                 solverScope.getScoreDirector(), runnablePartThreadLimit);
         int partCount = partList.size();
         phaseScope.setPartCount(partCount);
+        phaseStarted(phaseScope);
```
yurloc: Thanks to this change, listeners will be able to read the `partCount` when the phase starts.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
```diff
@@ -85,13 +86,13 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
...
-    public boolean isMoveDoable(ScoreDirector<Solution_>  scoreDirector) {
+    public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
         return true;
     }
 
     @Override
-    public PartitionChangeMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
-        throw new UnsupportedOperationException();
+    public NoChangeMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
+        return new NoChangeMove<>();
```
yurloc: Because of "PLANNER-611 Move.doMove() must return undo move" this method is now called in `AbstractMove.doMove()` so it must not throw the exception. Returning NoChangeMove seems to me like a good replacement.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/scope/DefaultSolverScope.java
```diff
@@ -256,12 +256,12 @@ public void initializeYielding() {
...
      * Similar to {@link Thread#yield()}, but allows capping the number of active solver threads
      * at less than the CPU processor count, so other threads (for example servlet threads that handle REST calls)
-     * and other processes (such as SSH) have access to uncontested CPU's and don't suffer any latency.
+     * and other processes (such as SSH) have access to uncontested CPUs and don't suffer any latency.
      * <p>
      * Needs to be called <b>before</b> {@link Termination#isPhaseTerminated(AbstractPhaseScope)},
      * so the decision to start a new iteration is after any yield waiting time has been consumed
      * (so {@link Solver#terminateEarly()} reacts immediately).
-     * Furthermore, this method will
+     * Furthermore, this method will TODO
```
yurloc: Seems like the sentence is unfinished.

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueueTest.java
```diff
@@ -31,6 +33,16 @@
...
 
+    @After
+    public void tearDown() {
+        executorService.shutdown();
+        try {
+            executorService.awaitTermination(1, TimeUnit.SECONDS);
+        } catch (InterruptedException ex) {
+            throw new RuntimeException(ex);
+        }
+    }
```
yurloc: Looks like it is a good idea to clean up threads after the test. For example PIT says:
```
More threads at end of test (6) addExceptionWithNonEmptyQueue(org.optaplanner.core.impl.partitionedsearch.queue.PartitionQueueTest) than start. (4)
```
IIUC this means that every time the test runs it leaves 2 additional threads (somewhere in memory?). I don't know how much resources these threads can consume but it's nicer to clean them up.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
ge0ffrey: Ow damn, I broke Partitioned Search without realizing it because we clearly have no test coverage for it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -146,6 +145,7 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
         } finally {
             // If a partition thread throws an Exception, it is propagated here
             // but the other partition thread won't finish any time soon, so we need to terminate them
+            // TODO move to catch block
```
ge0ffrey: What do you mean? Its' in a finally, so it always terminates the children

ge0ffrey: I remember doing manual tests around this, as not cleaning up threads can bring down a server to it's knees.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -48,7 +47,7 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solution_>
-        implements PartitionedSearchPhase<Solution_>, PartitionedSearchPhaseLifecycleListener<Solution_> {
+        implements PartitionedSearchPhase<Solution_> {
```
ge0ffrey: If we remove PartitionedSearchPhaseLifecycleListener, then it still needs to implement SolverLifecycleListener.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
ge0ffrey: We can't return NoChangeMove as that will not cause fail-fasts (but silent score corruption when used). Either we implement it or return null for now. Let's return null for now.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/scope/DefaultSolverScope.java
ge0ffrey: Absolutely no idea any more what I was thinking there. Let's just remove that line.

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueueTest.java
```diff
@@ -31,6 +33,16 @@
 
     private ExecutorService executorService = Executors.newFixedThreadPool(2);
 
+    @After
+    public void tearDown() {
+        executorService.shutdown();
+        try {
+            executorService.awaitTermination(1, TimeUnit.SECONDS);
+        } catch (InterruptedException ex) {
+            throw new RuntimeException(ex);
```
ge0ffrey: InterruptedException should never be handled this way because it eats the interrupted flag. Look for other catch InterruptedException's in the code on how to handle it properly.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/partitionedsearch/TestdataPartitionSearchEntity.java
```diff
@@ -0,0 +1,61 @@
...
+package org.optaplanner.core.impl.testdata.domain.partitionedsearch;
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+
+@PlanningEntity
+public class TestdataPartitionSearchEntity extends TestdataPartitionSearchObject {
```
ge0ffrey: Why do we need separate Testdata objects for this? I think we can reuse TestdataSolution, if we're willing to add PlanningId on TestdataObject.code.
I'd rather create a non-PlanningId testdata object to test the non-planning id code, that way we can reuse the entire set of testdata objecst (normal, extended, 3th party, etc) directly in Partitioned Search tests. And later in multithreaded-incremental score calculation tests.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/partitionedsearch/TestdataPartitionSearchSolution.java
```diff
@@ -0,0 +1,77 @@
...
+import org.optaplanner.core.api.domain.lookup.LookUpStrategyType;
+import org.optaplanner.core.api.domain.solution.PlanningEntityCollectionProperty;
+import org.optaplanner.core.api.domain.solution.PlanningScore;
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.solution.drools.ProblemFactCollectionProperty;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+@PlanningSolution(lookUpStrategyType = LookUpStrategyType.EQUALITY)
```
ge0ffrey: I doubt anyone will actually use equality. PlanningId should be the primary way to test.

ge0ffrey: I've talked to the field a few months ago, specifically about if they have a unique id per entity. Turns out they all have that.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/partitionedsearch/TestdataPartitionSearchEntity.java
yurloc: If you think it is a good idea, I'll go ahead and add PlanningId there. I'll just need to fix 1 test that fails after that change. Using the separate domain was just a quick workaround for that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,43 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
                 stepEnded(stepScope);
                 phaseScope.setLastCompletedStepScope(stepScope);
             }
-        } finally {
+        } catch (RuntimeException e) {
             // If a partition thread throws an Exception, it is propagated here
             // but the other partition thread won't finish any time soon, so we need to terminate them
             childThreadPlumbingTermination.terminateChildren();
+        } finally {
```
yurloc: Based on the comment I understand that it's only needed to manually terminate children if one of them threw exception. If that doesn't happen it should be guaranteed that all children have terminated simply by finishing iteration over the PartitionQueue. Terminating children in finally had one not-so-nice side-effect of printing the _"Terminating child solvers early."_ even if that's not the case.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,43 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
+            List<Runnable> partitionSolvers = threadPoolExecutor.shutdownNow();
```
yurloc: This should actually handle the thread clean-up you mentioned. This is even covered by the new test.

Sorry for the messy code below but that's the cost of being super-defensive.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -48,10 +50,10 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solution_>
-        implements PartitionedSearchPhase<Solution_>, PartitionedSearchPhaseLifecycleListener<Solution_> {
+        implements PartitionedSearchPhase<Solution_> {
 
     protected SolutionPartitioner<Solution_> solutionPartitioner;
-    protected ThreadPoolExecutor threadPoolExecutor;
+    protected ThreadFactory threadFactory;
```
yurloc: The phase needs to have the thread factory if `solver.solve()` should be repeatable. At the end of the phase, thread pool is shut down and so a new thread pool has to be created at the phase start.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new IllegalStateException("Thread pool shutdown was interrupted.", e);
+            }
         }
         phaseEnded(phaseScope);
     }
 
+    private ThreadPoolExecutor newThreadPoolExecutor() {
```
ge0ffrey: rename to createThreadPoolExecutor()

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -128,6 +132,7 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
                         partitionSolver.solve(part);
                         partitionQueue.addFinish(partIndex);
                     } catch (Throwable throwable) {
+                        // TODO is it a good idea to catch Throwables including OOME for example?
```
ge0ffrey: Remove todo, but clarify with a comment:
for example: // Transfer from the part thread to the solver thread, so the solver thread throws an exception

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
                 stepEnded(stepScope);
                 phaseScope.setLastCompletedStepScope(stepScope);
             }
-        } finally {
+        } catch (RuntimeException e) {
```
ge0ffrey: terminateChildren() goes into finally

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
+            logger.trace("{}         Awaiting termination of {} partition solvers.",
+                    logIndentation, partitionSolvers.size());
+            try {
+                long start = 0;
+                if (logger.isTraceEnabled()) {
+                    start = System.nanoTime();
+                }
+                if (threadPoolExecutor.awaitTermination(1, TimeUnit.MILLISECONDS)) {
+                    if (logger.isTraceEnabled()) {
+                        logger.trace("{}         Thread pool shutdown took {} nanos.",
```
ge0ffrey: this log message can be removed, so also long start, and nanotime() etc

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
+        } finally {
+            List<Runnable> partitionSolvers = threadPoolExecutor.shutdownNow();
+            logger.trace("{}         Awaiting termination of {} partition solvers.",
+                    logIndentation, partitionSolvers.size());
+            try {
+                long start = 0;
+                if (logger.isTraceEnabled()) {
+                    start = System.nanoTime();
+                }
+                if (threadPoolExecutor.awaitTermination(1, TimeUnit.MILLISECONDS)) {
```
ge0ffrey: Let's go for 1 second for now.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
             }
-        } finally {
+        } catch (RuntimeException e) {
             // If a partition thread throws an Exception, it is propagated here
             // but the other partition thread won't finish any time soon, so we need to terminate them
             childThreadPlumbingTermination.terminateChildren();
+            throw e;
+        } finally {
+            List<Runnable> partitionSolvers = threadPoolExecutor.shutdownNow();
+            logger.trace("{}         Awaiting termination of {} partition solvers.",
```
ge0ffrey: invalid because it's only the non started partitioned solvers, so remove this log message

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
+                if (logger.isTraceEnabled()) {
+                    start = System.nanoTime();
+                }
+                if (threadPoolExecutor.awaitTermination(1, TimeUnit.MILLISECONDS)) {
+                    if (logger.isTraceEnabled()) {
+                        logger.trace("{}         Thread pool shutdown took {} nanos.",
+                                logIndentation, System.nanoTime() - start);
+                    }
+                } else {
+                    logger.warn("{}Thread pool didn't terminate within the timeout.", logIndentation);
```
ge0ffrey: log message: "{}Partitioned Search thread pool shutdown failed with timeout (1 second)."

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -143,14 +148,44 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new IllegalStateException("Thread pool shutdown was interrupted.", e);
```
ge0ffrey: Maybe do this:

boolean a =Thread.currentThread().isInterrupted();
try {
  awaitTermination();
catch(InterruptedException) {
  if (!a) {
            Thread.currentThread().interrupt();
             throw new IllegalStateException("Thread pool shutdown was interrupted.", e);
  }

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -48,10 +50,10 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solution_>
-        implements PartitionedSearchPhase<Solution_>, PartitionedSearchPhaseLifecycleListener<Solution_> {
+        implements PartitionedSearchPhase<Solution_> {
```
ge0ffrey: Let's keep PartitionedSearchPhaseLifecycleListener for now, due to consistency across algorithms LocalSearchPhase, ConstructionHeuristicsPhase, ...

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
ge0ffrey: Lets add a comment above: // TODO not yet implemented, returns null to fail fast if it is used

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseTest.java
```diff
@@ -0,0 +1,127 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.optaplanner.core.impl.partitionedsearch;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadMXBean;
```
ge0ffrey: Is this JMX code portable?

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.optaplanner.core.impl.solver.thread.DefaultSolverThreadFactory;
+import org.optaplanner.core.impl.testdata.domain.partitionedsearch.TestdataPartitionSearchEntity;
+import org.optaplanner.core.impl.testdata.domain.partitionedsearch.TestdataPartitionSearchSolution;
+import org.optaplanner.core.impl.testdata.domain.partitionedsearch.TestdataPartitionedSearchValue;
+import org.optaplanner.core.impl.testdata.util.PlannerTestUtils;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+public class DefaultPartitionedSearchPhaseTest {
```
ge0ffrey: This is an integration test - replace it with CloudBalancingPartitionedSearchTest that mimics CloudBalancingExhaustiveSearchTest and CloudBalancingConstructionHeuristicTest

ge0ffrey: Might be worth keeping this test, to simulate what happens:
- if a part thread throws an exception (divide by zero). Does the solver thread throw an exception?
- if terminateEarly() is called after 1 seconds from another thread - with no termination set (=runs forever otherwise). Are all threads cleaned up?
- ...

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseTest.java
```diff
@@ -0,0 +1,127 @@
...
+            @Override
+            public void phaseStarted(AbstractPhaseScope<TestdataPartitionSearchSolution> phaseScope) {
+                // TODO? should be possible to phase.addPartitionPhaseLifecycleListener() to avoid the cast?
+                assertEquals(Integer.valueOf(1), ((PartitionedSearchPhaseScope) phaseScope).getPartCount());
+            }
+        });
+
+        scope.startingNow();
+        phase.solve(scope);
+        phase.solve(scope);
```
ge0ffrey: duplicate

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionQueueTest.java
```diff
@@ -20,16 +20,29 @@
...
 
 public class PartitionQueueTest {
 
-    private ExecutorService executorService = Executors.newFixedThreadPool(2);
+    private static final Logger logger = LoggerFactory.getLogger(PartitionQueueTest.class);
+    private final ExecutorService executorService = Executors.newFixedThreadPool(2);
+
+    @After
+    public void tearDown() throws InterruptedException {
+        executorService.shutdown();
```
ge0ffrey: call shutdowNow()

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/partitionedsearch/TestdataPartitionSearchEntity.java
ge0ffrey: So this class can be removed

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
yurloc: Sorry, this no longer makes sense to me. Here we are on the main thread. I can't see any way how it could be interrupted and THEN start to `awaitTermination()` (while being interrupted). Notice that if a partition thread is interrupted it won't interrupt the solverThread. I don't even have an idea how to test such code. I'll think about it but I suggest this PR be merged without the `isInterrupted()` check.

yurloc:
TODO:

- [ ] move the test setup code to Config API
- [ ] change the test from integration style to unit style and test various scenarios separately
- [ ] decide if it's necessary to check interrupted flag BEFORE beginning `threadPoolExecutor.awaitTermination()`

yurloc:
I consider this ready to be merged. There is a couple of things to do that I think are lower priority and can be done later:

- implement `PartitionChangeMove.toString()`
- add more tests for threads (exception throwing, interrupting, runnable thread count limit)

Apart from the test, this RP contains a few fixes so I'd prefer to merge it without the lower priority tasks. The fixes include:

- removed `UnsupportedOperationException` thrown by `PartitionChangeMove.createUndoMove()`
- thread pool shutdown
- possibility to set `SolutionPartitioner` via Config API

I'd prefer to squash commits after the PR is approved.

optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
```diff
@@ -71,11 +68,11 @@
     // Constructors and simple getters/setters
     // ************************************************************************
 
-    public Class<SolutionPartitioner> getSolutionPartitionerClass() {
+    public Class<? extends SolutionPartitioner<?>> getSolutionPartitionerClass() {
         return solutionPartitionerClass;
     }
 
-    public void setSolutionPartitionerClass(Class<SolutionPartitioner> solutionPartitionerClass) {
+    public void setSolutionPartitionerClass(Class<? extends SolutionPartitioner<?>> solutionPartitionerClass) {
```
yurloc: This is necessary in order to be able to use an implementation of the `SolutionPartitioner` interface.

ge0ffrey:
I 'll review and merge (and squash during the merge) later today. No need to do the squashing at your side (unless you really want to:)

PR #280, opened 2017-03-21
Updated all documentation chapter folder names, chapter names, and se…

gsheldon:
…ction file names to camel case.

Hi Geoffrey, 

This is for all docs folders and file names from Chapter-Chapter_name/Chapter-Chapter_name.adoc to ChapterName/ChapterName-chapter.adoc, as discussed on the call yesterday.

I have checked this using asciidoctor and found no issues. 

Thanks,
Gemma.


ge0ffrey:
@gsheldon Merged, thanks :)
Note: I presume you also always run a "mvn clean install" in "optaplanner-docs" directory and look at "target/generated-docs/" when submitting a PR.

ge0ffrey:
Optaplanner-website's slidedeck mimics these directories.
  https://github.com/kiegroup/optaplanner-website/tree/master/learn/slides/optaplanner-presentation
Now we can actually use the exact same directory names, I 'll make those changes.

ge0ffrey:
@gsheldon We have a directory called "QuickStart" but the section is named differently. This needs to be consistent. I think we need to go back to the name Quick Start, as that's what developers are looking for. The term "quick start" is also referenced on optaplanner.org frontpage although we can adjust that. Wdyt?

optaplanner-docs/src/main/asciidoc/index.adoc
```diff
@@ -1,7 +1,7 @@
 = OptaPlanner User Guide
 The OptaPlanner Team <https://www.optaplanner.org/community/team.html>
 :doctype: book
-:title-logo-image: image:Chapter-Planner_introduction/optaPlannerLogo.png[align="center"]
+:title-logo-image: image:PlannerIntroduction/optaPlannerLogo.png[align="center"]
```
ge0ffrey: Because the images dirs haven't changed yet, this image isn't found and a warning is shown in the "mvn clean install" log.
I've fixed this by reverting this line for now. The image dirs need to change too, of course, but let's first see if we can go to relative images.

PR #281, opened 2017-03-21
Eliminate raw types in MoveListFactory and MoveIteratorFactory interfaces

yurloc:
I'm not sure if this has any benefit for the user. The "nice" thing is that you have to declare the move type as the factory type parameter. This way it is guaranteed that `MoveIteratorFactory`'s `createOriginalMoveIterator()` and `createRandomMoveIterator()` both return the same type of moves. And there is less rawtypes warnings. 😄 On the other hand, any factory implementations need to be updated.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/factory/MoveIteratorFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/factory/MoveIteratorFactory.java
@@ -49,7 +49,7 @@ public interface MoveIteratorFactory<Solution_> {
      * @throws UnsupportedOperationException if only {@link #createRandomMoveIterator(ScoreDirector, Random)} is
      * supported
      */
-    Iterator<Move> createOriginalMoveIterator(ScoreDirector<Solution_> scoreDirector);
+    Iterator<? extends Move<Solution_>> createOriginalMoveIterator(ScoreDirector<Solution_> scoreDirector);
 
     /**
      * When it is called depends on the configured {@link SelectionCacheType}.
@@ -60,6 +60,7 @@ public interface MoveIteratorFactory<Solution_> {
      * @return never null, an {@link Iterator} that is allowed (or even presumed) to be never ending
      * @throws UnsupportedOperationException if only {@link #createOriginalMoveIterator(ScoreDirector)} is supported
      */
-    Iterator<Move> createRandomMoveIterator(ScoreDirector<Solution_> scoreDirector, Random workingRandom);
+    Iterator<? extends Move<Solution_>> createRandomMoveIterator(
+            ScoreDirector<Solution_> scoreDirector, Random workingRandom);
 
 }
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/investment/solver/move/factory/InvestmentQuantityTransferMoveIteratorFactory.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/investment/solver/move/factory/InvestmentQuantityTransferMoveIteratorFactory.java
@@ -43,12 +43,14 @@ public class InvestmentQuantityTransferMoveIteratorFactory implements MoveIterat
     }
 
     @Override
-    public Iterator<Move> createOriginalMoveIterator(ScoreDirector<InvestmentSolution> scoreDirector) {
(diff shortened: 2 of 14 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/factory/MoveIteratorFactory.java
```diff
@@ -49,7 +50,7 @@
      * @throws UnsupportedOperationException if only {@link #createRandomMoveIterator(ScoreDirector, Random)} is
      * supported
      */
-    Iterator<Move> createOriginalMoveIterator(ScoreDirector<Solution_> scoreDirector);
+    Iterator<Move_> createOriginalMoveIterator(ScoreDirector<Solution_> scoreDirector);
```
ge0ffrey: I was expecting this to change into
`Iterator<? extends Move<Solution_>> createOriginalMoveIterator(...)` to avoid needing an extra type on the interface (which is a pain).

Would that work?

yurloc: Yes, that works too. I started with this approach and it's covered by the first two commits. Then I realized it's possible to declare the move type like this. What I consider nice about this is that when you declare
```java
class MyMoveFactory implements MoveIteratorFactory<MySolution, MyMove> {}
```
and then let your IDE generate stubs implementing the interface, you get
```java
public Iterator<MyMove> createOriginalMoveIterator(ScoreDirector<MySolution> scoreDirector) {}
```
instead of
```java
public Iterator<? extends Move<MySolution>> createOriginalMoveIterator(ScoreDirector<MySolution> scoreDirector) {}
```
and I like the former better. But I understand why you consider the additional type parameter a pain so I'll remove commits 3 and 4.

ge0ffrey: Thanks Jiri.
It's a nice experiment, but the extra generic type isn't worth it's weight in complexity (and it also breaks backw. compatbility on the side although it's not public api so it doesn't factor into the decision much).
+1 to remove commit 3 & 4

yurloc: Notice that even just with commits 1 & 2 "backward compatibility" is broken, but not in all cases. If the implementing (overriding) method returns `List<MyMove>` and `MyMove extends AbstractMove<MySolution>` then the factory source will compile. If `MyMove extends AbstractMove` without the `Solution_` parameter then it will not compile. And if it returns `List<? extends Move>`, which was the return type declared by the interface before this change, then it won't compile either.

I'm going to update the upgrade recipe with this instructions.

optaplanner-examples/src/main/java/org/optaplanner/examples/investment/solver/move/factory/InvestmentQuantityTransferMoveIteratorFactory.java
```diff
@@ -88,7 +90,7 @@ public boolean hasNext() {
         }
 
         @Override
-        public Move next() {
+        public Move<InvestmentSolution> next() {
```
ge0ffrey: I think we can have this return InvestmentQuantityTransferMove instead of Move<InvestmentSolution>

ge0ffrey:
This does break backwards compatibility (on non public API's, but several users have custom move factories). So we need to add it in the upgrade recipe.

I still think we need to do it though. Better now, before 7.0.0.CR1.

optaplanner-examples/src/main/java/org/optaplanner/examples/investment/solver/move/factory/InvestmentQuantityTransferMoveIteratorFactory.java
yurloc: Yes we can. There is just a minor side effect with "Exporting non-public type through public API". So I'll just make the iterator class public (because it will be effectively public anyway).

yurloc:
No, the examples don't compile without adjusting. So yes, this is a backward compatibility issue. A minor one, though, as long as the `Move*Factory` interfaces are considered an internal API.

yurloc:
Upgrade recipe PR open: kiegroup/optaplanner-website#70.

PR #282, opened 2017-03-23
Updated with proposed changed levels to produce modularized sections.…

gsheldon:
… I will submit a separate PR with separate section files.

The second PR will have section levels modified again so that it fits with the level offset +1 and appears properly in the TOC.


ge0ffrey:
@gsheldon Thanks for making separate PR's between the restructuring and the content changes. Better one too many then one too few.
But let's make an exception for this particular types of changes (the "=" change due to splitting up the chapter files in sections), as this is just red tape for you and second pair of eyeballs won't have a different opinion. Feel free to always do those in the same PR.

I 'll close this PR and merge the one with has both commits. Thanks :)

PR #283, opened 2017-03-23
Quick start chapter topic files

gsheldon:
Includes changes to section levels == so that they appear in the right level when built.


ge0ffrey:
Merged directly, see last comment on https://github.com/kiegroup/optaplanner/pull/282.

PR #286, opened 2017-03-30
Review chapter use cases

gsheldon:
I think these are the only changes I would suggest for this chapter. I will send a modularization pull request separately, but not today.

Thanks,

Gemma.


optaplanner-docs/src/main/asciidoc/UseCasesAndExamples/UseCasesAndExamples-chapter.adoc
```diff
@@ -244,34 +242,35 @@ These realistic competitions provide an objective comparison of Planner with com
...
 
 Constraints:
 
 * Use a chessboard of _n_ columns and _n_ rows.
 * Place _n_ queens on the chessboard.
-* No 2 queens can attack each other. A queen can attack any other queen on the same horizontal, vertical or diagonal line.
+* No two queens can attack each other. A queen can attack any other queen on the same horizontal, vertical or diagonal line.
 
-This documentation heavily uses the 4 queens puzzle as the primary example.
+This documentation heavily uses the four queens puzzle as the primary example.
```
ge0ffrey: Yes, you're right with this change (as wikipedia says "eight queens puzzle" and not "8 queens puzzle").
In the entire manual, I regularly use the phrase "In 8 queens, <statement>" or "in the cloud balancing example, <statement>" etc. Those will need to be consistently fixed. Some proposals:
- In the eight queens example, <statement>
- In the eight queens puzzle, <statement>
- In eight queens, <statement>
Of course, it's often "four" or "n" instead of "eight".
I am fine with whatever proposal you prefer, as long as it's consistently used in the manual.

Because it's easy to find these with "find in path" (which finds all occurrences of a text in all files in a folder), I suggest to fix these first as a single PR, before reviewing the other chapters.

@gsheldon Wdyt?

optaplanner-docs/src/main/asciidoc/UseCasesAndExamples/UseCasesAndExamples-chapter.adoc
```diff
@@ -11,9 +11,17 @@
 [[examplesOverview]]
 == Examples Overview
 
-Planner has several examples.
-In this manual we explain mainly using the n queens example and cloud balancing example.
-So it's advisable to read at least those sections.
+Planner has several examples. In this manual we explain mainly using the _n_ queens example and cloud balancing example. So it is advisable to read at least those sections.
```
ge0ffrey: Note (history): The manual is started out focused on the n queens example, but these days: because cloud balancing is the quick start, we should explain every feature mainly with cloud balancing and actively get rid of the use of the n queens example.
The only exception to that is the optimization algorithms themselves, because only with 4 queens it is possible to draw enough solutions states to show the behaviour of the algorithm (see Simulated Annealing algorithm etc)

optaplanner-docs/src/main/asciidoc/UseCasesAndExamples/UseCasesAndExamples-chapter.adoc
gsheldon: Ack, I can do that @ge0ffrey. There are certain circumstances where leaving a numeral is correct (these are outlined in the IBM Style Guide), but generally speaking, and in the case of the Queens, it should be spelled out.

optaplanner-docs/src/main/asciidoc/UseCasesAndExamples/UseCasesAndExamples-chapter.adoc
gsheldon: I agree that we should have a consistent example used throughout where possible. If a particular concept is better explained with n queens, then obviously we should use the example that best suits, but where we can use cloud balancing instead we should for the sake of consistency. It is much easier to understand a concept when it's building on a familiar example.

ge0ffrey:
I had to use squash&merge instead of Rebase&merge to avoid merge conflict.

PR #287, opened 2017-03-30
Increase TemporalValueRange test coverage

yurloc:
Includes a fix that avoids an unexpected `DateFormatException` when creating ranges that use extreme upper bounds like `LocalDateTime.MAX`.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
@@ -80,7 +80,7 @@ public class TemporalValueRange<Temporal_ extends Temporal & Comparable<? super 
         // Fail fast if there's a remainder on amount (to be consistent with other value ranges)
         if (space % incrementUnitAmount > 0) {
             throw new IllegalArgumentException("The " + getClass().getSimpleName()
-                    + " 's incrementUnitAmount (" + incrementUnitAmount
+                    + "'s incrementUnitAmount (" + incrementUnitAmount
                     + ") must fit an integer number of times in the space (" + space
                     + ") between from (" + from + ") and to (" + to + ").");
         }
@@ -94,7 +94,7 @@ public class TemporalValueRange<Temporal_ extends Temporal & Comparable<? super 
         }
         if (typeRemainder) {
             throw new IllegalArgumentException("The " + getClass().getSimpleName()
-                    + " 's incrementUnitType (" + incrementUnitType
+                    + "'s incrementUnitType (" + incrementUnitType
                     + ") must fit an integer number of times in the space (" + space
                     + ") between from (" + from + ") and to (" + to + ").\n"
                     + "The to (" + from.plus(space, incrementUnitType) + ") is not the expected to (" + to + ").");
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.optaplanner.core.impl.domain.valuerange.buildin.temporal;
 
 import java.time.temporal.Temporal;
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
@@ -15,6 +15,7 @@
  */
 package org.optaplanner.core.impl.domain.valuerange.buildin.temporal;
 
+import java.time.DateTimeException;
 import java.time.temporal.Temporal;
 import java.time.temporal.TemporalAmount;
 import java.time.temporal.TemporalUnit;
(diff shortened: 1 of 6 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRangeTest.java
```diff
@@ -332,4 +359,118 @@ public void getAndCreateOriginalIteratorInSyncForLocalDate() {
...
+        assertNotNull(new TemporalValueRange<>(from, to, 1, ChronoUnit.HOURS));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void remainderOnIncrementTypeExceedsMaximumYear() {
+        Year from = Year.of(Year.MIN_VALUE);
+        Year to = Year.of(Year.MAX_VALUE - 0);
+        assertNotEquals(0, (to.getValue() - from.getValue()) % 10); // Maximum Year range is not divisible by 10
+        assertNotNull(new TemporalValueRange<>(from, to, 1, ChronoUnit.DECADES));
+    }
```
yurloc: This test would also throw `DateTimeException` without the fix. Now it "succeeds" with IllegalArgException explaining that Decades increment doesn't fit into the `[from, to)` by integer number of times.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -45,35 +63,43 @@ public TemporalValueRange(Temporal_ from, Temporal_ to, long incrementUnitAmount
         if (!from.isSupported(incrementUnitType) || !to.isSupported(incrementUnitType)) {
             throw new IllegalArgumentException("The " + getClass().getSimpleName()
                     + " must have a incrementUnitType (" + incrementUnitType
-                    + ") that is supported by its from (" + from + ") and to (" + to + ").");
+                    + ") that is supported by its from (" + from.getClass().getSimpleName()
+                    + ") and to (" + to.getClass().getSimpleName() + ").");
```
ge0ffrey: The labels (the part before the "(") should change too, so

    "... supported by its from class (" + from.getClass().getSimpleName()

ge0ffrey: I still think it wouldn't hurt to **also** show the from and to values too in the error message.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -45,35 +63,43 @@ public TemporalValueRange(Temporal_ from, Temporal_ to, long incrementUnitAmount
...
         }
         // We cannot use Temporal.until() to check bounds due to rounding errors
         if (from.compareTo(to) > 0) {
             throw new IllegalArgumentException("The " + getClass().getSimpleName()
                     + " cannot have a from (" + from + ") which is strictly higher than its to (" + to + ").");
         }
         long space = from.until(to, incrementUnitType);
-        if (to.equals(from.plus(space + 1, incrementUnitType))) {
+        if (!to.equals(from.plus(space, incrementUnitType))) {
```
ge0ffrey: I remember this being very sensitive.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -45,35 +63,43 @@ public TemporalValueRange(Temporal_ from, Temporal_ to, long incrementUnitAmount
...
                     + ") between from (" + from + ") and to (" + to + ").");
         }
         // Fail fast if there's a remainder on type (to be consistent with other value ranges)
-        if (!to.equals(from.plus(space, incrementUnitType))) {
+        boolean typeRemainder = true;
+        try {
+            typeRemainder = !to.equals(from.plus(space, incrementUnitType));
+        } catch (DateTimeException e) {
+            // When using extreme ranges, like [LocalDate.MIN-LocalData.MAX) the space may exceed maximum Year value
+            space--; // only needed for the exception message below (otherwise DateTimeException will be thrown)
```
ge0ffrey: We won't use exceptions for flow control.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRangeTest.java
ge0ffrey: Interesting test, very nice!

ge0ffrey:
I 'll pull this change locally and investigate. This PR raises some very valid points (the new unit tests) and we definitely need that test coverage, but there be dragons in the time stuff, so I want to take a closer look locally at the changes in the sensitive area.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
yurloc: +1 but I thought this is similar to catching `NumberFormatException` to see if a string contains a number (you can't do the check without try/catch AFAIK). I'll take a look if there's a way to check if `from.plus(space, incrementUnitType)` overflows maximum Year bound.

yurloc:
You can easily check out the PR with
```
$ git fetch origin pull/287/merge:pr287
```
This will create a local branch (pr287) containing the PR changes after merging my source branch into target (upstream master).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
```diff
@@ -45,35 +63,43 @@ public TemporalValueRange(Temporal_ from, Temporal_ to, long incrementUnitAmount
...
+                    + "'s incrementUnitAmount (" + incrementUnitAmount
                     + ") must fit an integer number of times in the space (" + space
                     + ") between from (" + from + ") and to (" + to + ").");
         }
         // Fail fast if there's a remainder on type (to be consistent with other value ranges)
-        if (!to.equals(from.plus(space, incrementUnitType))) {
+        boolean typeRemainder = true;
+        try {
+            typeRemainder = !to.equals(from.plus(space, incrementUnitType));
+        } catch (DateTimeException e) {
```
ge0ffrey: This DateTimeException e is eaten instead of changed. Fixing.

ge0ffrey:
Merged manually, in this commit:
  http://github.com/kiegroup/optaplanner/compare/eac6eac...46dee59

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
ge0ffrey: Change makes sense indeed, avoid overflow.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/valuerange/buildin/temporal/TemporalValueRange.java
ge0ffrey: This didn't actually use exceptions for flow control as it also throw an exception if the DateTimeException was thrown.

PR #288, opened 2017-03-30
Add Pitest plugin configuration

yurloc:
This allows to run Pitest mutation coverage analysis with a command
similar to:
```
mvn org.pitest:pitest-maven:mutationCoverage -Dpitest.classes=org.optaplanner.core.impl.score*
```
or just:
```
mvn org.pitest:pitest-maven:mutationCoverage
```
for analysis of the whole module.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -189,6 +189,30 @@
             </execution>
           </executions>
         </plugin>
+        <plugin>
+          <groupId>org.pitest</groupId>
+          <artifactId>pitest-maven</artifactId>
+          <version>1.1.11</version>
+          <configuration>
+            <reportsDirectory>local/pit-reports</reportsDirectory>
+            <timestampedReports>false</timestampedReports>
+            <timeoutConstant>1000</timeoutConstant>
+            <mutators>
+              <!--
+              <mutator>CONDITIONALS_BOUNDARY</mutator>
+              <mutator>NEGATE_CONDITIONALS</mutator>
+              <mutator>MATH</mutator>
+              <mutator>INCREMENTS</mutator>
+              <mutator>INVERT_NEGS</mutator>
+              <mutator>RETURN_VALS</mutator>
+              <mutator>VOID_METHOD_CALLS</mutator>
+              -->
+              <mutator>DEFAULTS</mutator>
+              <mutator>NON_VOID_METHOD_CALLS</mutator>
+            </mutators>
+            <avoidCallsTo>java.lang.StringBuilder</avoidCallsTo>
+          </configuration>
+        </plugin>
       </plugins>
     </pluginManagement>
   </build>


pom.xml
```diff
@@ -189,6 +193,33 @@
             </execution>
           </executions>
         </plugin>
+        <plugin>
+          <groupId>org.pitest</groupId>
+          <artifactId>pitest-maven</artifactId>
```
ge0ffrey: @psiroky Can we move this into kie-parent?

pom.xml
```diff
@@ -189,6 +193,33 @@
...
+          <version>1.1.11</version>
+          <configuration>
+            <reportsDirectory>local/pit-reports</reportsDirectory>
+            <timestampedReports>false</timestampedReports>
+            <timeoutConstant>1000</timeoutConstant>
+            <targetClasses>
+              <param>${pitest.classes}</param>
```
ge0ffrey: Does the pitest-maven plugin really need to know "org.optaplanner*"?

Can't it just grab all the classes from the current module? Suppose the module is optaplanner-examples, does pitests need to change optaplanner-core classes then?

yurloc: This allows to limit the analysis from the whole module to a selected package. The full analysis takes over 30 minutes on optaplanner-core so limiting to a package is handy. Unfortunately the plugin itself doesn't seem to support parameter tweaking through system properties.

pom.xml
psiroky: yes, we can!

pom.xml
ge0ffrey: Ah, so by default you'd like to run it without this configuration parameter, but then it's no longer possible to override on the command line?

ge0ffrey: Reading this
  http://pitest.org/quickstart/maven/
I wonder if -DtargetClasses or -Dpit.targetClasses doesn't work, because all "extra" properties do have command line switch, so why wouldn't the "non-extra" one have it?
If not, an issue needs to be raised in their issue tracker.

Either way, I 'd prefer to remove that module specific property "pitest.classes", even if that disabled the command line override for now... because that's a maintenance pain (on drools, optaplanner, optaplanner-wb, ...) as well as a distraction (less in the pom is more - pitest should take up pom space except in the parent). Wydt?

yurloc: You are correct. `-DtargetClasses` works perfectly. I removed the property.

Now that we got rid of the optaplanner-specific property the Pitest plugin config can be moved to a higher level (kie-parent). Do you prefer that over having the config in optaplanner-parent only?

ge0ffrey:
@yurloc Would you mind doing a similar PR for kie-parent? If that gets merged, I 'd like to close this PR.

ge0ffrey:
Closing this issue as this will be solved on kie-parent

yurloc:
Superceded by kiegroup/droolsjbpm-build-bootstrap#454.

PR #290, opened 2017-03-31
changed numerals to spelled out numbers where applicable throughout a…

gsheldon:
…ll chapters that have not been checked yet as part of the current review and modularization initiative.

All instances where numerals were used instead of spelled-out numbers should now be correct. Numerals have been left in place where it is the name of something (ie, a line in a diagram). I'm happy to check any of these against IBM if there is any contention here. I don't have my guide with me right now so please let me know if you feel any should remain numerals.


ge0ffrey:
They all looked good.
In code, numbers should remain numerals of course (and I didn't see any such changes), otherwise I am ok to follow the product docs style, for consistency.

PR #291, opened 2017-03-31
Checkstyle cleanup

yurloc:
### EqualsHashCode warning in TestdataObjectEqualsNoHashCode
This is a test file that intentionally implements `equals()` without implementing `hashCode()` so we don't want to check it.
### JavadocStyle in persistence modules
There is a frequent usage of `{@inheritDoc}` in class descriptions but Checkstyle thinks that `{@inheritDoc}` isn't allowed there so it emits a warning about a first sentence not ending with '.' even though it is able to handle it in methods descriptions. I disabled this check in all persistence modules.

### UnusedImports
This was possible after the recent import optimization.
### WhitespaceAround
Maybe this check used to be blocked by a false positive but now it passes even though the `allowEmptyBlocks` property is not available.

## Other
- Removed an obsolete TODO for Checkstyle version upgrade.
- Fixed a few valid warnings.
- The build now passes without any Checkstyle warnings \o/.


yurloc:
Feel free to squash when merging.

pom.xml
```diff
@@ -132,15 +123,18 @@
...
                       <property name="tabWidth" value="4"/>
                       <!-- Checks for Javadoc comments. -->
                       <module name="JavadocStyle"/>
                       <!-- TODO: Enable NonEmptyAtclauseDescription after https://github.com/checkstyle/checkstyle/issues/1064 is fixed -->
+                      <!-- Update: #1064 is resolved and the performance has greatly improved but this check still adds ~2 seconds -->
+                      <!-- to the build time -> always measure build time before enabling this module (use github.com/jcgay/maven-profiler) -->
                       <!--<module name="NonEmptyAtclauseDescription"/>-->
                       <!-- Checks for imports. -->
-                      <!--<module name="UnusedImports"/>-->
+                      <module name="UnusedImports"/>
```
ge0ffrey: Failing on unused imports is still controversial:
1) It slows down the build apparently
2) The rest of KIE doesn't do this yet:
  https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L495
3) It's easy to fix automatically before a CR1 release etc.
4) It doesn't impact the bits of the .class files.

I don't want to start turning down contributions from contributors because of this, so I'd rather leave this commented out, for now.

optaplanner-persistence/pom.xml
```diff
@@ -30,4 +30,27 @@
...
+          <artifactId>maven-checkstyle-plugin</artifactId>
+          <executions>
+            <execution>
+              <id>validate</id>
+              <phase>validate</phase>
+              <goals>
+                <goal>check</goal>
+              </goals>
+              <configuration>
+                <suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
```
ge0ffrey: This file isn't important enough to live in the root folder. Let's move it into src/(main/?)/checkstyle or something.

Actually, SupressWarning works on a class level (not a package level unfortunately), which is much closer to the actual code (and seen by codacy and other services that rate our quality, etc while this is not) - I 'd much rather just see a SurpressWarning annotation on all the affected classes (even though those are bunch...). Wdyt?

pom.xml
yurloc: +0.5 why not to remove the check completely?

ge0ffrey: +1 to remove indeed. Commenting out code we don't intend to add soon is better removed usually.

yurloc:
Review feedback implemented.

ge0ffrey:
Thanks Jiri!

PR #294, opened 2017-04-05
Add weight type param to weight factory and WF selection sorter

yurloc:
Removes some rawtype and unchecked warnings.

From the user's perspective it makes the type system more explicit. For example it makes it very clear from the start that weight is something different than the selection (entity/value).

Also it is clear that `MyWeight` is going to implement `Comparable<MyWeight>` and not `Comparable<MyEntity>`, which is of course a stupid mistake, but it's something I myself did and I wouldn't have done it with this more explicit type declaration.

So, I would highlight that the extra type parameter saves the time it takes to understand "_How am I going to implement the weight factory?_" for first-time users.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/decorator/WeightFactorySelectionSorter.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/decorator/WeightFactorySelectionSorter.java
@@ -34,13 +34,15 @@ import org.optaplanner.core.impl.score.director.ScoreDirector;
  * Sorts a selection {@link List} based on a {@link SelectionSorterWeightFactory}.
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  * @param <T> the selection type
+ * @param <W> the weight type
  */
-public class WeightFactorySelectionSorter<Solution_, T> implements SelectionSorter<Solution_, T> {
+public class WeightFactorySelectionSorter<Solution_, T, W extends Comparable<W>>
+        implements SelectionSorter<Solution_, T> {
 
-    private final SelectionSorterWeightFactory<Solution_, T> selectionSorterWeightFactory;
-    private final Comparator<Comparable> appliedWeightComparator;
+    private final SelectionSorterWeightFactory<Solution_, T, W> selectionSorterWeightFactory;
+    private final Comparator<W> appliedWeightComparator;
 
-    public WeightFactorySelectionSorter(SelectionSorterWeightFactory<Solution_, T> selectionSorterWeightFactory,
+    public WeightFactorySelectionSorter(SelectionSorterWeightFactory<Solution_, T, W> selectionSorterWeightFactory,
             SelectionSorterOrder selectionSorterOrder) {
         this.selectionSorterWeightFactory = selectionSorterWeightFactory;
         switch (selectionSorterOrder) {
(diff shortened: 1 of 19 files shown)


yurloc:
Note: this will require an upgrade recipe update.

optaplanner-examples/src/main/java/org/optaplanner/examples/curriculumcourse/domain/solver/LectureDifficultyWeightFactory.java
```diff
@@ -22,11 +22,13 @@
...
 import org.optaplanner.examples.curriculumcourse.domain.UnavailablePeriodPenalty;
+import org.optaplanner.examples.curriculumcourse.domain.solver.LectureDifficultyWeightFactory.LectureDifficultyWeight;
 
-public class LectureDifficultyWeightFactory implements SelectionSorterWeightFactory<CourseSchedule, Lecture> {
+public class LectureDifficultyWeightFactory
+        implements SelectionSorterWeightFactory<CourseSchedule, Lecture, LectureDifficultyWeight> {
 
     @Override
-    public Comparable createSorterWeight(CourseSchedule schedule, Lecture lecture) {
+    public LectureDifficultyWeight createSorterWeight(CourseSchedule schedule, Lecture lecture) {
```
ge0ffrey: Returning LectureDifficultyWeight is a good thing here. That can be done with or without adding that 3 generic parameter. So even if we withold this PR, we should still do that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/decorator/WeightFactorySelectionSorter.java
```diff
@@ -34,13 +34,15 @@
...
+ * @param <W> the weight type
  */
-public class WeightFactorySelectionSorter<Solution_, T> implements SelectionSorter<Solution_, T> {
+public class WeightFactorySelectionSorter<Solution_, T, W extends Comparable<W>>
+        implements SelectionSorter<Solution_, T> {
 
-    private final SelectionSorterWeightFactory<Solution_, T> selectionSorterWeightFactory;
-    private final Comparator<Comparable> appliedWeightComparator;
+    private final SelectionSorterWeightFactory<Solution_, T, W> selectionSorterWeightFactory;
+    private final Comparator<W> appliedWeightComparator;
```
yurloc: The impact on `WeightFactorySelectionSorter` is a nice side effect of the additional 3rd parameter.

The `WeightFactory` itself is complex because it's an abstract thing. If we add the 3rd parameter, the complexity remains the same but it becomes more visible and easier to understand, which is a good thing.

The downside is that it will slightly hurt users with existing weight factory implementations. They will need to add the weight type to the `implements` declaration and may need to change the return type of `createSorterWeight()`. On the other hand it will help users that are going to implement a new weight factory.

So in my opinion this is an improvement and it's better to make small improvements even at some cost (when the time is right).

yurloc: But I don't want to push you to merging this :-) Unless there is a consensus, it's OK to close this or continue discussion.

ge0ffrey:
Sorry for waiting so long with a decision on this.
I believe it's still trading one bad for another. Especially with 7.0 out now, I am effectively rejecting this PR :/

I 'll recycle the LectureDifficultyWeight improvement of this PR.

PR #295, opened 2017-04-06
Modularizing use cases chapter

gsheldon:
Completed modularizing topics and moving images into section folders. There appears to still be a number of images allocated to this chapter's folder that are not included here. I'm not sure if they are redundant images or if they were moved here by mistake when the images were moved out of a single directory into the chapter directories. 

I'm also a little unsure of the naming of these sections, ie, "Difficult Examples" I feel could deter some users. Perhaps "Advanced", "Complex" or something similar would be better? WDYT?


ge0ffrey:
Hi Gemma,

"There appears to still be a number of images allocated to this chapter's folder that are not included here." Can you make an inventory of them? These are probably images that should be included somewhere too, but I created them for the slides and didn't add them in the docs yet. If that's not the case, they should be removed. Let's go through that inventory in a separate issue.

"I'm also a little unsure of the naming of these sections, ie, "Difficult Examples" I feel could deter some users. Perhaps "Advanced", "Complex" or something similar would be better?" Yes, "Advanced Examples" is a much better term.

"This branch cannot be rebased due to conflicts" Looks like your branch didn't start fresh, so with a nice "git checkout master; git pull --rebase; git checkout -b modularizingUseCasesChapter", but instead continued on an older branch. So I can't "rebase and merge", only "squash and merge" (which I 'll do now as it seems fine), but to avoid trouble, please start from an updated master whenever it is practical (= whenever you wouldn't loose time or commits by doing so).

ge0ffrey:
I've merged this already, because it's step forward,
but I think we might want to have a section.adoc file per example, as the BasicExamples adoc already gets pretty big, it's hard to remember in which file to find the "vehicle routing example", etc. I also believe each example could use it's own directory, because some examples (vehicle routing) have a lot of images.

At that point the question becomes: do we want to keep the "Basic examples" "Advanced examples" etc separation in the docs TOC? If so, do we also want to mimic that in the directory structure (so "UseCasesAndExamples/MeetingScheduling/foo.png" or "UseCasesAndExamples/BasicExcamples/MeetingScheduling/foo.png"? WDYT?

gsheldon:
I know that there would be a lot of contention with nesting the files in
this way, as this was recently discussed with our new hire, Anna. I'll
forward you the thread so you can see what I mean.

I think we can and should definitely break up these files more, they are
very chunky and I can see problems down the line with maintaining them. I
am assuming that we would be breaking the rule of beginning files with more
than a single offset though if we tried to break them up into nested
section files?

I also don't really like the section titles we are using here either. I'm
going to think on those more because how we name these sections can impact
on how we structure the content too, and I'd like to do more work on that
in this chapter.

PR #297, opened 2017-04-06
Eliminate raw Move in LS move and step scopes and some tests

yurloc:
The most important change is the return type of `LocalSearchStepScope.getStep()` which makes it possible to observe steps through `PhaseLifecycleListener` so it makes sense to avoid the rawtype here.

Same changes made in `LocalSearchMoveScope` for consistency.

The changes in tests are not implied by the `LocalSearchStepScope` and `LocalSearchMoveScope` changes, they are just rawtype elimination in related code.

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/AbstractAcceptorTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/AbstractAcceptorTest.java
@@ -25,9 +25,10 @@ import static org.mockito.Mockito.*;
 
 public abstract class AbstractAcceptorTest {
 
-    protected LocalSearchMoveScope buildMoveScope(LocalSearchStepScope stepScope, int score) {
-        LocalSearchMoveScope moveScope = new LocalSearchMoveScope(stepScope);
-        Move move = mock(Move.class);
+    protected <Solution_> LocalSearchMoveScope<Solution_> buildMoveScope(
+            LocalSearchStepScope<Solution_> stepScope, int score) {
+        LocalSearchMoveScope<Solution_> moveScope = new LocalSearchMoveScope<>(stepScope);
+        Move<Solution_> move = mock(Move.class);
         moveScope.setMove(move);
         moveScope.setScore(SimpleScore.valueOf(score));
         return moveScope;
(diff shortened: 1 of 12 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/AbstractAcceptorTest.java
```diff
@@ -25,9 +25,9 @@
 
 public abstract class AbstractAcceptorTest {
 
-    protected LocalSearchMoveScope buildMoveScope(LocalSearchStepScope stepScope, int score) {
-        LocalSearchMoveScope moveScope = new LocalSearchMoveScope(stepScope);
-        Move move = mock(Move.class);
+    protected <T> LocalSearchMoveScope<T> buildMoveScope(LocalSearchStepScope<T> stepScope, int score) {
```
ge0ffrey: Can we replace T with Solution_ so it's clear that a solution class is expected there?

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/hillclimbing/HillClimbingAcceptorTest.java
```diff
@@ -29,20 +29,20 @@
 public class HillClimbingAcceptorTest extends AbstractAcceptorTest {
 
     @Test
-    public void hillClimbingEnabled() {
+    public <Solution_> void hillClimbingEnabled() {
         HillClimbingAcceptor acceptor = new HillClimbingAcceptor();
 
-        DefaultSolverScope solverScope = new DefaultSolverScope();
+        DefaultSolverScope<Solution_> solverScope = new DefaultSolverScope<>();
```
ge0ffrey: I'd argue that this test should use a specific generic type, so TestdataSolution:

`DefaultSolverScope<TestdataSolution> solverScope = new DefaultSolverScope<>();`

that would remove the need generifying a test method.

yurloc: +0. The exact type parameter has no effect. Even `Object` works. So I went with an unspecified parameter. And `Solution_` is shorter than `TestdataSolution`.

I'll change that as you request. I suppose that all tests should be updated.

yurloc:
Comments resolved.

PR #299, opened 2017-04-07
Clean up dependencies

yurloc:
Removed unused dependencies, adjusted some scopes from `compile` to `runtime`.

### DIFF
--- a/optaplanner-benchmark/pom.xml
+++ b/optaplanner-benchmark/pom.xml
@@ -59,6 +59,7 @@
     <dependency>
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
+      <scope>runtime</scope>
     </dependency>
     <!-- External dependencies -->
     <!-- Common utils -->
@@ -69,10 +70,6 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
     </dependency>
     <!-- Logging -->
     <dependency>
--- a/optaplanner-examples/pom.xml
+++ b/optaplanner-examples/pom.xml
@@ -83,7 +83,7 @@
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-persistence-xstream</artifactId>
     </dependency>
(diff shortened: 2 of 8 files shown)


optaplanner-benchmark/pom.xml
```diff
@@ -53,12 +53,9 @@
       <artifactId>optaplanner-persistence-xstream</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.optaplanner</groupId>
-      <artifactId>optaplanner-persistence-jaxb</artifactId>
-    </dependency>
-    <dependency>
```
ge0ffrey: once we add JaxbSolutionFileIO, this one will be needed, no?

optaplanner-examples/pom.xml
```diff
@@ -83,10 +83,6 @@
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-persistence-xstream</artifactId>
     </dependency>
-    <dependency><!-- Few examples use the JAXB integration -->
```
ge0ffrey: Let's change this into a TODO, that we need an example that uses it :) That way we have an integration test too.

optaplanner-test/pom.xml
```diff
@@ -29,25 +29,6 @@
...
-      <artifactId>guava</artifactId>
-    </dependency>
-    <!-- Logging -->
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>ch.qos.logback</groupId>
-      <artifactId>logback-classic</artifactId>
```
ge0ffrey: we'll need tests for the ScoreVerifier I fear, so we might need logging for those, no?

optaplanner-benchmark/pom.xml
yurloc: Correct.

optaplanner-test/pom.xml
yurloc: Yes, I eventually agree with keeping this. Other modules have this dependency even though it's unused because the tests don't log anything. But in this case I think it's better to have the logging enabled in the test scope so when you add tests and need logging you don't have to figure out how to enable it.

yurloc:
Comments resolved.

ge0ffrey:
jenkins retest this

PR #306, opened 2017-05-04
Change revapi-config.json comment to be more general

MarianMacik:
Hi, @ge0ffrey,

this is only a minor change to the revapi-config.json file, since these files are no more only on master branch, but on 7.0.x as well.

Marian

P.S. I still have in plan to fine tune revapi in benchmark config classes. Sadly, there isn't time for this now, but I will try to incorporate this in the near future. Thanks for your understanding.


ge0ffrey:
Thanks Marian!

PR #308, opened 2017-05-08
Chapter3 modularization and title changes

gsheldon:
Sorry for the delay with this part, was having trouble with the leveloffset.


ge0ffrey:
Looks good. I had committed and pushed some changes in the TTP example meanwhile, but those only affect the SVG/PNG file, so it all comes together well.

PR #309, opened 2017-05-13
PLANNER-793 Autodiscover problem facts without annotations

ge0ffrey:
through flag on @Planningsolution

 (AKA deprecate AbstractSolution)


ge0ffrey:
@yurloc This contains the basic unit tests, but it think we need a few more specified here:
https://issues.jboss.org/browse/PLANNER-793

ge0ffrey:
jenkins retest this (it didn't fail, it grayed)

PR #315, opened 2017-06-05
Revapi: filter only public fields and getters/setters

MarianMacik:
Hi, @ge0ffrey,

here is the promised custom config for OptaPlanner. It should check from now on only public fields + getters and setters, as you wanted.

Marian


ge0ffrey:
Wonderful, thanks

PR #317, opened 2017-06-09
PLANNER-793 Cover more autodiscovery use cases

yurloc:
Covered use cases requested in https://issues.jboss.org/browse/PLANNER-793:
- member is a subclass of an entity class
- member is a collection of entity subclass elements
- field/getter would be autodiscovered but it uses an annotation to override the autodiscovery result
- getter is overriden and the parent was autodiscovered but the child has an explicit annotation **(broken)**

Negative test cases:
- Autodiscovery will fail with `Collection<?> getFacts()` (unsupported element type)
- Autodiscovery will fail with `MyFactCollection getFacts()` (element type is unknown)

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptorTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptorTest.java
@@ -27,6 +27,7 @@ import org.optaplanner.core.impl.testdata.domain.TestdataValue;
 import org.optaplanner.core.impl.testdata.domain.collection.TestdataArrayBasedSolution;
 import org.optaplanner.core.impl.testdata.domain.collection.TestdataSetBasedSolution;
 import org.optaplanner.core.impl.testdata.domain.extended.TestdataAnnotatedExtendedSolution;
+import org.optaplanner.core.impl.testdata.domain.extended.TestdataUnannotatedExtendedEntity;
 import org.optaplanner.core.impl.testdata.domain.extended.abstractsolution.TestdataExtendedAbstractSolution;
 import org.optaplanner.core.impl.testdata.domain.extended.abstractsolution.TestdataExtendedAbstractSolutionOverridesGetScore;
 import org.optaplanner.core.impl.testdata.domain.extended.legacysolution.TestdataLegacySolution;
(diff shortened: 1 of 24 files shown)


ge0ffrey:
Jenkins retest this

yurloc:
@ge0ffrey The 1 failure is the test that demonstrates that one of the use cases doesn't work. At least if it should work the way I understand it. Let me know once you have the time to discuss it.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptorTest.java
```diff
@@ -195,6 +203,20 @@ public void extendedAbstractSolutionOverridesGetScore() {
         assertAllCodesOfCollection(solutionDescriptor.getAllFacts(solution), "e1", "e2", "v1", "v2", "extra");
     }
 
+    // ************************************************************************
+    // Autodiscovery
```
baldimir: Just a comment (not needed to do this): It implies to me that if some tests are marked with such header, they could be part of their own test class. E.g. in this case SolutionDescriptorAutodiscoveryTest.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptorTest.java
```diff
@@ -237,6 +259,103 @@ public void autoDiscoverGetters() {
...
+        List<String> listFact = new TestdataCodeAssertableArrayList<>("list1", Arrays.asList("x", "y"));
+        TestdataAutoDiscoverFieldOverrideSolution solution = new TestdataAutoDiscoverFieldOverrideSolution(
+                "s1", singleProblemFact, valueList, entityList, otherEntity, listFact);
+
+        assertAllCodesOfCollection(solutionDescriptor.getAllFacts(solution),
+                "otherE1", "list1", "p1", "e1", "e2", "v1", "v2");
+    }
+
+    @Test
+    public void autoDiscoverGettersFactCollectionOverridenToSingleProperty() {
```
baldimir: This test duplicates code of previous test. It can be made as one test method that gets solution instance as a parameter.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/solutionproperties/autodiscover/TestdataAutoDiscoverFieldOverrideSolution.java
```diff
@@ -0,0 +1,74 @@
...
+
+@PlanningSolution(autoDiscoverMemberType = AutoDiscoverMemberType.FIELD)
+public class TestdataAutoDiscoverFieldOverrideSolution extends TestdataObject {
+
+    public static SolutionDescriptor<TestdataAutoDiscoverFieldOverrideSolution> buildSolutionDescriptor() {
+        return SolutionDescriptor.buildSolutionDescriptor(TestdataAutoDiscoverFieldOverrideSolution.class, TestdataEntity.class);
+    }
+
+    private TestdataObject singleProblemFact;
+    @ValueRangeProvider(id = "valueRange")
```
baldimir: I think there should be an empty line before fields with annotations, so the code is more readable.

ge0ffrey: Not that I mind the comment, but this is a matter of taste (tabs vs spaces etc). We don't do this. 

Because it's a matter of taste ("color of the curtains"), it means neither of us is wrong. We consistently go for allowing this.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/extended/abstractsolution/TestdataExtendedAbstractSolutionOverridesGetScore.java
```diff
@@ -31,8 +31,9 @@
 @PlanningSolution
 public class TestdataExtendedAbstractSolutionOverridesGetScore extends AbstractSolution<BendableScore> {
```
ge0ffrey: I 'll rename this class so it ends with "Solution" as it's a solution. (Not really related to this PR)

ge0ffrey:
@yurloc That failing test points out a chunk of missing functionality, namely that subclassing SolutionPlanning classes doesn't work.
I did have to "fix" the test first, due to some hidden requirements that are only in my head up to now, so let me explain them:
- If A extends B then all A instances should be handled by a Descriptor for A, not a Descriptor for B. If A instances are handled by a Descriptor for B then they will be treated as B's, so their A specific annotations will be ignored. We can't fix as that would require scanning. We should fail fast on that (or at least warn), so I just created https://issues.jboss.org/browse/PLANNER-829 @yurloc would you mind making a unit test for that new issue?
- If A has a planner annotation on one of it's fields/methods, it should have a PlanningSolution/Entity annotation on it's class, even if it's parent class already has that. That is the current state I believe. (TODO: Is this true for PlanningEntity?) Do we want this requirement? We can probably change this if we want to. I am not sure if we want to. Wdyt?

yurloc:
@ge0ffrey I noticed the test still fails after your fix. Are you aware of that?

I think PLANNER-829 is a good opportunity to try the agile approach where the issue can be closed after both the fix and tests (and documentation) for it are implemented. And it should be all done in one sprint even though it requires collaboration of multiple people.

ge0ffrey:
I pushed something to early. In any case, it's fixed now, as part of the work for https://issues.jboss.org/browse/PLANNER-831 (which was a bug that was also exposed by your test), see https://github.com/kiegroup/optaplanner/commit/8f8d83792

PR #325, opened 2017-07-22
SLF4J-related fixes

dtrebbien:


ge0ffrey:
ok to test

ge0ffrey:
@dtrebbien Thank you, Daniel. Nice work.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/reproducer/TestGenCorruptedScoreReproducer.java
```diff
@@ -66,10 +66,10 @@ public boolean isReproducible(TestGenKieSessionJournal journal) {
         } catch (RuntimeException e) {
             if (e.getMessage() != null && e.getMessage().startsWith("No fact handle for ")) {
                 // this is common when removing insert of a fact that is later updated - not interesting
-                logger.debug("    Can't remove insert: {}", e.toString());
+                logger.debug("    Can't remove insert: {}", (Object) e);
```
yurloc: Unfortunately, this doesn't work as expected. The result looks like:
```
18:39:02.382 [main] DEBUG     Can't remove insert: {}
java.lang.IllegalStateException: No fact handle for cloudProcess_847
	at org.optaplanner.core.impl.score.director.drools.testgen.operation.TestGenKieSessionUpdate.invoke(TestGenKieSessionUpdate.java:80) ~[classes/:na]
	(full stack trace)
```
It seems that SLF4J is too [smart](https://www.slf4j.org/faq.html#paramException) and always detects whether the last parameter is an Exception even if you cast it to Object and even if it results in an unused placeholder.

ge0ffrey: @yurloc Thank you for noticing this. Those subset of changes need to be undone. Do you want to do the honors? (happy to do them myself too)

Is there a difference between e.toString() and e.getMessage()? Should we consider using getMessage() over toString()?

yurloc: @ge0ffrey I'll undo this together with some other work on testgen.

I was intentionally using e.toString() as I wanted to get both the message and FQCN. For the sake of avoiding the warning about implicit toString() calls I could rework it to `e.getClass().getName() + ": " + e.getMessage()` but that's not nice.

Another solution could be to fix SLF4J to support this use case...

ge0ffrey: Agreed to rollback to toString()

dtrebbien: This is strange.  I have a small test app with which I tried out various versions of SLF4J, including this pattern:

    LOGGER.info("here is a Throwable: {}", (Object) t);

For me, using the "simple" backend, the stack trace is not printed.

I wonder if this depends on the specific backend?

dtrebbien: This does appear to be a difference between backends.

The "simple" backend's implementation of debug(String, Object) passes an extra `null` parameter to MessageFormatter.format():  
https://github.com/qos-ch/slf4j/blob/1c13755b2520e511bc41b5c6fdc8bd6c58e529c0/slf4j-simple/src/main/java/org/slf4j/simple/SimpleLogger.java#L445

The "log4j12" backend does not:  
https://github.com/qos-ch/slf4j/blob/1c13755b2520e511bc41b5c6fdc8bd6c58e529c0/slf4j-log4j12/src/main/java/org/slf4j/log4j12/Log4jLoggerAdapter.java#L227

There is also a related bug report and open pull request: qos-ch/slf4j#166

When that PR is merged, it should be okay to cast to Object rather than call toString() because the exception will be "consumed" by the format string.

ge0ffrey: Very nice investigation, thank you for sharing.
The long term solution is indeed that SFL4J PR merged and released. Neither is the case today, so let's go a head with reverting those few lines (only that subset of these changes).

dtrebbien: Agreed. I will submit a PR.

dtrebbien: PR #331

PR #327, opened 2017-07-25
[DROOLS-1649] ServiceRegistry refactor

mariofusco:


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/kie/KieSolverAssemblerService.java
```diff
@@ -36,13 +35,7 @@ public ResourceType getResourceType() {
     }
 
     @Override
-    public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceType type, ResourceConfiguration configuration) {
+    public void addResource(Object kbuilder, Resource resource, ResourceType type, ResourceConfiguration configuration) {
```
ge0ffrey: A bit strange that this interface in drools-core is now weakly typed, but I presume you have your reasons :)
+1 to merge together with upstream changes

PR #331, opened 2017-08-03
Partially revert 8e0d114c528d3d0a47c3e6fcb179e7df0e169a4c

dtrebbien:
It appears that there is a difference in behavior of the "log4j12"
backend that causes the stack trace of exceptions to be printed, which
is not desired.

See kiegroup/optaplanner#325


ge0ffrey:
ok to test

@psiroky is anyone looking at the cause of the jenkins spam?

psiroky:
As mentioned in the other PR, this is the cause https://github.com/jenkinsci/ghprb-plugin/issues/267 and this is a potential fix https://github.com/jenkinsci/ghprb-plugin/pull/533. Unfortunately the plugin is not that actively maintained over the past few months and the PR has been ignored so far.

PR #332, opened 2017-08-18
RHBRMS-2784 - XStream: DoS when unmarshalling void type (#328)

mbiarnes:


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/XStreamXmlSolverFactory.java
```diff
@@ -49,6 +49,8 @@
      */
     public static XStream buildXStream() {
         XStream xStream = new XStream();
+        String[] voidDeny = {"void.class", "Void.class"};
```
ge0ffrey: Since this is a temporary workaround, we want to add a comment to make it easy to remove later, something like:
`
// TODO Workaround: remove when upgrading to XStream x.y.z or greater`

krisv:
@mbiarnes please make sure this one can be merged as well, as you already set the JIRA to resolved.

mbiarnes:
@ge0ffrey @krisv 
As master will be upgraded to jboss-ip-bom 7.0.0.Final which has version.com.thoughtworks.xstream:1.4.10
and 

Affected Versions:
==================
Currently all versions until and including version 1.4.9 are affected, but 
workarounds exist.

we should close this PR and revert on **master** the commits I did for
drools, droolsjbpm-integration, guvnor, kie-wb-common, jbpm and droolsjbpm-tools

drgzalex:
!

ge0ffrey:
Closing this PR as we've updated XStream and this is no longer an issue.

PR #334, opened 2017-08-30
PLANNER-834 Annotation overwriting

yurloc:
@ge0ffrey Asking for an early review.

This is test for scenarios with a parent entity and child entity that overrides `@PlanningVariable` in various ways.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
@@ -274,13 +274,18 @@ public class ConfigUtils {
 
     public static List<Class<?>> getAllAnnotatedLineageClasses(Class<?> bottomClass,
             Class<? extends Annotation> annotation) {
+        if (bottomClass == null) {
+            throw new IllegalArgumentException("The inspected class (bottomClass) cannot be null");
+        }
         if (!bottomClass.isAnnotationPresent(annotation)) {
             return Collections.emptyList();
         }
         List<Class<?>> lineageClassList = new ArrayList<>();
         lineageClassList.add(bottomClass);
         Class<?> superclass = bottomClass.getSuperclass();
-        lineageClassList.addAll(getAllAnnotatedLineageClasses(superclass, annotation));
+        if (superclass != null) { // superclass is null when bottomClass is an interface
+            lineageClassList.addAll(getAllAnnotatedLineageClasses(superclass, annotation));
+        }
         for (Class<?> superInterface : bottomClass.getInterfaces()) {
             lineageClassList.addAll(getAllAnnotatedLineageClasses(superInterface, annotation));
         }
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntityParent.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntityParent.java
@@ -1,0 +1,55 @@
+package org.optaplanner.core.impl.testdata.domain.anno.entity;
+
+import java.util.Collection;
+import java.util.Collections;
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.solution.PlanningEntityProperty;
+import org.optaplanner.core.api.domain.solution.PlanningScore;
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+
+@PlanningEntity
+public class TestdataAnnoEntityParent {
+
+    @PlanningSolution
+    public static interface Solution {
+
+        @PlanningScore
+        SimpleScore getScore();
+
+        void setScore(SimpleScore score);
+
+        @PlanningEntityProperty
+        TestdataAnnoEntityParent getEntityParent();
+    }
+
+    public static SolutionDescriptor<Solution> buildParentSolutionDescriptor() {
+        return SolutionDescriptor.buildSolutionDescriptor(Solution.class, TestdataAnnoEntityParent.class);
+    }
+
+    public static EntityDescriptor<Solution> buildParentEntityDescriptor() {
+        return buildParentSolutionDescriptor().findEntityDescriptorOrFail(TestdataAnnoEntityParent.class);
+    }
+
+    public static GenuineVariableDescriptor<Solution> buildParentVariableDescriptorForVar1() {
+        return buildParentEntityDescriptor().getGenuineVariableDescriptor("var1");
+    }
+
+    @PlanningVariable(valueRangeProviderRefs = "valrng1")
+    private String var1;
+
+    @ValueRangeProvider(id = "valrng1")
+    public Collection<String> valueRange1() {
+        return Collections.singleton("A");
+    }
+
+    public String getVar1() {
+        return var1;
+    }
+}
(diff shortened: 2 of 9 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntitySubChangeValRange.java
```diff
@@ -0,0 +1,63 @@
...
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+
+@PlanningEntity
+public class TestdataAnnoEntitySubChangeValRange extends TestdataAnnoEntityParent {
+
+    @PlanningSolution
+    public static interface Solution {
```
ge0ffrey: A new class also named Solution?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntitySubBrokenMissingEntityAnnotation.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.optaplanner.core.api.domain.solution.PlanningEntityProperty;
+import org.optaplanner.core.api.domain.solution.PlanningScore;
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+//@PlanningEntity // missing
```
ge0ffrey: If this is intentially not a PlanningEntity, can we change that line to `// Intentionally no @PlanningEntity annotation` instead?

Diagonally scanning this looks like something that was temporarly uncommented and forgetten to be commented it.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -274,13 +274,18 @@ public static int resolveThreadPoolSizeScript(String propertyName, String script
 
     public static List<Class<?>> getAllAnnotatedLineageClasses(Class<?> bottomClass,
             Class<? extends Annotation> annotation) {
+        if (bottomClass == null) {
+            throw new IllegalArgumentException("The inspected class (bottomClass) cannot be null");
```
ge0ffrey: error message should include relevant runtime information, in this case the bottomClass's name:

`throw new IllegalArgumentException("The bottomClass (" + bottomClass + ") cannot be null");`

manstis: But.... ```bottomClass``` is ```null``` so this'll show "The bottomClass (null) cannot be null"!!

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntityParent.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static EntityDescriptor<Solution> buildParentEntityDescriptor() {
+        return buildParentSolutionDescriptor().findEntityDescriptorOrFail(TestdataAnnoEntityParent.class);
+    }
+
+    public static GenuineVariableDescriptor<Solution> buildParentVariableDescriptorForVar1() {
+        return buildParentEntityDescriptor().getGenuineVariableDescriptor("var1");
+    }
+
+    @PlanningVariable(valueRangeProviderRefs = "valrng1")
```
ge0ffrey: Personally, I'd like to avoid abbreviations as they can be misinterpreted. We don't have to pay for using more characters.
I've seen too many abbreviations that seem obvious for one person, but non understandable for someone else, so they're not worth the risk in code. And besides, "rng" stands for "random number generator".

valrng1 -> primaryValueRange

ge0ffrey:
Looks good.

Class names are cryptic, but I don't see a way to make less cryptic. Unless we can somehow classify all these tests as "polymorfism" tests, so we can reuse that keyword in the class name to make the only difference between their names be their uniqueness. I am thinking of the pattern "TestData.*PolymorfismEntity" or something like that, which would give "TestDataRemoveVariablePolymorfismEntity".

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntitySubBrokenMissingEntityAnnotation.java
yurloc: Yes, exactly. The comment suggested by you is much better, +1.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
yurloc: What @ge0ffrey is suggesting here is a standard pattern (in optaplanner) of constructing exception messages whose purpose is to explain a programmer's mistake. Generally it's `The <description> <variableName> (<its value>) is invalid for some reason.`. I agree that including both the invalid value and the variable name is helpful when debugging and reading the code [1]. However in case of null checking this pattern starts to provide redundant information, exactly as @manstis pointed out. Besides that, some tools like FindBugs will report this as dereferencing null pointer.

[1] For example: `The argument inputArg (-9) cannot be negative because square root is not defined for negative numbers (when the codomain is real).`

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntityParent.java
yurloc: +1 this will be polished when preparing the PR for merging.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntitySubBrokenVarRemoved.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public String getVar2() {
+        return var2;
+    }
+
+    // trying to remove a planning variable (unsupported)
+    @Override
+    public String getVar1() {
+        return super.getVar1();
+    }
```
yurloc: The assumption is that subclass entity cannot remove a variable defined by its ancestor entity (we've agreed on that). The question is: should we "fail fast" and by that force the user to add the `@PlanningVariable` annotation and put him into the risk of unintentional overriding of original behavior unless he **repeats all the attributes from ancestor**? Or just print a warning?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/anno/entity/TestdataAnnoEntitySubChangeValRange.java
yurloc: Yeah I was just lazy to create a class for the solution. I though an interface would do, especially when Java 8 allows the default method implementations. After thinking twice about it, this actually looks like a kind of anti-pattern to use an interface as a planning solution, so I'm gonna change all of them into classes.

ge0ffrey:
jenkins retest this

ge0ffrey:
{code}
Test Result (3 failures )
org.optaplanner.core.config.anno.entity.TestdataAnnoTest.subEntityDescriptorWith2ndVariableField
org.optaplanner.core.config.anno.entity.TestdataAnnoTest.cannotRemoveAnnotation
org.optaplanner.core.config.anno.entity.TestdataAnnoTest.cannotOverrideToDifferentAnnotation
{code}

rsynek:
@ge0ffrey @yurloc this PR has been opened for quite a while, please consider either merging or closing it. Thanks.

ge0ffrey:
My bad, on my todo list, hopefully mid next week.

rsynek:
@ge0ffrey no problem, just thought it would be good to clean up the PR queue

triceo:
Jenkins please retest this.

triceo:
@yurloc and I have agreed that we'll close this without merging. The scope of this work was never agreed upon and since the PR was left untouched for 2 years, the issue obviously isn't important.

PR #340, opened 2017-10-03
PLANNER-683 Add exception propagation test for Partitioned Search

yurloc:
See https://issues.jboss.org/browse/PLANNER-683.

* Added test that verifies propagation of the original exception that caused one of child partition solvers to fail.
* Child thread termination is handled by `DefaultPartitionedSearchPhase`. This piece of code is hard to test but I think it's covered by `awaitTermination()` in the test. There are no thread sleeps and the tests are quick.

As a result of the small changes accompanying the new tests, optaplanner behaves nicely and terminates all children when one of them fails or when the main thread is interrupted.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
@@ -127,7 +127,7 @@ public class DefaultPartitionedSearchPhase<Solution_> extends AbstractPhase<Solu
                         partitionQueue.addFinish(partIndex);
                     } catch (Throwable throwable) {
                         // Any Exception or even Error that happens here (on a partition thread) must be stored
-                        // in the partitionQueue in order to be propagated to the solver thread
+                        // in the partitionQueue in order to be propagated to the solver thread.
                         partitionQueue.addExceptionThrown(partIndex, throwable);
                     }
                 });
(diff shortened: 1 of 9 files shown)


yurloc:
Squash advised.

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseTest.java
```diff
@@ -99,38 +135,92 @@ private static void setPartSize(SolverConfig solverConfig, int partSize) {
...
+
+        executor.shutdown();
+        assertTrue(executor.awaitTermination(100, TimeUnit.MILLISECONDS));
+        assertNotNull(solvedSolution.get());
+    }
+
+    @Test(timeout = 5000)
+    // FIXME rename, because this test interrupts the main thread in PQueue iterator!!
+    // TODO add another test that will interrupt one of the PartSolver threads (will require custom ThreadFactory that
+    // will provide access to created threads)
```
yurloc: Not sure if this is doable or desired.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -151,34 +144,61 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
-            childThreadPlumbingTermination.terminateChildren();
+            if (Thread.interrupted()) {
+                // 1a. If current thread is interrupted, propagate interrupt signal to children by initiating
+                // abrupt shutdown.
+                executor.shutdownNow();
+            } else {
+                // 1b. Otherwise, initiate graceful shutdown of the executor. This allows partition solvers to finish
+                // solving upon detecting the termination issued in the next step (2). Shutting down the executor
+                // service is important because the JVM cannot exit until all nondaemon threads have terminated.
+                executor.shutdown();
```
yurloc: It is noteworthy that this PR reintroduces thread pool shutdown at the beginning of the finally block, which is in contradiction to e88dd1670 where it was reverted. However this time the shutdown is justified and supported by accompanying tests and explained in comments above.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -151,34 +144,61 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
-                // First wait for solvers to terminate voluntarily (because we have just issued children termination)
-                if (!threadPoolExecutor.awaitTermination(1, TimeUnit.SECONDS)) {
+                final int awaitingSeconds = 1;
+                if (!executor.awaitTermination(awaitingSeconds, TimeUnit.SECONDS)) {
                     // Some solvers refused to complete. Busy threads will be interrupted in the finally block
-                    logger.warn("{}Partitioned Search threadPoolExecutor didn't terminate within timeout (1 second).",
-                            logIndentation);
+                    logger.warn("{}Partitioned Search threadPoolExecutor didn't terminate within timeout ({} second).",
+                                logIndentation,
+                                awaitingSeconds);
```
yurloc: Tests provide line coverage for this `if`'s block but they don't provide mutation coverage as it is impossible (hard) to detect logging statements. Should we throw an exception instead?

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -151,34 +144,62 @@ public void solve(DefaultSolverScope<Solution_> solverScope) {
...
-                // First wait for solvers to terminate voluntarily (because we have just issued children termination)
-                if (!threadPoolExecutor.awaitTermination(1, TimeUnit.SECONDS)) {
-                    // Some solvers refused to complete. Busy threads will be interrupted in the finally block
-                    logger.warn("{}Partitioned Search threadPoolExecutor didn't terminate within timeout (1 second).",
-                            logIndentation);
+
+            if (Thread.currentThread().isInterrupted()) {
+                // 2a. If the current thread is interrupted, propagate interrupt signal to children by initiating
+                // an abrupt shutdown.
+                executor.shutdownNow();
```
yurloc: I still think it might be worthy to await thread pool termination on this branch. I just don't have the supporting arguments at this moment.

ge0ffrey: Agreed. from shutdownNow() javadoc: "This method does not wait for actively executing tasks to
     * terminate.  Use {@link #awaitTermination awaitTermination} to
     * do that."

ge0ffrey: Please change indeed, I 'll wait for that change to merge

yurloc:
Please be aware that I've added a test dependency on AssertJ since the last review. I've done it to simplify checking a root cause of an expected exception. I suppose this is OK as migration of custom asserts to AssertJ was already requested in [PLANNER-649](https://issues.jboss.org/browse/PLANNER-649).

Fun note: Not even AssertJ is resilient to a `t.getCause() == t` situation which would cause an infinite loop. But that's no longer our problem :-) If we encounter such condition, we can contribute a patch AssertJ instead of maintaining our code.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
yurloc: Done.

ge0ffrey:
Hmm, sneaking in assertj :) However in PLANNER-649 it says "I think we decided not to pursue this further, especially in the light of what's coming with JUnit 5. If that's the case, we should probably resolved "won't fix""

ge0ffrey:
As discussed, merged for now, but the assertj usage will be reviewed and potentially rolled back.

PR #347, opened 2017-11-01
JDK 9 fixes

psiroky:
These should be last pieces to make the default `mvn clean install -Dfull` build work on JDK9.

### DIFF
--- a/optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
+++ b/optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
@@ -58,6 +58,10 @@
       <scope>runtime</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.spec.javax.xml.bind</groupId>
+      <artifactId>jboss-jaxb-api_2.2_spec</artifactId>
+    </dependency>
 
     <!-- Test -->
     <dependency>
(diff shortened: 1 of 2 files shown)


optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
```diff
@@ -58,6 +58,12 @@
       <scope>runtime</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <!-- Required by JDK9. Configured as 'provided' to avoid the inclusion in the transitive dependency tree. -->
+      <groupId>org.jboss.spec.javax.xml.bind</groupId>
+      <artifactId>jboss-jaxb-api_2.2_spec</artifactId>
+      <scope>provided</scope>
```
ge0ffrey: Are we sure this needs to be scope=provided and not optional=true?
https://stackoverflow.com/questions/5860426/optional-dependency-with-scope-provided-in-maven

Is there a stackoverflow question specifically around jaxb-api being in JDK 8 and not in JDK 9 and how to handle that in maven?

psiroky: I  don't think there is a question about that. My reasoning is the following:

For the JDK8 the jaxb-api is provided (and enabled) by the standard JRE installation, so users don't need to do anything.

For the JDK9 the jaxb-api is by default disabled and users need to provide that themselves (either by depending on the spec jar or by using `--add-modules java.se.ee`)

In case the jaxb-api is actually optional for this module, then we should definitely mark it as such.

PR #349, opened 2017-11-01
[DROOLS-1776] add Automatic-Module-Name MANIFEST attribute

psiroky:


psiroky:
This needs be merged together will the other related PRs:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/596
https://github.com/kiegroup/droolsjbpm-knowledge/pull/279
https://github.com/kiegroup/drools/pull/1575
https://github.com/kiegroup/optaplanner/pull/349
https://github.com/kiegroup/jbpm/pull/1054
https://github.com/kiegroup/droolsjbpm-integration/pull/1256
https://github.com/kiegroup/kie-uberfire-extensions/pull/59
https://github.com/kiegroup/kie-wb-common/pull/1274
https://github.com/kiegroup/drools-wb/pull/697
https://github.com/kiegroup/optaplanner-wb/pull/235
https://github.com/kiegroup/jbpm-designer/pull/696
https://github.com/kiegroup/jbpm-wb/pull/936
https://github.com/kiegroup/kie-wb-distributions/pull/648
https://github.com/kiegroup/kie-docs/pull/491

optaplanner-integration/optaplanner-core-gwt/pom.xml
```diff
@@ -22,6 +22,10 @@
   </description>
   <url>https://www.optaplanner.org</url>
 
+  <properties>
+    <java.module.name>org.optaplanner.core.gwt</java.module.name>
```
ge0ffrey: This one is a pain:
- optaplanner-core uses org.optaplanner.core. If this pom uses org.optaplanner.core.gwt, they both start with "org.optaplanner.core". I am not sure if JDK 9 even allows that if we would add code here - this would clearly be a split packages violation.
- On the other hand - optaplanner-core-gwt is a "gwt-lib" wrapper of optaplanner-core (the new tbroyer plugin), so maybe they should use the same module name? I doubt JDK will like that any better - but if there are no JDK 9 jars in play, it's could work.
- Can we get a way with not declaring a java.module.name for this one at all?

In any case, I am raising this question upstream at the tbroyer plugin - the recommended way needs to be decided there.

ge0ffrey: Upstream question raised here:
  https://github.com/tbroyer/gwt-maven-plugin/issues/98

ge0ffrey: @psiroky To allow this set of PR's to be merged, I suggest we change the module name to "org.optaplanner.coregwt" (so without the dot between core and gwt). That means no split package, a unique name, etc - and we can change it later we look at making the gwt stuff work on jdk 9 too.

psiroky: I don't think that `org.optaplanner.core` and `org.optaplanner.core.gwt` qualifies as split packages -- as long as as the `org.optaplanner.core.gwt` includes only classes in that package or its sub-packages.

I can of course change it as you suggested, but that could mean changing dozens of other names as well in other repos (if we want to be consistent) - e.g. there modules like `org.drools.examples.api`, `org.drools.examples.cdi` or `org.drools.examples.somethingelse`.

The reason why I specified the module name in all `<packaging>jar</packaging>` modules is that otherwise the attribute would be empty. I unfortunately have not found a way to only store the attribute if the property has non-empty value (not sure maven-jar-plugin can do that).

psiroky: Ah, I am sorry my fingers were faster than my head :) I now see your point with split packages. I'll think about this a bit more...

psiroky: OK, so I think we should go either with `org.optaplanner.coregwt` or `org.optaplanner.gwt.core` (but this kind of assumes that the modules is about core GWT stuff which is probably not what we want).

optaplanner-webexamples/pom.xml
```diff
@@ -25,6 +25,7 @@
   <properties>
     <version.org.wildfly>10.1.0.Final</version.org.wildfly>
   </properties>
+
   <build>
```
yurloc: @psiroky Was this intended for formatting consistency? Please confirm that the webexamples module doesn't need the `java.module.name` property.

psiroky: Probably not intended. The property is not needed uses packaging `war`. We only want module names for jars.

PR #351, opened 2017-11-06
PLANNER-804 Don't log early termination twice

yurloc:


ge0ffrey:
Out of curiosity, when is it possible to interrupt twice?

yurloc:
The interruption happens only once but termination check happens multiple times during solver lifecycle.

I launched Optashift on WildFly, started solving and then killed the server with `SIGTERM`. I modified the logging statement to include thread stack trace:
```java
logger.error("The solver thread got interrupted, so this solver is terminating early.",
             new Exception("DUMMY"));
```
which resulted in
```
15:20:14.481 ERROR [EE-ManagedExecutorService-default-Thread-1] The solver thread got interrupted, so this solver is terminating early.
java.lang.Exception: DUMMY
	at org.optaplanner.core.impl.solver.termination.BasicPlumbingTermination.isSolverTerminated(BasicPlumbingTermination.java:142)
	at org.optaplanner.core.impl.solver.termination.OrCompositeTermination.isSolverTerminated(OrCompositeTermination.java:46)
	at org.optaplanner.core.impl.solver.termination.PhaseToSolverTerminationBridge.isPhaseTerminated(PhaseToSolverTerminationBridge.java:78)
	at org.optaplanner.core.impl.localsearch.DefaultLocalSearchPhase.solve(DefaultLocalSearchPhase.java:66)
	at org.optaplanner.core.impl.solver.AbstractSolver.runPhases(AbstractSolver.java:87)
	at org.optaplanner.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:167)
	at org.optaplanner.openshift.employeerostering.server.solver.WannabeSolverManager.lambda$solve$2(WannabeSolverManager.java:82)
	...
15:20:14.480 INFO  [EE-ManagedExecutorService-default-Thread-1] Local Search phase (1) ended: time spent (8183), best score (0hard/240soft), score calculation speed (31665/sec), step total (16225).
15:20:14.481 ERROR [EE-ManagedExecutorService-default-Thread-1] The solver thread got interrupted, so this solver is terminating early.
java.lang.Exception: DUMMY
	at org.optaplanner.core.impl.solver.termination.BasicPlumbingTermination.isSolverTerminated(BasicPlumbingTermination.java:142)
	at org.optaplanner.core.impl.solver.termination.OrCompositeTermination.isSolverTerminated(OrCompositeTermination.java:46)
	at org.optaplanner.core.impl.solver.AbstractSolver.runPhases(AbstractSolver.java:85)
	at org.optaplanner.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:167)
	at org.optaplanner.openshift.employeerostering.server.solver.WannabeSolverManager.lambda$solve$2(WannabeSolverManager.java:82)
	...
```
The first message comes from LS phase that checks termination before each step. After that the phase ends but there's still the phase loop active in AbstractSolver, that also checks termination before continuing to the next phase.

So, termination is checked multiple times, which is both correct and necessary. The (cosmetic) problem is that the info about early termination should only be printed for the first time, when interruption is detected and not each time. Do you agree?

rsynek:
@ge0ffrey @yurloc this PR has been opened for quite a while, please consider either merging or closing it. Thanks.

PR #356, opened 2017-11-16
Simplify construction of uncaught exception message

yurloc:
- Eliminated duplicate code (`"Exception in thread " + ...`).
- Replaced complex stack trace construction with `e#printStackTrace(PrintWriter)`.
- The UI stack trace now preserves tabs in the Caused by section.


optaplanner-benchmark/src/main/java/org/optaplanner/swing/impl/SwingUncaughtExceptionHandler.java
```diff
@@ -49,13 +51,16 @@ public static void register() {
 
     @Override
     public void uncaughtException(Thread t, Throwable e) {
+        StringWriter sw = new StringWriter();
```
ge0ffrey: I 'll rename variable to stringWriter after merging.
(only well known variables can be very short, such as Object o, Exception e, int i, etc)

PR #358, opened 2017-11-22
 PLANNER-922 Variable Neighborhood Descent

ge0ffrey:


ge0ffrey:
@yurloc Please review :)

optaplanner-docs/src/main/asciidoc/LocalSearch/LocalSearch-chapter.adoc
```diff
@@ -619,6 +619,56 @@ The following ``finalistPodiumType``s are supported:
...
+[[variableNeighborhoodDescent]]
+== Variable Neighborhood Descent
+
+
+[[variableNeighborhoodDescentAlgorithm]]
+=== Algorithm Description
+
+Variable Neighborhood Descent iteratively tries multiple move selectors
+in original order (depleting each one entirely before trying the next one),
+accepting the first improving move (which also resets the move selector).
```
yurloc: I would `s/accepting/picking/` here to help understand this is done by a forager as part of the "pick early" strategy.

yurloc: I'd like to have a better understanding of what makes VND unique or different from simple Hill Climbing. Would it be possible to highlight what they have in common and what the differences are?

ge0ffrey: The big difference is that a hill_climbing by default will select the moves in random order. VND always does original order.
An academic hill climbing only had 1 moveSelector (=neighborhood), but we relaxed that limitation.
VND is about exploring more complex neighberhoods as simpler neighberhoods don't yield results.

yurloc: Thanks. I think such explanation will help readers understand the advantage of this algorithm and decide when it is appropriate to use it. I would also provide a reference link for those that want to learn more about the scientific background, e.g. at least https://en.wikipedia.org/wiki/Variable_neighborhood_search.

I'd also like to propose a naming change from "descent" which makes sense in the context of searching for a local minimum. Because OptaPlanner's goal is to maximize the score function I think that something like "ascent" or "climb" might make better sense. Or just `VARIABLE_NEIGHBORHOOD`. What do you think?

One more question: If VND is about exploring complex neighborhoods created by multiple selectors, shouldn't we emphasize the fact and perhaps encourage users to go for the advanced configuration? It seems to me that with the basic configuration there is only the default union of change and swap move selectors (same as with simple hill climbing configuration).

ge0ffrey: In future versions, as soon as pillar swaps etc also support original selection, VND will automatically add all generic move selectors. We also need to start writing k-swap selectors etc (swap is really just 2-swap at this point - a 3-swap would take 3 entities and rotate their values clockwise).

ge0ffrey: VNS != VND so I can't link that wikipedia and I think we should link the paper foremost. But I have a hard time identifying "the canonical paper" of VND. So I am avoiding that complication for now :/

agree on the "descent" naming problem, but I don't want to invent a different name than the academia, so I 'll just make a not of it the algo description.

I 've made an issue that includes adding more generic selectors in VND automatically: https://issues.jboss.org/browse/PLANNER-928

PR #360, opened 2017-12-01
Fix a few typos

yurloc:
Actually, the word 'therefor' exists, but it has a different meaning: https://english.stackexchange.com/questions/58616/is-there-a-difference-between-therefor-and-therefore/58617#58617. Didn't know that...


ge0ffrey:
Ironically, I already did a big "therefor" to "therefore" fix about a year ago, but I keep introducing the wrong spelling in new commits (if I forget to look it up first).

PR #361, opened 2017-12-05
Remove SolutionDao classes (replace by SolutionFileIO) to simplify examples

ge0ffrey:
A "small" refactor that had a few icebergs attached to it.
But this removes a long standing complexity in the examples code.

- SolutionDao has been removed
- Importer/Experter have been stripped of the converter responsibility, which is in a Converter class now

It's the 3th commit that's the iceberg.


yurloc:
1 test failing: org.optaplanner.examples.vehiclerouting.persistence.VehicleRoutingImporterTest.readSolution[54: data/vehiclerouting/import/usa/basic/air/usa-n100000-k500.vrp]

Stacktrace:
```
java.lang.NullPointerException: null
	at org.optaplanner.examples.vehiclerouting.persistence.VehicleRoutingImporter$VehicleRoutingInputBuilder.readSolution(VehicleRoutingImporter.java:107)
	at org.optaplanner.examples.vehiclerouting.persistence.VehicleRoutingImporter$VehicleRoutingInputBuilder.readSolution(VehicleRoutingImporter.java:69)
	at org.optaplanner.examples.common.persistence.AbstractTxtSolutionImporter.readSolution(AbstractTxtSolutionImporter.java:53)
	at org.optaplanner.examples.common.persistence.SolutionImporterTest.readSolution(SolutionImporterTest.java:81)
```

NPE in this statement:
```java
factorial(customerListSize + vehicleListSize - 1).divide(factorial(vehicleListSize - 1));
```
because `factorial()` returns null if the base is too big. Can we use [Stirling's_approximation](https://en.wikipedia.org/wiki/Stirling%27s_approximation) in order to never return null?

yurloc:
I think we don't want to use Stirling's approximation because it doesn't perform better than the precise calculation. At least it helped me find a bug in the current factorial implementation.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/AbstractSolutionImporter.java
```diff
@@ -25,107 +25,35 @@
...
     public static abstract class InputBuilder extends LoggingMain {
 
         public static BigInteger factorial(int base) {
+            if (base > 100000) {
+                // Calculation takes too long
+                return null;
+            }
             BigInteger value = BigInteger.ONE;
             for (int i = 1; i <= base; i++) {
                 value = value.multiply(BigInteger.valueOf(base));
```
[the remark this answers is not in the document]
ge0ffrey: oh crap, means ours docs in chapter are wrong too... fixing :)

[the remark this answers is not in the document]
yurloc: Please include a test with a few `[n, n!]` pairs if possible if you're going to fix the calculation.

ge0ffrey:
Suggested by Jiri fixed.

optaplanner-docs/src/main/asciidoc/UseCasesAndExamples/ExamplesOverview/ExamplesOverview-section.adoc
```diff
@@ -118,9 +118,9 @@ and also in git under [path]_optaplanner/optaplanner-examples_.
 |<<vehicleRouting,Vehicle routing>> with time windows
 |* All of Vehicle routing
 * 1 shadow variable
-|* Entity <= `1000`
-* Value <= 1250
-* Search space <= `10^3000`
+|* Entity <= `2740`
+* Value <= `2795`
+* Search space <= `10^8380`
```
yurloc: According to my observation, this search space size (10^8380) is a result of having 2740 customers and 55 vehicles. What I can see here looks more like that `entities = customers` and `values = customers + vehicles`? Since vehicle is both an entity and a value and customers are values only, I'd expect it to be `entities = vehicles` and `values = vehicles + customers`. So
```
Entity <= 55
Value <= 2795
```
Wdyt?

ge0ffrey: As it stands today:
- Customer is a genuine PlanningEntity
- Vehicle is a shadow PlanningEntity, so I don't count it towards the entities as those don't affect scalability

That being said, the new VRP model alternative with lists will impact this section somehow.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/AbstractSolutionImporter.java
```diff
@@ -25,117 +25,48 @@
 import org.optaplanner.core.api.domain.solution.PlanningSolution;
 import org.optaplanner.examples.common.app.LoggingMain;
 import org.optaplanner.examples.common.business.ProblemFileComparator;
+import org.optaplanner.persistence.common.api.domain.solution.SolutionFileIO;
```
yurloc: Unused import (+2 more in this file).

yurloc:
@ge0ffrey Please fix the unused imports (~21) if you want. Otherwise I'll open a general imports clean-up PR later. We should probably turn on Checkstyle soon to catch this automatically. The rules are there, we just need to fail the build on error.

yurloc:
Do we want to also rename all `*DaoTest`s? It may be misleading since there is now no trace of DAOs left. I haven't thought about a better name.

ge0ffrey:
I 'll deal with the unused imports in a separate PR (and adjust my settings and checkstyle then too). They're mostly harmless, but there are other instances across the code. (Either we fix them all or we don't.)
Renaming the DaoTests and merging

sgtlove:
Good day to you.

My name is Sgt.Leigh A. Hester,i will like to be your friend ,if you care please reply me back in my private email ( Hester.Ann.leigh87@gmail.com ) So that i will tell you more about my self .

PR #364, opened 2018-01-02
Fix drl accumulate bug reported on stackoverflow

yurloc:
1. There's one more score DRL that suffers from the bug. It's in the cheap time scheduling example.
2. I think the alternative answer by the.wizard is actually better as it doesn't require casting.


ge0ffrey:
@yurloc Might you have an url to the stackoverflow issue?

ge0ffrey:
Thanks Jiri, I already fixed the project one some time ago and now fixed the cheap one differently.
I prefer the new accumulate style becuase it's less verbose, so I 've gone with that approach

PR #366, opened 2018-01-08
PLANNER-935 Lambda's instead of reflection

ge0ffrey:
Open questions:
- When you run this locally, do new dirs appear in the working directory?
- Shouldn't we use a lambda variant for MethodMemberAccessor too?


ge0ffrey:
With this change, **PatientAdmissionScheduleBenchmarkConfigTest becomes 46 seconds**... that's because the code generation isn't cached. The DRL compilation isn't cached either AFAIK, but that didn't impact multitenancy nearly as bad.

optaplanner-docs/src/main/asciidoc/ConstructionHeuristics/ConstructionHeuristics-chapter.adoc
```diff
@@ -758,14 +758,14 @@ Scales better, at the cost of solution quality. The order of the planning variab
...
 ending the Construction Heuristic 3 times faster.
 Especially for three or more planning variables, the scaling difference is huge.
-For example, with three variables of 1000 values each,
-a cartesian product selects 1000000000 moves per entity (1 step per entity).
-A sequential approach only selects 240 moves per entity (3 steps per entity),
-ending the Construction Heuristic 300 000 times faster.
+For example, with three variables of 1{nbsp}000 values each,
+a cartesian product selects 1{nbsp}000{nbsp}000{nbsp}000 moves per entity (1 step per entity).
+A sequential approach only selects 3{nbsp}000 moves per entity (3 steps per entity),
+ending the Construction Heuristic 300{nbsp}000 times faster.
```
yurloc: This will definitely improve readability but shouldn't we use the en-US digit grouping style that uses a comma instead of a short space? E.g. 1,000,000,000.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactoryTest.java
```diff
@@ -0,0 +1,146 @@
...
+                MemberAccessorFactory.MemberAccessorType.FIELD_OR_GETTER_METHOD_WITH_SETTER, ProblemFactProperty.class);
+        assertEquals("privateField", memberAccessor.getName());
+        assertEquals(String.class, memberAccessor.getType());
+
+        TestdataVisibilityModifierSolution s1 = new TestdataVisibilityModifierSolution("s1",
+                "firstValue", "n/a",
+                "n/a", "n/a", "n/a", "n/a");
+        assertSame("firstValue", memberAccessor.executeGetter(s1));
+        memberAccessor.executeSetter(s1, "secondValue");
+        assertSame("secondValue", memberAccessor.executeGetter(s1));
```
yurloc: Is it guaranteed that these two String literals are represented by the same String instance? I think they are as a result of JVM optimization but I wouldn't rely on that. If I'm correct, this should be changed to `assertEquals()` or
```java
String v2 = "secondValue";
memberAccessor.executeSetter(s1, v2);
assertSame(v2, memberAccessor.executeGetter(s1));
```

ge0ffrey: Agreed (Bad copy paste from other test.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LamdbaBeanPropertyMemberAccessor.java
```diff
@@ -0,0 +1,156 @@
...
+import java.lang.reflect.Type;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+import org.optaplanner.core.impl.domain.common.ReflectionHelper;
+
+/**
+ * A {@link MemberAccessor} based on a getter and optionally a setter.
+ */
+public final class LamdbaBeanPropertyMemberAccessor implements MemberAccessor {
```
rsynek: Typo - "Lamdba" vs "Lambda"

optaplanner-docs/src/main/asciidoc/ConstructionHeuristics/ConstructionHeuristics-chapter.adoc
ge0ffrey: It's an interesting point. If the article follows a locale, it is indeed en_US (so "optimizer", not "optimiser" as in en_UK).
But the numbers are following the engineering notation currently: https://en.wikipedia.org/wiki/Engineering_notation, so "1 000 000.00". Dates follow the ISO notation: 2001-9-11 (instead of the very confusing 9/11/2001 which can also mean 9 november for many readers).

ge0ffrey: Hmm, wikepedia says for English: "Digits should be grouped and separated either by commas or by narrow gaps (never a period/full point)"
https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Dates_and_numbers
So no decision. I find gaps much more readable for international audiences, so let's stick with that (but you do have a good point, was worth bringing this up).

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+        TestdataVisibilityModifierSolution s1 = new TestdataVisibilityModifierSolution("s1",
+                "n/a", "firstValue",
+                "n/a", "n/a", "n/a", "n/a");
+        assertEquals("firstValue", memberAccessor.executeGetter(s1));
+        memberAccessor.executeSetter(s1, "secondValue");
+        assertEquals("secondValue", memberAccessor.executeGetter(s1));
+    }
+
+    @Test @Ignore("Non-public getters/setters are not supported")
```
yurloc: @ge0ffrey Why not remove the test instead of ignoring it? Is there any chance the test will be unignored in the future? I guess it won't be.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+public class MemberAccessorFactoryTest {
+
+    @Test
+    public void fieldAnnotatedEntity() throws NoSuchFieldException {
+        MemberAccessor memberAccessor = MemberAccessorFactory.buildMemberAccessor(
+                TestdataFieldAnnotatedEntity.class.getDeclaredField("value"),
+                MemberAccessorFactory.MemberAccessorType.FIELD_OR_GETTER_METHOD_WITH_SETTER, PlanningVariable.class);
+        assertEquals("value", memberAccessor.getName());
+        assertEquals(TestdataValue.class, memberAccessor.getType());
```
rsynek: The test for public property includes also assertion regarding the MemberAccessor class - `assertInstanceOf(...)`. It would be worth adding it into every test method in this test case class.

And just a suggestion, I personally prefer using AssertJ for assertions instead of plain jUnit. It provides very nice messages when a test fails, has really cool capabilities when it comes to e.g. (not only) checking collections. It's already among this project's dependencies, so we can use it with every new test, if you agree.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LambdaBeanPropertyMemberAccessor.java
```diff
@@ -0,0 +1,156 @@
...
+public final class LambdaBeanPropertyMemberAccessor implements MemberAccessor {
+
+    private final Class<?> propertyType;
+    private final String propertyName;
+    private final Method getterMethod;
+    private final Function getterFunction;
+    private final Method setterMethod;
+    private final BiConsumer setterFunction;
+
+    public LambdaBeanPropertyMemberAccessor(Method getterMethod) {
```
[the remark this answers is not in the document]
ge0ffrey: I extracted createGetterFunction and createSetterFunction.

Simply extracting the .getTarget().invokeExact() method isn't worth it to me, especially with all the "the compiler needs to understand what's going on to inline it" discussion I've been hearing from the jvm architects. So for this specific case, code reuse loses vs making it inline friendly (which is only the right call for less than 1% of any codebase).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactory.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public static MemberAccessor buildMemberAccessor(Member member, MemberAccessorType memberAccessorType,
+            Class<? extends Annotation> annotationClass) {
+        if (member instanceof Field) {
+            Field field = (Field) member;
+            return new ReflectionFieldMemberAccessor(field);
+        } else if (member instanceof Method) {
+            Method method = (Method) member;
+            MemberAccessor memberAccessor;
+            switch (memberAccessorType) {
```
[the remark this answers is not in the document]
ge0ffrey: I tried that, but I didn't think it made the logic easier to follow.
Agreed it could be cleaner, but I am not sure how currently.
I' d rather focus on getting rid of the diversification between ReflectionMethodMemberAccessor and ReflectionBeanPropertyMemberAccessor first. (and their lamdba counterparts), but that's a separate issue.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactoryTest.java
ge0ffrey: @yurloc and @triceo also asked to start using assertJ. There's a jira around it. I am ready to accept that, but I want to first see the junit 5 migration happen, which is 3 steps:
- Switch to the junit 5 platform runner on kie-parent (1 file change probably, very little work) and hook in junit-vintage by default.
- Migrate all optaplanner tests to junit 5. I've done the migration from 3 to 4 and it took me about a day's worth of work, by using regex tricks: https://stackoverflow.com/a/4535170/472109
- Remove junit-vintage from kie-parent, add it on all repo roots, except for optaplanner where we add junit-jupiter

ge0ffrey: If you want to take it on:
https://issues.jboss.org/browse/PLANNER-1003

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactoryTest.java
ge0ffrey: Done, thanks

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/MemberAccessorFactoryTest.java
ge0ffrey: Done assertInstanceOf() in other tests, thanks :)

PR #369, opened 2018-01-22
[Do not MERGE yet] RHDM-251 - wildfly 11 upgrade

mareknovotny:


optaplanner-webexamples/pom.xml
```diff
@@ -23,7 +23,7 @@
   <url>https://www.optaplanner.org</url>
 
   <properties>
-    <version.org.wildfly>10.1.0.Final</version.org.wildfly>
+    <version.org.wildfly>11.0.0.Final</version.org.wildfly>
```
ge0ffrey: Has it been tested that the optaplanner-webexamples still work? Especially their arquillain test?

We need to move this version.org.wildfly to kie-parent or ip-bom. But I am even more worried about the relationship with the arquillian versions - that stuff is really brittle.

mareknovotny:
hey @ge0ffrey, I am not aware of optaplanner-webexamples repository, please point me where it is to try. Btw i am on pto, so I would check that later next week that ;)

ge0ffrey:
@mareknovotny optaplanner-webexamples is a module in this repository. It builds a simple war file to deploy on WildFly and has an arquillian test.

mareknovotny:
ok, I ran full downstream build on CI. If the examples does not run in that I will look at it manual verification on Monday when I came back.

PR #371, opened 2018-02-02
PLANNER-983: Exposed Indictment, ConstraintMatch and ConstraintMatchTotal to GWT and added their Jackson Serializers/Deserializers

Christopher-Chianelli:


optaplanner-integration/optaplanner-core-gwt/src/main/module.gwt.xml
```diff
@@ -4,6 +4,9 @@
 <module>
   <!-- Starts from org.optaplanner.core -->
   <source path="api/score/">
-    <include name="**/*Score.java"/>
+      <include name="**/*Score.java"/>
```
ge0ffrey: xml is 2 space idents

optaplanner-persistence/optaplanner-persistence-jackson/pom.xml
```diff
@@ -58,6 +58,17 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
+    <!-- Testing -->
+      <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
```
ge0ffrey: We standarize on Mockito, so we don't want to use powermock, as that will increase complexity (harder for developers to read code as they now have to know 2 mocking frameworks) and decrease reuse opportunity (see PlannerAssert.java which already has Mockito helper methods specifically for optaplanner interfaces).

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonDeserializer.java
```diff
@@ -0,0 +1,43 @@
...
+                try {
+                    justificationList.add(DeserializationUtils.deserializePolymorphicProperty(context, t,
+                            "justification"));
+                } catch (Exception e) {
+                    throw new IllegalStateException(e);
+                }
+            });
+
+            return new ConstraintMatch(constraintPackage, constraintName, justificationList, score);
+        } catch (ClassNotFoundException e) {
```
ge0ffrey: try catch much too wide

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonDeserializer.java
```diff
@@ -0,0 +1,43 @@
...
+            String constraintName = tree.get("constraintName").asText();
+
+            Score<?> score = DeserializationUtils.deserializePolymorphicProperty(context, tree, "score");
+
+            List<Object> justificationList = new ArrayList<>();
+            tree.get("justificationList").forEach((t) -> {
+                try {
+                    justificationList.add(DeserializationUtils.deserializePolymorphicProperty(context, t,
+                            "justification"));
+                } catch (Exception e) {
```
ge0ffrey: We never catch Exception or RuntimeException (unless there is a very very very good reason to do so)

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/DeserializationUtils.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class DeserializationUtils {
+
+    @SuppressWarnings("unchecked")
+    public static <T> T deserializePolymorphicProperty(DeserializationContext context, JsonNode tree, String property)
+            throws ClassNotFoundException, JsonParseException, JsonMappingException, IOException {
+        if (!tree.has(property) || !tree.has(property + "Class")) {
+            throw new IllegalStateException("Tree does not have specified property and it class infomation! Tree:\n"
+                    + tree.toString());
```
ge0ffrey: doesn't follow standard exception message format (see optaplanner docs chapter 17):
"Tree" => "The tree"
"specified property" => "specified property (<value filled in>)"
"information!" => we don't yell, we're professionals, we don't get emotional ;)
tree.toString() -> "tree (" + tree + ").

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/DeserializationUtils.java
```diff
@@ -0,0 +1,30 @@
...
+        }
+        ObjectMapper mapper = new ObjectMapper();
```
ge0ffrey: A Deserializer that creats an ObjectMapper?? That that doesn't make sense?

optaplanner-persistence/optaplanner-persistence-jackson/pom.xml
Christopher-Chianelli: I removed Powermock as I found another solution when solving a different a different issue. However, Powermock shares Mockito API: All Mockito methods are defined in Powermock Mockito API (So you only need to know one mocking framework). All Powermock adds is the ability to mock constructors, final classes and static methods. Due to poor documentation on how to configure jackson in unit tests, I used it simulate ObjectMapper injection.

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/DeserializationUtils.java
Christopher-Chianelli: Fixed; learned about parser.getCodec() which supplies the ObjectMapper the parser using.

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonSerializerAndDeserializerTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+import org.optaplanner.core.api.score.constraint.ConstraintMatch;
+import org.optaplanner.persistence.jackson.api.score.ScoreJacksonJsonSerializer;
+import org.optaplanner.persistence.jackson.api.score.buildin.hardsoft.HardSoftScoreJacksonJsonDeserializer;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConstraintMatchJacksonJsonSerializerAndDeserializerTest {
+
+    ObjectMapper mapper;
```
ge0ffrey: code encapsulation: make it private or protected. declaring nothings means "friendly", which usually not what we want

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchTotalJacksonJsonDeserializer.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private ConstraintMatchJacksonJsonDeserializer constraintMatchDeserializer;
+
+    public ConstraintMatchTotalJacksonJsonDeserializer() {
+        constraintMatchDeserializer = new ConstraintMatchJacksonJsonDeserializer();
+    }
+
+    @Override
+    public ConstraintMatchTotal deserialize(JsonParser parser, DeserializationContext context) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
```
ge0ffrey: Isn't creating a new ObjectMapper as part of a deserialiation problematic, because of the bootstrap cost and because it doesn't inherit the original's ObjectMapper's configuration? For example, one of the justification objects might be a Foo object for which on the origional ObjectMapper a special (de)serializer was configured.

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonSerializerAndDeserializerTest.java
```diff
@@ -0,0 +1,82 @@
...
+        TestConstraintMatchWrapper wrapper = new TestConstraintMatchWrapper(constraintMatch);
+
+        try {
+            String json = mapper.writeValueAsString(wrapper);
+
+            TestConstraintMatchWrapper other = mapper.readValue(json, TestConstraintMatchWrapper.class);
+
+            assertEquals(constraintMatch, other.getConstraintMatch());
+        } catch (JsonParseException e) {
+            throw new IllegalStateException(e);
```
ge0ffrey: interestingly enough, for test code, we are totally fine with test helper methods that have a long throws method signature so I wouldn't bother wrapping these.

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonSerializerAndDeserializerTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        mapper = new ObjectMapper();
+        mapper.registerModule(new SimpleModule()
+                .addSerializer(HardSoftScore.class, new ScoreJacksonJsonSerializer<HardSoftScore>())
+                .addDeserializer(HardSoftScore.class, new HardSoftScoreJacksonJsonDeserializer()));
+    }
+
+    private void testSerializeAndDeserialize(String packageName, String name, Score score, Object... list) {
```
ge0ffrey: the prefix "test" in the name of this method is misleading, as in junit 4 era that were suppose to be test methods (and this doesn't have a @Test annotation because it's not a test)

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/IndictmentJacksonJsonSerializerAndDeserializerTest.java
```diff
@@ -0,0 +1,97 @@
...
+    @Test
+    public void testSerializeAndDeserialize() {
+        testSerializeAndDeserialize("Hello", S(0, 0));
+        testSerializeAndDeserialize(1, S(0, 1), CM("package", "name", 0, 10, "Bad"));
+        testSerializeAndDeserialize("What", S(0, 1), CM("package", "name", 0, 10, "Bad"), CM("package", "b", 10, 20,
+                "Very Bad"));
+    }
+
+    private static HardSoftScore S(int h, int s) {
+        return HardSoftScore.valueOf(h, s);
```
ge0ffrey: Is this method worth it's weight in characters written?

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchTotalJacksonJsonDeserializer.java
Christopher-Chianelli: Thought I changed all of them, forgot I also used the ObjectMapper techique outside of DeserializationUtils; thanks, fixing now.

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/IndictmentJacksonJsonSerializerAndDeserializerTest.java
Christopher-Chianelli: In this example no, since `CM(...)` calls it implicitly. In `ConstraintMatchTotalJacksonJsonSerializerAndDeserializerTest`, it is (otherwise line 66 would be much longer). If I make `CM(...)` accept a score rather than 2 integers, then it'll be worth its weight (plus we can test multiple score types).

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonSerializerAndDeserializerTest.java
Christopher-Chianelli: Will rename to `assertSerializeAndDeserialize` to match `AbstractScoreJacksonJsonSerializerAndDeserializerTest` method this is based on.

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonSerializerAndDeserializerTest.java
Christopher-Chianelli: That somewhat contradicts lines 36 to 42 of `AbstractScoreJacksonJsonSerializerAndDeserializer`:
```
try {
            ObjectMapper objectMapper = new ObjectMapper();
            jsonString = objectMapper.writeValueAsString(input);
            output = (W) objectMapper.readValue(jsonString, input.getClass());
        } catch (IOException e) {
            throw new IllegalStateException("Marshalling or unmarshalling for input (" + input + ") failed.", e);
        }
```

ge0ffrey:
This one truly is pandora's box. Especially the requirement that a ConstraintMatch's Score field should just delegate back to the normal serializer (which loaded the OptaPlannerModule), exposed a deep seeded problem with the difference of serializing a field of type Score or a field of type HardSoftScore, even if both contain a HardSoftScore instance:
https://stackoverflow.com/questions/49193869/jackson-serialize-animal-and-cat-with-2-different-serializers-without-field-ann/49197886#49197886

optaplanner-persistence/optaplanner-persistence-jackson/src/main/java/org/optaplanner/persistence/jackson/api/score/constraint/ConstraintMatchJacksonJsonDeserializer.java
```diff
@@ -0,0 +1,34 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.api.score.constraint.ConstraintMatch;
+import org.optaplanner.persistence.jackson.api.score.DeserializationUtils;
+
+public class ConstraintMatchJacksonJsonDeserializer extends JsonDeserializer<ConstraintMatch> {
+
+    @Override
+    public ConstraintMatch deserialize(JsonParser parser, DeserializationContext context) throws IOException {
+        JsonNode tree = parser.getCodec().readTree(parser);
```
ge0ffrey: readTree() is a DOM approach and it has 2 big disadvantages:
- it parses and buffers the entire subtree. This increases memory usage and lowers performance of the parsing
- the parser can't simply be given to the score deserializer (instead, a new parser is created on the subtree).

I am using the nextToken() approach to avoid this DOM approach.

Christopher, just mentioning this FYI,  to understand that in any type of parser (xml, json, ...) there are often 2 approaches: DOM (= user friendly) and non-DOM (= cpu & memory-friendly).

ge0ffrey:
Closing this PR without merging:
- It has been partially replaced by https://github.com/kiegroup/optaplanner/pull/380 (which was based on this PR's code). That one delegates polymorphic Score serialization back to the root object mapper (instead of doing it itself), it also never uses readTree() for performance reasons and is not exploitable.
- ConstraintMatch API itself is not yet Jackson supported, as for why see comments on https://issues.jboss.org/browse/PLANNER-1042

PR #372, opened 2018-02-05
PLANNER-706 fix performance drop caused by using computeIfAbsent method

rsynek:
Running the most severely impacted benchmarks locally shows improvement by ~25 %. We need to run complete benchmarks to be sure this is the right fix, but all the evidence point to the `computeIfAbsent` method.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
@@ -42,9 +42,8 @@ import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
+import java.util.concurrent.ConcurrentMap;
 import org.apache.commons.lang3.tuple.Pair;
 import org.optaplanner.core.api.domain.solution.PlanningSolution;
 import org.optaplanner.core.api.domain.solution.cloner.DeepPlanningClone;
@@ -52,6 +51,7 @@ import org.optaplanner.core.api.domain.solution.cloner.SolutionCloner;
 import org.optaplanner.core.impl.domain.common.ReflectionHelper;
 import org.optaplanner.core.impl.domain.common.accessor.MemberAccessor;
 import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.domain.solution.util.ConcurrentCache;
 
 /**
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
@@ -60,10 +60,10 @@ public class FieldAccessingSolutionCloner<Solution_> implements SolutionCloner<S
 
     protected final SolutionDescriptor<Solution_> solutionDescriptor;
 
-    protected final ConcurrentMap<Class<?>, Constructor<?>> constructorCache = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<Class<?>, List<Field>> fieldListCache = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<Pair<Field, Class<?>>, Boolean> deepCloneDecisionFieldCache = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<Class<?>, Boolean> deepCloneDecisionActualValueClassCache = new ConcurrentHashMap<>();
+    protected final ConcurrentMap<Class<?>, Constructor<?>> constructorCache = new ConcurrentCache<>();
+    protected final ConcurrentMap<Class<?>, List<Field>> fieldListCache = new ConcurrentCache<>();
+    protected final ConcurrentMap<Pair<Field, Class<?>>, Boolean> deepCloneDecisionFieldCache = new ConcurrentCache<>();
+    protected final ConcurrentMap<Class<?>, Boolean> deepCloneDecisionActualValueClassCache = new ConcurrentCache<>();
 
     public FieldAccessingSolutionCloner(SolutionDescriptor<Solution_> solutionDescriptor) {
         this.solutionDescriptor = solutionDescriptor;
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
@@ -1,0 +1,31 @@
+package org.optaplanner.core.impl.domain.solution.util;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Thread-safe basic cache implementation.
+ *
+ * @param <K> the type of keys for the cache
+ * @param <V> the type of values stored in the cache
+ * */
+public final class ConcurrentCache<K, V> extends ConcurrentHashMap<K, V> {
+
+    /**
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} should not be used to directly retrieve cached values
+     * as it incorporates locking even when the value is present and doesn't have to be computed. Thus, the value is
+     * retrieved by {@link ConcurrentHashMap#get(Object)} and if and only if the value is null,
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} is used to ensure thread-safety.
+     *
+     * {@inheritDoc}
+     * */
+    @Override
+    public V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
+        V value = get(key);
+        if (value == null) {
+            value = super.computeIfAbsent(key, mappingFunction);
+        }
+        return value;
+    }
+
+}
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -34,10 +34,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 import com.google.common.collect.Iterators;
+import java.util.concurrent.ConcurrentMap;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.optaplanner.core.api.domain.autodiscover.AutoDiscoverMemberType;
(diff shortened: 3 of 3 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/Cache.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.function.Function;
+
+/**
+ * Simple cache interface.
+ *
+ * @param <K> the type of keys for the cache
+ * @param <V> the type of values stored in the cache
+ * */
+public interface Cache<K, V> {
```
triceo: Did you consider extending `ConcurrentHashMap`, overriding `computeIfAbsent(...)`? Seems that introducing this interface is not strictly necessary.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
```diff
@@ -0,0 +1,41 @@
...
+ * @param <V> the type of values stored in the cache
+ * */
+public final class ConcurrentCache<K, V> implements Cache<K, V> {
+
+    private final ConcurrentMap<K, V> cache = new ConcurrentHashMap<>();
+
+    /**
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} should not be used to directly retrieve cached values
+     * as it incorporates locking even when the value is present and doesn't have to be computed. Thus, the value is
+     * retrieved by {@link ConcurrentHashMap#get(Object)} and iff the value is null,
```
triceo: Typo.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
```diff
@@ -60,10 +60,10 @@
...
     protected final SolutionDescriptor<Solution_> solutionDescriptor;
 
-    protected final ConcurrentMap<Class<?>, Constructor<?>> constructorCache = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<Class<?>, List<Field>> fieldListCache = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<Pair<Field, Class<?>>, Boolean> deepCloneDecisionFieldCache = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<Class<?>, Boolean> deepCloneDecisionActualValueClassCache = new ConcurrentHashMap<>();
+    protected final Cache<Class<?>, Constructor<?>> constructorCache = new ConcurrentCache<>();
+    protected final Cache<Class<?>, List<Field>> fieldListCache = new ConcurrentCache<>();
+    protected final Cache<Pair<Field, Class<?>>, Boolean> deepCloneDecisionFieldCache = new ConcurrentCache<>();
+    protected final Cache<Class<?>, Boolean> deepCloneDecisionActualValueClassCache = new ConcurrentCache<>();
```
triceo: A map to `Boolean` is effectively a `Set`. Shouldn't we have that instead?

triceo: (Consider the performance impact of boxing and unboxing those `Booleans` into `boolean`s in the methods below.)

triceo: Nevermind, realized my mistake. You actually need to cache the `false`s as they are a result of computation.

ge0ffrey: It's a Boolean, not a boolean:
TRUE: calculation result is true
FALSE: calculation result is false
null: calculation hasn't been done yet

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
ge0ffrey: if "iff" is the abbreviation of "if and only if" let's write it out the long way, not everyone is a math geek ;)

ge0ffrey:
@rsynek Very interesting research that it's the computeIfAbscent change, not the ConcurrentHashMap change!

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
triceo: Didn't even cross my mind that it could actually mean something. Well, I guess "not everyone is a math geek." :-)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/Cache.java
rsynek: I did - I didn't like changing semantic of JDK method a bit, but maybe it's not so bad thing. Thanks for suggestion, I will change it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
rsynek: I don't consider myself a math geek and all professors who used to give me lessons in Math would support my self-assessment :)

But yes, will write it in the full form.

rsynek:
@ge0ffrey actually it's a combination of both, plain `HashMap` doesn't do any locking in `computeIfAbsent`, also its implementation is a bit less complicated.

rsynek:
@triceo @ge0ffrey updated with your suggestions, please review.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * {@inheritDoc}
+     * */
+    @Override
+    public V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
+        V value = get(key);
+        if (value == null) {
+            value = super.computeIfAbsent(key, mappingFunction);
+        }
+        return value;
```
ge0ffrey: This looks a lot like the (broken) double locked idiom:
  https://www.javaworld.com/article/2074979/java-concurrency/double-checked-locking--clever--but-broken.html

What happens to performance if we make this method synchronized?

triceo: In this case, I think it doesn't matter. The cache update is properly synchronized - the worst thing we're risking is that the computation will happen twice. Once one thread stores the result, every other thread will see the cached result.

rsynek: We are not testing and updating a shared field - `value` is a local variable. If it's null and two threads will end up calling `super.computeIfAbsent()`, they will get synchronized inside the `computeIfAbsent` method, so as Lukas pointed out, both thread will compute the value in this situation. If the computed value depends on what thread executes it (which it shouldn't), the last thread wins.

ge0ffrey: Agreed, good point. Because value is a local variable it is not a double checked locking.

rsynek: Let me just correct myself, after looking with @yurloc into the ConcurrentHashMap implementation more carefully, even the computation cannot happen twice - the collection guarantees that.

rsynek: DO NOT MERGE yet - I would prefer running complete benchmarks first to make sure the fix addresses all the performance issues found.

ge0ffrey: I was just about to hit the merge button :)
I tried your approach locally and I can confirm that the performance notecably improves on the micro benchmark.
I think it's ready to merge, especially on master. I might nitpick on a few lines. I 'll then backport it to 7.5.x once it's merged.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Thread-safe basic cache implementation.
+ *
+ * @param <K> the type of keys for the cache
+ * @param <V> the type of values stored in the cache
+ * */
+public final class ConcurrentCache<K, V> extends ConcurrentHashMap<K, V> {
```
ge0ffrey: Renaming discussion:
- ConcurrentCache sounds like a standard class from the JDK. Can we make it less general?
- We use it to store calculations, not retrieve information for a remote service/database.
- We use this Cache as a lazy initialization only: once a value is calculated, it is never recalculated.
- After discussion with Mario, he rightfully suggested this is a Memoization, see https://en.wikipedia.org/wiki/Memoization

As most people don't know what a Memoization is, shall we call it MemoizationCache? :) I'd leave out the name Concurrent, as that seems implicit.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
```diff
@@ -0,0 +1,31 @@
+package org.optaplanner.core.impl.domain.solution.util;
```
ge0ffrey: Can you move it to the package `org.optaplanner.core.impl.domain.common`? I am not a big fan of `util` packages, as they are a one-name-fits-all package that attrack bloat.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
```diff
@@ -0,0 +1,31 @@
...
+public final class ConcurrentCache<K, V> extends ConcurrentHashMap<K, V> {
+
+    /**
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} should not be used to directly retrieve cached values
+     * as it incorporates locking even when the value is present and doesn't have to be computed. Thus, the value is
+     * retrieved by {@link ConcurrentHashMap#get(Object)} and if and only if the value is null,
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} is used to ensure thread-safety.
+     *
+     * {@inheritDoc}
+     * */
```
ge0ffrey: Remove the first *

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
```diff
@@ -0,0 +1,31 @@
...
+ * @param <K> the type of keys for the cache
+ * @param <V> the type of values stored in the cache
+ * */
+public final class ConcurrentCache<K, V> extends ConcurrentHashMap<K, V> {
+
+    /**
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} should not be used to directly retrieve cached values
+     * as it incorporates locking even when the value is present and doesn't have to be computed. Thus, the value is
+     * retrieved by {@link ConcurrentHashMap#get(Object)} and if and only if the value is null,
+     * {@link ConcurrentHashMap#computeIfAbsent(Object, Function)} is used to ensure thread-safety.
```
ge0ffrey: A good javadoc shouldn't discuss implementation details, just behavior guaranties towards the client.
The reason for this, if we want to change the code of this method, we'd want to avoid having to update the javadoc as that will be strictly-speaking a backwards incompatible change (yes this in not a public api, but still it is a bc breakage against all our internal classes using this, which is fixable of course, unlike public api bc breakage).

Anyway, I'd shorten this paragraph to something like:
"A computeIfAbsent() implementation that heavily favors read access over write access."

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/util/ConcurrentCache.java
rsynek: Sounds good

PR #377, opened 2018-02-20
PLANNER-1020 Add method Move.rebase(Scoredirector) for multithreaded incremental solving

ge0ffrey:
Ready to merge

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
@@ -82,6 +82,28 @@ public interface Move<Solution_> {
      */
     Move<Solution_> doMove(ScoreDirector<Solution_> scoreDirector);
 
+    /**
+     * Rebases a move from an origin {@link ScoreDirector} to another destination {@link ScoreDirector}
+     * which is usually on another {@link Thread} or JVM.
+     * The new move returned by this method mutates the entities and problem facts
+     * on the {@link PlanningSolution} of the destination {@link ScoreDirector},
+     * which should be (at least) a deep planning clone of the {@link PlanningSolution}
+     * that this move was generated from.
+     * That new move does the exact same change as this move,
+     * resulting in the same {@link PlanningSolution} state,
+     * presuming that destination {@link PlanningSolution} was in the same state as the original {@link PlanningSolution}.
+     * <p>
+     * Generally speaking, an implementation of this method iterates through every entity and fact instance in this move,
+     * translates each one to the destination {@link ScoreDirector} with {@link ScoreDirector#lookUpWorkingObject(Object)}
+     * and creates a new move instance of the same move type, using those translated instances.
+     * @param destinationScoreDirector never null, the
+     * @return never null,
+     */
+    default Move<Solution_> rebase(ScoreDirector<Solution_> destinationScoreDirector) {
+        throw new UnsupportedOperationException("The custom move class (" + getClass()
+                + ") doesn't implement the rebase() method, so multithreaded solving is impossible.");
+    }
+
     // ************************************************************************
     // Introspection methods
     // ************************************************************************
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
@@ -98,8 +98,9 @@ public interface Move<Solution_> {
      * Generally speaking, an implementation of this method iterates through every entity and fact instance in this move,
      * translates each one to the destination {@link ScoreDirector} with {@link ScoreDirector#lookUpWorkingObject(Object)}
      * and creates a new move instance of the same move type, using those translated instances.
-     * @param destinationScoreDirector never null, the
-     * @return never null,
+     * @param destinationScoreDirector never null, the {@link ScoreDirector#getWorkingSolution()}
+     * that the new move should change the planning entity instances of.
+     * @return never null, a new move that does the same change as this move on another solution instance
      */
     default Move<Solution_> rebase(ScoreDirector<Solution_> destinationScoreDirector) {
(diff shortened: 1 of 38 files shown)


rsynek:
Looks good, just few questions rather about the concept:
- will be the `PartitionChangeMove` removed when all "common" moves will support the `rebase` operation?
- why are the `getPlanningEntities()` and `getPlanningValues()` operations unsupported for the `PartitionChangeMove` ?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
```diff
@@ -82,6 +82,30 @@
...
+     * That new move does the exact same change as this move,
+     * resulting in the same {@link PlanningSolution} state,
+     * presuming that destination {@link PlanningSolution} was in the same state
+     * as the original {@link PlanningSolution} to begin with.
+     * <p>
+     * Generally speaking, an implementation of this method iterates through every entity and fact instance in this move,
+     * translates each one to the destination {@link ScoreDirector} with {@link ScoreDirector#lookUpWorkingObject(Object)}
+     * and creates a new move instance of the same move type, using those translated instances.
+     * @param destinationScoreDirector never null, the
+     * @return never null,
```
yurloc: Please complete the descriptions.

ge0ffrey:
> will be the PartitionChangeMove removed when all "common" moves will support the rebase operation?

No, they are still needed to update the main solver when a partition child solver finds a better solution.

>  why are the getPlanningEntities() and getPlanningValues() operations unsupported for the PartitionChangeMove ?

Those are only use for tabu search, so I don't think it can hurt, but you're right that they should be either implemented or thow IllegalStateException("Impossible situation" ...).

ge0ffrey:
@rsynek I've improved the javadocs/exceptions of PartitionChangeMove based on your comment. All moves are now implemented and tested. This PR is ready to merge.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
```diff
@@ -34,12 +34,13 @@
...
 
 /**
+ * Applies a new best solution from a partition child solver into the global working solution of the parent solver.
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public final class PartitionChangeMove<Solution_> extends AbstractMove<Solution_> {
 
     public static <Solution_> PartitionChangeMove<Solution_> createMove(InnerScoreDirector<Solution_> scoreDirector,
-                                                                        int partIndex) {
+            int partIndex) {
```
yurloc: @ge0ffrey Would you mind importing the latest version of [KIE Java Conventions](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/7.6.0.Final/ide-configuration/intellij-configuration/code-style) to avoid formatting inconsistency?

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/move/CompositeMoveTest.java
```diff
@@ -89,6 +89,44 @@ public void doMove() {
...
+                        {e2, destinationE2},
+                        {e3, destinationE3},
+                });
+
+        ChangeMove<TestdataSolution> a = new ChangeMove<>(e1, variableDescriptor, v2);
+        ChangeMove<TestdataSolution> b = new ChangeMove<>(e2, variableDescriptor, v1);
+        CompositeMove<TestdataSolution> rebaseMove = new CompositeMove<>(a, b).rebase(destinationScoreDirector);
+        Move<TestdataSolution>[] rebasedChildMoves = rebaseMove.getMoves();
+        assertEquals(2, rebasedChildMoves.length);
+        ChangeMove rebasedA = (ChangeMove) rebasedChildMoves[0];
```
yurloc: We can avoid raw type warning here by changing to
```java
ChangeMove<?> rebasedA = (ChangeMove) rebasedChildMoves[0];
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
ge0ffrey: This needs to happen on all sources of the optaplanner repository in one big bulk change - currently all sources follow the older version of the KIE conventions that I have locally.

The new conventions might have some readability trouble on long method signatures (especially with generic methods), but you're right - it's better to be consistent than to be ideal. Anyway, let's make a separate jira and investigate there.

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/move/generic/PillarChangeMoveTest.java
```diff
@@ -116,6 +119,52 @@ public void doMove() {
...
+        assertSameProperties(Arrays.asList(destinationE1, destinationE3), destinationV3,
+                new PillarChangeMove<>(Arrays.asList(e1, e3), variableDescriptor, v3).rebase(destinationScoreDirector));
+        assertSameProperties(Arrays.asList(destinationE2), destinationV1,
+                new PillarChangeMove<>(Arrays.asList(e2), variableDescriptor, v1).rebase(destinationScoreDirector));
+        assertSameProperties(Arrays.asList(destinationE1), destinationV2,
+                new PillarChangeMove<>(Arrays.asList(e1), variableDescriptor, v2).rebase(destinationScoreDirector));
+    }
+
+    public void assertSameProperties(List<Object> pillar, Object toPlanningVariable, PillarChangeMove<?> move) {
+        assertListElementsSameExactly(pillar, (List<Object>) move.getPillar());
```
yurloc: Can be replaced by AssertJ's `assertThat(move.getPillar()).hasSameElementsAs(pillar);` but we can change this in bulk later.

PR #378, opened 2018-02-21
PLANNER-1014 Introduce findbugs and fail build on violations

rsynek:
- findbugs now fails the build
- maxRank set to find violations of "Troubling" severity (findbugs distinguishes Scariest, Scary, Troubling, Of concern)
- current violations have been either fixed or excluded


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/descriptor/VariableDescriptor.java
```diff
@@ -28,7 +29,7 @@
 /**
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
-public abstract class VariableDescriptor<Solution_> {
+public abstract class VariableDescriptor<Solution_> implements Serializable {
```
yurloc: Interesting... What is the justification?

rsynek: Inherited classes are used in classes that implement Serializable. Should their fields that are not serializable be rather declared transient?

yurloc:
> ... integer value between 1 and 20, where 1 to 4 are scariest, 5 to 9 scary, 10 to 14 troubling, and 15 to 20 of concern bugs.

rsynek:
@yurloc updated with the rest of changes for other modules. The check should also pass now.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/SolutionConverter.java
```diff
@@ -107,6 +107,10 @@ private SolutionConverter(String dataDirName,
 
     public void convertAll() {
         File[] inputFiles = inputDir.listFiles();
+        if (inputFiles == null) {
+            inputFiles = new File[] {};
+        }
+
```
yurloc: Trailing whitespace.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/SolutionConverter.java
```diff
@@ -107,6 +107,10 @@ private SolutionConverter(String dataDirName,
 
     public void convertAll() {
         File[] inputFiles = inputDir.listFiles();
+        if (inputFiles == null) {
+            inputFiles = new File[] {};
```
yurloc: Unconventional space after `File[]`.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
```diff
@@ -92,6 +91,16 @@ public void writePlannerBenchmarkResult(File benchmarkReportDirectory,
         return plannerBenchmarkResultList;
     }
 
+    private File[] retrieveSortedBenchmarkReportDirectories(File benchmarkDirectory) {
+        File[] benchmarkReportDirectories = benchmarkDirectory.listFiles((FileFilter) DirectoryFileFilter.INSTANCE);
+        if (benchmarkReportDirectories == null) {
+            return new File[] {};
```
yurloc: Unconventional space.

rsynek: Another proof the checkstyle + KIE conventions is necessary. Thanks, will fix

rsynek:
@yurloc formatting fixed, feel free to merge.

rsynek:
Could we merge this, please?

rsynek:
@ge0ffrey could you please review and merge? Thanks.

optaplanner-core/src/build/findbugs-excludes.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FindBugsFilter>
+  <Match>
+    <!--
+    Uninitialized fields are intentional in this class - to make sure the uninitialized value is the same as JVM
+    would use for uninitialized fields in facts.
+    -->
+    <Class name="org.optaplanner.core.impl.score.director.drools.testgen.fact.TestGenPrimitiveValueProvider"/>
```
ge0ffrey: Can we do this with annotations instead of an external xml file?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/AbstractMove.java
```diff
@@ -27,7 +28,7 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  * @see Move
  */
-public abstract class AbstractMove<Solution_> implements Move<Solution_> {
+public abstract class AbstractMove<Solution_> implements Move<Solution_>, Serializable {
```
ge0ffrey: This is dangerous

optaplanner-core/src/build/findbugs-excludes.xml
ge0ffrey: it seems there is, but the question is if it introduces an unwelcome dependency (if we make it optional=true, I think it might be ok, if that works for users of course...)
https://stackoverflow.com/a/1829960/472109

ge0ffrey:
jenkins retest this

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
```diff
@@ -78,8 +78,7 @@ public void writePlannerBenchmarkResult(File benchmarkReportDirectory,
             throw new IllegalArgumentException("The benchmarkDirectory (" + benchmarkDirectory
                     + ") is not a directory.");
         }
-        File[] benchmarkReportDirectories = benchmarkDirectory.listFiles((FileFilter) DirectoryFileFilter.INSTANCE);
-        Arrays.sort(benchmarkReportDirectories);
+        File[] benchmarkReportDirectories = retrieveSortedBenchmarkReportDirectories(benchmarkDirectory);
```
ge0ffrey: 3 lines above there is a "benchmarkDirectory.exists()" check, so this will never return null (unless in a race condition at which time it should fail fast and NPE is ok enough for me).

Can we revert this change and tell Findbugs to exclude it?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/SolutionConverter.java
```diff
@@ -107,6 +107,10 @@ private SolutionConverter(String dataDirName,
 
     public void convertAll() {
         File[] inputFiles = inputDir.listFiles();
+        if (inputFiles == null) {
```
ge0ffrey: same principle here: there is already a inputDir.exists() check in the constructor. It's never null, unless there a race condition in which case NPE is fine.

optaplanner-core/src/build/findbugs-excludes.xml
yurloc: I think `provided` is what we want - it will be available in compile time but not required for runtime.

ge0ffrey: Either way, I am ok to go with an excludes file for now too.
But let's revert the listFiles() changes.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/SolutionConverter.java
yurloc: This is a null check for the `inputDir.listFiles()` return value.

> Returns null if this abstract pathname does not denote a directory, or if an I/O error occurs.

yurloc: We could perhaps improve the fail-fast approach and also check if `inputDir` is a directory in the constructor.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/AbstractMove.java
rsynek: @ge0ffrey I was pretty sure someone will notice this and raise concert during PR review - thanks :) The issue findbugs has is that classes being marked as Serializable should have all the Object fields either Serializable too or transient.

More general question, do we use any classes in OptaPlanner for Java serialization? Or are we forced to declare some classes as Serializable by some framework?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/SolutionConverter.java
rsynek: I like the idea of checking if it's really a directory - will change it.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
yurloc: Oh, isn't this the same situation as the other one in this PR that I already commented on?

If I understand it correctly, the null risk is not with `benchmarkDirectory` but with the result of `listFiles()` which is not guaranteed to be non-null. Replacing it with an empty array looks like an elegant solution that allows the code below to work with it safely without being enclosed with a null-checking `if` statement.

rsynek: @yurloc is correct - it's about the listFiles() result and listFiles() returns null instead of throwing IOException, so we need the null check even if we check for directory. I will leave this null check in place.

optaplanner-core/src/build/findbugs-excludes.xml
rsynek: I like the annotation approach, the justification is there right with the code, much more intuitive. The listFiles() won't be reverted - see the other comment.

rsynek:
@ge0ffrey update with your and @yurloc comments, short summary:
- introduced annotations instead of .xml file for exceptions
- removed the Serializable and marked the field findbugs complained about as `transient`
- the checking of listFiles() result stays, just added a fail fast check for a directory

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionChangedEvent.java
```diff
@@ -29,7 +29,7 @@
     private final int partIndex;
     private final long eventIndex;
     private final PartitionChangedEventType type;
-    private final PartitionChangeMove<Solution_> move;
+    private final transient PartitionChangeMove<Solution_> move;
```
ge0ffrey: Moves aren't Serializable. Sprinkling transient here to make FindBugs happy but not really supporting serializing Moves has no benefit for our users.

Solution proposals
A) Add transient in places to mislead findbugs bug detection. I veto this.
B) Properly make Move's etc Serializable. I think that's unrealistic. That's pandora's box.
C) Remove Serializable from classes that findbugs correctly says aren't really serializable. I like this: no one is serializing moves/descriptors/solvers etc anyway. Yes, this can make CDI-in-the-servlet-context unhappy, but meh.
D) Sprinkle supressWarning annotations or similar Findbugs annotations to avoid findbugs complains. I don't like this much: findbugs should only add to our code in a good way.
E) Exclude the Serializable rule in the findbugs rules to check. I like this one.

Wdyt? I prefer C) or E).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/descriptor/VariableDescriptor.java
ge0ffrey: Maybe those casses that use this and implement Serializable shouldn't implement Serializable?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/BenchmarkResultIO.java
ge0ffrey: @rsynek Let's jump on bluejeans to discuss this one

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/queue/PartitionChangedEvent.java
rsynek: For me C) is the best options. Although E) sounds reasonably, excluding the Serializable rule does not make the classes properly Serializable either. I would say we can either do it correctly or should not do it at all.

Is there any particular situation with CDI we need to care about? That would be in my opinion the only reason for E).

rsynek:
@ge0ffrey  any more concerns? If not, could you merge it, please? Thanks!

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenPrimitiveValueProvider.java
```diff
@@ -15,6 +15,10 @@
  */
 package org.optaplanner.core.impl.score.director.drools.testgen.fact;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+@SuppressFBWarnings(value = "UR_UNINIT_READ", justification = "Uninitialized fields are intentional in this class - "
+        + "to make sure the uninitialized value is the same as JVM would use for uninitialized fields in facts.")
```
ge0ffrey: The default values for primitive values are written in the Java Language Spec 1.0 for over than 20 years.
I think it's safe to hard code them :)

ge0ffrey: @yurloc Would you mind if I change this code according to this:


 ```
       if (value.getClass().equals(Byte.class)) {
            uninitialized = (byte) 0;
        } else if (value.getClass().equals(Short.class)) {
            uninitialized = (short) 0;
        } else if (value.getClass().equals(Integer.class)) {
            uninitialized = 0;
        } else if (value.getClass().equals(Long.class)) {
            uninitialized = 0L;
        } else if (value.getClass().equals(Float.class)) {
            uninitialized = 0.0F;
        } else if (value.getClass().equals(Double.class)) {
            uninitialized = 0.0;
        } else if (value.getClass().equals(Boolean.class)) {
            uninitialized = false;
        } else if (value.getClass().equals(Character.class)) {
            uninitialized = (char) 0;
```

File: pom.xml [hunk 11]
```diff
@@ -101,6 +118,14 @@
   <build>
     <pluginManagement>
       <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>findbugs-maven-plugin</artifactId>
+          <configuration>
+            <maxRank>14</maxRank>
+            <includeTests>true</includeTests>
+          </configuration>
```
ge0ffrey: Does findbugs activate with every build because of this? I don't see that happening?

pom.xml
```diff
@@ -80,9 +84,22 @@
...
+        <version>3.0.0</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
+  <dependencies>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>annotations</artifactId>
+      <optional>true</optional>
```
ge0ffrey: I was able to remove this, but getting rid of the need to do those suppresswarning annotations

ge0ffrey:
@rsynek I 've pulled everything except for the pom changes into this PR (where I did a number of small changes): https://github.com/kiegroup/optaplanner/pull/384
Please review, so I can merge that.

About the pom changes, let's talk about that on bluejeans. I believe the remaining part should go into kie-parent (unless it's activated by default, which I don't see it doing right now)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenPrimitiveValueProvider.java
yurloc: OK

pom.xml
yurloc: @ge0ffrey No it doesn't. Findbugs plugin execution is part of `fullProfile` defined in `kie-parent` and can be activated by `-Dfull` property. The point of this change is to tweak Findbugs configuration for this project. Without setting maxRank to this specific value, it doesn't any of the problems that this PR fixes.

PR #381, opened 2018-03-19
PLANNER-1064 Solver.getTimeMillisSpent() and getBestScore() are thread-safe

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/api/solver/Solver.java
```diff
@@ -61,12 +61,22 @@
...
+     * <p>
+     * This method is thread-safe.
      * @return null if the {@link PlanningSolution} is still uninitialized
      */
     Score getBestScore();
 
     /**
-     * @return the amount of millis spent between when this solver started (or last restarted) and ended
+     * Returns the amount of milliseconds spent solving since the last start.
+     * If it hasn't started it yet, it runs 0.
```
rsynek: runs -> returns ?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -109,11 +109,17 @@ public Score getBestScore() {
 
     @Override
     public long getTimeMillisSpent() {
```
rsynek: DefaultSolverScope already contains methods `calculateTimeMillisSpentUpToNow` and `getTimeMillisSpent`. Why not to reuse them?

ge0ffrey: Because calculateTimeMillisSpentUpToNow() isn't multi-thread safe. It is only solver-thread safe.
In a multi-threaded environment, start might be null, end might be null or might not.
And restarts might make a non-null end, null again.

PR #384, opened 2018-03-30
Fix findbugs warnings + don't make things Serializable that aren't

ge0ffrey:
Everything from Radovan's PR except the pom changes, with some changes.
Author is set to Radovan as he deserves the credit for this.


ge0ffrey:
@yurloc I didn't do the pom changes yet (noted in the PR's description), because I want to handle those in a separate discussion: they don't seem to do anything yet, in my experience, and in any case they need to go into kie-parent. Let's discuss with Radovan on bluejeans on Tuesday or Wednesday?

rsynek:
@ge0ffrey please add the configuration as we discussed and it's ready for merge. Thanks!

PR #386, opened 2018-04-12
PLANNER-1089 Use a java.util.Supplier instead of a Class for ScoreDirectorFactoryConfig

roshan:
A `Class` was used since it allowed instantiating multiple instances of
the `ScoreCalculator` to supply to calculating threads, etc.

Using a `Supplier` instead allows building `EasyScoreCalculator`s at
run-time while retaining the desired ability to instantiate multiple
`EasyScoreCalculator`s. This permits varying parameters in the same
process in a simple fashion.

For example:

```
scoreDirectorFactoryConfig.setEasyScoreCalculatorSupplier(
  () -> new ScoreCalculator(4.3, 2.1)
);
```

See: https://stackoverflow.com/a/49748882/3858681

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
@@ -108,7 +108,7 @@ public class ScoreDirectorFactoryConfig extends AbstractConfig<ScoreDirectorFact
     protected ScoreDirectorFactoryConfig assertionScoreDirectorFactory = null;
 
     protected Boolean generateDroolsTestOnError = null;
-    protected Supplier<? extends EasyScoreCalculator> easyScoreCalculatorSupplier;
+    protected Supplier<? extends EasyScoreCalculator> easyScoreCalculatorSupplier = null;
 
     /**
      * @return sometimes null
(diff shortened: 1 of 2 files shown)


ge0ffrey:
ok to test

roshan:
Thank you for looking @ge0ffrey . I'm happy to implement as you describe. I'll ask back about the docs once I have something.

ge0ffrey:
RH Summit stress until end of next week at the least. Afterwards there will be time to pick up things like this.

roshan:
Any luck finding time to look at this?

ge0ffrey:
I 've taken part of this PR and put it in this PR https://github.com/kiegroup/optaplanner/pull/401 which already allows dynamic parameter value configuration - as long those are ints, longs, strings, enums.

The Supplier approach is not included in PR 401. Can you please help understand why you need it in your use case? I am a bit in the status "when in doubt, leave it out (until you understand/proof that you need it)"

ge0ffrey:
@roshan #401 is merged. Check if it suits your need and I can close this PR.

ge0ffrey:
Closing this because #401 fixes this (mostly) and no response/motivation if the rest (Supplier) is really needed (worth it's API cost). Feel free to re-open with such a motivation.

roshan:
Sorry it took me some time to respond. Looks like this provides the functionality I need. I thought it was cleaner to go with a concrete instantiation instead of building an XML and then using that to configure the calculator but the latter approach does seem to be the approach that aligns with the rest of the API. Thanks for the feature.

PR #387, opened 2018-04-17
Multithreaded Local Search

ge0ffrey:
TODO:
- clean up threads
- manager thread lifecycle in solver
- fix race condition PLANNER-1088

@rsynek Out of about 30 runs, I saw 2 runs hanging, probably due to a race condition.
The first time, it was in a big 3-day benchmark, on the 3 subsingle of SA after 10 subsingles of TS.
The second time, it was on step 600 on a small CB TS configuration, so it's not that it ran out of resources.

https://issues.jboss.org/browse/PLANNER-1088
**I have not been able to reproducible reliably.** We need to first find a way to trigger it more frequently (I suspect playing with the buffer size, moveThreadCount, random sleeps, etc, might do it.)

I now suspect the hanging is due to a dead-lock.


optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/RearrangingBlockingQueue.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * Not thread-safe. Can only be called from the solver thread.
+     * @return never null
+     * @throws InterruptedException if interrupted
+     * @see BlockingQueue#take()
+     */
+    public MoveResult<Solution_> take() throws InterruptedException {
+        searchMoveIndex++;
```
ge0ffrey: @rsynek this causes a one-off issue that deadlocks moveThreadCount=1 with buffer=1, because the searchMoveIndex is initialized to 0, so the first one it's looking for is moveIndex 1, but the first moveIndex is 0.

ge0ffrey: The question is, if this is what triggered PLANNER-1088...

rsynek: @ge0ffrey truth, this might be just the first step to solve it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/thread/ThreadUtils.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ThreadUtils {
+
+    protected static final transient Logger logger = LoggerFactory.getLogger(ThreadUtils.class);
```
yurloc: Why protected? Why transient?

optaplanner-docs/src/main/asciidoc/OptimizationAlgorithms/OptimizationAlgorithms-chapter.adoc
```diff
@@ -759,17 +759,112 @@ To avoid those, configure a `NoChangePhase`:
...
+  <threadFactoryClass>...MyAppServerThreadFactory</threadFactoryClass>
+  ...
+</solver>
+----
+
+The following ``moveThreadCount``s are supported:
+
+* `NONE` (default): Don't run any move threads. Use the single threaded code.
+* ``AUTO``: Let Planner decide how many move threads to run in parallel.
+On machines or containers with little or no CPU's, this can fall back to the single threaded code.
```
yurloc: Should be "CPUs" (in this case is's plural).

optaplanner-docs/src/main/asciidoc/OptimizationAlgorithms/OptimizationAlgorithms-chapter.adoc
```diff
@@ -759,17 +759,112 @@ To avoid those, configure a `NoChangePhase`:
...
+----
+
+It is counter-effective to set a moveThreadCount that is higher than the number of available CPU cores,
+as that will slow down the score calculation speed.
+One good reason to do it anyway, is to reproduce a bug of a high-end production machine.
+
+[NOTE]
+====
+Multithreaded solving is _still reproducible_, as long as the resolved `moveThreadCount` is stable.
+A run of the same solver configuration on 2 machines with a different number of CPU's,
```
yurloc: CPUs.

optaplanner-docs/src/main/asciidoc/PartitionedSearch/PartitionedSearch-chapter.adoc
```diff
@@ -55,22 +55,21 @@ Advanced configuration:
...
     <localSearch>...</localSearch>
   </partitionedSearch>
 ----
 
-The `threadFactoryClass` allows to plug in a custom `ThreadFactory` for environments
-where arbitrary thread creation should be avoided, such as most application servers, Android, or Google App Engine.
-
 The `runnablePartThreadLimit` allows limiting CPU usage to avoid hanging your machine, see below.
 
+To run in a environment that doesn't like arbitrary thread creation,
```
yurloc: "an environment"

optaplanner-docs/src/main/asciidoc/PlannerConfiguration/PlannerConfiguration-chapter.adoc
```diff
@@ -136,7 +136,7 @@ They build the runtime components (of the package namespace ``org.optaplanner.co
 
 [IMPORTANT]
 ====
-The `SolverFactory` is only multi-thread safe after its configured.
+The `SolverFactory` is only thread-safe after its configured.
```
yurloc: "it's configured"

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
```diff
@@ -0,0 +1,407 @@
...
+                logger.trace("{}        Move index ({}) not doable, ignoring move ({}).",
+                        logIndentation, foragingMoveIndex, foragingMove);
+            } else {
+                moveScope.setScore(result.getScore());
+                boolean accepted = acceptor.isAccepted(moveScope);
+                moveScope.setAccepted(accepted);
+                logger.trace("{}        Move index ({}), score ({}), accepted ({}), move ({}).",
+                        logIndentation,
+                        foragingMoveIndex, moveScope.getScore(), moveScope.getAccepted(),
+                        foragingMove);
```
ron-murhammer: Would it be better to rewrap the lines of these logging parameters?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
```diff
@@ -0,0 +1,407 @@
...
+            OrderByMoveIndexBlockingQueue.MoveResult<Solution_> result;
+            try {
+                result = resultQueue.take();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                break;
+            }
+            if (stepIndex != result.getStepIndex()) {
+                throw new IllegalStateException("Impossible situation: the solverThread's stepIndex (" + stepIndex
+                        + ") differs from the result's stepIndex (" + result.getStepIndex() + ").");
```
ron-murhammer: Is this just some kind of sanity check or have you seen this happen? Just seems like a very interesting situation and wonder if its necessary.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
```diff
@@ -0,0 +1,407 @@
...
+                                    logIndentation, moveThreadIndex, stepIndex, moveIndex, score);
+                            // Deliberately add to fail fast if there is not enough capacity (which is impossible)
+                            resultQueue.addMove(moveThreadIndex, stepIndex, moveIndex, move, score);
+                        }
+                    } else {
+                        throw new IllegalStateException("Unknown operation (" + operation + ").");
+                    }
+                    // TODO checkYielding();
+                }
+                logger.trace("{}            Move thread finished.");
```
ron-murhammer: Is there a missing log parameter (logIndentation) here?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/thread/ThreadUtils.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public static void shutdownAwaitOrKill(ExecutorService executor, String logIndentation, String name) {
+        // Intentionally clearing the interrupted flag so that awaitTermination() in step 3 works.
+        if (Thread.interrupted()) {
+            // 2a. If the current thread is interrupted, propagate interrupt signal to children by initiating
```
ron-murhammer: Looks like this code was refactored into this new class from `DefaultPartitionedSearchPhase.java`. Probably should at least remove the numbers in the comments since there is no longer a step 1 otherwise its a big confusing.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -257,6 +293,37 @@ protected RandomFactory buildRandomFactory(EnvironmentMode environmentMode_) {
         return randomFactory;
     }
 
+    protected Integer resolveMoveThreadCount() {
```
rsynek: Could use unit tests

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
```diff
@@ -0,0 +1,407 @@
...
+
+        private int moveThreadIndex;
+        private InnerScoreDirector<Solution_> scoreDirector = null;
+
+        public MoveThreadRunner(int moveThreadIndex) {
+            this.moveThreadIndex = moveThreadIndex;
+        }
+
+        @Override
+        public void run() {
```
rsynek: Too long and complicated method, breaking it into private (or package private) methods would make it more readable and testable.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/thread/OrderByMoveIndexBlockingQueue.java
```diff
@@ -0,0 +1,206 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.heuristic.move.Move;
+
+public class OrderByMoveIndexBlockingQueue<Solution_> {
```
rsynek: Could use some unit testing

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
ge0ffrey: What do you mean with rewrap? As in logger.isTrace() ? The getScore() and getAccepted() methods are simple getters, they would be inlined normally.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
```diff
@@ -0,0 +1,407 @@
...
+    protected final ThreadFactory threadFactory;
+    protected final int moveThreadCount;
+    protected final int selectedMoveBufferSize;
+
+    protected boolean assertStepScoreFromScratch = false;
+    protected boolean assertExpectedStepScore = false;
+    protected boolean assertShadowVariablesAreNotStaleAfterStep = false;
+
+    protected List<MoveThreadRunner> moveThreadRunnerList;
+    protected BlockingQueue<MoveThreadOperation<Solution_>> operationQueue;
```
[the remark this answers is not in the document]
ge0ffrey: Agreed - it needs to be consistent. I've made Operation classes protected, so in case people want to experiment with them. Do note that all these classes are in the impl package, so no BC guarantee.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/thread/ThreadUtils.java
ge0ffrey: Agreed, static field loggers should be private and non-transient. I used that form for consistency. So now I was able to consistently fix that :) some benchmark classes adjusted too

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
ge0ffrey: It's a fail fast. More specifically, because it says "impossible situation", it's a fail fast that won't be triggered by a user doing something wrong (like most fail fasts), but by an intrinsic bug in optaplanner-core itself.
Because the if check is extremely cheap, it's ok to do this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
ge0ffrey: Yes, thanks for reporting :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/thread/ThreadUtils.java
ge0ffrey: Yes, we need to fix that, but I think we might need to simplify the whole thing here. It's probably overkilled.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -257,6 +293,37 @@ protected RandomFactory buildRandomFactory(EnvironmentMode environmentMode_) {
...
+            }
+        } else {
+            resolvedMoveThreadCount = ConfigUtils.resolveThreadPoolSizeScript(
+                    "moveThreadCount", moveThreadCount, MOVE_THREAD_COUNT_NONE, MOVE_THREAD_COUNT_AUTO);
+        }
+        if (resolvedMoveThreadCount < 1) {
+            throw new IllegalArgumentException("The moveThreadCount (" + moveThreadCount
+                    + ") resulted in a resolvedMoveThreadCount (" + resolvedMoveThreadCount
+                    + ") that is lower than 1.");
+        }
```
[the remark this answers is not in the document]
ge0ffrey: Yes, it could. Findbugs or PMD might complain about this too.
But we might add other options later (other if else's) and this is a fail-fast check that happens after the number has been resolved. I am going to leave it as-is. Thanks for reporting though.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -257,6 +293,37 @@ protected RandomFactory buildRandomFactory(EnvironmentMode environmentMode_) {
...
 
+    protected Integer resolveMoveThreadCount() {
+        int availableProcessorCount = Runtime.getRuntime().availableProcessors();
+        Integer resolvedMoveThreadCount;
+        if (moveThreadCount == null || moveThreadCount.equals(MOVE_THREAD_COUNT_NONE)) {
+            return null;
+        } else if (moveThreadCount.equals(MOVE_THREAD_COUNT_AUTO)) {
+            // Leave one for the Operating System and 1 for the solver thread, take the rest
+            resolvedMoveThreadCount = (availableProcessorCount - 2);
+            if (resolvedMoveThreadCount <= 1) {
```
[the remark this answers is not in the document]
ge0ffrey: 1 move thread != single threaded. In single-threaded there are no move threads. With 1 move thread, there are 2 threads, because there is also the solver thread.

For AUTO, we replace 1 move thread with single threaded. Comment improved to make this clearer:
// To deliberately enforce 1 moveThread, set the moveThreadCount explicitly to 1.

optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
```diff
@@ -136,10 +146,16 @@ public void setPhaseConfigList(List<PhaseConfig> phaseConfigList) {
...
+            threadFactory = ConfigUtils.newInstance(this, "threadFactoryClass", threadFactoryClass);
+        } else {
+            threadFactory = solverConfigPolicy.buildThreadFactory(ChildThreadType.PART_THREAD);
+        }
         DefaultPartitionedSearchPhase phase = new DefaultPartitionedSearchPhase(
                 phaseIndex, solverConfigPolicy.getLogIndentation(), bestSolutionRecaller,
                 buildPhaseTermination(phaseConfigPolicy, solverTermination),
-                buildSolutionPartitioner(), buildThreadFactory(), resolvedActiveThreadCount());
+                buildSolutionPartitioner(), threadFactory, resolvedActiveThreadCount());
         List<PhaseConfig> phaseConfigList_ = phaseConfigList;
```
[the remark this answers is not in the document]
ge0ffrey: If phaseConfigList is empty, then phaseConfigList_ get defaulted to a CH+LS. But if that would change the global variable, then the benchmarker's report would not contain the exact same solver config in XML form, as the solver config changed when one of the global vars in the config objects change.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/LocalSearchPhaseConfig.java
```diff
@@ -128,15 +131,31 @@ private LocalSearchDecider buildDecider(HeuristicConfigPolicy configPolicy, Term
...
+        if (moveThreadCount == null) {
+            decider = new LocalSearchDecider(configPolicy.getLogIndentation(),
+                    termination, moveSelector, acceptor, forager);
+        } else {
+            Integer moveThreadBufferSize = configPolicy.getMoveThreadBufferSize();
+            if (moveThreadBufferSize == null) {
+                // TODO Verify this is a good default by more meticulous benchmarking on multiple machines and JDK's
+                // If it's too low, move threads will need to wait on the buffer, which hurts performance
+                // If it's too high, more moves are selected that aren't foraged
+                moveThreadBufferSize = 10;
```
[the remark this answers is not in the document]
ge0ffrey: I think that would reduce vertical density
and as that value isn't used anywhere, I don't see any gain to counteract that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/multithreaded/MultiThreadedLocalSearchDecider.java
```diff
@@ -0,0 +1,407 @@
...
+        phaseScope.addChildThreadsScoreCalculationCount(childThreadsScoreCalculationCount);
+        moveThreadRunnerList = null;
+        operationQueue = null;
+        resultQueue = null;
+    }
+
+    private ExecutorService createThreadPoolExecutor() {
+        ThreadPoolExecutor threadPoolExecutor
+                = (ThreadPoolExecutor) Executors.newFixedThreadPool(moveThreadCount, threadFactory);
+        if (threadPoolExecutor.getMaximumPoolSize() < moveThreadCount) {
```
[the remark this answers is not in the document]
ge0ffrey: I have no idea. The JDK javadocs were ambiguous.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
ge0ffrey: Yes, let's make a jira around this? I 'll add some unit tests, but probably after summit (merging won't be immediate anyway).

ge0ffrey:
jenkins retest this

optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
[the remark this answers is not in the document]
rsynek: thanks

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/LocalSearchPhaseConfig.java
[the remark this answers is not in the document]
rsynek: that's a valid point, I agree

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
rsynek: https://issues.jboss.org/browse/PLANNER-1099

ge0ffrey:
jenkins retest this with the latest master

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/thread/OrderByMoveIndexBlockingQueueTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.junit.Assert.assertEquals;
+import static org.optaplanner.core.impl.testdata.util.PlannerAssert.*;
+
+public class OrderByMoveIndexBlockingQueueTest {
+
+    private static final Logger logger = LoggerFactory.getLogger(PartitionQueueTest.class);
```
yurloc: Should be `OrderByMoveIndexBlockingQueueTest.class`.

ge0ffrey:
@rsynek By writing OrderByMoveIndexBlockingQueueTest, I also thought by adding the test addExceptionIsNotEatenIfNextStepStartsBeforeTaken() - and guess what - there was an edge case that exceptions got eaten... Fixing now (and committing test then too)

ge0ffrey:
@rsynek I did a bunch more changes, to also enable it for Construction Heuristics. I'd like to merge this PR soon (early next week).

PR #388, opened 2018-04-27
Solver.explainBestScore()

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -107,6 +108,21 @@ public Score getBestScore() {
         return solverScope.getBestScore();
     }
 
+    @Override
+    public String explainBestScore() {
+        Solution_ bestSolution = getBestSolution();
+        if (bestSolution == null) {
+            return null;
+        }
+        if (solverScope.getSolutionDescriptor().getScore(bestSolution) == null) {
```
rsynek: The javadoc says 
> null if {@link #getBestScore()} returns null

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -247,6 +252,62 @@ protected void setCalculatedScore(Score score) {
...
+        final int CONSTRAINT_MATCH_LIMIT = 2;
+        Score workingScore = calculateScore();
+        Collection<ConstraintMatchTotal> constraintMatchTotals = getConstraintMatchTotals();
+        ConstraintMatchScoreComparator constraintMatchScoreComparator = new ConstraintMatchScoreComparator();
+        StringBuilder scoreExplanation = new StringBuilder((constraintMatchTotals.size() + 4 + 2 * INDICTMENT_LIMIT) * 80);
+        scoreExplanation.append("Explanation of score (").append(workingScore).append("):\n");
+        scoreExplanation.append("    Constraint match totals:\n");
+        constraintMatchTotals.stream()
+                .sorted(new ConstraintMatchTotalScoreComparator())
+                .forEach(constraintMatchTotal -> {
```
rsynek: Explaining entire score and explaining a single ConstraintMatch are two different level of abstraction; maybe the lambda could go into a private method.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -247,6 +252,62 @@ protected void setCalculatedScore(Score score) {
...
+                        scoreExplanation.append("            ...\n");
+                    }
+                });
+
+        Collection<Indictment> indictments = getIndictmentMap().values();
+        scoreExplanation.append("    Indictments (top ").append(INDICTMENT_LIMIT)
+                .append(" of ").append(indictments.size()).append("):\n");
+        indictments.stream()
+                .sorted(new IndictmentScoreComparator()).limit(INDICTMENT_LIMIT)
+                .forEach(indictment -> {
```
rsynek: ditto.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
```diff
@@ -64,6 +64,11 @@ public Score getBestScore() {
         return solverScope.getBestScore();
     }
 
+    @Override
+    public String explainBestScore() {
+        throw new UnsupportedOperationException();
```
rsynek: Does it mean it doesn't make sense to support this operation or it's just not supported yet? The exception could provide some explanation.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
ge0ffrey: Yes, but this style is intentional to avoid a race condition.
So I need to add a comment to explain that.

This method is thread-safe. Calling getBestScore() might yield a different score than the getBestSolution()'s because it was changed meanwhile.

PR #389, opened 2018-05-14
shadow entity

dmarrazzo:
when a shadow variable can refer a variable of another entity, but it should belong to entity class (potentially a shadow entity)


ge0ffrey:
Sorry for not dealing with this Pr earlier (summit etc).
I've improved the docs to make this pitfall clearer:
  http://github.com/kiegroup/optaplanner/commit/78fd43ce0

PR #392, opened 2018-06-08
docs: fix typo JAXB -> Jackson

mwkroening:
I just stumbled upon the Header "16.2.4.1. JAXB: Marshalling a Score" which is a subheader of "16.2.4. JSON: Jackson" and think, it might be a typo.


ge0ffrey:
ok to test

ge0ffrey:
Thanks :)

PR #401, opened 2018-07-08
Planner 1089 easy score calculator parameters + PLANNER-1161 enum in custom params

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirectorFactory.java
```diff
@@ -48,6 +49,8 @@ public IncrementalScoreDirectorFactory(Class<? extends IncrementalScoreCalculato
             boolean lookUpEnabled, boolean constraintMatchEnabledPreference) {
         IncrementalScoreCalculator<Solution_> incrementalScoreCalculator = ConfigUtils.newInstance(this,
                 "incrementalScoreCalculatorClass", incrementalScoreCalculatorClass);
+        ConfigUtils.applyCustomProperties(incrementalScoreCalculator, "incrementalScoreCalculatorClass",
```
rsynek: The String literal is repeated one line above; maybe it would be worth introducing a constant.

ge0ffrey:
Testing the XML config over the API is mostly just testing XStream (and XStream already does that). I don't think it's worth the invested time. We do have a few general tests to at least smoke test an xml, see SolverFactoryTest

rsynek:
@ge0ffrey I agree that testing XStream brings no value, but there are pieces outside XStream that are probably not covered - e.g. the `KeyAsElementMapConverter`. Looking at configurations used in the SolverFactoryTest - they are quite simple and this element mapping doesn't seem to be tested there.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirectorFactory.java
ge0ffrey: I'd prefer not, it's not going to improve the readability of the code.

ge0ffrey:
Agreed we need a test for KeyAsElementMapConverter, adding.

ge0ffrey:
@rsynek Requested changes implemented

rsynek:
@mbiarnes for some reason the build was aborted due to time out. Do you by any chance know what happened? It seems the build took twice as much time as one of the previous successful builds.

jhrcek:
@rsynek FYI yesterday Michael merged [this PR](https://github.com/kiegroup/kie-jenkins-scripts/pull/259). This changed the fixed 1hour timeout to elastic timeout which looks at average of last 3 successful builds and times out if the build time exceeds 2x that average. If there are not enough build to measure the duration, the default timeout of 90 minutes is used.

PR #404, opened 2018-07-12
Change revapi to check against 7.5.0.Final

MarianMacik:
depends on kiegroup/droolsjbpm-build-bootstrap#775

See parent commit for more info.


MarianMacik:
@ge0ffrey I am afraid all PRs should have been merged together because now it will probably fail since parent is not merged (version is not updated). Right now we are waiting for kiegroup/droolsjbpm-integration/pull/1513.

PR #406, opened 2018-07-20
Integration tests for multithreaded solving

rsynek:
- adds some unit coverage for related parts of SolverConfig
- enables solving with 2 threads for all examples
- adds a test verifying reproducibility with multithreaded solving
- adds tests covering some boundary values (high thread count, minimal dataset) 

`testSmallProblemMultithreadedSolving` shows an issues - OptaPlanner freezes.


rsynek:
@yurloc it seems I don't have a permission to set a reviewer, is it possible?

rsynek:
@ge0ffrey could you please review? Now thinking whether enabling the feature in all examples is not too premature. Maybe we should override the config just in tests and let users set whatever they wish. WDYT?

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private SolverConfig mockSolverConfigForMoveThreadCountAuto(int mockCpuCount) {
+        SolverConfig solverConfig = spy(SolverConfig.class);
+        when(solverConfig.getAvailableProcessors()).thenReturn(mockCpuCount);
+        solverConfig.setMoveThreadCount(SolverConfig.MOVE_THREAD_COUNT_AUTO);
+        return solverConfig;
+    }
+
+    @Test
+    public void testMoveThreadCountFixedPositive() {
```
ge0ffrey: test methods shouldn't start with "test" prefix, that's an artifact of JUnit 3 (and we're using JUnit 4, soon 5)

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
```diff
@@ -0,0 +1,119 @@
...
+        SolverConfig solverConfig = new SolverConfig();
+        solverConfig.setMoveThreadCount("2");
+        assertThat(solverConfig.resolveMoveThreadCount()).isEqualTo(2);
```
ge0ffrey: JUnit trumps AssertJ. In this case, assertEquals() of JUnit wins.

optaplanner-examples/pom.xml
```diff
@@ -149,6 +149,12 @@
       <groupId>com.thoughtworks.xstream</groupId>
       <artifactId>xstream</artifactId>
     </dependency>
+    <!-- Comfortable assertions -->
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
ge0ffrey: As we've discussed before: JUnit 5 migration before introducing AssertJ, so we can see what we're missing after that. I can fast track the JUnit 5 upgrade.

Mixing several technologies to do the same thing, is bad. It interferes with code unity and code consistency.

optaplanner-examples/src/main/resources/org/optaplanner/examples/cheaptime/solver/cheapTimeSolverConfig.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <solver>
   <!--<environmentMode>FULL_ASSERT</environmentMode>--><!-- To slowly prove there are no bugs in this code -->
-  <!--<moveThreadCount>AUTO</moveThreadCount>--><!-- To solve faster by saturating multiple CPU cores -->
+  <!-- To solve faster by saturating multiple CPU cores. Use value "AUTO" to let OptaPlanner calculate optimal
+       move thread count.-->
+  <moveThreadCount>2</moveThreadCount>
```
ge0ffrey: I am not in favor of this change due to the verbosity. Yes, its more accurate, but now it's 3 lines instead of 1. Let's discuss.

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultithreadedReproducibilityTest.java
```diff
@@ -0,0 +1,80 @@
...
+                phaseConfig.setTerminationConfig(terminationConfig);
+            }
+        });
+        solverConfig.setMoveThreadCount(MOVE_THREAD_COUNT);
+        solverConfig.setEnvironmentMode(EnvironmentMode.REPRODUCIBLE);
+    }
+
+    @Test
+    public void testMultithreadedSolvingIsReproducible() {
+        IntStream.range(0, REPETITION_COUNT).forEach(iteration -> solveAndCompareWithPrevious(iteration));
```
ge0ffrey: we don't use forEach() when a simple for loop suffices (has no disadvantage). For most java programmers, the normal for loop is much easier to read.

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultithreadedReproducibilityTest.java
```diff
@@ -0,0 +1,80 @@
...
+        for (int i = 0; i < REPETITION_COUNT; i++) {
+            File dataSetFile = new File(CommonApp.determineDataDir(vehicleRoutingApp.getDataDirName()), DATA_SET);
+            VehicleRoutingSolution solution = importer.readSolution(dataSetFile);
+            vehicleRoutingSolutions[i] = solution;
+        }
+
+        solverFactory = SolverFactory.createFromXmlResource(vehicleRoutingApp.getSolverConfig());
+        SolverConfig solverConfig = solverFactory.getSolverConfig();
+        TerminationConfig terminationConfig = new TerminationConfig();
+        terminationConfig.setStepCountLimit(STEP_LIMIT);
```
ge0ffrey: take a look at TerminationConfig.withStepCountLimit(...)

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultithreadedReproducibilityTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        solverConfig.getPhaseConfigList().forEach(phaseConfig -> {
+            if (LocalSearchPhaseConfig.class.isAssignableFrom(phaseConfig.getClass())) {
+                phaseConfig.setTerminationConfig(terminationConfig);
```
ge0ffrey: I am not in favor of using the same config instance in multiple other config instances.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
rsynek: I use name that can read as a sentence; that said, I may use passive form and omit the "test" prefix. As majority of existing tests already do so, will follow this convention.

optaplanner-examples/pom.xml
rsynek: I am afraid JUnit 5 introduces nothing new to assertions, apart from asserting that exception has been thrown using lambda. The fluent API which reads very well and nice, descriptive failure messages are not provided by JUnit. The way I look at JUnit vs. AssertJ is that JUnit provides test execution, AssertJ assertions. I agree with you that we should not mix them in a single purpose - for assertions, we should either completely switch to AssertJ or stick with JUnit.

optaplanner-examples/src/main/resources/org/optaplanner/examples/cheaptime/solver/cheapTimeSolverConfig.xml
rsynek: Actually, I think that switching to 2 threads by default is not very user-friendly. AUTO is much better default value. I will put there "AUTO" and override to "2" in tests setup - to have it consistent across testing machines.

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultithreadedReproducibilityTest.java
rsynek: Ok, will change

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultithreadedReproducibilityTest.java
rsynek: Thanks for the tip!

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultithreadedReproducibilityTest.java
rsynek: Good catch, thanks!

optaplanner-examples/pom.xml
yurloc: @rsynek You speak my mind :-)

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test(timeout = 5000L)
+    public void testSolvingWithTooHighThreadCount() {
+        runSolvingAndVerifySolution(10, 20, "256");
+    }
+
+    @Test(timeout = 5000L)
+    public void testSmallProblemMultithreadedSolving() {
+        runSolvingAndVerifySolution(1, 1, "2");
```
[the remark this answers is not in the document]
rsynek: +1 I will ignore this test - the issue has been filed.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
yurloc: Not so recently I started to like the idea to make test methods as descriptive as possible even at the cost of breaking Java naming conventions. But that's OK, tests are not genuine Objects whose methods' purpose is sending messages.

For example:
```java
public void resolvedMoveThreadCountShouldEqualToConfig()
public void negativeMoveThreadCountShouldThrowException()
```
or even this more readable syntax:
```java
public void resolved_move_thread_count_should_equal_to_config()
public void negative_move_thread_count_should_throw_exception()
```

When reading the test method it is immediately clear what the test does and what's the intended behavior. It is also helpful when reading test report and looking at failures.

I understand that keeping the tests consistent might be a good reason for not adopting this style but I think it was worth mentioning.

rsynek: +1 for making method names as much descriptive as possible. There are several different conventions described, many of them working with "expected behaviour" and "state under test" [1]. However, if we adopt any of these conventions, we should rename the existing tests to ensure consistency.

[1] https://dzone.com/articles/7-popular-unit-test-naming

optaplanner-examples/pom.xml
jhrcek: Just taking part with @rsynek and @yurloc :) In QE's main test repo we have [a checkstyle rule](https://github.com/jboss-integration/bxms-qe-tests/blob/master/pom.xml#L803-L807) that makes using `import org.junit.Assert.*` forcing people to use assertj consistently.

rsynek:
@ge0ffrey @yurloc changes we agreed on in comments have been worked in. I didn't remove the AssertJ - it's used by optaplanner-core anyway, this brings it one more module. If we seek consistency, let's plan a migration from JUnit assert to AssertJ, do you agree?

rsynek:
jenkins retest this

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    public void moveThreadCountAutoIsPositive() {
+        final int cpuCount = 16;
+        assertThat(mockSolverConfigForMoveThreadCountAuto(cpuCount).resolveMoveThreadCount()).isEqualTo(cpuCount - 2);
+    }
+
+    @Test
+    public void moveThreadCountAutoIsNegative() {
+        final int cpuCount = -2;
+        assertThat(mockSolverConfigForMoveThreadCountAuto(cpuCount).resolveMoveThreadCount()).isNull();
```
yurloc: I can't help it, the test name is misleading. `move thread count auto` IS `negative`??? Actually `resolved move thread count` IS `null` WHEN `the resolution strategy` IS `auto` AND `available cpu count` IS `negative`.

yurloc:
Jenkins retest this. Let's see if the build gets aborted again. If it does, I'll increase the timeout (and maybe investigate why it's slower than usual).

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
rsynek: @yurloc although being much longer, this variant is much more descriptive. Updated the naming according to your suggestion. Thanks!

optaplanner-examples/pom.xml
ge0ffrey: A PR cannot introduce a new technology alternative for *some* files, leaving the rest of the codebase in the old way. That's how you end up with XLS decision tables that use POI and XLS DRL templates that use jexcelAPI (and this was still fixable in reasonable time, the MVC vs MVP vs pre-MVP GWT forms in the workbench is another matter).
Furthermore, a PR should not mix 2 non-trivial things such as introducing multithreaded solving tests and introducing assertj in every JUnit test.

Do or do not. There is not try...ing to add assertj technology in some files as part of another issue. Let's revert this PR to JUnit style and merge this. Let's then upgrade to JUnit 5. And let's then do the AssertJ discussion, based on experience and data. Again, I am open to fast-track the JUnit 5 work to move forward on this.

rsynek: I agree with the point regarding consistency. I will change this PR to use plain jUnit assertions and do the big bang migration to AssertJ later.

rsynek: @ge0ffrey just for the record, there is already AssertJ being used in optaplanner-core. Take a look at DefaultPartitionedSearchPhaseTest.

rsynek:
jenkins retest this

rsynek:
jenkins retest this

rsynek:
@ge0ffrey @yurloc can we merge the tests to see what happens on OSxJDK certification matrix? Once corresponding issues are fix, I will enable multi-threaded solving on remaining examples.

optaplanner-examples/src/main/resources/org/optaplanner/examples/nurserostering/solver/nurseRosteringSolverConfig.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <solver>
   <!--<environmentMode>FULL_ASSERT</environmentMode>--><!-- To slowly prove there are no bugs in this code -->
-  <!--<moveThreadCount>AUTO</moveThreadCount>--><!-- To solve faster by saturating multiple CPU cores -->
+  <!-- To solve faster by saturating multiple CPU cores. Use value "AUTO" to let OptaPlanner calculate optimal
+     move thread count.-->
+  <moveThreadCount>AUTO</moveThreadCount>
```
ge0ffrey: Why the different style?

optaplanner-examples/src/main/resources/org/optaplanner/examples/tsp/solver/tspSolverConfig.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <solver>
   <!--<environmentMode>FULL_ASSERT</environmentMode>--><!-- To slowly prove there are no bugs in this code -->
-  <!--<moveThreadCount>AUTO</moveThreadCount>--><!-- To solve faster by saturating multiple CPU cores -->
+  <!--<moveThreadCount>AUTO</moveThreadCount> To solve faster by saturating multiple CPU cores -->
```
ge0ffrey: The goal is that a simple "ctrl-/" in intellij enables it. This would create invalid xml

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Test
+    public void moveThreadCountThrowsExceptionWhenValueIsNegative() {
+        SolverConfig solverConfig = new SolverConfig();
+        solverConfig.setMoveThreadCount("-6");
+        try {
+            solverConfig.resolveMoveThreadCount();
+            fail("IllegalArgumentException should have been thrown.");
+        } catch (IllegalArgumentException expectedException) {
+            //expected
```
ge0ffrey: Is this possible? 
@Test(expectedException = IllegalArgumentException.class)

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/testutil/TestThreadFactory.java
```diff
@@ -0,0 +1,25 @@
+package org.optaplanner.core.config.solver.testutil;
+
+import java.util.concurrent.ThreadFactory;
+
+public class TestThreadFactory implements ThreadFactory {
```
ge0ffrey: Can we use the word Mock somewhere in the name?

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
yurloc: I think that, even though it is possible, it's the old style of checking exceptions. The way Radek wrote it, although it requires some boilerplate code, it is clearly visible which statement throws the exception, it's possible to check exception message and even continue with the test after the exception is thrown.

Once we migrate to JUnit 5, it will be possible to put it in a more elegant way:
```java


    @Test
    void exceptionTesting() {
        Throwable exception = assertThrows(IllegalArgumentException.class, () -> {
            solverConfig.resolveMoveThreadCount();
        });
        assertThat(exception).hasMessageContaining("negative");
    }
```
and Radek's style will make the migration easier.

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/app/VehicleRoutingMultiThreadedReproducibilityTest.java
```diff
@@ -0,0 +1,78 @@
...
+        VehicleRoutingSolution bestSolution = solver.solve(vehicleRoutingSolutions[iteration]);
+        vehicleRoutingSolutions[iteration] = bestSolution;
+
+        if (iteration > 0) {
+            VehicleRoutingSolution previousBestSolution = vehicleRoutingSolutions[iteration - 1];
+            assertEquals(bestSolution.getScore(), previousBestSolution.getScore());
+        }
+    }
+
+}
```
yurloc: @rsynek Thank you for keeping the limit of 128 chars per line. It makes the review on GitHub comfortable.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
rsynek: @yurloc Geoffrey's motivation for this change is to make searching for all occurrences of error checking easier - in case we have only @Test(expectedException=...), there is 0 chance to miss it.
I have checked the current source code and found out that we already mix both style - now it doesn't make sense to change it here. I will keep the test as is.

rsynek:
@ge0ffrey comments incorporated, feel free to merge.

ge0ffrey:
jenkins retest this

optaplanner-examples/src/test/java/org/optaplanner/examples/common/TestProperties.java
```diff
@@ -0,0 +1,28 @@
...
+package org.optaplanner.examples.common;
+
+public final class TestProperties {
+
+    public static final String MOVE_THREAD_COUNT = "moveThreadCount";
+
+    public static final String RUN_TURTLE_TESTS = "runTurtleTests";
+
+    private TestProperties() {
+        throw new UnsupportedOperationException();
```
ge0ffrey: for consistency with other private constructors, I'll remove this UnsupportedOperationException because it's a private constructor anyway

ge0ffrey:
jenkins retest this

ge0ffrey:
Good to merge, just waiting for jenkins to agree :)

optaplanner-examples/src/test/java/org/optaplanner/examples/common/TestProperties.java
yurloc: For consistency maybe remove it, for super-defensive programming it makes sense to keep it this way. Some people do it like that to prevent other people from doing whatever they like by using Reflections API :)

rsynek: @ge0ffrey as @yurloc said, it was for defensive programming reasons; a matter of habit. You're right there is no way to completely prevent such a misuse - it's about how far the user is willing to go.

PR #407, opened 2018-07-20
PLANNER-1168 conference scheduling devoxx rest api

MusaTalluzi:
For now, to import a conference open conference scheduling example, click on any of the files and you'll see an "import" button, when you click on it the conference data will be imported from the url specified in ConferenceSchedulingImporter.java. Once conference data is imported you can check it by opening the solution instance in an excel file.


yurloc:
No need to investigate the build failure. The reason was:
```
16:28:36 The following 2 dependencies are NOT using a managed version:
16:28:36 
16:28:36   - javax.json:javax.json-api:jar
16:28:36   - org.glassfish:javax.json:jar
```
and it has been already addressed.

optaplanner-examples/pom.xml
```diff
@@ -175,6 +175,32 @@
...
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>javax.json</groupId>
+      <artifactId>javax.json-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.json</artifactId>
+    </dependency>
```
MusaTalluzi: @jhrcek could you please check the build failure?
enforce-direct-dependencies on optaplanner-examples is failing after I added these two dependencies although the pom tree doesn't show any transitive dependencies.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,511 @@
...
+        try (InputStream inputStream = openConnectionCheckRedirects(new URL(url).openConnection())) {
+            JsonReader jsonReader = Json.createReader(inputStream);
+            return jsonReader.readArray();
+        } catch (IOException e) {
+            throw new IllegalStateException(
+                    "Import failed on URL (" + url + ").", e);
+        }
+    }
+
+    private InputStream openConnectionCheckRedirects(URLConnection connection) throws IOException { // credits for https://www.cs.mun.ca/java-api-1.5/guide/deployment/deployment-guide/upgrade-guide/article-17.html
```
rsynek: This is a very general method, suitable for some utility class. There is no reason `ConferenceSchedulingImporter` should care about details how to follow http(s) redirects.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,511 @@
...
+            }
+        }
+
+        return talkTypeSet;
+    }
+
+    private JsonObject readJsonObject(String url) {
+        try (InputStream inputStream = openConnectionCheckRedirects(new URL(url).openConnection())) {
+            JsonReader jsonReader = Json.createReader(inputStream);
+            return jsonReader.readObject();
```
rsynek: Soft suggestion: this line and the return type are the only differences between this method and `readJsonArray`. Maybe passing there  lambda and using type parameter would avoid duplication.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingRESTEndpoints.java
```diff
@@ -0,0 +1,71 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.conferencescheduling.persistence;
+
+public class ConferenceSchedulingRESTEndpoints {
```
rsynek: Given the fact you already hard-code the URLs in `ConferenceSchedulingImporter`, how about making this an Enum?

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
rsynek: It would also be nice to have a unit test for that.  E.g. http://www.mock-server.com/ might help with that - you can easily mock expected response from a server.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,511 @@
...
+    private Map<String, TalkType> talkTypeNameToTalkTypeMap;
+    private Map<String, Room> roomIdToRoomMap;
+    private Map<String, Speaker> speakerNameToSpeakerMap;
+    private Map<String, Talk> talkCodeToTalkMap;
+    private Set<String> talkUrlSet;
+
+    private ConferenceSolution solution;
+
+    public ConferenceSchedulingImporter() {
+        this.endpoints = new ConferenceSchedulingRESTEndpoints();
```
rsynek: See my comment below, this can be replaced by an enum.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingRESTEndpoints.java
MusaTalluzi: The goal is to eventually get these values from the user, but I haven't implemented the GUI interface for this yet.

rsynek: Ok

rsynek:
@MusaTalluzi thanks for incorporating changes based on the comments. One thing remains - could you please add a test for the `ConnectionFollowRedirects` class?

optaplanner-examples/pom.xml
```diff
@@ -175,6 +175,45 @@
...
+    <dependency>
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-netty</artifactId>
```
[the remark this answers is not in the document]
MusaTalluzi: I declared it under <dependencyManagement>

optaplanner-examples/src/test/java/org/optaplanner/examples/conferencescheduling/persistence/ConnectionFollowRedirectsTest.java
```diff
@@ -0,0 +1,116 @@
...
+                );
+
+        try {
+            ConnectionFollowRedirects connectionFollowRedirects = new ConnectionFollowRedirects(getUrl("path"));
+            assertTrue(connectionFollowRedirects.getConnection() instanceof HttpURLConnection);
+            connectionFollowRedirects.getInputStream().close();
+            assertEquals(connectionFollowRedirects.getRedirects(), 1);
+            assertEquals(((HttpURLConnection) connectionFollowRedirects.getConnection()).getResponseCode(), 200);
+        } catch (IOException e) {
+            e.printStackTrace();
```
rsynek: e.printStackTrace as well as System.out.println cannot be easily logged into file. Please use either logging framework or throw an exception.

optaplanner-examples/src/test/java/org/optaplanner/examples/conferencescheduling/persistence/ConnectionFollowRedirectsTest.java
```diff
@@ -0,0 +1,116 @@
...
+                );
+
+        try {
+            ConnectionFollowRedirects connectionFollowRedirects = new ConnectionFollowRedirects(getUrl("path"));
+            assertTrue(connectionFollowRedirects.getConnection() instanceof HttpURLConnection);
+            connectionFollowRedirects.getInputStream().close();
+            assertEquals(connectionFollowRedirects.getRedirects(), 0);
+            assertEquals(((HttpURLConnection) connectionFollowRedirects.getConnection()).getResponseCode(), 200);
+        } catch (IOException e) {
+            e.printStackTrace();
```
rsynek: Please see below

optaplanner-examples/pom.xml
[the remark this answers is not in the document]
ge0ffrey: `<scope>` need to go under `<dependency>` because:
- the dependencyManagement stuff will be moved to a parent pom, so other pom.xml can use that dependency too
- other poms might use it in a different scope. So to avoid all confusion, we have a convention to have scope in the child pom (and never in the dependency management section), so it's explicitly clear without lookup.

optaplanner-examples/pom.xml
```diff
@@ -175,6 +175,38 @@
...
+    </dependency>
   </dependencies>
 
+  <dependencyManagement>
```
ge0ffrey: move dependencyManagement to the optaplanner root pom, we'll work in a separate PR/issue to move it from there to where it needs to be.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingRESTEndpoints.java
ge0ffrey: baseUrl needs to come from the user, for example it could be the REST url of devoxx belgium, devoxx france, voxxed zurich or any of the other conferences that use the devoxx cfp technology.

roomsEndpoint etc aren't variable: this entire code is against the "cfp-devoxx" rest api, for example the structure of the expected JSON files and therefore YAGTNI for the paths to be variable. Given a url "foo.com/rest/api/bar/boo/rooms" the path is "bar/boo/rooms", hard code that "bar/boo/rooms", directly as a string where you need it. If you need it at 2 placed, make it a constant, otherwise don't.

Rename the classes to signal that this an importer specifically for the rest api of cfp-devoxx, and add a link to https://github.com/nicmarti/cfp-devoxx in the javadoc of that class.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConnectionFollowRedirects.java
```diff
@@ -0,0 +1,91 @@
...
+
+package org.optaplanner.examples.conferencescheduling.persistence;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLConnection;
+
+public class ConnectionFollowRedirects {
```
ge0ffrey: Let's talk about this redirects implementation during our call.
I agree we need a way to mock out the rest service.
But I am not keen on adding all this complexity and especially the dependency into dep tree.
I hope we can simply supply a "file://" protocol as the baseUrl to mock it out magically.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,496 @@
...
+import org.optaplanner.examples.conferencescheduling.domain.TalkType;
+import org.optaplanner.examples.conferencescheduling.domain.Timeslot;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConferenceSchedulingImporter {
+
+    private static final Logger logger = LoggerFactory.getLogger(ConferenceSchedulingImporter.class);
+    private static final String zoneId = "Europe/Paris";
+    private static final String[] smallRoomsTypeNames = {"Quickie Sessions", "Quickie", "Hands-on Labs", "BOF (Bird of a Feather)"};
```
ge0ffrey: static finals are constants, constanstans need to use capitals, so SMALL_ROOMS_TYPE_NAMES instead

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,496 @@
...
+import org.optaplanner.examples.conferencescheduling.domain.Talk;
+import org.optaplanner.examples.conferencescheduling.domain.TalkType;
+import org.optaplanner.examples.conferencescheduling.domain.Timeslot;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConferenceSchedulingImporter {
+
+    private static final Logger logger = LoggerFactory.getLogger(ConferenceSchedulingImporter.class);
+    private static final String zoneId = "Europe/Paris";
```
ge0ffrey: Ow, this is a pain: devoxx poland and devoxx UK and devoxx morocco are in other timezones.

Options:

- Simply default it to the default locale's timezones, so `ZoneId.systemDefault()`.
-- There's a JVM -D system property to override this per java process.
-- That presumes that the conference organizer scheduler is in the same timezone as the conference itself. That's usually true, it's a good default.
-- this is good enough for phase 1.
- We make the zoneId user input, like the restBaseUrl, defaulted to system default.
- Or we somehow get it from the rest service somehow:
-- ask stephan for a rest api to know it?
-- Look at HTTP headers to figure out the servers's timezone?
--- This is often unreliable, especially now that everything's hosted in the cloud
--- There's a plan to host the UK and FR/BE stuff on the same server (simply different tenants), IIRC.
--

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,496 @@
...
+    private Map<String, Talk> talkCodeToTalkMap;
+    private Set<String> talkUrlSet;
+    private Set<String> trackIdSet;
+
+    private ConferenceSolution solution;
+
+    public ConferenceSchedulingImporter() {
+        this.endpoints = new ConferenceSchedulingRESTEndpoints();
+        //TODO: Add a panel to get these from the user
+        this.endpoints.setBaseUrl("https://dvbe18.confinabox.com/api/conferences/DVBE18");
```
ge0ffrey: If you get all the files locally, you should be able to use something like this here for mocking out purposes:
`  setBaseUrl("file:///home/musa/projects/optaplanner/optaplanner-examples/local/cfpDevoxxMock/")`

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
```diff
@@ -0,0 +1,496 @@
...
+            trackIdSet.add(tracksArray.getJsonObject(i).getString("id"));
+        }
+    }
+
+    private void importRoomList() {
+        this.roomIdToRoomMap = new HashMap<>();
+        List<Room> roomList = new ArrayList<>();
+
+        logger.info("Sending a request to: " + endpoints.getBaseUrl() + endpoints.getRoomsEndpoint());
+        JsonObject rootObject = readJson(endpoints.getBaseUrl() + endpoints.getRoomsEndpoint(), JsonReader::readObject);
```
ge0ffrey: remove endpoints.getRoomEndpoint() and simply use

String roomsUrl = endpoints.getBaseUrl() + "/rooms/";

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
ge0ffrey: Also, when getting the baseUrl supplied from the user in a dialog, there should be some code to deal with the presence or absence of the trailing "/". Later code presumes it's not there (it probably should presume it's there), so the input needs to automatically add/remove the trailing "/" accordingly.

pom.xml
```diff
@@ -84,6 +84,24 @@
...
       </dependency>
+      <dependency>
+        <groupId>org.glassfish</groupId>
+        <artifactId>javax.json</artifactId>
+        <!-- Old version until this issue is resolved: https://github.com/ImmobilienScout24/illegal-transitive-dependency-check/issues/30 -->
+        <version>1.0.4</version>
+      </dependency>
+      <dependency>
+        <groupId>org.mock-server</groupId>
+        <artifactId>mockserver-netty</artifactId>
```
ge0ffrey: Under discussion - addding @rsynek  - wether this dependency it's worth it's weight. I don't think it is.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConnectionFollowRedirects.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private URLConnection connection;
+    private boolean isRedirect;
+    private int redirects = 0;
+
+    public ConnectionFollowRedirects(String url) throws IOException {
```
ge0ffrey: Do we need a redirect on every URL or only on the first of the rest api?

MusaTalluzi: We need it on every URL, since the url references in some of the data returns 301 status code and others don't.

For example: the first data entry in this endpoint https://dvbe18.confinabox.com/api/conferences/DVBE18/speakers contains a reference to the speaker data http://dvbe18.confinabox.com/api/conferences/DVBE18/speakers/f8caf20b7f6f226bf0637bc8e077d95161561439 which returns a response with code 301 (it redirects to https).

pom.xml
MusaTalluzi: For reference, this is the dependency tree of mockserver-netty:
```
[INFO] +- org.mock-server:mockserver-netty:jar:5.3.0:test
[INFO] |  +- org.mock-server:mockserver-client-java:jar:5.3.0:test
[INFO] |  +- org.mock-server:mockserver-core:jar:5.3.0:test
[INFO] |  |  +- javax.servlet:javax.servlet-api:jar:3.1.0:test
[INFO] |  |  +- io.netty:netty-codec-socks:jar:4.1.17.Final:test
[INFO] |  |  +- com.jcraft:jzlib:jar:1.1.3:test
[INFO] |  |  +- org.bouncycastle:bcmail-jdk15on:jar:1.58:test
[INFO] |  |  +- org.bouncycastle:bcpkix-jdk15on:jar:1.58:test
[INFO] |  |  +- org.apache.velocity:velocity-engine-scripting:jar:2.0:test
[INFO] |  |  +- org.apache.velocity:velocity-engine-core:jar:2.0:test
[INFO] |  |  +- org.skyscreamer:jsonassert:jar:1.2.3:test
[INFO] |  |  |  \- org.json:json:jar:20090211:test
[INFO] |  |  +- com.github.java-json-tools:json-schema-validator:jar:2.2.8:test
[INFO] |  |  |  +- com.github.java-json-tools:json-schema-core:jar:1.2.8:test
[INFO] |  |  |  |  +- org.mozilla:rhino:jar:1.7R4:test
[INFO] |  |  |  |  +- com.github.fge:jackson-coreutils:jar:1.8:test
[INFO] |  |  |  |  |  \- com.github.fge:msg-simple:jar:1.1:test
[INFO] |  |  |  |  |     \- com.github.fge:btf:jar:1.2:test
[INFO] |  |  |  |  \- com.github.fge:uri-template:jar:0.9:test
[INFO] |  |  |  +- javax.mail:mailapi:jar:1.4.3:test
[INFO] |  |  |  +- joda-time:joda-time:jar:2.9.7:test
[INFO] |  |  |  +- com.googlecode.libphonenumber:libphonenumber:jar:8.0.0:test
[INFO] |  |  |  +- com.google.code.findbugs:jsr305:jar:3.0.1:test
[INFO] |  |  |  \- net.sf.jopt-simple:jopt-simple:jar:5.0.3:test
[INFO] |  |  +- org.apache.commons:commons-text:jar:1.1:test
[INFO] |  |  \- commons-collections:commons-collections:jar:3.2.2:test
[INFO] |  +- io.netty:netty-buffer:jar:4.1.17.Final:test
[INFO] |  +- io.netty:netty-codec:jar:4.1.17.Final:test
[INFO] |  +- io.netty:netty-codec-http:jar:4.1.17.Final:test
[INFO] |  +- io.netty:netty-common:jar:4.1.17.Final:test
[INFO] |  +- io.netty:netty-handler:jar:4.1.17.Final:test
[INFO] |  +- io.netty:netty-transport:jar:4.1.17.Final:test
[INFO] |  |  \- io.netty:netty-resolver:jar:4.1.17.Final:test
[INFO] |  +- org.bouncycastle:bcprov-jdk15on:jar:1.58:test
[INFO] |  +- janino:janino:jar:2.5.10:test
[INFO] |  \- org.mock-server:mockserver-logging:jar:5.3.0:test
```

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingImporter.java
yurloc: Just for the record, not all static final variables are constants. For example you can have a static final collection, which, unless you create it as unmodifiable, can be easily changed (added and removed to) and therefore is not a constant and shouldn't use the all capitals convention.

The deciding factor of course is the author's intention about the purpose of the variable.

Another example is the logger field, which is always kept lower case in optaplanner code, although it is a static final variable. I think that is correct, since, no matter if the logger has an internal state that's being modified or not, it is not intended to be used as a constant.

optaplanner-examples/pom.xml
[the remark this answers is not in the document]
ge0ffrey: Can we remove the mockserver-netty usage?

pom.xml
ge0ffrey: We'd like to have the mocknetty-server dependency removed - even if that impacts test coverage negatively. The quality of that dependency isn't high enough:
- bloat (too many dependencies)
- very old unmaintained dependencies (velocity, bcprov-jdk15, ...)

ge0ffrey:
Merged (I had to squash merge because rebase merge didn't work)

PR #408, opened 2018-07-23
PLANNER-1175 Fail fast if scriptLanguage is unknown

ge0ffrey:


ge0ffrey:
The case were this happens is a really exotic JVM (a scala build hacked thing as far as I understand). The "Maybe ..." sentences are about pointing to where to start looking. I'd start looking at what JVM is being run - I fear that pointing to the JSR 223, will just lead them to start looking on how to include the JSR 223 dependency jar (and it's not a dependencies problem).

ge0ffrey:
Jenkins failed because of a timeout. Jenkins is unreliable and it's a big pain.

PR #409, opened 2018-07-30
Planner 1193 conf scheduling test xlsx

MusaTalluzi:
Improve test.xlsx file so it's readable by business people


optaplanner-examples/src/test/java/org/optaplanner/examples/conferencescheduling/solver/ConferenceSchedulingScoreRulesXlsxTest.java
```diff
@@ -48,53 +52,66 @@
...
+        this.constraintName = constraintName;
+        this.expectedScore = expectedScore;
+        this.solution = solution;
+        this.testSheetName = testSheetName;
+    }
+
+    @Parameterized.Parameters(name = "{4}")
+    public static Collection testSheetParameters() {
+        List<Object[]> parametersList = new ArrayList<>();
+        Object[] currentParameterList;
```
rsynek: This is just a nitpicking to improve readability a bit - the `currentParameterList` could be declared right before the while loop, as this is the place where it's used for the first time.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingXlsxFileIO.java
```diff
@@ -68,6 +68,18 @@
...
 public class ConferenceSchedulingXlsxFileIO extends AbstractXlsxSolutionFileIO<ConferenceSolution> {
 
+    private static boolean isTest;
+
+    public ConferenceSchedulingXlsxFileIO() {
+        super();
+        this.isTest = false;
+    }
+
+    public ConferenceSchedulingXlsxFileIO(boolean isTest) {
```
rsynek: @MusaTalluzi One more question, what's the reason for introducing the variable "isTest" and throwing exceptions only if the `isTest = false` ?

MusaTalluzi: The xlsx test file doesn't necessarily contain all the information required for a complete conference schedule, just the ones necessary to test one or more score rules. So some of the fields (e.g. audience level for Talks) is required for a real life schedule but not if you're testing room conflict.
Also the test file is most likely written (and reviewed) by business people, so it does't need to stick to the regex for codes and ids.
In general it's to make rules more lenient for test file, especially if you're testing one constraint.

rsynek: @MusaTalluzi thanks for clarification. Mixing these rules with reading an input makes the code even more complicated, especially if in every step you need to decide whether throw an exception or not.

Let me offer you a different solution:
Decoupling the checking logic from reading the file. You can check the resulting domain object tree instead of doing it "on-the-fly". Even if you for some reason really need to do it "on-the-fly", you can still delegate the checking to another class.

MusaTalluzi: Thanks @rsynek, I think this is a good idea, and instead of throwing an exception at the first violation the checker class logs all the violations and throws an exception at the end. So if there are too many violations they're all fixed at once instead of running the program too many times and fixing them one by one.

However, this contradicts the fail-fast approach we've been following, what do think @ge0ffrey ?

rsynek: @MusaTalluzi if you truly need to preserve both fail-fast and violation report after reading, you can e.g. have two different implementations of the checker and pass the impl. via a constructor.

ge0ffrey: @MusaTalluzi (I'll merge anyway, so in a seperate PR): I understand Radovan's confusion. Instead of explaining or commenting it, let's see if we can just make the code clearer. What if we rename "isTest" to "strict" (and inverse all it's usages)? A boolean "strict" is a very common way differentiation between a fail-fast parser and a lenient parser.
The default should be strict=true of course, and the no-arg constructor should call this(strict = true)

ge0ffrey: @rsynek I am not convinced the suggestion of "splitting up the validation in a separate class" will work or will be more readable. But it's worth an experiment. I am merging this PR so that that experiment is a refactor PR on top of this one.

@MusaTalluzi Please try Radavon's suggestion in a seperate PR. Also do my rename "isTest" to "strict" (and inverse it) request (you can do that fast in a PR and ping me and I 'll merge quickly, before you start Radovan's suggestion).

PR #414, opened 2018-08-17
PLANNER-1204 and PLANNER-1226 Multithreaded Incremental Solving improvements

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/thread/MoveThreadRunner.java
```diff
@@ -46,6 +46,7 @@
     private final boolean assertShadowVariablesAreNotStaleAfterStep;
 
     private InnerScoreDirector<Solution_> scoreDirector = null;
+    private volatile long calculationCount = -1L;
```
rsynek: If multiple threads can write to this variable, `AtomicLong` should be used instead of `volatile`.

ge0ffrey: true. long's are not guaranteed to have atomic read/writes, unlike int's.
only 1 thread writes to it, but that doesn't excuse it.

ge0ffrey: Done

PR #415, opened 2018-08-19
Add logic to prevent arithmetic overflow of nearby entity count limit

Woodz:
Resolve: PLANNER-1225


kiereleaseuser2:

kiereleaseuser2:

Woodz:
This is a hotfix for 7.0.0 but should also be applied to master and any other legacy hotfix branches

ge0ffrey:
Thanks @Woodz - nice work

ge0ffrey:
@rsynek I think the combination of BetaDistributionNearbyRandom and discardNearbyIndexZero=true isn't unit tested, otherwise we'd have seen this already. Overall, our unit testing of nearby selection is weak (none of the examples have it activated).

The main reason it's not activated in the examples, is because we don't want it to happen for small datasets. So we need an auto-tuning feature - which I haven't had time to work on yet. None the less, regardless, it's probably a good idea that we build some integration like tests around this for like we have for multithreaded solving? Wdyt?

rsynek:
@ge0ffrey I'm not sure this issue could be caught by an integration test. Best start is usually writing unit tests guided by a coverage report. We already have it in our backlog https://issues.jboss.org/browse/PLANNER-544. Covering many different combinations by integration tests is expensive, thus I would rather build it on top of a solid unit coverage.

PR #420, opened 2018-09-04
Planner-1238 supply Move.rebase bug

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
```diff
@@ -38,20 +38,124 @@
...
+    protected final Object rightAnchor;
+
+    protected final boolean sameAnchor;
+    protected final Object leftNextEntity;
+    protected final Object rightNextEntity;
+    protected final boolean reverseAnchorSide;
+    protected final Object lastEntityInChain;
+    protected final Object entityAfterAnchor;
 
     public TailChainSwapMove(GenuineVariableDescriptor<Solution_> variableDescriptor,
```
rsynek: The constructor now includes a non-trivial logic, would it be possible to include a unit test for it? The current tests focus on existing constructor with the supply.

optaplanner-examples/src/main/resources/org/optaplanner/examples/tsp/benchmark/tspBenchmarkConfig.xml
```diff
@@ -47,7 +47,7 @@
         <unionMoveSelector>
           <changeMoveSelector/>
           <swapMoveSelector>
-            <cacheType>PHASE</cacheType>
+            <cacheType>STEP</cacheType>
```
rsynek: @ge0ffrey could you please shortly summarize why TSP needs this change of cacheType?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
ge0ffrey: This is the constructor with the supply's - that's the one heavily used by the existing unit test.
Do you mean the other constructors?

optaplanner-examples/src/main/resources/org/optaplanner/examples/tsp/benchmark/tspBenchmarkConfig.xml
ge0ffrey: Yes, this is the trade-off I mentioned earlier: we can no longer do phase caching on chained changeMoves and chained swapMoves: now that fails at bootstrap instead of working. (In the past it had bugs on other types and those now fail fast during bootstrap too).
The benefit of this is that we can use those chained change moves and swap moves in multithreaded solving (just like all other moves).

I 've mentioned this in the upgrade recipe: https://github.com/kiegroup/optaplanner-website/commit/b32be369bb380f48129a4d7e9cd490678af032d5

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
rsynek: Sorry, I confused the constructors. Truth, this one is tested in unit tests, the new constructors are not. And there is not much to test in them - maybe in connection with `rebase()`.

optaplanner-examples/src/main/resources/org/optaplanner/examples/tsp/benchmark/tspBenchmarkConfig.xml
rsynek: Thanks!

ge0ffrey:
@rsynek Docs added :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
```diff
@@ -38,20 +38,124 @@
...
-        this.anchorVariableSupply = anchorVariableSupply;
         this.leftEntity = leftEntity;
+        leftValue = variableDescriptor.getValue(leftEntity);
+        leftAnchor = anchorVariableSupply.getAnchor(leftEntity);
+        rightEntity = inverseVariableSupply.getInverseSingleton(rightValue);
+        this.rightValue = rightValue;
+        rightAnchor = variableDescriptor.isValuePotentialAnchor(rightValue) ? rightValue
+                : anchorVariableSupply.getAnchor(rightValue);
+        sameAnchor = leftAnchor == rightAnchor;
+        if (!sameAnchor) {
```
rsynek: It's more natural to have 'positive path' in the `if` branch and 'negative' in the `else` branch. Basically, if there is an if-else statement, you shouldn't use negation in the condition.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
```diff
@@ -110,52 +205,49 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
-            if (rightEntity != null) {
-                return new TailChainSwapMove<>(variableDescriptor, inverseVariableSupply, anchorVariableSupply,
-                        rightEntity, rightValue);
-            } else {
+            if (rightEntity == null) {
                 // TODO Currently unsupported because we fail to create a valid undoMove... even though doMove supports it
+                // https://issues.jboss.org/browse/PLANNER-1250
                 throw new IllegalStateException("Impossible state, because isMoveDoable() should not return true.");
             }
+            if (!reverseAnchorSide) {
```
rsynek: ditto

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
```diff
@@ -192,9 +272,39 @@ protected void reverseChain(ScoreDirector scoreDirector, Object fromValue, Objec
 
     @Override
     public TailChainSwapMove<Solution_> rebase(ScoreDirector<Solution_> destinationScoreDirector) {
-        return new TailChainSwapMove<>(variableDescriptor, inverseVariableSupply, anchorVariableSupply,
-                destinationScoreDirector.lookUpWorkingObject(leftEntity),
-                destinationScoreDirector.lookUpWorkingObject(rightValue));
+        if (!sameAnchor) {
```
rsynek: ditto

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
```diff
@@ -192,9 +272,39 @@ protected void reverseChain(ScoreDirector scoreDirector, Object fromValue, Objec
...
+            return new TailChainSwapMove<>(variableDescriptor,
+                    destinationScoreDirector.lookUpWorkingObject(leftEntity),
+                    destinationScoreDirector.lookUpWorkingObject(leftValue),
+                    destinationScoreDirector.lookUpWorkingObject(leftAnchor),
+                    destinationScoreDirector.lookUpWorkingObject(rightEntity),
+                    destinationScoreDirector.lookUpWorkingObject(rightValue),
+                    destinationScoreDirector.lookUpWorkingObject(rightAnchor));
+        } else {
+            if (!reverseAnchorSide) {
```
rsynek: ditto

PR #425, opened 2018-09-06
[7.11.x] Skip failing test for moveThreadCount > 0

rsynek:
The test fails due to score corruption in RockTour example, which manifests with multiple move threads.

@yurloc could you please take a look?


rsynek:
@yurloc could you please merge this PR?

PR #427, opened 2018-09-11
Reproducer for rhdm-730

rsynek:
@ge0ffrey a reproducer for https://issues.jboss.org/browse/RHDM-730


rsynek:
@ge0ffrey big thanks for helping me to find the root cause.

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMoveTest.java
```diff
@@ -278,4 +280,39 @@ public void toStringTest() {
                 inverseVariableSupply, anchorVariableSupply, a2, a0).toString());
     }
 
+    @Test
+    public void retrievedPlanningEntitiesContainNeverNull() {
```
ge0ffrey: It's nice if the test method name starts with (or equals) the method being tested.
Would you mind renaming it to getPlanningEntitiesWithRightEntityNull() ?

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMoveTest.java
```diff
@@ -278,4 +280,39 @@ public void toStringTest() {
...
+        assertPlanningEntitiesDoNotContainNull(move);
+
+        move.doMoveOnGenuineVariables(scoreDirector);
+        assertPlanningEntitiesDoNotContainNull(move);
+
+        Move undoMove = move.createUndoMove(scoreDirector);
+        assertPlanningEntitiesDoNotContainNull(undoMove);
+    }
+
+    private void assertPlanningEntitiesDoNotContainNull(final Move move) {
```
ge0ffrey: soft suggestion: I'd inline this method - it's easier to read that way.

rsynek:
@ge0ffrey suggestions incorporated

PR #435, opened 2018-09-25
Conference scheduling diagrams

ge0ffrey:


ge0ffrey:
Misha, it's not hooked up in the docs yet, but I needed to make these for the slides/website. I 'll hook them up in the docs in some point probably.

PR #439, opened 2018-09-26
Planner 1297 import conference button

MusaTalluzi:


MusaTalluzi:
@rsynek For the first suggestion, @ge0ffrey and I discussed it before but decided not to go with it because it's a bit complicated and requires redesigning many components of the UI.

Using a separate thread for importing the data allows users to play with the UI afterwards, at the end of the import the thread will change the workingSolution in the Panel (which will override any actions the user has taken while the importing was in progress), so I think freezing the UI makes sense in this case, unless we can prevent executing other actions while the importing is in progress.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/swingui/ConferenceSchedulingPanel.java
```diff
@@ -64,14 +82,23 @@ public ConferenceSchedulingPanel() {
                 throw new IllegalStateException("Failed to show temp file (" + tempFile + ") in LibreOffice or Excel.", e);
             }
         });
-        add(importConferenceButton);
-        add(publishButton);
-        add(button);
-        add(new JLabel("Changes to that file are ignored unless you explicitly save it there and open it here."));
+
+        setLayout(new GridLayout(12, 1));
```
ge0ffrey: GridLayout with 12 rows. That works well in the original's windows size, but try resizing it to something much smaller or bigger. It needs to be "responsive", so handle differenet window sizes.
Use a vertical BoxLayout instead.

MusaTalluzi: The problem with BoxLayout is when window size is too large the panels containing the buttons are stretched too much to fit the window (which results in too much space between the buttons). Also using VerticalGlue didn't reduce the distance enough, that's why I used GridLayout, wdyt?

ge0ffrey: Alternatively, what you can do is use a BorderLayout for the other panel and add a 2th panel in there BorderLayout.NORTH, That 2th panel can uses a 2-row GridLayout. That should make the resizing behave as it should.

rsynek:
@MusaTalluzi regarding the counter-intuitive import button, I won't hold of merging because of this. But it's a pity this feature will stay hidden to users who are not curious enough to click through xlsx import to get there.

Currently the freezing of UI is not safe - some actions that impatient user does take effect once the freeze is over; often leading to accidentally opening dialogs, etc. One possible way of solving it might be introducing a modal dialog informing about the fact that import is in progress and preventing user from doing any action in UI during import.

rsynek:
@MusaTalluzi here is an issue for the hidden import button https://issues.jboss.org/browse/PLANNER-1303

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/swingui/ConferenceSchedulingPanel.java
```diff
@@ -16,40 +16,62 @@
...
+            JTextField cfpRestUrlTextField = new JTextField("https://dvbe18.confinabox.com/api/conferences/DVBE18");
+            Object[] dialogue = {
+                    "Choose conference:", cfpConferenceBox,
+                    "Enter CFP REST Url:", cfpRestUrlTextField
+            };
+
+            int option = JOptionPane.showConfirmDialog(this, dialogue, "Import", JOptionPane.OK_CANCEL_OPTION);
+            if (option == JOptionPane.OK_OPTION) {
+                String conferenceBaseUrl = cfpRestUrlTextField.getText();
+                newSingleThreadExecutor().execute(new ConferenceImporter(conferenceBaseUrl, solutionBusiness));
```
rsynek: ExecutorService needs to be shutdown to properly clean up running threads. Please see https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html for details.

rsynek:
@MusaTalluzi Thanks! The progress feedback and blocking are good ideas, I am fine with doing it in a separate Jira ticket.

Please see my comment regarding executor service shutdown.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/swingui/ConferenceSchedulingPanel.java
MusaTalluzi: It looks too much complication for executing one task here, is there a reason not to use new Thread().start() instead?

rsynek: @MusaTalluzi I am not aware of any, new Thread().start() should work too.

ge0ffrey: Don't use new Thread().start() but use SwingWorker (which wraps that) instead.
See SolveWorker in SolverAndPersistenceFrame

ge0ffrey:
LGTM, merged

PR #440, opened 2018-09-26
PLANNER-1298: Import talks from REST API instead of local files

MusaTalluzi:


optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingCfpDevoxxImporter.java
```diff
@@ -221,128 +219,36 @@ private void importSpeakerList() {
...
 
-/* TODO: Uncomment this when the REST api exposes the talks
     private void importTalkList() {
         this.talkCodeToTalkMap = new HashMap<>();
-        List<Talk> talkList = new ArrayList<>();
-        Long talkId = 0L;
+        solution.setTalkList(new ArrayList<>());
+
+        String talksUrl = conferenceBaseUrl + "/talks";
+        LOGGER.debug("Sending a request to: " + talksUrl);
```
ge0ffrey: Use logger.debug("Sending a request to: {}", talksUrl);
The difference is that the string concatenation won't happen if the logger isn't set to debug, saving performance. Not that you'll notice the perf difference, but it's a good practice to do it right everywhere.

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingCfpDevoxxImporter.java
```diff
@@ -354,35 +260,35 @@ private void importTalkList() {
...
 
         return speakerList;
     }
 
-    private Speaker getSpeakerOrCreateOneIfNull(String confType, String code, String title, String speakerName) {
+    private Speaker getSpeakerOrCreateOneIfNull(String code, String title, String speakerName) {
         Speaker speaker = speakerNameToSpeakerMap.get(speakerName);
         if (speaker == null) {
-            LOGGER.warn("The talk (" + code + ": " + title + ") of type ( " + confType
+            LOGGER.warn("The talk (" + code + ": " + title
```
ge0ffrey: see previous comment

PR #442, opened 2018-10-01
PLANNER-1200 subtract weekly budget from total seconds

rsynek:
When driving seconds are over weekly budget, the budget should be
subtracted instead of putting the driving seconds to 0.

@ge0ffrey would you kindly review?

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/rocktour/domain/solver/RockShowVariableListener.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/rocktour/domain/solver/RockShowVariableListener.java
@@ -118,13 +118,15 @@ public class RockShowVariableListener implements VariableListener<RockShow> {
                 timeOfDay = RockTimeOfDay.EARLY;
             }
         }
-        hosWeekDrivingSecondsTotal += drivingSeconds;
+
+        hosWeekDrivingSecondsTotal += show.getDrivingTimeFromPreviousStandstill();
+
         // HOS driving time per week limits: add weekend rest period if driving for too many hour or too many days
         if (hosWeekDrivingSecondsTotal > hosWeekDrivingSecondsBudget
                 || hosWeekStart.getDepartureDate().until(arrivalDate, DAYS) > hosWeekConsecutiveDrivingDaysBudget) {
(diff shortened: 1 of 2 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/rocktour/domain/solver/RockShowVariableListener.java
```diff
@@ -118,13 +118,15 @@ private Arrival calculateArrival(RockTourSolution solution, RockShow show, RockS
...
+
+        hosWeekDrivingSecondsTotal += show.getDrivingTimeFromPreviousStandstill();
+
         // HOS driving time per week limits: add weekend rest period if driving for too many hour or too many days
         if (hosWeekDrivingSecondsTotal > hosWeekDrivingSecondsBudget
                 || hosWeekStart.getDepartureDate().until(arrivalDate, DAYS) > hosWeekConsecutiveDrivingDaysBudget) {
             arrivalDate = arrivalDate.plusDays(hosWeekRestDays);
             hosWeekStart = show;
-            hosWeekDrivingSecondsTotal = 0L;
+            hosWeekDrivingSecondsTotal -= hosWeekDrivingSecondsBudget;
```
ge0ffrey: +1 This makes complete sense: if a you can not drive more than 50 hours per week and you're driving 56 hours, you should rest for 2 days after driving 50 hours, but you still need to drive those 6 hours in next week's budget.

This could indeed cause the score corruption.

On note, for further enhancement, that is not a priority right now:
- What happens if a driver can only drive 40 hours in one week and a single trip takes 90 hours? This is impossible in the US dataset, as a cross-country trip takes less time. That "if" would need to be in a loop to fix that.

ge0ffrey:
Great work!!

optaplanner-examples/src/main/java/org/optaplanner/examples/rocktour/domain/solver/RockShowVariableListener.java
rsynek: You're right; ideally it should be in a loop, but it needs reviewing the condition - I think I played with this and it didn't work properly. As you say, the dataset does not contain drives that would span across more than 2 weeks.

PR #444, opened 2018-10-02
PLANNER-1286 Make tests pass on JDK 11

rsynek:
@ge0ffrey for the web examples, I saw no failures/warning from the enforcer plugin as your comment there suggested. So I guess we are fine there with the general javaee-api dependency.


optaplanner-examples/src/main/java/org/optaplanner/examples/curriculumcourse/domain/Lecture.java
```diff
@@ -60,7 +60,8 @@ public boolean isPinned() {
         return pinned;
     }
 
-    public void setPinned(boolean pinned) {
+    // Object type Boolean required due too Lambda creating on JDK 11
+    public void setPinned(Boolean pinned) {
```
ge0ffrey: That's a pain - users will want to use a boolean instead of a Boolean here.

Looks like a regression in JDK 11 for LamdbaMetaFactory. We must create a ticket for that.

ge0ffrey: @rsynek Would you mind creating a new git repo with a reproducer that succeeds in JDK 8 but not in JDK 11?
Basically it's a copy paste of this file:
https://github.com/ge0ffrey/ge0ffrey-presentations/blob/master/code/fasterreflection/fasterreflection-framework/src/main/java/be/ge0ffrey/presentations/fasterreflection/framework/LambdaMetafactoryBeanPropertyReader.java
with
```
private final BeanPropertyReader lambdaMetafactoryBeanPropertyReader = new LambdaMetafactoryBeanPropertyReader(Foo.class, "pinned");
lambdaMetafactoryBeanPropertyReader.executeGetter(foo);
```

Once we have such a bug report, let's mail it to Andrew and ask what's the best way forward to bring them to the openJDK's team's attention.

triceo: @ge0ffrey It's not a regression - they just deliberately added type checking. I think we can work around this on our end by creating extra methods that process primitive types and not just `Object`.

rsynek: @ge0ffrey +1, changing primitive types to Object type is backward incompatible workaround, not a solution.  @triceo thanks for making me realize that.

Given BiConsumer does not accept primitive types, we may need to map to BiConsumers for every primitive type, something like https://docs.oracle.com/javase/8/docs/api/java/util/function/ObjIntConsumer.html

ge0ffrey: @rsynek @triceo Agreed, please create a jira for this work.
It will involve fixing this the LambdaMetafactory.metafactory() calls in this class:
https://github.com/kiegroup/optaplanner/blob/5c33d3c0490b5d42a40178b76003874a363b720b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LambdaBeanPropertyMemberAccessor.java#L87

That's 5 lines of headache code. The first thing to know that is any set of method parameters, the first type is the return type. Not that the JDK javadocs mention that anywhere...

rsynek:
Ok, so the enforcer on PR check still complains about duplicate classes from javaee-api artifact; I didn't run this detection locally as it's defined in fullProfile.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterAccessor.java
```diff
@@ -0,0 +1,101 @@
...
+import java.lang.reflect.Method;
+
+public abstract class PropertySetterAccessor<T> {
+
+    private final Method setterMethod;
+    private final Class<?> propertyType;
+    private final MethodHandles.Lookup lookup;
+    private final T setterFunction;
+
+    public static PropertySetterAccessor create(final Method setterMethod,
```
triceo: Just out of curiosity, @rsynek - do we have a reason to worry about `char[]`, `int[]` etc.? Would be interesting to see if this code works for them as well. (Although they should probably hide under the `Object` case.)

rsynek: yes, tried with int [] and it falls under Object

rsynek:
@triceo good point, added unit tests. Thanks!
@ge0ffrey feel free to merge

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterAccessorTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        int integerProperty;
+        boolean booleanProperty;
+        String objectProperty;
+        double doubleProperty;
+        float floatProperty;
+        short shortProperty;
+        char charProperty;
+        byte byteProperty;
+        char[] arrayProperty;
```
ge0ffrey: let's make these fields private

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterAccessorTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        public void setCharProperty(final char charProperty) {
+            this.charProperty = charProperty;
+        }
+
+        public void setByteProperty(final byte byteProperty) {
+            this.byteProperty = byteProperty;
+        }
+
+        public void setArrayProperty(final char[] arrayProperty) {
```
ge0ffrey: We don't use the "final" keyword before method parameters normally. So for consistency with the rest of the code, would you mind removing it?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/ShortPropertySetterAccessor.java
```diff
@@ -0,0 +1,53 @@
...
+package org.optaplanner.core.impl.domain.common.accessor.lambda;
+
+import java.lang.invoke.CallSite;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.Method;
+
+import org.optaplanner.core.impl.domain.common.accessor.lambda.consumer.ObjectShortConsumer;
+
+final class ShortPropertySetterAccessor extends PropertySetterAccessor<ObjectShortConsumer> {
```
ge0ffrey: we don't use the "friendly" modifier normally, lets make this class public?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/ShortPropertySetterAccessor.java
```diff
@@ -0,0 +1,53 @@
...
+
+    ShortPropertySetterAccessor(final Method setterMethod,
```
ge0ffrey: we don't use the friendly modifier normally. Let's make this constructor public. It's in the "impl" package, so it's not part of the API.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterAccessorTest.java
rsynek: I tend to rather keep it as is, if you don't mind; changing to private would require the test to call setAccessible(true).

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterAccessorTest.java
rsynek: For the sake of consistency, ok, will do

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterAccessor.java
```diff
@@ -0,0 +1,101 @@
...
+                                                final Class<?> propertyType,
+                                                final MethodHandles.Lookup lookup) {
+        if (Object.class.isAssignableFrom(propertyType)) {
+            return new ObjectPropertySetterAccessor(setterMethod, propertyType, lookup);
+        } else if (boolean.class.isAssignableFrom(propertyType)) {
```
ge0ffrey: Are you sure the isAssignableFrom() call is correct/fast? I am thinking unboxing etc (although the Object isAssignable would catch that).

Class has method isPrimitive(), which javadocs mentions Long.TYPE, etc.

In ConfigUtils, we do:

                } else if (propertyType.equals(Boolean.TYPE) || propertyType.equals(Boolean.class)) {
                    typedValue = Boolean.parseBoolean(valueString);
                } else if (propertyType.equals(Integer.TYPE) || propertyType.equals(Integer.class)) {
                    typedValue = Integer.parseInt(valueString);

So, just thinking aloud, would it make more sense to replace:
   `} else if (boolean.class.isAssignableFrom(propertyType)) {`
with
   `} else if (propertyType.equals(Boolean.TYPE)) {`

or vica versa in ConfigUtils? I'd argue the second approach is faster.

rsynek: I didn't consider which approach might be faster; this code is invoked once when the lambda is being created. Thanks for this hint, will replace it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/ShortPropertySetterAccessor.java
rsynek: I am inclined to leave it package-private; the idea is to isolate this set of classes as much as possible. Classes outside the package should not be aware about such details.

You're right that what's inside impl package is not a part of a public API, on the other hand, this just follows the general least privilege principle.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/ShortPropertySetterAccessor.java
rsynek: the same argument as above

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LambdaBeanPropertyMemberAccessor.java
```diff
@@ -127,7 +104,7 @@ public boolean supportSetter() {
 
     @Override
     public void executeSetter(Object bean, Object value) {
-        setterFunction.accept(bean, value);
+        propertySetterAccessor.executeSetter(bean, value);
```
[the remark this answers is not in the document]
ge0ffrey: Ideally, in the Object path scenario should just have setterFunction here.
For primitives, we can decorate the setterFunction with a cast (following in decorater design pattern) in the worst case scenario.

Let's set up a meeting tomorrow morning to discuss.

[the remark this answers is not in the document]
ge0ffrey: How about approach:
```

if (type == Long.TYPE) {
  setterFunction = new LongUnboxingBiConsumer(setterFunction);
}

LongUnboxingSetterFunction implements BiConsumer<Object, Object> {

   void accep(Object a, Object b) {
      setterFunction.accept(((Long) b).longValue());
   }

```

rsynek:
@ge0ffrey so the approach with UnboxingSetterFunction as a wrapper of the generated Consumer works, which effectively turns the Accessor classes into just factories.

Sorry for making review less pleasant - the renaming has broken the diff.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/lambda/PropertySetterFactory.java
```diff
@@ -0,0 +1,102 @@
...
+        } else if (propertyType.equals(Double.TYPE)) {
+            propertySetterFactory = new DoublePropertySetterFactory(setterMethod, propertyType, lookup);
+        } else if (propertyType.equals(Character.TYPE)) {
+            propertySetterFactory = new CharPropertySetterFactory(setterMethod, propertyType, lookup);
+        } else if (propertyType.equals(Byte.TYPE)) {
+            propertySetterFactory = new BytePropertySetterFactory(setterMethod, propertyType, lookup);
+        } else if (propertyType.equals(Short.TYPE)) {
+            propertySetterFactory = new ShortPropertySetterFactory(setterMethod, propertyType, lookup);
+        } else {
+            throw new IllegalArgumentException("Unsupported property type " + propertyType);
```
ge0ffrey: All optaplanner's exceptions use this style:

`throw new IllegalArgumentException("Unsupported property type (" + propertyType + ").);`

rsynek: ok

PR #448, opened 2018-10-08
PLANNER-1315 Score corruption analysis with multithreaded solving

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -507,14 +512,71 @@ public void assertExpectedWorkingScore(Score expectedWorkingScore, Object comple
...
     public void assertShadowVariablesAreNotStale(Score expectedWorkingScore, Object completedAction) {
+        Map<ShadowVariableDescriptor, List<String>> shadowVariableDescriptorListMap = calculateShadowVariableViolationListMap();
+        if (!shadowVariableDescriptorListMap.isEmpty()) {
+            StringBuilder message = new StringBuilder();
+            message.append(VariableListener.class.getSimpleName()).append(" corruption after completedAction (")
+                    .append(completedAction).append("):\n");
+            shadowVariableDescriptorListMap.forEach((shadowVariableDescriptor, violationList) -> {
+                message.append(violationList.get(0));
+                if (violationList.size() > 1) {
+                    message.append("  ...\n");
```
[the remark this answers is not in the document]
ge0ffrey: Will also show 3 and show how many more in the "..." line

ge0ffrey:
I've added a test for shadow variable corruption. I don't immediately see a practical way to add one for score corruption :/

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/easy/EasyScoreDirectorTest.java
```diff
@@ -45,4 +56,33 @@ public void constraintMatchTotalsUnsupported() {
...
+        TestdataCorruptedShadowedSolution solution = new TestdataCorruptedShadowedSolution("s1");
+        TestdataValue v1 = new TestdataValue("v1");
+        TestdataValue v2 = new TestdataValue("v2");
+        solution.setValueList(Arrays.asList(v1, v2));
+        TestdataCorruptedShadowedEntity e1 = new TestdataCorruptedShadowedEntity("e1");
+        TestdataCorruptedShadowedEntity e2 = new TestdataCorruptedShadowedEntity("e2");
+        solution.setEntityList(Arrays.asList(e1, e2));
+        scoreDirector.setWorkingSolution(solution);
+
+        scoreDirector.assertShadowVariablesAreNotStale(SimpleScore.valueOfUninitialized(-2, 0), "NoChange");
```
rsynek: This line should never throw an exception, right? While the last line of this test should. If the test uses `@Test(expected = IllegalStateException.class)`, it will pass regardless where the exception has been thrown. I would suggest using try-catch with assertion of the `IllegalStateException` object for the last line of this test.

ge0ffrey: Ok, I'll make the change, but I am not a fan of how the code will look like.
We really need to upgrade to JUnit 5, which has nice support for that.

ge0ffrey:
Fixed, merged, thanks Radovan :)

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/easy/EasyScoreDirectorTest.java
rsynek: I don't like this code either, but until we migrate to jUnit5 or AssertJ, this is the only safe way to do such a checking.

PR #449, opened 2018-10-10
Planning parametrization

ge0ffrey:
It's not finished yet, left todo:

- docs
- Add methods to Score and ScoreHolder in all scores other than HardMediumSoftScore
- Migrate all examples that use Parameterization, currently only meeting scheduling has been handled

@rsynek @MusaTalluzi Besides an entire review, I am particularly interested in your opinion on the impact for the user, so the changes on the Meeting Scheduling example. Is this a step forward?


optaplanner-examples/src/main/resources/org/optaplanner/examples/meetingscheduling/solver/meetingSchedulingScoreRules.drl
```diff
@@ -112,50 +111,45 @@ end
 
 rule "Do all meetings as soon as possible"
     when
-        MeetingParametrization($weight: doAllMeetingsAsSoonAsPossible != 0)
         MeetingAssignment(startingTimeGrain != null, $lastTimeGrainIndex : lastTimeGrainIndex)
     then
-        scoreHolder.addSoftConstraintMatch(kcontext, - $lastTimeGrainIndex * $weight);
+        scoreHolder.match(kcontext, $lastTimeGrainIndex);
 end
```
ge0ffrey: This rule is a good example of the impact for the user.

optaplanner-examples/src/test/java/org/optaplanner/examples/meetingscheduling/solver/MeetingSchedulingScoreConstraintTest.java
```diff
@@ -37,14 +38,14 @@
...
 
     private HardMediumSoftScoreVerifier<MeetingSchedule> scoreVerifier = new HardMediumSoftScoreVerifier<>(
-            SolverFactory.createFromXmlResource("org/optaplanner/examples/meetingscheduling/solver/meetingSchedulingSolverConfig.xml"));
+            SolverFactory.createFromXmlResource(MeetingSchedulingApp.SOLVER_CONFIG));
 
     private MeetingSchedule getMeetingSchedule(int numberOfEntities) {
         // After getting the solution, need to set AttendanceList for it. And for every meeting Required & Preferred attendance lists
         MeetingSchedule solution = new MeetingSchedule();
-        MeetingParametrization parametrization = new MeetingParametrization();
+        MeetingWeightPack parametrization = new MeetingWeightPack();
```
MusaTalluzi: Shouldn't the variable name "parametrization" also be renamed? (e.g. to weightPack)

yurloc:
@ge0ffrey could you fix the Findbugs errors so that we could get more info from the PR build?

ge0ffrey:
jenkins retest this
@rsynek ready to merge (docs will be in seperate PR for Misha's convenience)

ge0ffrey:
07:36:47 [INFO] Total bugs: 2
07:36:47 [INFO] Nullcheck of packAnnotation at line 103 of value previously dereferenced in org.optaplanner.core.impl.domain.constraintweight.descriptor.ConstraintConfigurationDescriptor.processPackAnnotation(DescriptorPolicy) [org.optaplanner.core.impl.domain.constraintweight.descriptor.ConstraintConfigurationDescriptor, org.optaplanner.core.impl.domain.constraintweight.descriptor.ConstraintConfigurationDescriptor] At ConstraintConfigurationDescriptor.java:[line 103]Redundant null check at ConstraintConfigurationDescriptor.java:[line 107] RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE
07:36:47 [INFO] Null pointer dereference of rule in org.optaplanner.core.impl.score.director.drools.DroolsScoreDirectorFactory.createRuleToConstraintWeightExtractorMap(KieBase) [org.optaplanner.core.impl.score.director.drools.DroolsScoreDirectorFactory] Dereferenced at DroolsScoreDirectorFactory.java:[line 145] NP_ALWAYS_NULL

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/constraintweight/ConstraintConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+package org.optaplanner.core.api.domain.constraintweight;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.*;
+
+/**
+ * Specifies that the class holds a number {@link ConstraintWeight} annotated members.
```
yurloc:
```suggestion
 * Specifies that the class holds a number of {@link ConstraintWeight} annotated members.
```

yurloc:
@ge0ffrey One more Findbugs error to go
```
14:52:40 [INFO] Possible null pointer dereference of AbstractXlsxSolutionFileIO$AbstractXlsxWriter.score in
org.optaplanner.examples.common.persistence.AbstractXlsxSolutionFileIO$AbstractXlsxWriter.writeScoreView()
[org.optaplanner.examples.common.persistence.AbstractXlsxSolutionFileIO$AbstractXlsxWriter,
org.optaplanner.examples.common.persistence.AbstractXlsxSolutionFileIO$AbstractXlsxWriter] Dereferenced at
AbstractXlsxSolutionFileIO.java:[line 462]Known null at AbstractXlsxSolutionFileIO.java:[line 456] NP_NULL_ON_SOME_PATH
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintConfigurationDescriptor.java
```diff
@@ -0,0 +1,159 @@
...
+                    .forEach(member -> potentiallyOverwritingMethodList.add((Method) member));
+        }
+        if (constraintWeightDescriptorMap.isEmpty()) {
+            throw new IllegalStateException("The constraintConfigurationClass (" + constraintConfigurationClass
+                    + ") must have at least 1 member with a "
+                    + ConstraintWeight.class.getSimpleName() + " annotation.");
+        }
+    }
+
+    private void processPackAnnotation(DescriptorPolicy descriptorPolicy) {
```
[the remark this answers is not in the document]
ge0ffrey: Yes, but it's a design consistency to pass it along during annotation processing (meta model parsing).

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/ScoreDefinition.java
```diff
@@ -67,6 +67,22 @@
...
+    default boolean isPositive(S score) {
+        return score.compareTo(getZeroScore()) >= 0;
+    }
+
+    /**
+     * @param score never null
+     * @return true if the score is lower or equal to {@link #getZeroScore()}
+     */
+    default boolean isNegative(S score) {
+        return score.compareTo(getZeroScore()) <= 0;
```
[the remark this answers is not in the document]
ge0ffrey: In arithmetic, zero is both positive and negative, but neither strictly positive or strictly negative. OptaPlanner follows those conventions.

Score weights are positive, which includes zero to turn that constraint off.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/AbstractXlsxSolutionFileIO.java
```diff
@@ -411,6 +434,57 @@ protected void writeLongConstraintLine(String name, Supplier<Long> supplier, Str
...
+            nextSheet("Score view", 1, 3, true);
+            nextRow();
+            nextHeaderCell("Score");
+            nextCell().setCellValue(score == null ? "Not yet solved" : score.toShortString());
+            nextRow();
+            nextRow();
+            nextHeaderCell("Constraint match");
+            nextHeaderCell("Match score");
+            nextHeaderCell("Total score");
+            if (!score.isSolutionInitialized()) {
```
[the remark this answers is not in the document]
ge0ffrey: Agreed, good call.

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/constraintweight/ConstraintConfiguration.java
ge0ffrey: +1, these suggestions are intresting - I would have applied them directly but I have local changes that I fear might conflict later. Next time, I 'll make sure to look a the Pr review first to merge any of these.

ge0ffrey:
Thanks Jiri, suggestions fixed

optaplanner-core/src/test/java/org/optaplanner/core/api/score/holder/AbstractScoreHolderTest.java
```diff
@@ -28,8 +28,7 @@
 import org.kie.api.runtime.rule.RuleContext;
 import org.optaplanner.core.api.score.constraint.ConstraintMatchTotal;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
ge0ffrey: I've got "import star" exceptions for Assert.*, Mockito.* and PlannerAssert.*. They're all over optaplanner's tests.

My diversion from KIE Java conventions (which diverts from the Sun/Oracle/JBoss conventions - which my diversion is trying to fix sometimes) is creating tension indeed. Lead by example and all that. I can't prioritize this before end of november, but I 've made a jira for it and I 'll try to add it to in a sprint or 2-3: https://issues.jboss.org/browse/PLANNER-1337

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/ScoreDefinition.java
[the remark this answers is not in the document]
yurloc: In [arithmetic](http://mathworld.wolfram.com/Arithmetic.html), positive integers are {1, 2, 3, ...}. Zero is [neither negative nor positive](http://mathworld.wolfram.com/Zero.html). Diverging from this convention may be very confusing. I mean, I can imagine that in score context it might be more convenient to bend the definition slightly but I fear that for most people the terms are so deeply rooted that this would only raise confusion.

Let's take a look at the only usage of the `isPositive()` method:
```java
if (!getScoreDefinition().isPositive(constraintWeight)) {
    throw new IllegalArgumentException("The constraintWeight () for constraintPackage () \
            and constraintName () of constraintConfigurationClass () must have \
            a positive or zero constraintWeight ().");
```
Even the exception message suggests that "positive"  means strictly positive. Additionally, if we correct the logic of the the two methods the piece of code becomes even clearer:
```java
if (getScoreDefinition().isNegative(constraintWeight)) {
    throw new IllegalArgumentException("The constraintWeight () ... must have \
            a positive or zero constraintWeight ().");
```
I strongly recommend to change the method name (or the condition) to be consistent with the mathematical definition (e.g. `isNonNegative()` and `isNonPositive()` and also to be consistent with usage and understanding of the terms "positive" and "negative" in the rest of the code and its documentation.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/DroolsScoreDirector.java
```diff
@@ -75,15 +79,64 @@ private void resetKieSession() {
...
+                        + "Maybe validate your " + constraintConfigurationClass.getSimpleName() + " data input.");
+            }
+            if (!getScoreDefinition().isPositive(constraintWeight)) {
+                Class<?> constraintConfigurationClass = getSolutionDescriptor().getConstraintConfigurationDescriptor()
+                        .getConstraintConfigurationClass();
+                throw new IllegalArgumentException("The constraintWeight (" + constraintWeight
+                        + ") for constraintPackage (" + rule.getPackageName()
+                        + ") and constraintName (" + rule.getName()
+                        + ") of constraintConfigurationClass (" + constraintConfigurationClass
+                        + ") must have a positive or zero constraintWeight (" + constraintWeight + ").\n"
```
yurloc: Duplicate occurrence of the `constraintWeight` parameter. I think it would be sufficient and more natural to say "The constraintWeight (-1) ... must be positive or zero."

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintConfigurationDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+            throw new IllegalStateException("The constraintConfigurationClass (" + constraintConfigurationClass
+                    + ") has been specified as a " + ConstraintConfigurationProvider.class.getSimpleName()
+                    + " in the solution class (" + solutionDescriptor.getSolutionClass() + ")," +
+                    " but does not have a " + ConstraintConfiguration.class.getSimpleName() + " annotation.");
+        }
+        // If a @ConstraintConfiguration extends a @ConstraintConfiguration, their constraintPackage might differ.
+        // So the ConstraintWeightDescriptors parse this too.
+        constraintPackage = packAnnotation.constraintPackage();
+        if (constraintPackage.equals("")) {
+            constraintPackage = constraintConfigurationClass.getPackage().getName();
```
yurloc: The `constraintPacakage` field is only assigned to and never read. This might be a bug. If it's not a bug, the field is unused and should be removed.

optaplanner-core/src/build/revapi-config.json
```diff
@@ -207,8 +207,547 @@
...
+          "code": "java.method.returnTypeChanged",
+          "old": "method org.optaplanner.core.api.score.Score org.optaplanner.core.api.score.buildin.bendable.BendableScoreHolder::extractScore(int)",
+          "new": "method org.optaplanner.core.api.score.buildin.bendable.BendableScore org.optaplanner.core.api.score.buildin.bendable.BendableScoreHolder::extractScore(int)",
+          "oldType": "org.optaplanner.core.api.score.Score",
+          "newType": "org.optaplanner.core.api.score.buildin.bendable.BendableScore",
+          "package": "org.optaplanner.core.api.score.buildin.bendable",
+          "classSimpleName": "BendableScoreHolder",
+          "methodName": "extractScore",
+          "elementKind": "method",
+          "justification": "ScoreHolder now has at the Score type as a generic type parameter."
```
yurloc:
```suggestion
          "justification": "ScoreHolder now has the Score type as a generic type parameter."
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/DroolsScoreDirector.java
ge0ffrey: I prefer to have the -1 very close to the "must be positive or zero", so the reason is more obvious. But you're right, the duplication here is unneeded, changing it to:

```

                throw new IllegalArgumentException("The @" + ConstraintWeight.class.getSimpleName()
                        + " annotation for constraintPackage (" + rule.getPackageName()
                        + ") and constraintName (" + rule.getName()
                        + ") of constraintConfigurationClass (" + constraintConfigurationClass
                        + ") must have a positive or zero constraintWeight (" + constraintWeight + ").\n"
                        + "Maybe validate your " + constraintConfigurationClass.getSimpleName() + " data input.");
```

optaplanner-core/src/test/java/org/optaplanner/core/api/score/buildin/simplelong/SimpleLongScoreHolderTest.java
```diff
@@ -49,11 +50,47 @@ public void addConstraintMatch(boolean constraintMatchEnabled) {
...
+        Rule constraint2 = mockRule("constraint2");
+        scoreHolder.putConstraintWeight(constraint2, SimpleLongScore.of(100L));
+
+        scoreHolder.penalize(mockRuleContext(constraint1));
+        assertEquals(SimpleLongScore.of(-10L), scoreHolder.extractScore(0));
+
+        scoreHolder.penalize(mockRuleContext(constraint2), 2L);
+        assertEquals(SimpleLongScore.of(-210L), scoreHolder.extractScore(0));
+
+        scoreHolder = new SimpleLongScoreHolder(constraintMatchEnabled);
```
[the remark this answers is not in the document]
ge0ffrey: Doesn't it do that? Line 77 to 82:

```
        scoreHolder.penalize(mockRuleContext(constraint1));
        assertEquals(SimpleScore.of(-10), scoreHolder.extractScore(0));

        scoreHolder.penalize(mockRuleContext(constraint2), 2);
        assertEquals(SimpleScore.of(-210), scoreHolder.extractScore(0));
```

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/domain/MeetingConstraintConfiguration.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+package org.optaplanner.examples.meetingscheduling.domain;
+
+import org.optaplanner.core.api.domain.constraintweight.ConstraintConfiguration;
+import org.optaplanner.core.api.domain.constraintweight.ConstraintWeight;
+import org.optaplanner.core.api.score.buildin.hardmediumsoft.HardMediumSoftScore;
+import org.optaplanner.examples.common.domain.AbstractPersistable;
+
+// @ConstraintWeights
```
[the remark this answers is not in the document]
ge0ffrey: Fixed

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintWeightDescriptor.java
```diff
@@ -0,0 +1,90 @@
...
+    // ************************************************************************
+
+    public ConstraintWeightDescriptor(ConstraintConfigurationDescriptor<Solution_> constraintConfigurationDescriptor,
+            MemberAccessor memberAccessor) {
+        this.constraintConfigurationDescriptor = constraintConfigurationDescriptor;
+        ConstraintWeight constraintWeightAnnotation = memberAccessor.getAnnotation(ConstraintWeight.class);
+        String constraintPackage = constraintPackage = constraintWeightAnnotation.constraintPackage();
+        if (constraintPackage.isEmpty()) {
+            // If a @ConstraintConfiguration extends a @ConstraintConfiguration, their constraintPackage might differ.
+            ConstraintConfiguration packAnnotation = memberAccessor.getDeclaringClass().getAnnotation(ConstraintConfiguration.class);
```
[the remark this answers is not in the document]
ge0ffrey: Fixed

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintWeightDescriptor.java
```diff
@@ -0,0 +1,90 @@
...
+
+    // ************************************************************************
+    // Constructors and simple getters/setters
+    // ************************************************************************
+
+    public ConstraintWeightDescriptor(ConstraintConfigurationDescriptor<Solution_> constraintConfigurationDescriptor,
+            MemberAccessor memberAccessor) {
+        this.constraintConfigurationDescriptor = constraintConfigurationDescriptor;
+        ConstraintWeight constraintWeightAnnotation = memberAccessor.getAnnotation(ConstraintWeight.class);
+        String constraintPackage = constraintPackage = constraintWeightAnnotation.constraintPackage();
```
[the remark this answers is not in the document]
ge0ffrey: Fixed

optaplanner-examples/src/main/resources/org/optaplanner/examples/rocktour/solver/rockTourScoreRules.drl
```diff
@@ -22,72 +22,66 @@ import org.optaplanner.core.api.score.buildin.hardmediumsoftlong.HardMediumSoftL
...
+// Soft constraints usually
 // ############################################################################
 
-rule "Avoid long driving times"
+rule "Shorten driving time per millisecond squared"
     when
         RockShow(previousStandstill != null, $drivingTime : drivingTimeFromPreviousStandstill)
     then
-        scoreHolder.addSoftConstraintMatch(kcontext, - $drivingTime* $drivingTime);
+        scoreHolder.penalize(kcontext, $drivingTime * $drivingTime);
```
yurloc: @ge0ffrey This is a bit outside the regular PR review but why don't we generalize the concept of constraint weight to a constraint weight *function*, so for example:
```java
// simpler, allows a second parameter (we can have BiFunction)
scoreHolder.penalize(kcontext, $drivingTime, [optional 2nd param]);

@ConstraintWeight(SHORTEN_DRIVING_TIME_PER_MILLISECOND_SQUARED)
private Function<Long, HardMediumSoftLongScore> shortenshortenDrivingTimePerMillisecondSquared
        // any kind of "magic"
        = drivingTime -> HardMediumSoftLongScore.ofSoft(drivingTime * drivingTime);
```

Just an idea...

ge0ffrey:
@rsynek Thanks for the review, comments fixed except the one that confuses me (SimpleScoreHolderTest).

As for the suggested extra corner cases tests: good idea, but out of scope. We invest a lot in fail fasting with decent error messages, and go much, much further than other libraries. Ideally those fail fasts need to be unit tested, too - and we unit test some of them - but it's not time budget efficient to test all of them. Most of the time, it's the error message that isn't clear enough - and we can't unit test that, just watch the stackoverflow questions and customer tickets to see what confuses our users.
As for the 1 level BendableScoreHolderTest, that's probably a good idea, yes.

optaplanner-examples/src/main/resources/org/optaplanner/examples/rocktour/solver/rockTourScoreRules.drl
rsynek: I like this idea, but introducing it makes the boundary between constraint rules and weights blurry. You can do the same in rules and even combine it (maybe unintentionally).

optaplanner-core/src/test/java/org/optaplanner/core/api/score/buildin/simplelong/SimpleLongScoreHolderTest.java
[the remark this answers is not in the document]
ge0ffrey: I did that originally, but it made the test code less obvious. It would be an ok thing to do, but I don't think it's necessary - I 'll leave it as is.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintConfigurationDescriptor.java
ge0ffrey: That's a bug indeed. the @ConstraintWeight should inherit it.

optaplanner-examples/src/main/resources/org/optaplanner/examples/rocktour/solver/rockTourScoreRules.drl
yurloc: I wouldn't be worried about duplicate calculations - users can and should use ScoreVerifier to unit test their rules.

The main problem with this idea, I think, is that it might become difficult to change the weight from a UI.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintConfigurationDescriptor.java
ge0ffrey: Well, it turns out it's not a bug - the ConstraintWeightDescriptor goes through the annotations again, due to extends complexity, but indeed we don't need this field for now. Removing it

optaplanner-examples/src/main/resources/org/optaplanner/examples/rocktour/solver/rockTourScoreRules.drl
ge0ffrey: Interesting idea.
I definitely want to support "simple" @ConstraintWeights that are just a score too. Even if it's only because it's a PITA to store lamdba's in a database. Notice that shortenshortenDrivingTimePerMillisecondSquared is probably stored in a column in a table in the database.

I also did a performance trick to "unravel" the score weight multiplication if there is only one non-zero score level. That trick wouldn't be possible here, but that unravel doesn't happen if the score weight is for example 0hard/1medium/1soft (2 non-zero's) either (but this is a rare weight).

So, this is an interesting RFE, good for a separate jira.

Ps: note that ConstraintStreams will also support a lamdba parameter in the penalize() and reward() methods, but they 'll be used to extract the weightMultiplier from the matched facts. So that's unrelated.

yurloc: Yes, weight persistence is an important requirement to keep in mind. Didn't realize that.

rsynek:
@ge0ffrey regarding the tests of fail fasting, which one of them is the most important? Which one of the fail fast checks would cause users the biggest trouble if it didn't work? For me it would be having multiple `ConstraintConfigurationProvider` methods/fields without being aware of it. If you agree, let's focus on testing this one.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/ScoreDefinition.java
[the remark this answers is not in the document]
ge0ffrey: "Is 0 A positive number? It really depends on context. In common use in English language, zero is unsigned, that is, it is neither positive nor negative. In typical French mathematical usage, zero is both positive and negative."

This is a very great good catch, thanks for pointing it out. I 'll rename the method to avoid any ambiguity
 isPositiveOrZero()
 isStrictlyPositive()
 the naming isPostive() is ambiguous and therefore banned

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintConfigurationDescriptor.java
rsynek: @ge0ffrey probably another candidate for additional test coverage - default and custom package in ConstraintConfiguration with constraints written in Drools. Another scenario - overriding a constraintPackage on the ConstraintWeight.

ge0ffrey:
@rsynek Please make a jira for that, to spec out which exactly are missing (there are a few tests in this PR related to failing fast).

PR #450, opened 2018-10-16
fixing links that gave a privacy error

mramendi:
Links to mat.gsia.cmu.edu give privacy errors, the fix is to change to mat.tepper.cmu.edu, it is the same website


ge0ffrey:
Nice work :)

PR #452, opened 2018-10-24
PLANNER-1338 ConstraintMatchTotal should also know the @ConstraintWeight

ge0ffrey:
- Add ConstraintMatchTotal.getConstraintWeight() and show it in constraint match dialogs and XSLX score view output
- Prepoluate all ConstraintMatchTotals if @ConstraintWeight are used
- Fix constraint id javadoc bug
- Fix published coloring in conference scheduling example
- Some slides for Devoxx


optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -110,7 +124,7 @@ private ConstraintMatchTotal findConstraintMatchTotal(RuleContext kcontext) {
         String constraintName = rule.getName();
         String constraintId = constraintPackage + "/" + constraintName;
         return constraintMatchTotalMap.computeIfAbsent(constraintId,
-                k -> new ConstraintMatchTotal(constraintPackage, constraintName, zeroScore));
+                k -> new ConstraintMatchTotal(constraintPackage, constraintName, null, zeroScore));
```
rsynek: Let's use the constructor taking only 3 parameters to minimize passing null values.

optaplanner-docs/src/main/asciidoc/UseCasesAndExamples/ConferenceScheduling/ConferenceScheduling-section.adoc
```diff
@@ -6,8 +6,13 @@
 [[conferenceSchedulingProblemDescription]]
 == Problem Description
 
-Assign each conference talk to a timeslot and a room.
-Timeslots can overlap. Read/write to/from an `*.xlsx` file that can be edited with LibreOffice or Excel too.
+Assign each conference talk to a timeslot and a room, after the talks have been accepted.
+
+image::UseCasesAndExamples/ConferenceScheduling/conferenceSchedulingMilestonesTimeline.png[align="center"]
+
+Timeslots can overlap. It reads/writse to/from an `*.xlsx` file that can be edited with LibreOffice or Excel.
```
rsynek: typo "writse"

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
ge0ffrey: I want to explicitly use the null parameter here so it sticks out like a sore thumb, because I hope it's possible to get rid of it some day.

PR #458, opened 2018-11-13
Remove useless property override

jhrcek:
optaplanner-webexamples overrides 'version.org.wildfly' defined in kie-parent from '14.0.1.Final' to the same version '14.0.1.Final'
This override is unnecessary and will require changes in more places when we update wildfly in the future (or will be forgotten and will cause inconsistencies).

@ge0ffrey please check and consider merging.


jhrcek:
ping @ge0ffrey Can you please check this?

ge0ffrey:
Thanks Jan!

PR #461, opened 2018-11-20
PLANNER-1312 Two tests take too long + new blueprint type

ge0ffrey:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/blueprint/SolverBenchmarkBluePrintType.java
```diff
@@ -28,6 +28,10 @@
 import org.optaplanner.core.config.solver.SolverConfig;
 
 public enum SolverBenchmarkBluePrintType {
+    /*
+     * Run the default {@link ConstructionHeuristicType} with and without the default {@link LocalSearchType}.
+     */
+    CONSTRUCTION_HEURISTIC_WITH_AND_WITHOUT_LOCAL_SEARCH,
```
rsynek: revapi warning failed the build for me:
"The enum constant was defined on position 0 but is now on 1. The user code can break if it relies on the return value of the "ordinal()" method."
I would suggest moving this entry to the end of the enumeration list

ge0ffrey: It's a valid argument, but I find a logical order in the enum constants (in this case from few to many) more important, as it reduces the learning curve.

I cannot think of any scenario where they would depend on the ordinal() method - except maybe if they marshall it like that and the client and server use a different optaplanner version - which is very unlikely. Or if they store the benchmark configuration in the database (nobody does that afaik) and use the ordinal instead of the enum name as the value in database column - in which case they deserve what they get... So changing ordinal() is the lesser evil.

I'll fix revapi of course, I forgot to run mvn too, not just the tests - but that's why we have jenkins :)

ge0ffrey:
@rsynek Yes, they aren't perfect yet - still above the threshold I suggested. Let's merge this step forward though.

Note that the threshold is meant on a modern developer machine, anything between brand new or no older than 3 years. If the jenkins/CI stuff uses much slower machines, than it's ok to break those thresholds - we can't take that into account.

The meeting scheduling test will become better once constrainstreams are here. Otherwise we just need to create a smaller dataset for it.

The nqueens benchmark is a different beast. It always takes that amount of time, regardless if the machine is fast or slow. It's a poorly designed test, really. It will need to be refactored further...

rsynek:
@ge0ffrey I agree, it's definitely an improvement, so let's merge it

ge0ffrey:
Caused by: java.io.IOException: Server returned HTTP response code: 403 for URL: https://api.github.com/user

Jenkins retest this.

PR #463, opened 2018-11-27
Add zeroEntity test method to DefaultSolverTest

Fonifan:
According to requirements from this [Pull request](https://github.com/kiegroup/optaplanner/pull/429).

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
@@ -23,6 +23,7 @@ import org.junit.Test;
 import org.optaplanner.core.api.solver.Solver;
 import org.optaplanner.core.api.solver.SolverFactory;
 import org.optaplanner.core.config.constructionheuristic.ConstructionHeuristicPhaseConfig;
+import org.optaplanner.core.config.localsearch.LocalSearchPhaseConfig;
 import org.optaplanner.core.config.phase.custom.CustomPhaseConfig;
 import org.optaplanner.core.config.score.definition.ScoreDefinitionType;
 import org.optaplanner.core.config.solver.termination.TerminationConfig;
@@ -30,6 +31,9 @@ import org.optaplanner.core.impl.phase.custom.NoChangeCustomPhaseCommand;
 import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
 import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
 import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+import org.optaplanner.core.impl.testdata.domain.chained.TestdataChainedAnchor;
+import org.optaplanner.core.impl.testdata.domain.chained.TestdataChainedEntity;
+import org.optaplanner.core.impl.testdata.domain.chained.TestdataChainedSolution;
 import org.optaplanner.core.impl.testdata.domain.extended.legacysolution.TestdataLegacySolution;
 import org.optaplanner.core.impl.testdata.util.PlannerTestUtils;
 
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
```diff
@@ -111,4 +115,25 @@ public void solveStopsWhenPartiallyInitialized() {
...
+        Solver<TestdataChainedSolution> solver = solverFactory.buildSolver();
+
+        TestdataChainedSolution solution = new TestdataChainedSolution("1");
+        solution.setChainedEntityList(Collections.singletonList(new TestdataChainedEntity("3")));
+        solution.setChainedAnchorList(Collections.singletonList(new TestdataChainedAnchor("4")));
+
+        try {
+            solver.solve(solution);
+        } catch (Exception exception) {
+            assertTrue(exception instanceof IllegalStateException);
```
rsynek: Please make the assertion more accurate as it was in the original test.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
```diff
@@ -111,4 +115,25 @@ public void solveStopsWhenPartiallyInitialized() {
...
+        SolverFactory<TestdataChainedSolution> solverFactory = PlannerTestUtils.buildSolverFactory(
+                TestdataChainedSolution.class, TestdataChainedEntity.class);
+
+        LocalSearchPhaseConfig phaseConfig = new LocalSearchPhaseConfig();
+        phaseConfig.setTerminationConfig(new TerminationConfig().withStepCountLimit(1));
+        solverFactory.getSolverConfig().setPhaseConfigList(Collections.singletonList(phaseConfig));
+
+        Solver<TestdataChainedSolution> solver = solverFactory.buildSolver();
+
+        TestdataChainedSolution solution = new TestdataChainedSolution("1");
```
rsynek: The domain configuration is a bit different in the original test

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
```diff
@@ -116,7 +116,7 @@ public void solveStopsWhenPartiallyInitialized() {
     }
 
     @Test(timeout = 600_000)
-    public void zeroEntity() {
+    public void solveThrowsWhenZeroEntity() {
```
rsynek: How about `solveThrowsExceptionWhenZeroEntity`? I find it easier to read as a sentence.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
```diff
@@ -127,13 +127,15 @@ public void zeroEntity() {
...
         TestdataChainedSolution solution = new TestdataChainedSolution("1");
-        solution.setChainedEntityList(Collections.singletonList(new TestdataChainedEntity("3")));
+        solution.setChainedEntityList(Collections.EMPTY_LIST);
         solution.setChainedAnchorList(Collections.singletonList(new TestdataChainedAnchor("4")));
 
         try {
             solver.solve(solution);
         } catch (Exception exception) {
-            assertTrue(exception instanceof IllegalStateException);
+            assertEquals(true,exception instanceof IllegalStateException);
```
rsynek: If you assert logical expressions, use just `assertTrue` or `assertFalse`. Let's keep the `assertEquals` for something more complex.

Fonifan: I was trying to stick to the way assertions were done in the test before: e.g.
`assertEquals(false, solution.getScore().isSolutionInitialized());` in `solve()`, `solveStopsWhenUninitialized()` or `solveStopsWhenPartiallyInitialized()` tests.

rsynek: @Fonifan please take it just as recommendation, from my point of view, it doesn't block merging.

As we talk about consistency, could you please add a space after a comma in `true,exception` ?

Fonifan: Thanks for noticing, fixed. :+1:

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
```diff
@@ -111,4 +115,27 @@ public void solveStopsWhenPartiallyInitialized() {
...
+        solverFactory.getSolverConfig().setPhaseConfigList(Collections.singletonList(phaseConfig));
+
+        Solver<TestdataChainedSolution> solver = solverFactory.buildSolver();
+
+        TestdataChainedSolution solution = new TestdataChainedSolution("1");
+        solution.setChainedEntityList(Collections.EMPTY_LIST);
+        solution.setChainedAnchorList(Collections.singletonList(new TestdataChainedAnchor("4")));
+
+        try {
+            solver.solve(solution);
```
ge0ffrey: If it doesn't throw an exception, this test should fail.
One way to do that is to add a 
            fail("There was no RuntimeException thrown.");
after this line. There are other tests that do that, see PartitionQueueTest.

Besides that, this is a great PR!

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
ge0ffrey: Is that a problem?

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
Fonifan: Thanks! Fixed that.

PR #466, opened 2018-11-30
Configure sonarcloud and travis-ci

rsynek:
- configures jacoco code coverage and reporting to sonarcloud.io
- enables travis-ci to update the sonarcloud.io with every PR

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,12 @@
 
   <properties>
     <findbugs.failOnViolation>true</findbugs.failOnViolation>
+    <jacoco.agent.line/>
+    <surefire.argLine>
+      -Dfile.encoding=${project.build.sourceEncoding}
+      ${jacoco.agent.line}
+    </surefire.argLine>
+    <version.jacoco.plugin>0.8.2</version.jacoco.plugin>
   </properties>
 
   <repositories>
@@ -90,6 +96,11 @@
         <!-- Old version until this issue is resolved: https://github.com/ImmobilienScout24/illegal-transitive-dependency-check/issues/30 -->
         <version>1.0.4</version>
       </dependency>
+      <dependency>
+        <groupId>org.jacoco</groupId>
+        <artifactId>org.jacoco.agent</artifactId>
+        <version>${version.jacoco.plugin}</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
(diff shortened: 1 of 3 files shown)


rsynek:
@ge0ffrey please review

pom.xml
```diff
@@ -90,6 +96,11 @@
         <!-- Old version until this issue is resolved: https://github.com/ImmobilienScout24/illegal-transitive-dependency-check/issues/30 -->
         <version>1.0.4</version>
       </dependency>
+      <dependency>
+        <groupId>org.jacoco</groupId>
+        <artifactId>org.jacoco.agent</artifactId>
+        <version>${version.jacoco.plugin}</version>
```
ge0ffrey: I'd expect that jacoco is a compile/build/tool dependency, not a real normal dependency.
Is there any way we can restrict jacoco to one (or more) plugin classpaths?

pom.xml
```diff
@@ -106,6 +117,84 @@
...
+              <artifactId>gwt-maven-plugin</artifactId>
+              <configuration>
+                <skipCompilation>true</skipCompilation>
+              </configuration>
+            </plugin>
+            <plugin>
+              <artifactId>maven-surefire-plugin</artifactId>
+              <configuration>
+                <argLine>${surefire.argLine}</argLine>
+              </configuration>
```
ge0ffrey: What if we add a dependencies element here with the jacoco dependency?

pom.xml
```diff
@@ -28,6 +28,12 @@
 
   <properties>
     <findbugs.failOnViolation>true</findbugs.failOnViolation>
+    <jacoco.agent.line/>
+    <surefire.argLine>
+      -Dfile.encoding=${project.build.sourceEncoding}
+      ${jacoco.agent.line}
+    </surefire.argLine>
+    <version.jacoco.plugin>0.8.2</version.jacoco.plugin>
```
ge0ffrey: Can we only overrides these properties in the jacoco profile?

rsynek: +1, even as just a temporary solution - it will end up in kie-parent

pom.xml
rsynek: It makes sense, will look into it.

rsynek:
@ge0ffrey changed the jacoco to the surefire plugin's dependency as you suggested. Thanks for this idea. Ready to merge, once the jenkins build passes too.

PR #467, opened 2018-12-04
fix compilation problem on jdk 11

mariofusco:


optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/AbstractXlsxSolutionFileIO.java
```diff
@@ -507,17 +507,22 @@ protected void writeScoreView(Function<List<Object>, String> justificationListFo
             });
             nextRow();
             nextRow();
-            constraintMatchTotals.stream().sorted(Comparator.<ConstraintMatchTotal, Score>comparing(ConstraintMatchTotal::getScore)
-                    .thenComparing(ConstraintMatchTotal::getConstraintPackage)
-                    .thenComparing(ConstraintMatchTotal::getConstraintName)).forEach(constraintMatchTotal -> {
+
+            Comparator<ConstraintMatchTotal> c1 = Comparator.<ConstraintMatchTotal, Score>comparing(ConstraintMatchTotal::getScore);
```
rsynek: Just out of curiosity, what was the problem with using the extractor function?

PR #473, opened 2018-12-19
PLANNER-1394 Better error message than ArrayIndexOutOfBoundsException…

ge0ffrey:
… due to difference Bendable score levels

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
@@ -25,6 +25,7 @@ import org.kie.api.definition.rule.Rule;
 import org.kie.api.runtime.rule.RuleContext;
 import org.optaplanner.core.api.domain.constraintweight.ConstraintConfiguration;
 import org.optaplanner.core.api.domain.constraintweight.ConstraintWeight;
+import org.optaplanner.core.api.domain.solution.PlanningScore;
 import org.optaplanner.core.api.score.holder.AbstractScoreHolder;
 
 /**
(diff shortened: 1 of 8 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
```diff
@@ -218,6 +219,11 @@ public void reward(RuleContext kcontext, int[] hardWeightsMultiplier, int[] soft
      * @param weight higher is better, negative for a penalty, positive for a reward
      */
     public void addHardConstraintMatch(RuleContext kcontext, int hardLevel, int weight) {
+        if (hardLevel >= hardScores.length) {
+            throw new IllegalArgumentException("The hardLevel (" + hardLevel
```
rsynek: Soft suggestion: use a template for this repetitive error message and just pass the parameters that change.

rsynek: That might be an option; this does not block merging, though.

Please see also my other comment regarding unit testing.

ge0ffrey:
> Although we cannot test every possible fail fast situation, here it seems to be achievable. Let's please add unit coverage to avoid regressions in the future.

Agreed, unit tests added :)

optaplanner-core/src/test/java/org/optaplanner/core/api/score/buildin/bendablebigdecimal/BendableBigDecimalScoreHolderTest.java
```diff
@@ -130,4 +130,18 @@ public void rewardPenalize(boolean constraintMatchEnabled) {
         assertEquals(BendableBigDecimalScore.of(new BigDecimal[]{new BigDecimal("-210.0")}, new BigDecimal[]{new BigDecimal("-90.0"), new BigDecimal("310.0")}), scoreHolder.extractScore(0));
     }
 
+    @Test(expected = IllegalArgumentException.class)
```
rsynek: Thanks for adding the unit coverage. In my experience, this way of checking exceptions is sometimes not precise enough. Let's say the exception will inform about "hardWeights" instead of "softWeights". More fine-grained checking can be very easily achieved with the AssertJ library; or by a try-catch block.

ge0ffrey: or JUnit 5 :)

ge0ffrey: Once we are able to use JUnit 5, I 'll go thought the entire code base to replace all "@Test(excepted = ...)" instances

PR #477, opened 2019-01-09
PLANNER-1405 Fail fast when LookUp returns null

ge0ffrey:


rsynek:
@ge0ffrey some tests will probably require updating - they seem to fail after this change. See the travis log

kiereleaseuser2:
Build finished. 2397 tests run, 295 skipped, 6 failed.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/lookup/LookUpManager.java
```diff
@@ -74,4 +75,20 @@ public void clearWorkingObjects() {
...
 
+    /**
+     * As defined by {@link ScoreDirector#lookUpWorkingObjectOrReturnNull(Object)}.
+     * @param externalObject sometimes null
+     * @return null if externalObject is null or if there is no workingObject for externalObject
+     * @throws IllegalArgumentException if it cannot be looked up or if the externalObject's class is not supported
+     * @throws IllegalStateException if it cannot be looked up
+     * @param <E> the object type
+     */
+    public <E> E lookUpWorkingObjectOrReturnNull(E externalObject) {
```
rsynek: Given the minimal supported version is Java 8, we should start using Optional instead of returning null value from public methods. Optional makes it very difficult (unfortunately not impossible) to retrieve the value without thinking about corner cases first.

ge0ffrey:
Nah, Optional is for streams only. Notice how the JDK API's don't use it anywhere else.

Besides, if we do intend to use Optional, we'd need to do it consistently in ALL our api's. Do or do not. Don't it a bit here and there but not everywhere.

rsynek:
@ge0ffrey why do you think Optional is suitable only for streams? The reason why it's not used in JDK API might be backward compatibility. Which is the reason why we won't be ever able to use it consistently across OptaPlanner.

PR #478, opened 2019-01-21
BAQE-814 - Change revapi to check against 7.14.0.Final

MarianMacik:
depends on kiegroup/droolsjbpm-build-bootstrap#914


rsynek:
@ge0ffrey Looking at the changes regarding adding type parameter to Score, do we have it documented in the upgrade recipe? I cannot find it.

MarianMacik:
Travis CI failed because it cannot pick up changes from other repositories.

ge0ffrey:
@MarianMacik Sorry for the delay. Yes, good to merge.

ge0ffrey:
> @ge0ffrey Looking at the changes regarding adding type parameter to Score, do we have it documented in the upgrade recipe? I cannot find it.

IIRC Score already had a type parameter, but ScoreHolder didn't. I don't recall making any changes in the examples, (ScoreHolder is only used in the DRL) so I believe no upgrade recipe note is needed. @rsynek Did I miss anything? I do remember that was part of a salvage of a refactor that went haywire, so it's more prone to a mistake.

MarianMacik:
I can include these classes if you don't consider them as APIs. Although it's true that they are in the api package.

rsynek:
> I can exclude these classes if you don't consider them as APIs. Although it's true that they are in the api package.

We probably cannot completely remove this class from revapi checking - due to the usage in DRLs in case of drools score calculation. This particular change is a sort of exception, as DRLs don't care about generic types.

rsynek:
Anyway, nothing holds off merging this PR.

PR #480, opened 2019-01-29
Fixes related to Mockito upgrade

yurloc:
`org.mockito.Matchers` is deprecated. Matchers are now obtained by statically importing `Mockito` which extends `ArgumentMatchers`.

`any(Class)` no longer accepts `null` -> needed to set up a few additional mocks.


yurloc:
AppFormer repository is broken on master -> we can't have a green build at this moment.

PR #483, opened 2019-02-05
DROOLS-1781: replaces findbugs-maven-plugin

mbiarnes:


pom.xml
```diff
@@ -194,14 +194,6 @@
   <build>
     <pluginManagement>
       <plugins>
-        <plugin>
```
rsynek: This part that was removed provided more precise threshold for this project. Could you please do the same config via spotbugs? It appears the maven plugin supports the same configuration parameters [1] as findbugs maven plugin does.

[1] https://spotbugs.github.io/spotbugs-maven-plugin/check-mojo.html

ge0ffrey: Yes, keeping this config and just using findbugs with spotbugs should do the trick.

Or there's even a better way:
  https://spotbugs.github.io/spotbugs-maven-plugin/check-mojo.html#maxRank
`Just add the properties <spotbugs.maxRank>14</> and <spotbugs.includeTests>true</>.`

mareknovotny: I agree if there is not used default config we should keep it here.
Eventually there is a question why we need to use different config ;)

rsynek: e.g. the `maxRank` parameter is something user has to experiment with - to find a good balance between not seeing many false positives and not overlooking real failures. It may be different for each project.

mareknovotny: ok, that is ok. It always helps to comment a difference in the pom.xml to not forget or help others to understand why that property is there ;). But thanks for explanation here

mbiarnes: done

mbiarnes:
@mareknovotny @rsynek @ge0ffrey done so far

PR #484, opened 2019-02-05
Add ability to import timeslots from /slots endpoint

MusaTalluzi:
devoxx-cfp exposes timeslots via two endpoints:
- `/slots`: Timeslots as problem facts. It displays all available timeslots without additional data.
- `/schedules/`: The conference solution, i.e. it displays days, timeslots within each day and the talk scheduled at this timeslot.

If users choose to import data using `/slots` endpoint, they'll get an empty problem to solve from scratch, however if they choose `/schedules/`, they'll get the current schedule to either view it or continue solving.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingCfpDevoxxImporter.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingCfpDevoxxImporter.java
@@ -67,6 +67,7 @@ public class ConferenceSchedulingCfpDevoxxImporter {
     private static final String[] IGNORED_SPEAKER_NAMES = {"Devoxx Partner"};
 
     private String conferenceBaseUrl;
+    private String timeslotsEndpoint;
     private Map<String, TalkType> talkTypeIdToTalkTypeMap;
     private Map<String, Room> roomIdToRoomMap;
     private Map<String, Speaker> speakerNameToSpeakerMap;
@@ -77,8 +78,9 @@ public class ConferenceSchedulingCfpDevoxxImporter {
     private Map<String, Integer> timeslotTalkTypeToTotalMap = new HashMap<>();
     private Map<String, Integer> talkTalkTypeToTotalMap = new HashMap<>();
 
-    public ConferenceSchedulingCfpDevoxxImporter(String conferenceBaseUrl) {
+    public ConferenceSchedulingCfpDevoxxImporter(String conferenceBaseUrl, String timeslotsEndpoint) {
         this.conferenceBaseUrl = conferenceBaseUrl;
+        this.timeslotsEndpoint = timeslotsEndpoint;
     }
 
     public ConferenceSolution importSolution() {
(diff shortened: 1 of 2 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/persistence/ConferenceSchedulingCfpDevoxxImporter.java
```diff
@@ -259,7 +262,7 @@ private void importTalkList() {
         Set<String> themeTrackSet = new HashSet<>(Arrays.asList(talkObject.getJsonObject("track").getString("id")));
         if (!trackIdSet.containsAll(themeTrackSet)) {
             throw new IllegalStateException("The talk (" + title + ") with id (" + code
-                    + ") contains trackId + (" + trackIdSet + ") that doesn't exist in the trackIdSet.");
+                    + ") contains trackId " + themeTrackSet.toString() + " that doesn't exist in the trackIdSet.");
```
ge0ffrey: toString() is obsolete and use () to delimate start and end of value like in all other exceptions (even despite Set adding [] around it), so
"... trackId (" + themeTrackSet + ") ..."

optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/swingui/ConferenceCFPImportAction.java
```diff
@@ -54,15 +55,20 @@ public String getName() {
             String[] cfpArray = {"cfp-devoxx"};
             JComboBox<String> cfpConferenceBox = new JComboBox<>(cfpArray);
             JTextField cfpRestUrlTextField = new JTextField("https://dvbe18.confinabox.com/api/conferences/DVBE18");
+            String[] timeslotsEndpointArray = {"/slots", "/schedules/"};
+            JComboBox<String> timeslotsEndpointBox = new JComboBox<>(timeslotsEndpointArray);
             Object[] dialogue = {
                     "Choose conference:", cfpConferenceBox,
-                    "Enter CFP REST Url:", cfpRestUrlTextField
+                    "Enter CFP REST Url:", cfpRestUrlTextField,
+                    "Choose timeslots endpoint", timeslotsEndpointBox,
```
ge0ffrey: Why does the user need to get a choice? Choosing is losing.
Does /slots work for DevoxxBe? If so, let's standarize on that.
If not, maybe we can first try "/slots" and then fall through to "/tuesday" etc.

MusaTalluzi: In the cfp-devoxx API documentation `/slots` is never mentioned, although I found out that DevoxxBE implements it but I never knew about it before. So I'm not sure if it'll be implemented for other conferences as well.

Also `/schedules/` expose more data than /slots (e.g. if there's a talk scheduled at a timeslot, to be pinned for example).

MusaTalluzi: The rationale behind giving the user a choice, if they want to start from scratch they choose `/slots`, but if they wanna import the data with the current schedule (if published) they choose `/schedules`. What do you think>

ge0ffrey: The complexity of that choice is lost in that "what url do you want to use?". Keep it simple and stupid:
- **No choice during the CFP input.** We'll already lose some Voxxed organizers if we ask too many questions there. They give us 5 minutes of their time, starting from scratch, to show a solved schedule. Only once they believe it can do it, they 'll want to know about details (such as pinning and non-disruptive rescheduling).
- Timeslots list comes from `/slots`. That seems to be the canonical url available on all instances of the latest version. That is always available if the timeslots are configured.
- If `/schedules/` is available, we read that too to enrich the data:
-- Fill in the assigned room+timeslot of talks if that are already assigned to something in cfp-devoxx.
--- Do NOT set them as pinned. Very few talks are pinned and that user must do that in our xlsx, because cfp-devoxx has no "pinned" field.
--- I would not set them as published either. The user can click the "publish" button immediatly after importing them.
---- Click in the publshed button with half the talks assigned should only publish those talks. Please check that (if you haven't already).

Wdyt?

ge0ffrey: I don't know. Try reaching out to nicmarti (mail or github message with Stephan in CC - I don't know him personally) with a short, simple question such as to see if /slots is good to use, even though it's not in that github reference. By asking the question, we'll also signal to him that we're relying on it. Meanwhile, we can continue.

PR #487, opened 2019-02-21
Time bucket pattern

ge0ffrey:
Needed by Bilgin/Donato etc for customer conversations


ge0ffrey:
ok to review now

optaplanner-docs/src/main/asciidoc/DesignPatterns/DesignPatterns-chapter.adoc
```diff
@@ -118,21 +118,43 @@ Choose the right representation type for your use case:
...
+when the starting time has been decided already in an earlier planning stage.
+
+* If the starting time is not fixed, it is a planning variable (genuine or shadow).
+
+** If all planning entities have the same duration,
+use the <<timeslotPattern,Timeslot pattern>>.
 *** For example in course scheduling, all lectures take one hour. Therefore, each timeslot is one hour.
-** If the duration differs and time is rounded to a specifc time granularity (for example 5 minutes) use the <<timeGrainPattern,TimeGrain pattern>>.
+*** Even if the planning entities have different durations, but the same duration per type, it's often appropriate.
+**** For example in conference scheduling, breakout talks take one hour and lab talks take 2 hours.
```
rsynek: If you render the page, it shows you a rather complicated list reaching 4 levels. My first impression is that this chapter looks like a decision tree converted into a text.

@mramendi do you have any idea how to present such information in more natural way?

ge0ffrey: Agreed, I think we can handle this in a separate PR, if there is indeed a better way. Misha, would you mind looking at that?

PR #491, opened 2019-03-07
Add with methods on the important Config classes

ge0ffrey:
These makes Graal experiments far less painful, because Graal doesn't like XStream.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
@@ -240,6 +240,7 @@ public class AcceptorConfig extends AbstractConfig<AcceptorConfig> {
     // With methods
     // ************************************************************************
 
+    @Deprecated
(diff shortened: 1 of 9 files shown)


ge0ffrey:
jenkins rebuild this

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
```diff
@@ -240,6 +240,7 @@ public void setStepCountingHillClimbingType(StepCountingHillClimbingType stepCou
     // With methods
     // ************************************************************************
 
+    @Deprecated
```
rsynek: Nitpicking: would be nice to include `@deprecated` javadoc tag informing users about which operation replaces this one.

It should probably also appear in the upgrade recipe, right?

ge0ffrey: Yes, you're right in theory, but in practice no one is using this because Acceptor was the only Config class with with methods, it's a deep class and it wasn't documented (still isn't).
This field was deprecated a long time ago and when the with methods were added recently, the deprecated was forgotten to be added for this one.
So I am going to leave it as is, in this case.

PR #493, opened 2019-03-19
PLANNER-1451 End solver scope before firing solving ended event

yurloc:
https://issues.jboss.org/browse/PLANNER-1451

Getting score calculation speed from solver scope in `PhaseLifecycleListener#solvingEnded()` shouldn't throw NPE. The fix is simple, we just need to call `solverScope.endingNow()` just before `AbstractSolver#solvingEnded()`, which fires the lifecycle event.

Unblocks gathering Prothean metrics.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -220,8 +220,8 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
 
     @Override
     public void solvingEnded(DefaultSolverScope<Solution_> solverScope) {
-        super.solvingEnded(solverScope);
         solverScope.endingNow();
+        super.solvingEnded(solverScope);
     }
 
     public void outerSolvingEnded(DefaultSolverScope<Solution_> solverScope) {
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -220,8 +220,8 @@ public void solvingStarted(DefaultSolverScope<Solution_> solverScope) {
 
     @Override
     public void solvingEnded(DefaultSolverScope<Solution_> solverScope) {
-        super.solvingEnded(solverScope);
         solverScope.endingNow();
+        super.solvingEnded(solverScope);
```
ge0ffrey: This makes me nervous. The lifecycle event calls have a very predictable order: first superclasses, then subclasses.

The good news is we can have our cake and eat it too:
Both PartionedSolver and AstractSolver call solverScope.edningNow() immediatly after calling the super method, so it can be moved into the super method. I must have looked over that when splitting up the 2 solver implementations.
Once it's there, we can discuss moving it. 

With brings open the side-effect that if we move it up, any actions after the calling endingNow no longer affect the endingSystemTimeMillis... but the contract with the user is that they should.

ge0ffrey: Let's discuss real-time next week or after VRP meeting.

vblagoje: Either way, the decision is made please lmk. If we can have this PR in, then I can update https://github.com/kiegroup/droolsjbpm-integration/pull/1776 to include score calculation speed metric.

yurloc: I agree with @ge0ffrey that the change might impact the "contract" of the lifecycle methods. On the other hand it doesn't make sense to fire events that are not in a consistent state. I hope we'll be able to find a solution this Friday.

yurloc:
Superseded by #500.

PR #495, opened 2019-03-22
Spring Boot vs Drools warning

ge0ffrey:


mramendi:
Is the term "fire", related to rules, used regularly in this document? Otherwise no comment.

rsynek:
@mramendi the term "to fire rules" is definitely used in drools documentation, I have seen it several times.

PR #498, opened 2019-03-29
Recognize Temporal classes as immutable for multithreaded sovling look up strategy

ge0ffrey:


rsynek:
https://issues.jboss.org/browse/PLANNER-1467

mwkroening:
I suppose, recognizing immutable classes is beneficial for performance? Is there any documentation on this? Is there an API to mark custom classes as immutable? For example if one uses [ThreeTen-Extra](https://www.threeten.org/threeten-extra/) for additional time classes.

PR #499, opened 2019-03-29
Rename withTermination() to withTerminationConfig() + move endingNow(…

ge0ffrey:
…) up to parent + release notes

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
@@ -135,7 +135,6 @@ public class PartitionSolver<Solution_> extends AbstractSolver<Solution_> {
     @Override
     public void solvingEnded(DefaultSolverScope<Solution_> solverScope) {
         super.solvingEnded(solverScope);
-        solverScope.endingNow();
         solverScope.getScoreDirector().close();
         // TODO log?
     }
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
@@ -71,11 +71,7 @@ public class PartitionSolver<Solution_> extends AbstractSolver<Solution_> {
 
     @Override
     public long getTimeMillisSpent() {
-        Long endingSystemTimeMillis = solverScope.getEndingSystemTimeMillis();
-        if (endingSystemTimeMillis == null) {
-            endingSystemTimeMillis = System.currentTimeMillis();
-        }
-        return endingSystemTimeMillis - solverScope.getStartingSystemTimeMillis();
+        throw new UnsupportedOperationException();
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


ge0ffrey:
Sorry for lumping this together.

ge0ffrey:
Travis complains about revapi, but it's not correct (related to untouched code in this PR) and Jenkins doesn't complain about it. Weird

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
```diff
@@ -71,11 +71,7 @@ public String explainBestScore() {
 
     @Override
     public long getTimeMillisSpent() {
-        Long endingSystemTimeMillis = solverScope.getEndingSystemTimeMillis();
-        if (endingSystemTimeMillis == null) {
-            endingSystemTimeMillis = System.currentTimeMillis();
-        }
-        return endingSystemTimeMillis - solverScope.getStartingSystemTimeMillis();
+        throw new UnsupportedOperationException();
```
rsynek: Would be nice to tell a user why the operation is not supported to guide him towards a correct solution.

optaplanner-docs/src/main/asciidoc/Development/Development-chapter.adoc
```diff
@@ -65,4 +69,25 @@ Maybe the member (getFooRange) should return CountableValueRange.
...
 
-. Generics. The `Solution` class is often passed as a generic type parameter to subsystems. The `PlanningEntity` class(es) are rarely passed as a generic type parameter.
+=== Generics
+
+. The `Solution` class is often passed as a generic type parameter to subsystems.
+. The `PlanningEntity` class(es) are rarely passed as a generic type parameter because there could be multiple planning entities.
+
+=== Lifecycle
+
+One of the biggest challenges in multi-algorithm implementations (such as Planner)
```
rsynek: Should have been "OptaPlanner" ?

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
ge0ffrey: The user never uses PartitionedSolver directly, it's a internal trick of PartitionedSearch to spawn multiple Solver. The other methods of PartitionedSolver that are inherited from Solver, also throw this exception. The difference is that this method will be deprecated with the new event.

optaplanner-docs/src/main/asciidoc/Development/Development-chapter.adoc
ge0ffrey: All docs use "Planner" due to old product docs writers not wanting to see the word "OptaPlanner" when the product version was still called "Business Resource Planner". For consistency, I've continued to use that here, but ideally we should replace all "Planner" with "OptaPlanner" - or by a conditional that resolves to OptaPlanner in community and "Business Optimizer" in product.
In any case, different issue - so for now "keep it consistent" means doing it like this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
ge0ffrey: Normally I'd agree with this comment - in fact the PartitionedSearch code needs a review for such matters - and consistently change it - but separate issue, so I am going to merge that as is.

rsynek: @ge0ffrey would you please file an issue for the review so that we don't forget about it? Thanks

PR #501, opened 2019-04-02
Fix methematical terminology (addend)

yurloc:
https://en.wikipedia.org/wiki/Addition


ge0ffrey:
Good point, thank you Jiri :)

PR #510, opened 2019-05-29
[PLANNER-1483] Add change move micro-benchmark, profile for running benchmarks

Fonifan:
Depends on kiegroup/droolsjbpm-build-bootstrap/pull/965.

### DIFF
--- a/optaplanner-core/pom.xml
+++ b/optaplanner-core/pom.xml
@@ -94,8 +94,102 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- Benchmarking -->
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
-
+  <profiles>
+    <profile>
+      <id>benchmark</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>build-helper-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>add-test-source</id>
+                <phase>generate-test-sources</phase>
+                <goals>
+                  <goal>add-test-source</goal>
+                </goals>
+                <configuration>
+                  <sources>
+                    <source>src/benchmark/java</source>
+                  </sources>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>testCompile</goal>
+                </goals>
+                <configuration>
+                  <annotationProcessorPaths>
+                    <path>
+                      <groupId>org.openjdk.jmh</groupId>
+                      <artifactId>jmh-generator-annprocess</artifactId>
+                      <version>${jmh.version}</version>
+                    </path>
+                  </annotationProcessorPaths>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <artifactId>maven-assembly-plugin</artifactId>
+            <configuration>
+              <descriptors>
+                <descriptor>src/benchmark/assembly/benchmarks.xml</descriptor>
+              </descriptors>
+            </configuration>
+            <executions>
+              <execution>
+                <id>make-assembly</id>
+                <phase>package</phase>
+                <goals>
+                  <goal>single</goal>
+                </goals>
+                <configuration>
+                  <attach>true</attach>
+                  <archive>
+                    <manifest>
+                      <mainClass>org.openjdk.jmh.Main</mainClass>
+                    </manifest>
+                  </archive>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+      <dependencies>
+        <dependency>
+          <groupId>org.openjdk.jmh</groupId>
+          <artifactId>jmh-core</artifactId>
+          <scope>test</scope>
+        </dependency>
+        <dependency>
+          <groupId>org.openjdk.jmh</groupId>
+          <artifactId>jmh-generator-annprocess</artifactId>
+          <scope>test</scope>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
   <build>
     <plugins>
       <plugin>
--- a/optaplanner-core/pom.xml
+++ b/optaplanner-core/pom.xml
@@ -142,7 +142,7 @@
                     <path>
                       <groupId>org.openjdk.jmh</groupId>
                       <artifactId>jmh-generator-annprocess</artifactId>
-                      <version>${jmh.version}</version>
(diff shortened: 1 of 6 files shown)


Fonifan:
@rsynek , @ge0ffrey could you review this, please.

ge0ffrey:
I won't have time before my PTO, but I 'll look at it afterwards. Meanwhile Radovan can take a first pass.

pom.xml
```diff
@@ -28,6 +28,7 @@
 
   <properties>
     <spotbugs.failOnViolation>true</spotbugs.failOnViolation>
+    <jmh.version>1.21</jmh.version>
```
rsynek: From the maintainability reasons, we define versions of dependencies in the `kie parent ` https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml. This way we keep dependencies of all repositories in sync.

Would you please open a PR to the droolsjbpm-build-bootstrap, defining the jmh version there? At the same moment you can also remove it from kie-benchmarks.

optaplanner-core/pom.xml
```diff
@@ -94,8 +94,102 @@
...
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
-
+  <profiles>
+    <profile>
+      <id>benchmark</id>
```
rsynek: How difficult it is to run the benchmark locally?

At the very least, let's please document it in a readme. Let's also think about how to simplify it to a single step, if possible.

pom.xml
Fonifan: Good point, done.

optaplanner-core/pom.xml
Fonifan: Added small paragraph to README about starting benchmarks locally.

optaplanner-core/src/benchmark/assembly/benchmarks.xml
```diff
@@ -0,0 +1,27 @@
...
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.0.0 http://maven.apache.org/xsd/assembly-2.0.0.xsd">
+  <id>benchmarks</id>
+  <formats>
+    <format>jar</format>
+  </formats>
+  <includeBaseDirectory>false</includeBaseDirectory>
+  <dependencySets>
+    <dependencySet>
+      <outputDirectory>/</outputDirectory>
```
yurloc: Can we replace `/` with `${file.separator}` (or `.` or empty value or remove `<outputDirectory>` completely) to avoid the following warning?
```
[INFO] --- maven-assembly-plugin:3.1.0:single (make-assembly) @ optaplanner-core ---
[INFO] Reading assembly descriptor: src/benchmark/assembly/benchmarks.xml
[WARNING] The assembly descriptor contains a filesystem-root relative reference, which is not cross platform compatible /
```
https://stackoverflow.com/questions/28500401/maven-assembly-plugin-warning-the-assembly-descriptor-contains-a-filesystem-roo/

Fonifan: Nice catches! Fixed!

optaplanner-core/src/benchmark/java/org/optaplanner/core/impl/heuristic/move/AbstractPlannerMoveMicroBenchmark.java
```diff
@@ -0,0 +1,64 @@
...
+import org.openjdk.jmh.annotations.Warmup;
+import org.optaplanner.core.impl.score.director.ScoreDirector;
+
+@State(Scope.Thread)
+@BenchmarkMode(Mode.AverageTime)
+@Warmup(iterations = 5, time = 3)
+@Measurement(iterations = 5, time = 5)
+@Fork(value = 5)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+public abstract class AbstractPlannerMoveMicroBenchmark<Solution> {
```
rsynek: @Fonifan Running it locally I see following results:
Benchmark                            Mode  Cnt     Score      Error  Units
ChangeMoveBenchmark.benchmarkDoMove  avgt   25  4043.934 ± 1203.210  ns/op
ChangeMoveBenchmark.benchmarkRebase  avgt   25   451.929 ±   96.274  ns/op

The error looks scary; 4043.934 ± 1203.210 is not much reliable result. What numbers do you have locally?

pom.xml
```diff
@@ -90,6 +90,18 @@
         <!-- Old version until this issue is resolved: https://github.com/ImmobilienScout24/illegal-transitive-dependency-check/issues/30 -->
         <version>1.0.4</version>
       </dependency>
+      <dependency>
```
rsynek: Soft suggestion: the dependency management is usually part of the kie-parent too, so that you can just use the dependency and the version property is hidden for you.

optaplanner-core/src/benchmark/java/org/optaplanner/core/impl/heuristic/move/AbstractPlannerMoveMicroBenchmark.java
Fonifan: I have the same settings on my machine, but it seemed to be more stable (situation like this would occur once in 10 runs, so I thought it will be more stable on perf-machine and thus these settings will be a nice compromise between time and stability). 

Increasing number of warmup iterations to 10 seems to stabilize the benchmark, but it takes around 10 minutes for a single run. Are these optimal numbers?

pom.xml
Fonifan: Merge of pr on `kie-parent` takes a lifetime to complete.

There is no problem for me to move these dependencies to parent, but can you, please, advise me how to get my pr reviewed&merged?

optaplanner-core/src/benchmark/java/org/optaplanner/core/impl/heuristic/move/AbstractPlannerMoveMicroBenchmark.java
rsynek: Depends - does it improve if you increase warmup iterations? Or warmup time? Do we need so many forks? Without having reliability of results below 5%, there is no point in having such a benchmark. On laptop, the reliability may be worse, let's say between 5 and 10%, but anything more makes the benchmark unusable for a developer who would like to run it locally.

pom.xml
rsynek: Requested review from one of the owners of the repository there. Useful tip to avoid PR check failing when you have a dependency on change in other repository: create both PRs from branches that have the same name. This way, the PR check will build PRs from both repositories.

Fonifan: That's really nice! Thank's for the tip. That PR is approved, yet still not merged...

optaplanner-core/src/benchmark/java/org/optaplanner/core/impl/heuristic/move/AbstractPlannerMoveMicroBenchmark.java
Fonifan: After experimenting with settings, I've managed to find a good compromise -> 3 forks, 10 warmup iterations.
Results from 3 runs:
* ChangeMoveBenchmark.benchmarkDoMove  avgt   15  2591.209 ± 85.654  ns/op
* ChangeMoveBenchmark.benchmarkDoMove  avgt   15  2573.370 ± 85.696  ns/op
* ChangeMoveBenchmark.benchmarkDoMove  avgt   15  2534.195 ± 196.403  ns/op

PR #512, opened 2019-06-12
Increase timeouts for flaky tests

rsynek:
1 second timeout is too strict; it may randomly fail on Jenkins machines.


yurloc:
Replacing random numbers with bigger random numbers can't hurt.

PR #513, opened 2019-06-12
Great deluge impl

BorisKocian:
Implementation of Great Deluge local search algorithm with tests and AcceptroConfig class modified to support it.


kie-ci2:

triceo:
@rsynek This is the implementation we talked about. After this is merged, we'll introduce a new bunch of documentation, as well as some benchmarking to compare this to other LS algorithms.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,115 @@
...
+    private int levelsLength = -1;
+    private double[] initialLevelScoreLevels;
+    private double[] initialLevelScoreLevelsNegative;
+    private double[] levelScoreLevels;
+
+    private double levelMinimum = 0;
+    private final double THRESHOLD = .0000001;
+
+    // Good value to come out from
+    private final double DEFAULTRAINSPEEDRATIO = 0.99999995;
```
triceo: Let's change this to `DEFAULT_RAIN_SPEED_RATIO`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private int levelsLength = -1;
+    private double[] initialLevelScoreLevels;
+    private double[] initialLevelScoreLevelsNegative;
+    private double[] levelScoreLevels;
+
+    private double levelMinimum = 0;
+    private final double THRESHOLD = .0000001;
+
+    // Good value to come out from
```
triceo: As long as you're mentioning it's a good value, it'd be a good thing to reference a source for that claim.

rsynek:
ok to test

rsynek:
> @rsynek This is the implementation we talked about. After this is merged, we'll introduce a new bunch of documentation, as well as some benchmarking to compare this to other LS algorithms.

I am personally fine with introducing documentation in a separate PR. By benchmarking you mean using the optaplanner-benchmark and including the comparison into the documentation?

triceo:
> > @rsynek This is the implementation we talked about. After this is merged, we'll introduce a new bunch of documentation, as well as some benchmarking to compare this to other LS algorithms.
> 
> By benchmarking you mean using the optaplanner-benchmark and including the comparison into the documentation?

Not sure if including it in the doc is necessary. But yes, we will run `optaplanner-benchmark` and compare against some SAs and tabus.

rsynek:
jenkins retest this

rsynek:
Please see the coverage report on SonarCloud, namely the coverage for null `initialLevel` has some space for improvement.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
yurloc: Maybe also use underscores to make it clear how many decimal places are there (`0.999_999_95` or some other grouping style that makes most sense in this case, for example `0.99_999_995` would make sense if it means 99.999,995%).

triceo:
@BorisKocian Plenty for you to do here so far. Looking forward to the changes.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,113 @@
...
+        if (initialLevel != null) {
+            for (double initialLevelLevel : ScoreUtils.extractLevelDoubles(initialLevel)) {
+                if (initialLevelLevel < 0.0) {
+                    throw new IllegalArgumentException("The initial level (" + initialLevel
+                            + ") cannot have negative level (" + initialLevelLevel + ").");
+                }
+            }
+            initialLevelScoreLevels = ScoreUtils.extractLevelDoubles(initialLevel);
+            levelScoreLevels = initialLevelScoreLevels;
+        } else {
```
rsynek: This branch contains juicy logic which hasn't been covered by any test yet.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,113 @@
...
+            initialLevelScoreLevelsNegative = ScoreUtils.extractLevelDoubles(phaseScope.getBestScore());
+            logger.info("{}", phaseScope.getBestScore().toShortString());
+            levelScoreLevels = initialLevelScoreLevelsNegative;
+
+            for (int i = 0; i < levelScoreLevels.length; i++) {
+                if (Math.abs(levelScoreLevels[i]) < THRESHOLD) {
+                    logger.info("{}", Double.toString(levelScoreLevels[i]));
+                    continue;
+                }
+                levelScoreLevels[i] = -levelScoreLevels[i]+5;
```
rsynek: I probably overlooked this in the first review. I assume the value "5" has some special meaning which is best to express by a name of a constant holding the value.

BorisKocian: I am really sorry about this one, but it should not be there at all.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
BorisKocian: Everything should be covered by tests now.

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptorTest.java
```diff
@@ -0,0 +1,275 @@
...
+        DefaultSolverScope<TestdataSolution> solverScope = new DefaultSolverScope<>();
+        solverScope.setBestScore(HardMediumSoftScore.of(-5, 0, -24));
+        LocalSearchPhaseScope<TestdataSolution> phaseScope = new LocalSearchPhaseScope<>(solverScope);
+        LocalSearchStepScope<TestdataSolution> lastCompletedStepScope = new LocalSearchStepScope<>(phaseScope, -1);
+        lastCompletedStepScope.setScore(HardMediumSoftScore.of(-5, 0, -24));
+        phaseScope.setLastCompletedStepScope(lastCompletedStepScope);
+        acceptor.phaseStarted(phaseScope);
+        double[] actualScore = {5, 0, 24};
+
+        assertEquals(true, Arrays.equals(actualScore, acceptor.getLevelScoreLevels()));
```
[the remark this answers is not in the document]
triceo: Also, it is possible - although I'm not exactly sure - that the parent class will already have some code to do that. If it doesn't, consider adding it. :-)

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptorTest.java
```diff
@@ -0,0 +1,275 @@
...
+        LocalSearchMoveScope<TestdataSolution> moveScope1 = new LocalSearchMoveScope<>(stepScope0, 0, mock(Move.class));
+        moveScope1.setScore(HardMediumSoftScore.of(0,-180,-500));
+        LocalSearchMoveScope<TestdataSolution> moveScope2 = new LocalSearchMoveScope<>(stepScope0, 0, mock(Move.class));
+        moveScope2.setScore(HardMediumSoftScore.of(0,-50,-800));
+        LocalSearchMoveScope<TestdataSolution> moveScope3 = new LocalSearchMoveScope<>(stepScope0, 0, mock(Move.class));
+        moveScope3.setScore(HardMediumSoftScore.of(-5,-50,-100));
+        LocalSearchMoveScope<TestdataSolution> moveScope4 = new LocalSearchMoveScope<>(stepScope0, 0, mock(Move.class));
+        moveScope4.setScore(HardMediumSoftScore.of(0,-180,-450));
+
+        assertEquals(true, acceptor.isAccepted(moveScope0));
```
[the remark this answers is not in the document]
rsynek: There is: http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertTrue(boolean)

Also seeing that multiple move scopes are asserted below, soft assertions could do a better job, so that these can be verified in a bulk.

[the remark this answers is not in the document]
triceo: Can we do AssertJ in OptaPlanner? I thought we can not. Or is there some way to do that in pure JUnit?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,115 @@
...
+            logger.info("{}", phaseScope.getBestScore().toShortString());
+            levelScoreLevels = initialLevelScoreLevelsNegative;
+
+            for (int i = 0; i < levelScoreLevels.length; i++) {
+                if (Math.abs(levelScoreLevels[i]) < THRESHOLD) {
+                    logger.info("{}", Double.toString(levelScoreLevels[i]));
+                    continue;
+                }
+                levelScoreLevels[i] = -levelScoreLevels[i];
+                logger.info("{}", Double.toString(levelScoreLevels[i]));
```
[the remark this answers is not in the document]
BorisKocian: These logging statements probably do not need to be there. I used them for checking code functionality.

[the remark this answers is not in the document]
rsynek: I checked other acceptors and they don't log anything. So if for nothing else, at least for the sake of consistency, let's avoid `INFO` level logging here. `DEBUG` will be more suitable for the purpose. Please consider also adding some explanation of what you are logging - printing just the score levels without any context is not much helpful.

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptorTest.java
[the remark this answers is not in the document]
BorisKocian: I did it this way, because of SA acceptorTest, where it is used like this: `assertEquals(true, acceptor.isAccepted(moveScope1));`
Also, I used it for all tests. May change it.

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptorTest.java
```diff
@@ -11,9 +11,12 @@
 import org.optaplanner.core.impl.solver.scope.DefaultSolverScope;
 import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
 
+import java.util.Arrays;
+
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verifyZeroInteractions;
```
rsynek: This import does not seem to be used.

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptorTest.java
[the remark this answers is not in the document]
BorisKocian: Values are `double`s because of `ScoreUtils.extractLevelDoubles` but they are whole numbers. May also change it if needed.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,115 @@
...
+    // change water level at the beginning of the step
+    public void stepStarted(LocalSearchStepScope stepScope) {
+        super.stepEnded(stepScope);
+        for (int i = 0; i < levelsLength; i++) {
+            if (rainSpeed != null) {
+                levelScoreLevels[i] = levelScoreLevels[i] - rainSpeed;
+            } else if (rainSpeedRatio != null) {
+                levelScoreLevels[i] = levelScoreLevels[i] * rainSpeedRatio;
+            } else {
+                levelScoreLevels[i] = levelScoreLevels[i] * DEFAULT_RAIN_SPEED_RATIO;
```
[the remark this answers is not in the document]
BorisKocian: It makes more sense.

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptorTest.java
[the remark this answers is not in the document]
rsynek: optaplanner-core already depends on AssertJ and there are some tests using it. 

@ge0ffrey 's intention was to migrate to junit5 first before we migrate to AssertJ, so that there is only a single source of assertions in OptaPlanner.

However, this test is a perfect fit for soft assertions and every tutorial recommends using AssertJ for that. In opinion it's reasonable to use it here.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
rsynek: thanks

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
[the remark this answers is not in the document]
BorisKocian: I rather deleted them for consistency among acceptors.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
```diff
@@ -495,6 +524,27 @@ public Acceptor buildAcceptor(HeuristicConfigPolicy configPolicy) {
                     stepCountingHillClimbingSize_, stepCountingHillClimbingType_);
             acceptorList.add(acceptor);
         }
+        if ((acceptorTypeList!= null && acceptorTypeList.contains(AcceptorType.GREAT_DELUGE))
+                || greatDelugeRainSpeed != null || greatDelugeRainSpeedRatio != null) {
+            GreatDelugeAcceptor acceptor = new GreatDelugeAcceptor();
+
+            if (greatDelugeRainSpeed != null && greatDelugeRainSpeedRatio != null) {
+                throw new IllegalArgumentException("You can use only rainSpeed or rainSpeedRatio at a time.");
```
ge0ffrey: Let's not blame the user, so no "You ...". Let's keep it consistent with the other exceptions thrown when 2 fields are used together that can't, some something like "The greatDelugeRainSpeed (" + greatDelugeRainSpeed ") and ... cannot be both non null.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.localsearch.decider.acceptor.AbstractAcceptor;
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchMoveScope;
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchPhaseScope;
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchStepScope;
+import org.optaplanner.core.impl.score.ScoreUtils;
+
+public class GreatDelugeAcceptor extends AbstractAcceptor {
+
+    private Score initialLevel = null;
```
ge0ffrey: the term "level" is used both by as score level and as water level. That creates ambiguity/confusion.
Powerthesaurus didn't help though, so I don't have a good alternative.
I did find "gauge height". Maybe we can use that?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+    public void setInitialLevels(Score initialLevel) { this.initialLevel = initialLevel; }
+
+    public Score getInitialLevels() { return this.initialLevel; }
+
+    public void setRainSpeed(Double rainSpeed) {
+        this.rainSpeed = rainSpeed;
+        this.rainSpeedRatio = null;
+    }
+
+    public Double getRainSpeed() { return this.rainSpeed; }
```
ge0ffrey: code style: the getter before the setter.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+                if (initialLevelLevel < 0.0) {
+                    throw new IllegalArgumentException("The initial level (" + initialLevel
+                            + ") cannot have negative level (" + initialLevelLevel + ").");
+                }
+            }
+            initialLevelScoreLevels = ScoreUtils.extractLevelDoubles(initialLevel);
+            levelScoreLevels = initialLevelScoreLevels;
+        } else {
+            double[] initialLevelScoreLevelsNegative;
+            initialLevelScoreLevelsNegative = ScoreUtils.extractLevelDoubles(phaseScope.getBestScore());
```
ge0ffrey: collapse line 57 and 58 together, we only declare a local var without initailizing it if there are multiple ways of initializing it

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+            levelScoreLevels = initialLevelScoreLevelsNegative;
```
ge0ffrey: The variable initialLevelScoreLevelsNegative is redudant. (highlight it to notice)

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private Double rainSpeed = null;
+    private Double rainSpeedRatio = DEFAULT_RAIN_SPEED_RATIO;
+
+    private int levelsLength = -1;
+    private double[] initialLevelScoreLevels;
+    private double[] levelScoreLevels;
+
+    private double levelMinimum = 0;
+    private static final double THRESHOLD = .0000001;
```
ge0ffrey: code style: in doubles we put the 0 in front too, so 0.0000001

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+            initialLevelScoreLevels = ScoreUtils.extractLevelDoubles(initialLevel);
+            levelScoreLevels = initialLevelScoreLevels;
+        } else {
+            double[] initialLevelScoreLevelsNegative;
+            initialLevelScoreLevelsNegative = ScoreUtils.extractLevelDoubles(phaseScope.getBestScore());
+            levelScoreLevels = initialLevelScoreLevelsNegative;
+
+            for (int i = 0; i < levelScoreLevels.length; i++) {
+                if (Math.abs(levelScoreLevels[i]) < THRESHOLD) {
+                    continue;
```
ge0ffrey: if this "continue"; occurs, the levelScoreLevels[i] isn't filled in, which means it sticks with java's default of 0.0.
That's hard to read - can we not simply do:
levelScoreLevels[i] = Math.abs(levelScoreLevels[i]) < THRESHOLD ? 0.0 : -levelScoreLevels[i]

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
ge0ffrey: What threshold is this? Can we give it a better self-explaining name?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+
+public class GreatDelugeAcceptor extends AbstractAcceptor {
+
+    private Score initialLevel = null;
+
+    private Double rainSpeed = null;
+    private Double rainSpeedRatio = DEFAULT_RAIN_SPEED_RATIO;
+
+    private int levelsLength = -1;
+    private double[] initialLevelScoreLevels;
```
ge0ffrey: Does this field need to be global? I think the double[] version of the initial waterlevel score can be a local variable in phaseStarted()

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+            levelScoreLevels = initialLevelScoreLevelsNegative;
+
+            for (int i = 0; i < levelScoreLevels.length; i++) {
+                if (Math.abs(levelScoreLevels[i]) < THRESHOLD) {
+                    continue;
+                }
+                levelScoreLevels[i] = -levelScoreLevels[i];
+            }
+        }
+        levelsLength = levelScoreLevels.length;
```
ge0ffrey: I'd remove the levelsLength global field and just use levelScoreLevels.length everywhere, for readability. Let the compile optimize inlining that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+        for (int i = 0; i < levelsLength; i++) {
+            double moveScoreLevel = moveScoreLevels[i];
+            double levelScoreLevel = levelScoreLevels[i];
+
+            double isZeroIfEqual = Math.abs(moveScoreLevel + levelScoreLevel);
+
+            if (moveScoreLevel > -levelScoreLevel) {
+                // Checks if given score is better than water level, if so, move is accepted.
+                return true;
+            } else if (!(isZeroIfEqual < THRESHOLD)) {
```
ge0ffrey: !(a < b) is the same as (a >= b)

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+                // Returns false when above condition is not met and, new score and water level are not equal.
+                return false;
+            } else if (i == levelsLength - 1){
```
ge0ffrey: Skip this if, there's a return true at line 101 after the for loop.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+    public boolean isAccepted(LocalSearchMoveScope moveScope) {
+
+        Score moveScore = moveScope.getScore();
+        double[] moveScoreLevels = ScoreUtils.extractLevelDoubles(moveScore);
+
+        for (int i = 0; i < levelsLength; i++) {
+            double moveScoreLevel = moveScoreLevels[i];
+            double levelScoreLevel = levelScoreLevels[i];
+
+            double isZeroIfEqual = Math.abs(moveScoreLevel + levelScoreLevel);
```
ge0ffrey: inline this into line 93 - so it's calculated where it's used

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+
+            if (moveScoreLevel > -levelScoreLevel) {
+                // Checks if given score is better than water level, if so, move is accepted.
```
ge0ffrey: This comment does not add any extra information that the if statement on line 90 already does.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,118 @@
...
+    private Score initialLevel = null;
+
+    private Double rainSpeed = null;
+    private Double rainSpeedRatio = DEFAULT_RAIN_SPEED_RATIO;
+
+    private int levelsLength = -1;
+    private double[] initialLevelScoreLevels;
+    private double[] levelScoreLevels;
+
+    private double levelMinimum = 0;
```
ge0ffrey: (serious issue) The levelMinimum is the same for all levels - this basically presumes that all constraints are the same sign and are somehow in the same scale. In reality, most use cases have wildly different scales between hard and soft (for example 7 hard broken and 1234567 soft broken - the latter representing money) so that won't give a problem as long as the minimum is 0. As for the sign: some example use positive constraints and have a positive soft score.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
ge0ffrey: (serious issue) The threshold is to deal with floating point numerical stability.
But why not avoid the entire problem, but never using doubles, but Score instances?
This works:
- score.subtract(rainSpeedScore)
- score.multiply(rainRatioDouble)

ScoreUtils.extractLevelDoubles() is inaccurate. Try a SimpleLongScore of 9007199254740993L. [1], the double will be a different number. Try a SimpleBigDecimalScore of "0.01" with "0.09", similar problem. By keeping them in their original Score type (using getScoreDefinition() methods where needed), that numerical instability goes away.

[1] https://ge0ffrey.github.io/ge0ffrey-presentations/cornerCaseCheatSheet/cheatSheetJava.html

Using

ge0ffrey:
It's a good start. Besides the 2 "(serious issue)" that are important (but might need discussion to explain the problem better), we'll also want a small section in LocalSearch-chapter.adoc, so users can start using it too.

triceo:
@ge0ffrey We're working on docs in a separate PR, if that's OK with you.
@BorisKocian This is the final review we've been expecting. Let me know if you have any issues with anything, let's get it done.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
BorisKocian: Yes we can. Good point

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
BorisKocian: I should leave the variable isZeroIfEqual as self-documenting code?

ge0ffrey: No need.
Besides, a variable called "is..." that is a not a boolean is a bit weird to me.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
BorisKocian: This was a good suggestion and I reworked everything according to this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
BorisKocian: I didn't know about that. I thought the scoring function value is always negative and tries to maximize (approach 0). So what should be used as a breakpoint? There have to be some. For example, a function with the initial score of -10hard / -1000soft must increase of hard score end on 0, otherwise, none of the next steps can be accepted, because it can find best solutions with 0hard/...soft.
Can you please give me some direction to go?

triceo: Wouldn't it make sense to do the same thing that SA does? I'm assuming SA has to solve the same problem - correct, @ge0ffrey ?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public boolean isAccepted(LocalSearchMoveScope moveScope) {
+
+        Score moveScore = moveScope.getScore();
+
+        return moveScore.compareTo(currentWaterLevel.negate()) >= 0;
+    }
+
+    public void stepStarted(LocalSearchStepScope stepScope) {
+        super.stepEnded(stepScope);
```
ge0ffrey: stepEnded being called in stepStarted: not ok
I believe the water level should change at the end of a step, in the stepEnded method.

ge0ffrey: This bug is also in SimulatedAnnealing, we're fixing it there...

ge0ffrey: Also use  @Override when possible

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,87 @@
...
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchStepScope;
+import org.optaplanner.core.impl.score.ScoreUtils;
+
+public class GreatDelugeAcceptor extends AbstractAcceptor {
+
+    private Score initialWaterLevels = null;
+    private Score currentWaterLevel = null;
+
+    private Double rainSpeedRatio = DEFAULT_RAIN_SPEED_RATIO;
+    private Score rainSpeedScore = null;
```
ge0ffrey: naming: rainSpeedScore is addend part of addition. Let's call it rainSpeedIncrement.

ge0ffrey: Nitpick: put rainSpeedIncrement before rainSpeedRatio, because the stepEnded method handles rainSpeedIncrement before handling rainSpeedRatio

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,87 @@
...
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.localsearch.decider.acceptor.AbstractAcceptor;
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchMoveScope;
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchPhaseScope;
+import org.optaplanner.core.impl.localsearch.scope.LocalSearchStepScope;
+import org.optaplanner.core.impl.score.ScoreUtils;
+
+public class GreatDelugeAcceptor extends AbstractAcceptor {
+
+    private Score initialWaterLevels = null;
```
ge0ffrey: Naming: singular like next line?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,87 @@
...
+
+        return moveScore.compareTo(currentWaterLevel.negate()) >= 0;
+    }
+
+    public void stepStarted(LocalSearchStepScope stepScope) {
+        super.stepEnded(stepScope);
+        if (rainSpeedScore != null) {
+            currentWaterLevel = currentWaterLevel.subtract(rainSpeedScore);
+        } else {
+            currentWaterLevel = currentWaterLevel.multiply(rainSpeedRatio);
```
ge0ffrey: Note (no changes needed): this inherently presumes a optimal score of Score.ZERO. 
In a future PR, after this one is merged, we'll want to relax this presumption, by adding a targetScore parameter.

- In most cases the perfect score [1] is ZERO, because most cases only have negative constraints. For example, in course scheduling, we'd love to see 0 hard constraints and 0 soft constraints broken. There are cases that mix in positive constraints (for example conference scheduling) and the optimal score could be 0hard/700soft. Possible fix: adding a targetScore parameter that defaults to Score.ZERO.

- In most cases, the optimal score is not the perfect score. For example, in VRP, a score 0hard/0soft is impossible, because the driving time factors into the soft score: delivering at 10 locations is impossible in 0 seconds. Possible fix: adding a targetScore parameter that defaults to Score.ZERO.

The code would look like this:

```
// Notice that the rainSpeedRatio's meaning changes: Instead of setting it to 0.95, we 'll set it to 0.05. [2]
Score increment = targetScore.subtract(currentWaterLevel).multiply(rainSpeedRatio);
currentWaterLevel += increment;
```

We don't know the optimal score in advance, so the targetScore could be a pita parameter if the default of ZERO isn't any good.

[1] perfect score is not general academic terminology. It's NOT a synonym for optimal score or lower/higher bound score.
[2] @triceo we need to figure this out before merging. I would expect that the ratio is the relative amount of the water that is added, between current water level and highest possible water level.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
```diff
@@ -495,6 +525,31 @@ public Acceptor buildAcceptor(HeuristicConfigPolicy configPolicy) {
...
             acceptorList.add(acceptor);
         }
+        if ((acceptorTypeList!= null && acceptorTypeList.contains(AcceptorType.GREAT_DELUGE))
+                || greatDelugeRainSpeedRatio != null || greatDelugeRainSpeedScore != null) {
+            GreatDelugeAcceptor acceptor = new GreatDelugeAcceptor();
+
+            if (greatDelugeRainSpeedScore != null && greatDelugeRainSpeedRatio != null) {
+                throw new IllegalArgumentException("The greatDelugeRainSpeedScore (" + greatDelugeRainSpeedScore
+                        + ") and greatDelugeRainSpeedRatio (" + greatDelugeRainSpeedRatio
+                        + ") cannot be both non null.");
```
ge0ffrey: Great work. One more thing to think of: can we set a default ratio? What would be a good default ratio? 0.05?

Motivation: SA is underused by our users because it requires a domain-specific parameter to be set.
LA and TS do not. They are used far more due to this, because they are easier to use.
Our users often do a blueprint benchmark and use the best one in production.
SA isn't blueprint friendly, so it's often not even considered by many users. That's a pity.
A good default ratio would make it blueprint-friendly, see docs "17.2.6. Benchmark Blueprint: A Predefined Configuration"

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
ge0ffrey: On point we need to define clearly before merging What is the ratio? Is it 0.05 for 0.95? I would argue 0.05, as we increase the water level be 5% towards the final state of great deluge at every step. The idea of we keep 95% not flooded is inverse thinking. So I would argue that 95% is implementation friendly, and 5% is user friendly. We should use 5%.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
triceo: The default ratio is here, incl. a comment for where the number comes from:
https://github.com/kiegroup/optaplanner/pull/513/files#diff-85928658d71f0163f269403e0d64432bR19

Did you mean something else?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
ge0ffrey: Lukas meeting decision:
- waterLevelIncrementScore
- waterLevelIncrementRatio

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
ge0ffrey: TODO for next pr: better name targetScore

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,93 @@
...
+    public void stepEnded(LocalSearchStepScope stepScope) {
+        super.stepEnded(stepScope);
+
+        if (waterLevelIncrementScore != null) {
+            currentWaterLevel = currentWaterLevel.subtract(waterLevelIncrementScore);
+        } else {
+            Score increment = currentWaterLevel.multiply(waterLevelIncrementRatio);
+            currentWaterLevel = currentWaterLevel.subtract(increment);
+        }
+        logger.debug(currentWaterLevel.toShortString());
```
triceo: @BorisKocian I think that this debug line needs to go away.

BorisKocian:
> So, @BorisKocian, here's the minutes of the final meeting between Geoffrey and me. Here's what we need you to do, and then we're done:
> ...
> ...
> ...

It should be all done by now, but there is no solution for "waterLevelMinimum". I am not sure, but I think it is needed if an algorithm is not stopped in time by any termination condition. 

Also, I did some here not requested changes in LocalSearchType, to be possible to use: `<localSearchType>GREAT_DELUGE</localSearchType>` in config.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
BorisKocian: I'm sorry, my bad.

ge0ffrey:
@BorisKocian Thanks for adding <localSearchType>GREAT_DELUGE. If you do a blueprint benchmark on ALL_LOCAL_SEARCH_TYPES, it should now show up (and yes, SA will fail there).

As for waterLevelMinimum: running without termination condition happens in production during real-time planning: there is an ad-hoc termination from another thread through solver.terminateEarly(). So no need to require a termination.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/LocalSearchPhaseConfig.java
```diff
@@ -258,6 +261,9 @@ protected LocalSearchForager buildForager(HeuristicConfigPolicy configPolicy) {
                     // Fast stepping algorithm
                     foragerConfig_.setAcceptedCountLimit(1);
                     break;
+                case GREAT_DELUGE:
+                    // Fast stepping algorithm
+                    foragerConfig_.setAcceptedCountLimit(4);
```
ge0ffrey: Question: Why acceptedCountLimit 4? I doubt if the canonical Great Deluge paper suggest a "tournament factor" of 4.
In the past, I've overfitted some examples's benchmarks to use an acceptedCountLimit of 4 for SA (cloud balancing for one). This legacy still lingers in many of the benchmark configs. There is no reliable basis for this, except that on the first example where I did it, nurse rostering, it worked better than 1 for the 10 medium datasets, but a neglectable margin. I 've not confirmed this decently with other use cases.

Let's go with acceptedCountLimit 1, just like SA in this code, unless there is proof on at least 3 examples for at least 5 datasets each that 4 works better. @triceo Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
```diff
@@ -74,6 +75,10 @@
     protected Integer stepCountingHillClimbingSize = null;
     protected StepCountingHillClimbingType stepCountingHillClimbingType = null;
 
+    protected String greatDelugeInitialWaterLevel = null;
+    protected Double greatDelugeWaterLevelIncrementRatio = null;
+    protected String greatDelugeWaterLevelIncrementScore = null;
```
ge0ffrey: Nitpick: greatDelugeWaterLevelIncrementScore before greatDelugeWaterLevelIncrementRatio here too, also getter/setters and in the 2 methods that handle them.
Motivation:
- the ordering should be consistent in the code, so it's easier to spot when something is missing. Bugs are easier to spot this way.
- incrementScore is the simple one, so it should be first and incrementRatio is the more dynamic better one, so it should be second

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final double DEFAULT_WATER_LEVEL_INCREMENT_RATIO = 0.00_000_005;
+
+
+    public Score getWaterLevelIncrementScore() {
+        return this.waterLevelIncrementScore;
+    }
+
+    public void setWaterLevelIncrementScore(Score waterLevelIncrementScore) {
+        this.waterLevelIncrementScore = waterLevelIncrementScore;
+        this.waterLevelIncrementRatio = null;
```
ge0ffrey: Use a simple setter (notice that the setWaterLevelIncrementRatio doesn't have this side effect) and have either the AcceptorConfig set the default value, similar to how it sets 400 for Late Acceptance.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/acceptor/greatdeluge/GreatDelugeAcceptor.java
```diff
@@ -0,0 +1,93 @@
...
+            currentWaterLevel = phaseScope.getBestScore().negate();
+        }
+    }
+
+    @Override
+    public void phaseEnded(LocalSearchPhaseScope phaseScope) {
+        super.phaseEnded(phaseScope);
+        initialWaterLevel = null;
+        waterLevelIncrementRatio = DEFAULT_WATER_LEVEL_INCREMENT_RATIO;
+        waterLevelIncrementScore = null;
```
ge0ffrey: Bug: running the phase twice (not really reproducible right now) will make it forget it's settings. remove these 2 lines.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/LocalSearchPhaseConfig.java
ge0ffrey: btw Lukas, we have all these examples, all these datasets, all this algorithms and a benchmarker to mine this gold mine of research information. Maybe we should spend some time in doing that - and meanwhile clean up the benchmark configs to avoid magic numbers like acceptedCountLimit=4, tabuSize=11, etc in those benchmark configs - unless they can be proven. Furthermore, it might affect our defaults too (such as tabuSize=7)

triceo: I agree we should not be making decisions unless we have solid factual basis for them.
Boris and I actually tried to mine our own datasets, and we started comparing local searches in various configs against each other - unfortunately, we hit a wall with SA, the config of which is too problem-specific to allow any generalizations.

Let's discuss this elsewhere. This PR already has 90+ comments. :-)

ge0ffrey: Yes, it's great work though. I look forward to having this in our arsenal of algorithms. Thanks Boris keeping up with all these hard reviews :)

triceo:
I think we're good now, @ge0ffrey .

ge0ffrey:
Thanks Boris & Lukas, great work! :)

ge0ffrey:
@BorisKocian if you 'd like your name in the contributors page, please add a PR on optaplanner-website to add yourself in this file: https://github.com/kiegroup/optaplanner-website/blob/master/_config/users.yml

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/AcceptorConfig.java
```diff
@@ -495,6 +524,35 @@ public Acceptor buildAcceptor(HeuristicConfigPolicy configPolicy) {
...
             acceptorList.add(acceptor);
         }
+        if ((acceptorTypeList!= null && acceptorTypeList.contains(AcceptorType.GREAT_DELUGE))
+                || greatDelugeInitialWaterLevel != null
+                || greatDelugeWaterLevelIncrementScore != null
+                || greatDelugeWaterLevelIncrementRatio != null) {
+            GreatDelugeAcceptor acceptor = new GreatDelugeAcceptor();
+
+            if (greatDelugeWaterLevelIncrementScore != null && greatDelugeWaterLevelIncrementRatio != null) {
+                throw new IllegalArgumentException("The greatDelugeWaterLevelIncrementScore ("
```
ge0ffrey: I don't think the default has been tested. Fixing that now, I 'll submit a separate PR for your review.

ge0ffrey:
@BorisKocian @triceo Please review this follow up PR - https://github.com/kiegroup/optaplanner/pull/522

ge0ffrey:
the initial water level can be a postive or negative, depending on the use case.

PR #518, opened 2019-07-04
Support java.util.Date in TestGen

rsynek:
This modification was necessary for me to proceed in isolation of NPE in Drools.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenDateValueProvider.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenDateValueProvider.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.score.director.drools.testgen.fact;
+
+import java.util.Date;
+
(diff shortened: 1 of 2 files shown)


ge0ffrey:
Would we want similar support for LocalDate, LocalDateTime and LocalTime? (or Temporal in general)
Doesn't need to be in this PR - just thinking aloud.
This PR is good to merge.

rsynek:
@ge0ffrey `LocalDate` and others work because they have `parse` method and some types of fields are created by having that method. See https://github.com/kiegroup/optaplanner/blob/master/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenValueFact.java#L98

I would also prefer waiting for @yurloc before I finish this PR (there is no coverage for the new code).

ge0ffrey:
Given that users don't use the testgen framework, I don't mind there being no coverage. It's always nice to have of course.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/testgen/fact/TestGenDateValueProvider.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.score.director.drools.testgen.fact;
+
+import java.util.Date;
+
+public class TestGenDateValueProvider extends TestGenAbstractValueProvider<Date> {
```
yurloc: Should be package-private.

ge0ffrey: Done through github suggestion.

yurloc:
For the record, this shouldn't have been merged. I suspect that the new `Date` provider class behaves incorrectly. We agreed with @rsynek that he would add a test that prove my concern and he would change the implementation accordingly.

Since TestGen is very low priority, let's address this later (when someone hits the bug in `TestGenDateValueProvider` :slightly_smiling_face:).

ge0ffrey:
@yurloc My apologies for jumping the gun. I am trying to go through the PR list and getting the PR's merged (good is the enemy of perfect - sometimes). Further issues can be handled in incremental PR's.

rsynek:
My bad - I should have updated this conversation. Nevermind, I will prepare a new PR with the remaining work.

rsynek:
Finished in https://github.com/kiegroup/optaplanner/pull/523

PR #520, opened 2019-07-08
BAQE-1001 Unify reporting to SonarCloud under KIE parent

rsynek:
Depends on https://github.com/kiegroup/kie-jenkins-scripts/pull/492 and https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/979

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,39 +163,6 @@
         </pluginManagement>
       </build>
     </profile>
-
-    <profile>
-      <!--
-        See .travis.yml for how to run the coverage report
-      -->
-      <id>report-code-coverage</id>
-      <properties>
-        <jacoco.master.exec.file>${maven.multiModuleProjectDirectory}/target/jacoco.exec</jacoco.master.exec.file>
-        <sonar.jacoco.reportPaths>${jacoco.master.exec.file}</sonar.jacoco.reportPaths>
-      </properties>
-      <build>
-        <pluginManagement>
-          <plugins>
-            <plugin>
-              <groupId>org.jacoco</groupId>
-              <artifactId>jacoco-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -156,46 +156,14 @@
                   <groupId>org.jacoco</groupId>
                   <artifactId>org.jacoco.agent</artifactId>
                   <version>${version.jacoco.plugin}</version>
+                  <classifier>runtime</classifier>
```
yurloc: What is the difference between regular agent JAR and `-runtime` one? Why didn't you add the `runtime` classifier to drools repo as well? Can we move the Surefire plugin configuration up to the parent too?

rsynek: ad the runtime JAR:
the documentation suggests this is the agent, while the regular one is "API to get a local copy of the agent" https://www.jacoco.org/jacoco/trunk/doc/repo.html. If you download it, the regular jar contains the agent jar inside plus some wrapper class. I added the classifier to make it consistent with the agent line, when the runtime jar is referred.

ad drools repo: good catch. As I already put the change in this PR (although it's not directly connected to moving the reporting to KIE parent), I will do it for drools too.

ad the Surefire plugin:
unfortunately, not every repository in KIE group can use the same configuration. Some (primarily engines) rely upon the runtime agent jar, while others (wb-related, containing GWT code) require using Jacoco maven plugin. For these two cases the Surefire plugin's configuration is different.

rsynek: The classifier has been added into the PR to Drools .

ge0ffrey:
@rsynek because this is a multi-repo PR, I presume you'll do the merge once it's all approved? Or do you want me to do it?

rsynek:
@ge0ffrey changes to kie-parent and jenkins scripts have been just merged, feel free to merge this one too. Thanks!

PR #526, opened 2019-07-23
PLANNER-1565 Optimize subpillar selection

triceo:
This PR brings an observed 25 - 50 % score calculation speed increase on Machine Reassignment and 2 - 5 % improvement on Cloud Balancing.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
@@ -18,11 +18,14 @@ package org.optaplanner.core.impl.heuristic.selector.entity.pillar;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.optaplanner.core.config.heuristic.selector.common.SelectionCacheType;
 import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
```diff
@@ -252,22 +255,29 @@ public RandomSubPillarIterator() {
...
+            if (basePillarSize == 1) { // no subpillar to select
+                return basePillar;
+            }
             // Known issue/compromise: Every subPillar should have same probability, but doesn't.
             // Instead, every subPillar size has the same probability.
-            int basePillarSize = basePillar.size();
-            int min = (minimumSubPillarSize > basePillarSize) ? basePillarSize : minimumSubPillarSize;
-            int max = (maximumSubPillarSize > basePillarSize) ? basePillarSize : maximumSubPillarSize;
+            int min = Math.min(minimumSubPillarSize, basePillarSize);
+            int max = Math.min(maximumSubPillarSize, basePillarSize);
```
triceo: I replaced the original code with `Math.min()` so that the JVM can replace it with its intrinsic implementation, leading to an even more improved performance.

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelectorTest.java
```diff
@@ -312,11 +312,10 @@ public void randomWithSubs_Size2() {
...
         // Expected pillar cache: [a], [b, d], [c, e, f]
         when(workingRandom.nextInt(anyInt())).thenReturn(
-                0, 0, 0, // [a] If the pillar is less than the minimum, the minimum is reduced
-                2, 0, 0, 0, // [c, e, f]
-                1, 0, 0, 0, // [b, d]
-                1, 0, 1, 0); // [b, d]
-        assertCodesOfNeverEndingPillarSelector(pillarSelector, "[a]", "[c, e]", "[b, d]", "[d, b]");
+                0, // [a]
+                2, 0, 0, 1, // [c, e, f]
+                1, 0); // [b, d]
```
triceo: The last option (`[d, b]` instead of `[b, d]`) was removed, as one of the shortcuts makes it impossible to generate this permutation. It has been agreed beforehand that removing permutations is not an issue.

triceo:
@ge0ffrey All green including Sonar. From my point of view, this is ready to go.

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelectorTest.java
```diff
@@ -226,11 +226,11 @@ public void randomWithSubs() {
         // nextInt pattern: pillarIndex, subPillarSize, element 0, element 1, element 2, ...
         // Expected pillar cache: [a], [b, d], [c, e, f]
         when(workingRandom.nextInt(anyInt())).thenReturn(
-                0, 0, 0, // [a]
-                2, 1, 0, 0, // [c, e, f]
```
rsynek: A suggestion to make the test more readable:
would you please explain on one example, what is the relation of mocked random numbers to the subpillar being selected? Even with the comment on lines 226 and 227 it's difficult to see the connection without detailed knowledge of the implementation.

ge0ffrey: This part of the test is painful indeed, in the original code as much as it is now.
It's mocking random numbers to trigger certain selections.

But this is the lesser evil.
The other 2 options I am aware of are:
- not testing it (or not testing parts of it)
- statistical testing that randomly fails sometimes (we don't have the setup to accommodate this)
So, we stick with this approach until something better comes along.

ge0ffrey: @triceo @triceo Just thinking further on this (already merged, so a different issue):
What if we do statistical testing with a fixed random seed? That means we don't have random failing tests.

For example:
Given 4 entities, we select 100 subpillars and we validate that every combination of those 4 is selected.

triceo: I'm not sure, @ge0ffrey. It will make the tests run much longer, and I have no idea how to pick a random number and a data set that will cause it to trigger exactly what we need.

Too much effort for too little value?

rsynek: Let's just provide detailed explanation of what's happening with random numbers in the test and it will be much better than what we have now.

PR #527, opened 2019-07-29
PLANNER-1270 Provide tests and fix for unimproved time spent score diff termination

rsynek:
- adds a new constructor taking a `Clock` instance as a parameter to enable testing
- provides unit coverage for both the `UnimprovedTimeMillisSpentTermination` and `UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination`
- fixes a bug that solving terminates despite the fact that score improved within a time limit over a given threshold

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination.java
@@ -16,6 +16,7 @@
 
 package org.optaplanner.core.impl.solver.termination;
 
+import java.time.Clock;
 import java.util.ArrayDeque;
 import java.util.Iterator;
 import java.util.Queue;
@@ -31,6 +32,7 @@ public class UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination extend
 
     private final long unimprovedTimeMillisSpentLimit;
     private final Score unimprovedScoreDifferenceThreshold;
+    private final Clock clock;
 
     private Queue<Pair<Long, Score>> bestScoreImprovementHistoryQueue;
     // safeTimeMillis is until when we're safe from termination
@@ -38,13 +40,20 @@ public class UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination extend
     private long phaseSafeTimeMillis = -1L;
 
     public UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination(long unimprovedTimeMillisSpentLimit,
-            Score unimprovedScoreDifferenceThreshold) {
+                                                                        Score unimprovedScoreDifferenceThreshold) {
+        this(unimprovedTimeMillisSpentLimit, unimprovedScoreDifferenceThreshold, Clock.systemUTC());
+    }
+
+    protected UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination(long unimprovedTimeMillisSpentLimit,
+                                                                           Score unimprovedScoreDifferenceThreshold,
+                                                                           Clock clock) {
         this.unimprovedTimeMillisSpentLimit = unimprovedTimeMillisSpentLimit;
         this.unimprovedScoreDifferenceThreshold = unimprovedScoreDifferenceThreshold;
         if (unimprovedTimeMillisSpentLimit <= 0L) {
             throw new IllegalArgumentException("The unimprovedTimeMillisSpentLimit (" + unimprovedTimeMillisSpentLimit
-                    + ") cannot be negative.");
+                                                       + ") cannot be negative.");
         }
+        this.clock = clock;
     }
 
     public long getUnimprovedTimeMillisSpentLimit() {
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTerminationTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+public class UnimprovedTimeMillisSpentScoreDifferenceThresholdTerminationTest {
+
+    private static final long START_TIME_MILLIS = 0L;
+
+    @Test
+    public void forZeroUnimprovedTimeMillis_exceptionIsThrown() {
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> new UnimprovedTimeMillisSpentTermination(0L))
+                .withMessageContaining("cannot be negative");
```
yurloc: Slightly confusing message. Although I agree it might depend on the context, it seems that in most cases, zero is considered to be neither negative nor positive. Let's use an unambiguous message, for example "must be greater than zero" or "must not be negative or zero".

It's already been [discussed](https://github.com/kiegroup/optaplanner/pull/449#discussion_r227286991) and [dealt with](e1017c77a3a25598ce2491ce1fafd3efae372e40) in the past.

Useful resources:
- http://mathworld.wolfram.com/Zero.html
- https://math.stackexchange.com/questions/26705/is-zero-positive-or-negative

rsynek: That's a good point, the error messages are not clear enough in this regard. Will improve them.

rsynek: Interestingly, the `UnimprovedTimeMillisSpentTermination` checks only for negative numbers (zero is allowed). @ge0ffrey is this difference expected?

There is the same kind of difference, i.e. `>` vs. `>=` also in the `isTerminated()` method.

ge0ffrey: The acceptance of zero was intentional IIRC. It's a general philosophy: 0 time available, empty entity list, etc - it happens in dynamic usage of optaplanner and it's strange but not invalid. In fact, we violate that philisophy a few times (empty customer list for vrp IIRC) and it's a big pita, but we intend to fix those.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination.java
```diff
@@ -86,9 +95,13 @@ public void stepEnded(AbstractStepScope stepScope) {
             for (Iterator<Pair<Long, Score>> it = bestScoreImprovementHistoryQueue.iterator(); it.hasNext(); ) {
                 Pair<Long, Score> bestScoreImprovement = it.next();
                 Score scoreDifference = bestScore.subtract(bestScoreImprovement.getValue());
-                if (scoreDifference.compareTo(unimprovedScoreDifferenceThreshold) >= 0) {
+                final boolean timeLimitNotYetReached =
+                        bestScoreImprovement.getKey() + unimprovedTimeMillisSpentLimit >= bestSolutionTimeMillis;
```
ge0ffrey: Can't this also reuse bestSolutionTimeMillis directly instead of getKey()?

ge0ffrey: Also, optaplanner's code doesn't add "final" for local variables (it's a trade-off between readability (less is more) and avoiding bugs (final to be sure)).
For consistency's sake, I 'd prefer not to do have them on some code, and not on other code, so let's remove the final keywords here.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTerminationTest.java
rsynek: +1 , I will adjust the fail-fast checks to allow `0` millis.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination.java
rsynek: ad `final` : ok
ad reusing `bestSolutionTimeMillis` : there can be no reuse, as `bestSolutionTimeMillis` are taken from the current step, whereas `bestScoreImprovement.getKey()` comes from the last step that improved a score.

ge0ffrey:
Great work btw Radovan :) Nice bug found!

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination.java
```diff
@@ -31,20 +32,28 @@
...
 
     public UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination(long unimprovedTimeMillisSpentLimit,
-            Score unimprovedScoreDifferenceThreshold) {
+                                                                        Score unimprovedScoreDifferenceThreshold) {
+        this(unimprovedTimeMillisSpentLimit, unimprovedScoreDifferenceThreshold, Clock.systemUTC());
+    }
+
+    protected UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination(long unimprovedTimeMillisSpentLimit,
+                                                                           Score unimprovedScoreDifferenceThreshold,
+                                                                           Clock clock) {
```
yurloc: Suggestion: if you want to avoid the "align method parameters/call arguments when multiline or not" dilemma. format it like this:
```java
protected UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination(
        long unimprovedTimeMillisSpentLimit,
        Score unimprovedScoreDifferenceThreshold,
        Clock clock
) {
   // body
}
```

This is compatible with both "align ... when multiline" that's part of KIE Java Conventions and Geoffrey's IDE settings that keep ignoring KIE Java Conventions in this regard :wink:.

I don't like the aligning either. I chose to follow KIE conventions so I have that rule checked but I always format like this which keeps the rule from applying.

rsynek:
Thanks for your review, comments have been addressed with one exception - the usage of `bestScoreImprovement.getKey()` stays - see an explanation in the original comment.

ge0ffrey:
Thanks Radovan!

PR #532, opened 2019-08-13
Constraint streams

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -0,0 +1,201 @@
...
+            constraint.createNodes(buildPolicy, declaredClassToNodeMap, constraintWeight);
+        });
+        this.nodeOrderSize = buildPolicy.getNodeOrderMaximum() + 1;
+        constraintIdToScoringNodeMap = buildPolicy.getConstraintIdToScoringNodeMap();
+        effectiveClassToNodeListMap = new HashMap<>(declaredClassToNodeMap.size());
+        nodeOrderedQueueList = new ArrayList<>(nodeOrderSize);
+        for (int i = 0; i < nodeOrderSize; i++) {
+            nodeOrderedQueueList.add(new ArrayDeque<>(1000));
+        }
+        fromTupleListMap = new IdentityHashMap<>(1000);
```
triceo: A thousand as a default is a bit too much for my taste. Are you sure we don't just want to let this structure grow naturally?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -0,0 +1,201 @@
...
+
+    public List<BavetFromUniNode<Object>> findFromNodeList(Class<?> factClass) {
+        return effectiveClassToNodeListMap.computeIfAbsent(factClass, key -> {
+            List<BavetFromUniNode<Object>> nodeList = new ArrayList<>();
+            declaredClassToNodeMap.forEach((declaredClass, declaredNode) -> {
+                if (declaredClass.isAssignableFrom(factClass)) {
+                    nodeList.add(declaredNode);
+                }
+            });
+            return nodeList;
```
triceo: IMO another good candidate for a method reference.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetAbstractBiTuple.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.optaplanner.core.impl.score.stream.bavet.bi;
+
+import org.optaplanner.core.impl.score.stream.bavet.common.BavetAbstractTuple;
+
+public abstract class BavetAbstractBiTuple<A, B> extends BavetAbstractTuple {
+
+    @Override
+    public Object[] getFacts() {
+        return new Object[] {getFactA(), getFactB()};
```
triceo: How about memoization here? Or do you expect A/B to be changing?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetTuple.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Arrays;
+
+public interface BavetTuple {
+
+    BavetNode getNode();
+
+    Object[] getFacts();
+
+    default String getFactsString() {
+        return String.join(", ", Arrays.stream(getFacts())
```
triceo: You can do the same, and avoid creating an array inbetween, by using `Collectors.joining()` on the stream.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetGroupBridgeUniNode.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public String toString() {
+        return "GroupBridge()";
+    }
+
+    // ************************************************************************
+    // Getters/setters
+    // ************************************************************************
+
+    private class GroupKeyData {
```
triceo: This is a weird class. Doesn't seem to do anything. Holds data and possibly lets them be mutated from the outside. If this is actually being used anywhere, it looks like a `Map.Entry`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
```diff
@@ -0,0 +1,60 @@
...
+
+public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
+
+    private final List<SingleBiJoiner<A, B>> joinerList;
+
+    public CompositeBiJoiner(List<SingleBiJoiner<A, B>> joinerList) {
+        this.joinerList = joinerList;
+        if (joinerList.isEmpty()) {
+            throw new IllegalArgumentException("The joinerList (" + joinerList + ") must not be empty.");
+        }
```
triceo: Perhaps switch the store and the check?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
```diff
@@ -0,0 +1,60 @@
...
+    // ************************************************************************
+
+    @Override
+    public Function<A, Object[]> getLeftCombinedMapping() {
+        return buildCombinedMappingUni(joinerList, SingleBiJoiner::getLeftMapping);
+    }
+
+    @Override
+    public JoinerType[] getJoinerTypes() {
+        return joinerList.stream()
```
triceo: Assuming this is important for more than once, you might consider memoization.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/NoneBiJoiner.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.score.stream.bi;
+
+import java.util.function.Function;
+
+import org.optaplanner.core.impl.score.stream.common.JoinerType;
+
+public final class NoneBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
```
triceo: The things returned here might as well all be constants. (Assuming they are not mutated later.)

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -406,38 +441,40 @@ public ScoreDefinition buildDeprecatedScoreDefinition() {
...
         AbstractScoreDirectorFactory<Solution_> easyScoreDirectorFactory = buildEasyScoreDirectorFactory(solutionDescriptor);
+        AbstractScoreDirectorFactory<Solution_> constraintStreamScoreDirectorFactory = buildConstraintStreamScoreDirectorFactory(solutionDescriptor);
         AbstractScoreDirectorFactory<Solution_> incrementalScoreDirectorFactory = buildIncrementalScoreDirectorFactory(solutionDescriptor);
         AbstractScoreDirectorFactory<Solution_> droolsScoreDirectorFactory = buildDroolsScoreDirectorFactory(configContext, solutionDescriptor);
-        AbstractScoreDirectorFactory<Solution_> scoreDirectorFactory;
-        if (easyScoreDirectorFactory != null) {
+        if (Stream.of(easyScoreDirectorFactory, constraintStreamScoreDirectorFactory,
+                incrementalScoreDirectorFactory, droolsScoreDirectorFactory)
+                .filter(Objects::nonNull).count() > 1) {
+            List<String> scoreDirectorFactoryPropertyList = new ArrayList<>(4);
```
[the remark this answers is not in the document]
ge0ffrey: Yes, I just tried it and the code became less readable.
Because 4 will always be greater or equal to the count - and it's a temp list - I am leaving it as is.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -100,6 +100,11 @@
     public EntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
         this.solutionDescriptor = solutionDescriptor;
         this.entityClass = entityClass;
+        isInitializedPredicate = this::isInitialized;
+    }
+
+    public Predicate<Object> getIsInitializedPredicate() {
+        return isInitializedPredicate;
```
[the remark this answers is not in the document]
ge0ffrey: Yes, this is strange. It's due to node sharing: doing it the more obvious way creates 2 different lamdba instances.

I'll add a comment to point that out, good catch!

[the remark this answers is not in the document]
ge0ffrey:
```
    /**
     * Using entityDescriptor::isInitialized directly breaks node sharing
     * because it creates multiple instances of this {@link Predicate}.
     * @return never null, always the same {@link Predicate} instance to {@link #isInitialized(Object)}
     */
    public Predicate<Object> getIsInitializedPredicate() {
        return isInitializedPredicate;
    }
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreInliner.java
```diff
@@ -0,0 +1,116 @@
...
+            int levelWeight = constraintWeight.getHardOrSoftScore(singleLevel);
+            if (singleLevel < constraintWeight.getHardLevelsSize()) {
+                int level = singleLevel;
+                return (int matchWeight, Consumer<Score<?>> matchScoreConsumer) -> {
+                    int hardImpact = levelWeight * matchWeight;
+                    this.hardScores[level] += hardImpact;
+                    if (constraintMatchEnabled) {
+                        matchScoreConsumer.accept(BendableScore.ofHard(hardScores.length, softScores.length, level, hardImpact));
+                    }
+                    return () -> this.hardScores[level] -= hardImpact;
```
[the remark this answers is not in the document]
ge0ffrey: Ironically, it doesn't rely on the return value of `-=`:
it returns a Consumer lambda that does a `-=`, so:

```
                    return () -> {
                        this.softScores[level] -= softImpact;
                    };
```

Point proven though - this is not easy to read :)
My main concern is that constraint match totals got bolted on, causing the ScoreImpacter to have a matchScoreConsumer - this had a noticeable performance impact even when constraint matches are disabled. I suspect much of this code will change when I take a second stab at it. Right now, the behaviour is correct - but it's too complex.
I 've created a jira to simplify that complexity later: https://issues.jboss.org/browse/PLANNER-1578

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardmediumsoft/HardMediumSoftScoreInliner.java
```diff
@@ -0,0 +1,102 @@
...
+    public IntWeightedScoreImpacter buildWeightedScoreImpacter(HardMediumSoftScore constraintWeight) {
+        if (constraintWeight.equals(HardMediumSoftScore.ZERO)) {
+            throw new IllegalArgumentException("The constraintWeight (" + constraintWeight + ") cannot be zero,"
+                    + " this constraint should have been culled during node creation.");
+        }
+        int hardConstraintWeight = constraintWeight.getHardScore();
+        int mediumConstraintWeight = constraintWeight.getMediumScore();
+        int softConstraintWeight = constraintWeight.getSoftScore();
+        if (mediumConstraintWeight == 0 && softConstraintWeight == 0) {
+            return (int matchWeight, Consumer<Score<?>> matchScoreConsumer) -> {
```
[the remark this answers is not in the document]
ge0ffrey: Agreed that it's too hard to follow. I am not opening that pandora box for this PR (behaviour is correct), but I've created that jira (https://issues.jboss.org/browse/PLANNER-1578). If you want to take a stab at it, I can explain the need for the consumer in a meeting.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetTuple.java
ge0ffrey: Done, thanks :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetGroupBridgeUniNode.java
ge0ffrey: Oops - should have been deleted already :) Done

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
ge0ffrey: Agreed, done

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
ge0ffrey: Its only called once per scoredirector that is created. If we do do memoization, we should do it for getLeftCombinedMapping() etc too. I doubt it will make any noticable difference: for kogito-serverless it will happen only once per biJoiner (so no gain). For non-kogito, we're probably talking nanoseconds gain per solver run (= minutes).
Nice to have.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardsoftdouble/HardSoftDoubleScoreDefinition.java
```diff
@@ -68,6 +69,12 @@ public HardSoftDoubleScore fromLevelNumbers(int initScore, Number[] levelNumbers
         return HardSoftDoubleScore.ofUninitialized(initScore, (Double) levelNumbers[0], (Double) levelNumbers[1]);
     }
 
+    @Override
+    public ScoreInliner<HardSoftDoubleScore> buildScoreInliner(boolean constraintMatchEnabled) {
+        throw new IllegalStateException("ConstraintStreams don't support a " + HardSoftDoubleScore.class.getSimpleName()
```
triceo: Has `HardSoftDoubleScore` been `@Deprecated`? If so, perhaps it should be.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simpledouble/SimpleDoubleScoreDefinition.java
```diff
@@ -63,6 +64,13 @@ public SimpleDoubleScore fromLevelNumbers(int initScore, Number[] levelNumbers)
         return SimpleDoubleScore.ofUninitialized(initScore, (Double) levelNumbers[0]);
     }
 
+    @Override
+    public ScoreInliner<SimpleDoubleScore> buildScoreInliner(boolean constraintMatchEnabled) {
+        throw new IllegalStateException("ConstraintStreams don't support a " + SimpleDoubleScore.class.getSimpleName()
```
triceo: Ditto on deprecation.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardsoftdouble/HardSoftDoubleScoreDefinition.java
ge0ffrey: Agreed, nobody uses it, we don't want people to use it - it's just there so people can find it and figure out that they don't want to use it... Deprecated it.

ge0ffrey: Not adding that to this PR, tracking it here for later: https://issues.jboss.org/browse/PLANNER-1579

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/buildin/simple/SimpleScoreInlinerTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.score.inliner.IntWeightedScoreImpacter;
+import org.optaplanner.core.impl.score.inliner.UndoScoreImpacter;
+
+import static org.junit.Assert.*;
+
+public class SimpleScoreInlinerTest {
+
+    @Test
+    public void buildIntWeightedScoreImpacter() {
```
rsynek: The test is a bit difficult to follow, especially with variables like impacter1, impacter2.

Suggestions:
- splitting the test into two with self-explaining names?
- commenting the main steps?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simplebigdecimal/SimpleBigDecimalScoreInliner.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public SimpleBigDecimalScore extractScore(int initScore) {
+        return SimpleBigDecimalScore.ofUninitialized(initScore, score);
+    }
+
+    @Override
+    public String toString() {
+        return SimpleBigDecimalScore.class.getSimpleName() + " inliner";
+    }
```
triceo: We could save some repeating code if `toString()` moves to `ScoreInliner` and `SimpleBigDecimalScore.class` becomes `getClass()`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/DroolsScoreDirector.java
```diff
@@ -190,12 +148,24 @@ public boolean isConstraintMatchEnabled() {
...
+        return workingScoreHolder.getConstraintMatchTotalMap();
+    }
+
     @Override
     public Map<Object, Indictment> getIndictmentMap() {
         if (workingSolution == null) {
             throw new IllegalStateException(
                     "The method setWorkingSolution() must be called before the method getIndictmentMap().");
         }
+        // Notice that we don't trigger the variable listeners
```
triceo: Noticed. Perhaps you could also explain why it is important to notice that?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.optaplanner.core.impl.score.stream.bavet.BavetConstraintSession;
+
+public class BavetNodeBuildPolicy<Solution_> {
+
+    private final BavetConstraintSession session;
+
+    private int nodeOrderMaximum = 0;
```
triceo: This variable is never used in this class. It has a getter and a setter, but the class itself doesn't use it anywhere.
That suggests to me that, even though it may still be useful, it perhaps doesn't belong here?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetFilterTriConstraintStream.java
```diff
@@ -0,0 +1,55 @@
...
+public final class BavetFilterTriConstraintStream<Solution_, A, B, C> extends BavetAbstractTriConstraintStream<Solution_, A, B, C> {
+
+    private final TriPredicate<A, B, C> predicate;
+
+    public BavetFilterTriConstraintStream(BavetConstraint<Solution_> bavetConstraint, TriPredicate<A, B, C> predicate) {
+        super(bavetConstraint);
+        this.predicate = predicate;
+        if (predicate == null) {
+            throw new IllegalArgumentException("The predicate (null) cannot be null.");
+        }
```
triceo: Order of the check and the assignment could be switched here.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetFilterTriTuple.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final BavetFilterTriNode<A, B, C> node;
+    private final BavetAbstractTriTuple<A, B, C> parentTuple;
+
+    protected List<BavetAbstractTriTuple<A, B, C>> childTupleList = null;
+
+    public BavetFilterTriTuple(BavetFilterTriNode<A, B, C> node, BavetAbstractTriTuple<A, B, C> parentTuple) {
+        this.node = node;
+        this.parentTuple = parentTuple;
+        childTupleList = new ArrayList<>();
```
triceo: Any reason why this is not initialized directly in-line on line 27?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetJoinTriTuple.java
```diff
@@ -0,0 +1,89 @@
...
+    private final BavetJoinBridgeUniTuple<C> cTuple;
+
+    protected List<BavetAbstractTriTuple<A, B, C>> childTupleList = null;
+
+    public BavetJoinTriTuple(BavetJoinTriNode<A, B, C> node,
+            BavetJoinBridgeBiTuple<A, B> abTuple, BavetJoinBridgeUniTuple<C> cTuple) {
+        this.node = node;
+        this.abTuple = abTuple;
+        this.cTuple = cTuple;
+        childTupleList = new ArrayList<>();
```
triceo: Could be initialized in-line on line 33.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFilterUniConstraintStream.java
```diff
@@ -0,0 +1,56 @@
...
+public final class BavetFilterUniConstraintStream<Solution_, A> extends BavetAbstractUniConstraintStream<Solution_, A> {
+
+    private final Predicate<A> predicate;
+
+    public BavetFilterUniConstraintStream(BavetConstraint<Solution_> bavetConstraint, Predicate<A> predicate) {
+        super(bavetConstraint);
+        this.predicate = predicate;
+        if (predicate == null) {
+            throw new IllegalArgumentException("The predicate (null) cannot be null.");
+        }
```
triceo: Possibly switch the assignment with the check.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFilterUniTuple.java
```diff
@@ -0,0 +1,64 @@
...
+    private final BavetFilterUniNode<A> node;
+    private final BavetAbstractUniTuple<A> parentTuple;
+
+    protected List<BavetAbstractUniTuple<A>> childTupleList;
+
+    public BavetFilterUniTuple(BavetFilterUniNode<A> node, BavetAbstractUniTuple<A> parentTuple,
+            int childTupleListSize) {
+        this.node = node;
+        this.parentTuple = parentTuple;
+        childTupleList = new ArrayList<>();
```
triceo: Could be initialized in-line on line 27.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFromUniConstraintStream.java
```diff
@@ -0,0 +1,68 @@
...
+public final class BavetFromUniConstraintStream<Solution_, A> extends BavetAbstractUniConstraintStream<Solution_, A> {
+
+    private final Class<A> fromClass;
+
+    public BavetFromUniConstraintStream(BavetConstraint<Solution_> bavetConstraint, Class<A> fromClass) {
+        super(bavetConstraint);
+        this.fromClass = fromClass;
+        if (fromClass == null) {
+            throw new IllegalArgumentException("The fromClass (null) cannot be null.");
+        }
```
triceo: Ditto (order switch).

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFromUniTuple.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private final BavetFromUniNode<A> node;
+    private final A factA;
+
+    protected List<BavetAbstractUniTuple<A>> childTupleList;
+
+    public BavetFromUniTuple(BavetFromUniNode<A> node, A factA, int childTupleListSize) {
+        this.node = node;
+        this.factA = factA;
+        childTupleList = new ArrayList<>();
```
triceo: Ditto (init on line 27)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/AbstractBiJoiner.java
```diff
@@ -0,0 +1,54 @@
...
+                // Ignore it
+            } else if (joiner instanceof SingleBiJoiner) {
+                joinerList.add((SingleBiJoiner<A, B>) joiner);
+            } else if (joiner instanceof CompositeBiJoiner) {
+                joinerList.addAll(((CompositeBiJoiner<A, B>) joiner).getJoinerList());
+            } else {
+                throw new IllegalArgumentException("The joiner class (" + joiner.getClass() + ") is not supported.");
+            }
+        }
+        if (joinerList.size() == 0) {
```
triceo: `isEmpty()` may be better here. Alternatively, cache the `size()` and reuse here and in the condition below.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/tri/AbstractTriJoiner.java
```diff
@@ -0,0 +1,55 @@
...
+                // Ignore it
+            } else if (joiner instanceof SingleTriJoiner) {
+                joinerList.add((SingleTriJoiner<A, B, C>) joiner);
+            } else if (joiner instanceof CompositeTriJoiner) {
+                joinerList.addAll(((CompositeTriJoiner<A, B, C>) joiner).getJoinerList());
+            } else {
+                throw new IllegalArgumentException("The joiner class (" + joiner.getClass() + ") is not supported.");
+            }
+        }
+        if (joinerList.size() == 0) {
```
triceo: Same `size()`-related comment as with the BiJoiner.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/tri/CompositeTriJoiner.java
```diff
@@ -0,0 +1,61 @@
...
+
+public final class CompositeTriJoiner<A, B, C> extends AbstractTriJoiner<A, B, C> {
+
+    private final List<SingleTriJoiner<A, B, C>> joinerList;
+
+    public CompositeTriJoiner(List<SingleTriJoiner<A, B, C>> joinerList) {
+        this.joinerList = joinerList;
+        if (joinerList.isEmpty()) {
+            throw new IllegalArgumentException("The joinerList (" + joinerList + ") must not be empty.");
+        }
```
triceo: Possibly throw before assigning.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -839,6 +840,9 @@ public MemberAccessor getConstraintConfigurationMemberAccessor() {
         return constraintConfigurationMemberAccessor;
     }
 
+    /**
+     * @return sometimes null
```
triceo: Perhaps there would be a better description than `sometimes`?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendablelong/BendableLongScoreInliner.java
```diff
@@ -0,0 +1,116 @@
...
+        softScores = new long[softLevelsSize];
+    }
+
+    @Override
+    public LongWeightedScoreImpacter buildWeightedScoreImpacter(BendableLongScore constraintWeight) {
+        if (constraintWeight.equals(BendableLongScore.zero(hardScores.length, softScores.length))) {
+            throw new IllegalArgumentException("The constraintWeight (" + constraintWeight + ") cannot be zero,"
+                    + " this constraint should have been culled during node creation.");
+        }
+        Integer singleLevel = null;
```
triceo: This code probably deserves a couple comments. Took me 5 minutes to figure out how and why this ever becomes null and non-null, and what needs to happen afterwards.

The need for comments could be slightly reduced, if in the condition on line 54, the different branches were converted into descriptively named methods.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetScoringUniConstraintStream.java
```diff
@@ -0,0 +1,169 @@
...
+        this.longMatchWeigher = longMatchWeigher;
+        this.bigDecimalMatchWeigher = bigDecimalMatchWeigher;
+    }
+
+    // ************************************************************************
+    // Node creation
+    // ************************************************************************
+
+    @Override
+    protected BavetScoringUniNode<A> createNode(BavetNodeBuildPolicy<Solution_> buildPolicy,
```
rsynek: The same comment as for the `BavetScoringTriNode`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetGroupBridgeUniNode.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.optaplanner.core.api.score.stream.uni.UniConstraintCollector;
+import org.optaplanner.core.impl.score.stream.bavet.BavetConstraintSession;
+import org.optaplanner.core.impl.score.stream.bavet.bi.BavetGroupBiNode;
+import org.optaplanner.core.impl.score.stream.bavet.bi.BavetGroupBiTuple;
+import org.optaplanner.core.impl.score.stream.bavet.common.BavetTupleState;
+
+public final class BavetGroupBridgeUniNode<A, GroupKey_, ResultContainer_, Result_> extends BavetAbstractUniNode<A> {
+
+    private final BavetAbstractUniNode<A> parentNode;
```
rsynek: Unused

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetScoringBiConstraintStream.java
```diff
@@ -0,0 +1,166 @@
...
+        this.longMatchWeigher = longMatchWeigher;
+        this.bigDecimalMatchWeigher = bigDecimalMatchWeigher;
+    }
+
+    // ************************************************************************
+    // Node creation
+    // ************************************************************************
+
+    @Override
+    protected BavetScoringBiNode<A, B> createNode(BavetNodeBuildPolicy<Solution_> buildPolicy,
```
rsynek: The same comment as for the same method across other `BavetScoring*Node` classes

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -0,0 +1,126 @@
...
+        this.solutionDescriptor = solutionDescriptor;
+    }
+
+    @Override
+    public Constraint newConstraint(String constraintName) {
+        return newConstraint(null, constraintName);
+    }
+
+    @Override
+    public Constraint newConstraint(String constraintPackage, String constraintName) {
```
rsynek: This method does not seem to be covered by unit tests. As the examples do not have constraint streams turned on by default, it's not tested at all from the perspective of regression testing.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/Joiners.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import org.optaplanner.core.api.score.stream.bi.BiJoiner;
+import org.optaplanner.core.api.score.stream.tri.TriJoiner;
+import org.optaplanner.core.impl.score.stream.bi.SingleBiJoiner;
+import org.optaplanner.core.impl.score.stream.common.JoinerType;
+import org.optaplanner.core.impl.score.stream.tri.SingleTriJoiner;
+
+public final class Joiners {
```
rsynek: public API, should include Javadoc

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
```diff
@@ -0,0 +1,194 @@
...
+                CountDistinctResultContainer::new,
+                (resultContainer, a) -> {
+                    Object value = groupValueMapping.apply(a);
+                    long[] objectCount = resultContainer.objectCountMap.computeIfAbsent(value, k -> new long[1]);
+                    if (objectCount[0] == 0L) {
+                        resultContainer.count++;
+                    }
+                    objectCount[0]++;
+                    return (() -> {
+                        long[] objectCount2 = resultContainer.objectCountMap.get(value);
```
rsynek: poor variable name

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
```diff
@@ -0,0 +1,194 @@
...
+                },
+                resultContainer -> resultContainer.count);
+    }
+
+    public static <A> UniConstraintCollector<A, ?, Long> countDistinctLong(Function<A, ?> groupValueMapping) {
+        class CountDistinctResultContainer {
+            long count = 0L;
+            Map<Object, long[]> objectCountMap = new HashMap<>();
+        }
+        return new UniConstraintCollector<>(
```
rsynek: A constructor call, that passes there a method reference, a lambda containing a few more lambdas and as a last argument, another lambda, this time a simple one. This really does not read well, does it?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.function.Function;
+import java.util.function.ToIntBiFunction;
+import java.util.function.ToIntFunction;
+import java.util.function.ToLongBiFunction;
+import java.util.function.ToLongFunction;
+
+import org.optaplanner.core.api.score.stream.bi.BiConstraintCollector;
+import org.optaplanner.core.api.score.stream.uni.UniConstraintCollector;
+
+/**
```
rsynek: empty javadoc

rsynek:
Done with the review. A couple of remarks:
- please give a priority to the comments in *.api.* packages
- I didn't strive to completely understand individual pieces of implementation code, nor a connection between multiple implementation classes.
- in my opinion merging will introduce a significant amount of technical debt we should plan to address

optaplanner-core/src/main/java/org/optaplanner/core/api/function/TriPredicate.java
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.optaplanner.core.api.function;
+
+import java.util.Objects;
+import java.util.function.Predicate;
+
+/**
+ * Represents a predicate (boolean-valued function) of three arguments.  This is
```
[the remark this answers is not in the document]
ge0ffrey: fixed locally

optaplanner-examples/src/main/java/org/optaplanner/examples/cloudbalancing/optional/score/CloudBalancingConstraintProvider.java
```diff
@@ -0,0 +1,79 @@
...
+import org.optaplanner.core.api.score.stream.ConstraintFactory;
+import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.examples.cloudbalancing.domain.CloudProcess;
+
+import static org.optaplanner.core.api.score.stream.common.ConstraintCollectors.*;
+
+public class CloudBalancingConstraintProvider implements ConstraintProvider {
+
+    // WARNING: The ConstraintStreams/ConstraintProvider API is TECH PREVIEW.
+    // It is stable but it has many API gaps.
```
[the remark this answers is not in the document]
ge0ffrey: Pushing the commit in a few minutes that turns this into:


    // WARNING: The ConstraintStreams/ConstraintProvider API is TECH PREVIEW.
    // It works but it has many API gaps.
    // Therefore, it is not rich enough yet to handle complex constraints.

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProvider.java
```diff
@@ -0,0 +1,70 @@
...
+        descendingDiagonalConflict(constraintFactory);
+    }
+
+    protected void horizontalConflict(ConstraintFactory constraintFactory) {
+        Constraint c = constraintFactory.newConstraintWithWeight(
+                "Horizontal conflict", SimpleScore.of(1));
+        c.fromUniquePair(Queen.class,
+                equal(Queen::getRowIndex)
+        ).penalize();
+        // fromUniquePair() is syntactic sugar for from().join(..., lessThan(getId())
```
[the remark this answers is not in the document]
ge0ffrey: Yes, I do, for tutorial/learning purposes. I like having fromUniquePair(), but it's just a facade method.
@triceo Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirector.java
```diff
@@ -0,0 +1,179 @@
...
+    public void close() {
+        super.close();
+        session = null;
+    }
+
+    // ************************************************************************
+    // Entity/variable add/change/remove methods
+    // ************************************************************************
+
+    // public void beforeEntityAdded(EntityDescriptor entityDescriptor, Object entity) // Do nothing
```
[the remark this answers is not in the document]
ge0ffrey: Yes, because DroolsScoreDirector does it too :)
I mean - because to make it clear we're not doing anything in the before method :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simplelong/SimpleLongScoreInliner.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public LongWeightedScoreImpacter buildWeightedScoreImpacter(SimpleLongScore constraintWeight) {
+        if (constraintWeight.equals(SimpleLongScore.ZERO)) {
+            throw new IllegalArgumentException("The constraintWeight (" + constraintWeight + ") cannot be zero,"
+                    + " this constraint should have been culled during node creation.");
+        }
+        long simpleConstraintWeight = constraintWeight.getScore();
+        return (long matchWeight, Consumer<Score<?>> matchScoreConsumer) -> {
```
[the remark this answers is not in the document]
ge0ffrey: Ok, I believe you might have fixed this meanwhile? or it's in progress? In any case, this is related https://issues.jboss.org/browse/PLANNER-1578

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/DroolsScoreDirector.java
ge0ffrey: Yes, I agree. But only my former self currently knows why... (which really proves your point).
This is not an issue introduced in this PR.
Deferred into https://issues.jboss.org/browse/PLANNER-1584

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendablebigdecimal/BendableBigDecimalScoreInliner.java
```diff
@@ -0,0 +1,119 @@
...
+    public BendableBigDecimalScoreInliner(boolean constraintMatchEnabled, int hardLevelsSize, int softLevelsSize) {
+        super(constraintMatchEnabled);
+        hardScores = new BigDecimal[hardLevelsSize];
+        Arrays.fill(hardScores, BigDecimal.ZERO);
+        softScores = new BigDecimal[softLevelsSize];
+        Arrays.fill(softScores, BigDecimal.ZERO);
+    }
+
+    @Override
+    public BigDecimalWeightedScoreImpacter buildWeightedScoreImpacter(BendableBigDecimalScore constraintWeight) {
```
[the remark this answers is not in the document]
ge0ffrey: Deferred into https://issues.jboss.org/browse/PLANNER-1584

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreInliner.java
```diff
@@ -0,0 +1,118 @@
...
+        for (int i = 0; i < constraintWeight.getLevelsSize(); i++) {
+            if (constraintWeight.getHardOrSoftScore(i) != 0L) {
+                if (singleLevel != null) {
+                    singleLevel = null;
+                    break;
+                }
+                singleLevel = i;
+            }
+        }
+        if (singleLevel != null) {
```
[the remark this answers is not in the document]
ge0ffrey: Deferred into https://issues.jboss.org/browse/PLANNER-1584

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/buildin/simple/SimpleScoreInlinerTest.java
ge0ffrey: Deferred into https://issues.jboss.org/browse/PLANNER-1584

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
ge0ffrey: After building the nodes, we need the nodeOrderMaximum for performance (list capacity). The BuildPolicy is there during building all nodes, so it's easy to just check every nodeOrder number to figure out the largest.
That being said, we could also just look at the constraint list - after building the nodes - and grab the largest there. Although Constraint's don't know their ScoringStream as a parent...

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetFilterTriConstraintStream.java
ge0ffrey: You mean "if (null == predicate)"?
The rest of optaplanner doesn't follow that convention.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetFilterTriTuple.java
ge0ffrey: The goal is to get a good starting capacity for these array lists, for performance reasons,
by using the childNodeList.size() of the node, multiplied by a good constant.
But there's another perf improvement that will heavily impact this.
Anyway, for now I added in createTuple() methods:
        // TODO Use childNodeList.size() to the tuple's childTupleList's capacity

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/AbstractBiJoiner.java
ge0ffrey: Good catch. Fixed locally.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/tri/CompositeTriJoiner.java
ge0ffrey: Fixed locally

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
ge0ffrey: There are basically 2 levels of javadocs:
- Full Javadocs (needed on at least all api package public methods and classes - in theory)
- NonNull Javadocs. Like in Kotlin, it should be clear which methods might return null and which will never return null. The words "never null" and "sometimes null" are used, as a stand in for @NonNull annotations and the like, once we become kotlin-friendly (and java IDE-friendly). Once we use those annotations, those javadocs won't mention it any more.

This impl method falls into the second category - but it really clearly needs to point out that it can return null sometimes.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendablelong/BendableLongScoreInliner.java
ge0ffrey: Deferring to https://issues.jboss.org/browse/PLANNER-1584 - I suspect that PLANNER-1578 might influence this.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintCollector.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.optaplanner.core.api.score.stream.bi;
+
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import org.optaplanner.core.api.function.TriFunction;
+
+public final class BiConstraintCollector<A, B, ResultContainer_, Result_> {
```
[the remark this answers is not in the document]
ge0ffrey: Very good catch. It shouldn't be par of the API :)

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
```diff
@@ -0,0 +1,480 @@
...
+        super(constraintMatchEnabled);
+    }
+
+    // ************************************************************************
+    // Filter
+    // ************************************************************************
+
+    @Test
+    public void filter_entity() {
+        TestdataLavishSolution solution = TestdataLavishSolution.generateSolution(1, 0, 1, 0);
```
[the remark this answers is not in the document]
ge0ffrey: I wanted to that originally, but notice that the generateSolutions parameter change from test method to test method. Having it behind a generateSolution already helps a lot.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
```diff
@@ -0,0 +1,480 @@
...
+        scoreDirector.setWorkingSolution(solution);
+        assertEquals(SimpleBigDecimalScore.of(new BigDecimal("2.4")), scoreDirector.calculateScore());
+    }
+
+    // ************************************************************************
+    // Combinations
+    // ************************************************************************
+
+//    @Test
+//    public void globalNodeOrder() {
```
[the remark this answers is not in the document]
ge0ffrey: Changed into

    @Test @Ignore("Not yet implemented") // TODO
    public void globalNodeOrder() {

    }

:)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/JoinerType.java
```diff
@@ -0,0 +1,44 @@
...
+    LESS_THAN,
+    LESS_THAN_OR_EQUAL,
+    GREATER_THAN,
+    GREATER_THAN_OR_EQUAL,
+    CONTAINING,
+    INTERSECTING,
+    DISJOINT;
+
+    public JoinerType flip() {
+        switch (this) {
```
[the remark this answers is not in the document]
ge0ffrey: EQUAL too

Changes exception to:

                throw new IllegalStateException("The joinerType (" + this + ") cannot be flipped.");

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFilterUniTuple.java
```diff
@@ -0,0 +1,64 @@
...
+
+public final class BavetFilterUniTuple<A> extends BavetAbstractUniTuple<A> {
+
+    private final BavetFilterUniNode<A> node;
+    private final BavetAbstractUniTuple<A> parentTuple;
+
+    protected List<BavetAbstractUniTuple<A>> childTupleList;
+
+    public BavetFilterUniTuple(BavetFilterUniNode<A> node, BavetAbstractUniTuple<A> parentTuple,
+            int childTupleListSize) {
```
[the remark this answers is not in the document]
ge0ffrey: Fixed locally

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetScoringTriConstraintStream.java
```diff
@@ -0,0 +1,166 @@
...
+        this.longMatchWeigher = longMatchWeigher;
+        this.bigDecimalMatchWeigher = bigDecimalMatchWeigher;
+    }
+
+    // ************************************************************************
+    // Node creation
+    // ************************************************************************
+
+    @Override
+    protected BavetScoringTriNode<A, B, C> createNode(BavetNodeBuildPolicy<Solution_> buildPolicy,
```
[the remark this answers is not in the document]
ge0ffrey: Deferred to https://issues.jboss.org/browse/PLANNER-1584

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetJoinTriNode.java
```diff
@@ -0,0 +1,161 @@
...
+                + ") can't have a parentTuple (" + parentTuple + ");");
+    }
+
+    public BavetJoinTriTuple<A, B, C> createTuple(
+            BavetJoinBridgeBiTuple<A, B> abTuple, BavetJoinBridgeUniTuple<C> cTuple) {
+        return new BavetJoinTriTuple<>(this, abTuple, cTuple);
+    }
+
+    public void refresh(BavetJoinTriTuple<A, B, C> tuple) {
+        A a = tuple.getFactA();
```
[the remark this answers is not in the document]
ge0ffrey: Fixed locally

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/tri/TriConstraintStreamTest.java
```diff
@@ -0,0 +1,341 @@
...
+    public TriConstraintStreamTest(boolean constraintMatchEnabled) {
+        super(constraintMatchEnabled);
+    }
+
+    // ************************************************************************
+    // Filter
+    // ************************************************************************
+
+    @Test
+    public void filter_entity() {
```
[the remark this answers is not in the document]
ge0ffrey: I don't really see how that would work. Would you mind doing a proof of concept of such a test? Then let's meet up about it.

Having good test coverage of the ConstraintStreams is critical. Whatever we can do to make them more maintanable - without loosing reliability is welcome. I prefer avoiding mocks if we can use the real thing.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetGroupBridgeUniNode.java
ge0ffrey: All nodes know their parent currently (just like the tuples know their parents). That's by design and for consistently currently. Once this work fleshes out and this turns out to be unneeded - we can discuss removing them.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintFactory.java
ge0ffrey: This method is now gone due the side-effects refactor,
but the Conference Scheduling example will test the equivalent method.

ge0ffrey: Ah yes, you're right. We need unit tests for constraint configuration in optaplanner-core...

ge0ffrey: Put in https://issues.jboss.org/browse/PLANNER-1584, at the top.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/Joiners.java
ge0ffrey: Class javadoc done, rest deferred into https://issues.jboss.org/browse/PLANNER-1584 but I 'll try to do those later this week already.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
ge0ffrey: Any suggestions? :)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
ge0ffrey: No. It's a poor excuse, but this is from OpenJDK:

    static class CollectorImpl<T, A, R> implements Collector<T, A, R> {
        private final Supplier<A> supplier;
        private final BiConsumer<A, T> accumulator;
        private final BinaryOperator<A> combiner;
        private final Function<A, R> finisher;
        private final Set<Characteristics> characteristics;

        CollectorImpl(Supplier<A> supplier,
                      BiConsumer<A, T> accumulator,
                      BinaryOperator<A> combiner,
                      Function<A,R> finisher,
                      Set<Characteristics> characteristics) {
            this.supplier = supplier;
            this.accumulator = accumulator;
            this.combiner = combiner;
            this.finisher = finisher;
            this.characteristics = characteristics;
        }

How can we improve our naming?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
ge0ffrey: Fixed locally

ge0ffrey:
I've pushed the side-effects change, the PR improvements that aren't deferred to https://issues.jboss.org/browse/PLANNER-1584 (= the technical debt we should plan to address). I am currently working on not letting the ConstraintCollector escape into the wild in it's current form.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/common/ConstraintCollectors.java
rsynek: `updatedObjectCount` vs. `retrievedObjectCount`?

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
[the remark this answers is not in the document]
rsynek: ok

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetFilterTriConstraintStream.java
rsynek: No, if I understand it correctly, it's about checking for null first and doing the assignment `this.predicate = predicate;` just after that.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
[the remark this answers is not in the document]
rsynek: As far as I can tell, the generateSolution is called only with following parameters: `TestdataLavishSolution.generateSolution(1, 0, 1, 0)`
Perhaps we should add it on the list and look into possible refactoring later.

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProvider.java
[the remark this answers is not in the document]
rsynek: Maybe a wild idea: how about describing all the interesting nuances in a blog post and putting a link for the post here in the exaple?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirector.java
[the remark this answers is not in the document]
rsynek: ok

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simplelong/SimpleLongScoreInliner.java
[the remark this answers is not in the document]
rsynek: Not yet, added a note to the Jira in order not to forget about it.

ge0ffrey:
Thanks Lukas and Radovan for approving the draft.
 I've added the missing javadocs moved Joiners and ConstraintCollectors out the "common" package, removing that.
It's ready to be merged now I believe, presuming those last few commits are fine.
@rsynek you want to do the honors of merging?

rsynek:
jenkins execute full downstream build

rsynek:
> Thanks Lukas and Radovan for approving the draft.
> I've added the missing javadocs moved Joiners and ConstraintCollectors out the "common" package, removing that.
> It's ready to be merged now I believe, presuming those last few commits are fine.
> @rsynek you want to do the honors of merging?

The new commits look good. Running full downstream build to make sure we haven't broken anything.

rsynek:
The two failures in full downstream build are not connected to OptaPlanner. Merging.

PR #534, opened 2019-08-22
Address subpillar performance regression (follow-up to PLANNER-1565)

triceo:
@rsynek's benchmarks have shown that subpillars have actually regressed in raw Java performance. For that reason, I have gone back and [ran the new algorithm in JMH against its old implementation](https://docs.google.com/spreadsheets/d/1Szqv2bMRZaker15v7TtU7aj2q7XQ2Gs2cXttwMNFHnY/edit#gid=0) in a very wide range of circumstances.

- The new implementation (`STREAM+SHORTCUT`) is indeed better than the original loop (`ORIGINAL`) in a lot of places.
- However, this new data also shows that it is **only better in places where the code uses the shortcuts** [1]. The stream actually sucks in places where the shortcuts are not in effect. [2]
- Therefore we take the original loop, apply the same shortcuts, and we get the best results (`LOOP+SHORTCUT`).

I am submitting this PR, which implements `LOOP+SHORTCUT`. It is **the best solution in literally every circumstance tried**. Now that we have an algorithm that performs better even at raw Java level, we **should be seeing further score improvements**.

[1] It delivers a better score, though - which was the only thing that we actually measure via the Benchmarker. We did not actually measure its raw Java performance. 

[2] The stream suffers from generating *a lot* of random numbers. The more it generates, the harder it becomes to generate new distinct ones. Therefore, the stream solution begins to suck more and more as the `subPillarSize` starts reaching `basePillarSize`. The effect eventually becomes so pronounced that even the `basePillar.toArray()` copying isn't as big of a deal.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
@@ -24,8 +24,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 import org.optaplanner.core.config.heuristic.selector.common.SelectionCacheType;
 import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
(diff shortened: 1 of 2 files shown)


triceo:
@ge0ffrey Let's try to get this in before the Friday cut-off.

rsynek:
@ge0ffrey could you please quickly review and merge? Would be good to get it into this release.

ge0ffrey:
Thanks guys (both for the regression perf test proof and the adjustement of the code - even better than the original code).

PR #537, opened 2019-09-02
PLANNER-1586 Mention classloader in reflection-related exceptions

triceo:
This gives a useful hint, but that hint may not be applicable at all times. [See here](https://stackoverflow.com/questions/57711906/redhat-decision-manager-7-3-opta-cloud-solver-example-not-working) for a situation where the hint would mislead people - therefore I am not adding the same message on setters.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LambdaBeanPropertyMemberAccessor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LambdaBeanPropertyMemberAccessor.java
@@ -102,10 +102,7 @@ public final class LambdaBeanPropertyMemberAccessor implements MemberAccessor {
                     MethodType.methodType(void.class, Object.class, Object.class),
                     lookup.findVirtual(declaringClass, setterMethod.getName(), MethodType.methodType(void.class, propertyType)),
                     MethodType.methodType(void.class, declaringClass, propertyType));
-        } catch (IllegalAccessException e) {
-            throw new IllegalStateException("Lambda creation failed for setterMethod (" + setterMethod + "). You may " +
-                    "consider supplying a ClassLoader parameter to your SolverFactory.create...() method call.", e);
-        } catch (LambdaConversionException | NoSuchMethodException e) {
+        } catch (LambdaConversionException | NoSuchMethodException | IllegalAccessException e) {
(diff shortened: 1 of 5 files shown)


triceo:
@ge0ffrey Would you mind reviewing?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/LambdaBeanPropertyMemberAccessor.java
```diff
@@ -76,7 +76,10 @@ private Function createGetterFunction(MethodHandles.Lookup lookup) {
                     MethodType.methodType(Object.class, Object.class),
                     lookup.findVirtual(declaringClass, getterMethod.getName(), MethodType.methodType(propertyType)),
                     MethodType.methodType(propertyType, declaringClass));
-        } catch (LambdaConversionException | NoSuchMethodException | IllegalAccessException e) {
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException("Lambda creation failed for getterMethod (" + getterMethod + "). You may " +
+                    " consider supplying a ClassLoader parameter to your SolverFactory.create...() method call.", e);
```
ge0ffrey: Good change, this will help a lot of spring users. One thing for consistency:
For fail fast exception messages, optaplanner-core puts suggestions or conjectures on a second line (so after a newline) starting with the word "Maybe" (even if it's 90% sure) to signal that that could solve the user's problem but we're not a 100% sure.
Also nice to have is active voice (remove "consider", remove "your"), see https://developers.google.com/style/voice

In this case that would give:

            throw new IllegalStateException("Lambda creation failed for getterMethod (" + getterMethod + ").\n" +
                    "Maybe add getClass().getClassLoader() as a parameter to the " + SolverFactory.class.getSimpleName() + ".create...() method call.", e);

Thinking aloud: Should we mention "spring boot" in this exception message explicitly? Probably not - if a user runs into this - interrupting their flow - the maybe will set them back on track in a minute - (much better than the hours they loose now).

triceo:
@ge0ffrey Reworded. 

(Also extracted the wording to a constant, as it's too much to keep duplicated everywhere. My guess is that you won't like it though.)

ge0ffrey:
@triceo You're right, I don't like the constant :) Probably because I look at code inside the fail-fast if condition differently than the rest of the code. But it makes sense, it reduces DRY, I can't find any good reason not to do it, so it's merged. Thanks Lukas!

PR #536, opened 2019-08-27
Update to TspWorldPanel.java to be Mac friendly

JimTyrrell:
Simple change to enable control clicks to work on Macs.  OptaPlanner enablement is fun.

PR #538, opened 2019-09-03
PLANNER-1954 TSP example support for right-click emulation on Mac OS X

triceo:
Replaces #536 

The old code also used to react on middle button. (And that action was the same as right button.) In the interest of clarity, and because we do not know how middle buttons work on Mac, I have removed the middle button functionality.

The resulting code was tested on Linux. Left and right buttons' behavior remains unchanged.


triceo:
@ge0ffrey This is ready.

PR #539, opened 2019-09-03
PLANNER-1595 Constraint stream implementation for Task Assignment

triceo:
The following results were achieved in a 5-minute run on `500tasks-20employees` with `FULL_ASSERT`:

- DRL score: 0hard/-49205soft/... with 15701 steps.
- Constraint streams score: 0hard/-48844soft/... with 33313 steps.

**Constraint streams appears to be a clear winner here.** However, I am leaving a couple comments in the code where **we could still improve the API** for even better results.

@ge0ffrey @rsynek Please review.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
@@ -1,0 +1,80 @@
+package org.optaplanner.examples.taskassigning.solver.score;
+
+import org.optaplanner.core.api.score.buildin.bendable.BendableScore;
+import org.optaplanner.core.api.score.stream.Constraint;
+import org.optaplanner.core.api.score.stream.ConstraintFactory;
+import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.core.api.score.stream.uni.UniConstraintStream;
+import org.optaplanner.examples.taskassigning.domain.Priority;
+import org.optaplanner.examples.taskassigning.domain.Task;
+
+public final class TaskAssigningConstraintProvider implements ConstraintProvider {
+
+    private static final int BENDABLE_SCORE_HARD_LEVELS_SIZE = 1;
+    private static final int BENDABLE_SCORE_SOFT_LEVELS_SIZE = 4;
+
+    private static UniConstraintStream<Task> getInitializedTaskUniStream(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Task.class)
+                .filter(task -> task.getEmployee() != null);
+    }
+
+    private static Constraint noMissingSkills(ConstraintFactory constraintFactory) {
+        return getInitializedTaskUniStream(constraintFactory)
+                .filter(task -> task.getMissingSkillCount() > 0)
+                .penalize("No missing skills",
+                        BendableScore.ofHard(BENDABLE_SCORE_HARD_LEVELS_SIZE, BENDABLE_SCORE_SOFT_LEVELS_SIZE, 0, 1),
+                        Task::getMissingSkillCount);
+    }
+
+    private static Constraint criticalPriorityBasedTaskEndTime(ConstraintFactory constraintFactory) {
+        return getInitializedTaskUniStream(constraintFactory)
+                .filter(task -> task.getPriority() == Priority.CRITICAL)
+                .penalize("Critical priority task end time",
+                        BendableScore.ofSoft(BENDABLE_SCORE_HARD_LEVELS_SIZE, BENDABLE_SCORE_SOFT_LEVELS_SIZE, 0, 1),
+                        Task::getEndTime);
+    }
+
+    private static Constraint minimizeMakespan(ConstraintFactory constraintFactory) {
+        return getInitializedTaskUniStream(constraintFactory)
+                .filter(task -> task.getNextTask() == null)
+                .penalize("Minimize makespan, latest ending employee first",
+                        BendableScore.ofSoft(BENDABLE_SCORE_HARD_LEVELS_SIZE, BENDABLE_SCORE_SOFT_LEVELS_SIZE, 1, 1),
+                        task -> task.getEndTime() * task.getEndTime());
+        /*
+         * TODO ^^^ Some constraints need exponential weighting. perhaps we should have dynamic
+         *  constraint weights, inferred from the entity?
+         */
+    }
+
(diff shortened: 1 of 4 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+    private static Constraint minorPriorityTaskEndTime(ConstraintFactory constraintFactory) {
+        return getInitializedTaskUniStream(constraintFactory)
+                .filter(task -> task.getPriority() == Priority.MINOR)
+                .penalize("Minor priority task end time",
+                        BendableScore.ofSoft(BENDABLE_SCORE_HARD_LEVELS_SIZE, BENDABLE_SCORE_SOFT_LEVELS_SIZE, 3, 1),
+                        Task::getEndTime);
+    }
+
+    @Override
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
```
ge0ffrey: Can we put this method first like in the other Providers? (Or move it last in all other providers - but I prefer having it first, it's a good TOC)

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private static Constraint minimizeMakespan(ConstraintFactory constraintFactory) {
+        return getInitializedTaskUniStream(constraintFactory)
+                .filter(task -> task.getNextTask() == null)
+                .penalize("Minimize makespan, latest ending employee first",
+                        BendableScore.ofSoft(BENDABLE_SCORE_HARD_LEVELS_SIZE, BENDABLE_SCORE_SOFT_LEVELS_SIZE, 1, 1),
+                        task -> task.getEndTime() * task.getEndTime());
+        /*
+         * TODO ^^^ Some constraints need exponential weighting. perhaps we should have dynamic
+         *  constraint weights, inferred from the entity?
```
ge0ffrey: besides .penalize(), we 'll also need to support .penalizeLoadBalanceFairness() or something like that.
The squared technique isn't even watertight.
See https://www.optaplanner.org/blog/2017/02/03/FormulaForMeasuringUnfairness.html and tennis example.

ge0ffrey: @triceo It's really hard for people to understand that blog post, but they do need exactly that. So if we build a .penalizeUnfairness() or something like that - that follows the conclusions of that blog post - than we take that pain instead of them.

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
triceo: @ge0ffrey I agree, but don't we have something about this in code style? It's at the end, because that's where IDE auto-placed it.

(The other methods are static, and therefore go above.)

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
+        return new Constraint[]{
+                noMissingSkills(constraintFactory),
+                minimizeMakespan(constraintFactory),
+                /*
+                 * TODO if we had a way of determining constraint weight from the entity, different weight for different
+                 *  priority, then the following 3 constraints could have been one constraint, with associated
+                 *  performance benefits.
+                 */
+                criticalPriorityBasedTaskEndTime(constraintFactory),
```
triceo: @ge0ffrey What do you think about this comment? This really seems like a useful addition to the API.

Instead of three constraints:

- filter(task.getPriority() == Priority.CRITICAL)) => add penalty X
- filter(task.getPriority() == Priotiry.MAJOR)) => add penalty Y
- filter(task.getPriority() == Priority.MINOR)) = > add penalty Z

We could have just one constraint:

- task -> if (Priority.CRITICAL) => add penalty X, else if (Priority.MAJOR) => ...

rsynek: Either having a "free for all" penalize would do the job, or - in this particular case - something like `penalizeBendable` would help too. Currently, the result of `ToIntFunction<Task>` is just multiplied by each bendable score weight level passed in the previous parameter.

triceo:
@ge0ffrey This is ready to merge now.

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
ge0ffrey: The other methods shouldn't be static. They're not in the other examples.
Either we make them static in all examples, or not at all.

There are very-long idea's to allow the dependency inject instances of ConstraintProvider maybe some day, so let's not make them static.

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning/solver/score/TaskAssigningConstraintProvider.java
ge0ffrey: It's a good idea - but I 'd handle this as a separate issue. This change would need to be applied to the scoreRules.drl too. Right now, I 'd like have an apples to apples comparison for drl vs constraintProvider

ge0ffrey:
@triceo Can you remove the "static" modifiers? +1 to merge then.

If it actually runs (when commented in) - maybe we should add a ...ScoreDirectorBenchmarkConfig.xml ? See https://github.com/kiegroup/optaplanner/blob/master/optaplanner-examples/src/main/resources/org/optaplanner/examples/vehiclerouting/benchmark/vehicleRoutingScoreDirectorBenchmarkConfig.xml

triceo:
@ge0ffrey Added link to a newly reported JIRA, creates score director benchmark config. I also ran it - constraint streams perform marginally better and are much faster.

If possible, squash before merging - this sort of commit history doesn't help anyone.

triceo:
@ge0ffrey Ready to merge, confirmed with FULL_ASSERT.

Please squash before merging.

PR #540, opened 2019-09-04
PLANNER-1597 Provide move selectors for sequential subpillars

triceo:
There will be follow-up PRs for:

- Documentation,
- Nurse rostering cleanup,
- and perhaps `AbstractPersistable` refactor.


triceo:
[First benchmark results](https://github.com/kiegroup/optaplanner/files/3574748/first.tar.gz).

See that:
- FCS_DEFAULT is better than FCS_SEQUENTIAL.
- NR_CUSTOM is better than NR_JUST_GENERIC.

Not by much, but still.... the original configurations are all better.

triceo:
OK, with [further improvements](https://github.com/kiegroup/optaplanner/files/3575675/second.tar.gz), I got it to a point where:

- sequential is consistently better on Nurse Rostering
- and still underperforms (a little) on Flight Crew Scheduling.

However, this is likely not due to the awesomeness of sequential moves, but due to the fact that - on nurse rostering - they are 5 times as fast as the original custom moves.

This observation is confirmed by Flight Crew Scheduling, where they perform nearly identically, and the scores are nearly identical as well.

triceo:
Attached is a [new benchmark](https://github.com/kiegroup/optaplanner/files/3582255/third.tar.gz). This one ran very long.

- Results on Flight Crew Scheduling are nearly identical.
- Results on Nurse Rostering are, on average, 4 % worse than the custom moves.

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/PillarSelectorConfig.java
```diff
@@ -52,10 +57,18 @@ public void setEntitySelectorConfig(EntitySelectorConfig entitySelectorConfig) {
         this.entitySelectorConfig = entitySelectorConfig;
     }
 
+    /**
+     * @see PillarType and its uses in pillar move selectors.
```
triceo: Add @params, @return and @deprecated Javadoc tags.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/ShiftAssignmentComparator.java
```diff
@@ -0,0 +1,15 @@
+package org.optaplanner.examples.nurserostering.domain;
+
+import java.util.Comparator;
+
+public class ShiftAssignmentComparator implements Comparator<ShiftAssignment> {
+
+    private static final Comparator<ShiftAssignment> COMPARATOR =
+            Comparator.comparing((ShiftAssignment a) -> a.getShiftDate().getDate())
+            .thenComparing(ShiftAssignment::getIndexInShift);
```
triceo: This is wrong, provide a correct Comparator.

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/PillarSelectorConfig.java
triceo: (@deprecated in favor of {@link ...})

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/domain/FlightAssignmentOrderComparator.java
```diff
@@ -0,0 +1,16 @@
+package org.optaplanner.examples.flightcrewscheduling.domain;
+
+import java.util.Comparator;
+
+public class FlightAssignmentOrderComparator implements Comparator<FlightAssignment> {
```
triceo: AbstractPersistable should not be Comparable. Entities will be Comparable on-demand.

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/move/generic/PillarType.java
```diff
@@ -0,0 +1,19 @@
+package org.optaplanner.core.config.heuristic.selector.move.generic;
+
+import java.util.Comparator;
+
+public enum PillarType {
```
triceo: <subPillarType>SEQUENCE</pillarType><subPillarSequenceComparatorClass>org.optaplanner.examples.flightcrewscheduling.domain.FlightAssignmentOrderComparator</>

Types - NONE, SEQUENCE, ALL.

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/PillarSelectorConfig.java
```diff
@@ -26,17 +26,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import static org.apache.commons.lang3.ObjectUtils.*;
+import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 @XStreamAlias("pillarSelector")
 public class PillarSelectorConfig extends SelectorConfig<PillarSelectorConfig> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(PillarSelectorConfig.class);
```
ge0ffrey: From ScoreDirectorFactoryConfig.java:

    private static final Logger logger = LoggerFactory.getLogger(ScoreDirectorFactoryConfig.class);

As it is a "static final", it makes sense to make it uppercase according to typical java guidelines - but I'd like that change to be done consistently for all optaplanner-core code - in a separate PR I guess.

Do or do not - when we change something in optaplanner-core, we either do it consistently everywhere or we don't do it at all. I don't want to build car that has modern rubber wheels on the left side and still wooden wheels on the right side.

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/PillarSelectorConfig.java
```diff
@@ -52,10 +57,18 @@ public void setEntitySelectorConfig(EntitySelectorConfig entitySelectorConfig) {
         this.entitySelectorConfig = entitySelectorConfig;
     }
 
+    /**
+     * @see PillarType and its uses in pillar move selectors.
+     */
+    @Deprecated(/* forRemoval = true */)
```
ge0ffrey: Please add a line in the description of this jira so we don't forget to remove this in 8.0:
  https://issues.jboss.org/browse/PLANNER-750

Please also create a PR to add a note in the upgrade recipe:
  https://www.optaplanner.org/download/upgradeRecipe/upgradeRecipe7.html
  https://github.com/kiegroup/optaplanner-website/blob/master/download/upgradeRecipe/upgradeRecipe7.adoc

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/PillarSelectorConfig.java
```diff
@@ -82,41 +95,93 @@ public void setMaximumSubPillarSize(Integer maximumSubPillarSize) {
...
-    public PillarSelector buildPillarSelector(HeuristicConfigPolicy configPolicy,
-            SelectionCacheType minimumCacheType, SelectionOrder inheritedSelectionOrder,
-            List<String> variableNameIncludeList) {
+    public PillarSelector buildPillarSelector(HeuristicConfigPolicy configPolicy, PillarType pillarType,
+            Class<? extends Comparator> pillarOrderComparatorClass, SelectionCacheType minimumCacheType,
+            SelectionOrder inheritedSelectionOrder, List<String> variableNameIncludeList) {
+        if (subPillarEnabled != null) {
+            if (pillarType == null) {
+                LOGGER.warn("Property subPillarEnabled ({}) on pillarSelectorConfig ({}) is deprecated for removal.\n" +
+                        "Use property pillarType on the parent MoveSelectorConfig.", subPillarEnabled, this);
```
ge0ffrey: The compiler logs a warning when using deprecated methods.
Developers hopefully notice it there - or when following the upgrade recipe.
The developers who don't notice the compilation message, nor look at the upgrade recipe, will feel it when migrating to 8.0 and fix it then.

Generally, optaplanner-core doesn't add a warn log message for deprecated methods at runtime (especially if the deprecated method still behave equally well), because it could cause sys ops to reduce logging verbosity to ignore it (and that means they go blind on what optaplanner is doing, which is worse).

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/SubPillarConfigPolicy.java
```diff
@@ -0,0 +1,106 @@
...
+        this.entityComparator = null;
+    }
+
+    private SubPillarConfigPolicy(int minimumSubPillarSize, int maximumSubPillarSize, Comparator<?> entityComparator) {
+        this.subPillarEnabled = true;
+        this.minimumSubPillarSize = minimumSubPillarSize;
+        this.maximumSubPillarSize = maximumSubPillarSize;
+        validateSizes();
+        if (entityComparator == null) {
+            throw new IllegalStateException("entityComparator must not be null.");
```
ge0ffrey: error messages start with a capital
error messages are sentances so they have "The" and "A" etc.
=> "The entityComparator must not be null.")

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/SubPillarConfigPolicy.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public int getMaximumSubPillarSize() {
+        return maximumSubPillarSize;
+    }
+
+    /**
+     * @return Only present if the subpillars are to be treated as sequential. Never present if
+     * {@link #isSubPillarEnabled()} is false.
+     */
+    public Optional<Comparator<?>> getEntityComparator() {
```
ge0ffrey: Let's not use Optional. They don't make the code easier to read. Optionals are for Stream purposes.
Just rreturn the Compartor and have the javadoc state ist's null if subpillarenable is false

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/SubPillarConfigPolicy.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "{" +
+                "entityComparator=" + entityComparator +
+                ", maximumSubPillarSize=" + maximumSubPillarSize +
+                ", minimumSubPillarSize=" + minimumSubPillarSize +
+                ", subPillarEnabled=" + subPillarEnabled +
+                '}';
```
ge0ffrey: optaplanner-core doesn't use verbose toStrings.
Our philophy (which works really well in practice with log messages) is that toStrings() should not be used to debug - the IntelliJ/VsCode debugger has plenty of ways to look at the fields of a instance.
Instead, they must give an indication of what it is, in essence. Less is more. See toStrings() on other classes (ChangeMoveSelector, ChangeMove, FilterConstraintStream, ...)

PS: why the + '}' (single quotes)?

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/move/generic/PillarType.java
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * Pillars will only be affected in their entirety, in the order in which they're constructed.
+     */
+    FULL_ONLY,
+    /**
+     * Pillars may also be affected partially, but still only in the order in which they're constructed.
```
ge0ffrey: "the order in which they're constructed" is ambiguous (some people could interpret it as meaning sequential).

More importantly, it's an implementation detail that we don't want to guarantee for future versions by writing it in the javadoc. It's like how OpenJDK doesn't write the actual formula for string or Integer hashcode() in the hashcode javadoc - it prohibits them from ever improving it.

optaplanner-examples/src/main/resources/org/optaplanner/examples/flightcrewscheduling/solver/flightCrewSchedulingSolverConfig.xml
```diff
@@ -19,9 +19,8 @@
     <unionMoveSelector>
       <changeMoveSelector/>
       <swapMoveSelector/>
-      <!-- TODO Use a sorted pillar to heavily improve feasibility and solution quality https://issues.jboss.org/browse/PLANNER-1217 -->
-      <!--<pillarChangeMoveSelector/>-->
-      <!--<pillarSwapMoveSelector/>-->
+      <pillarChangeMoveSelector/>
+      <pillarSwapMoveSelector/>
```
ge0ffrey: do we want to enable SEQUENTIAL here?
It depends on your benchmark results - if this way is better (I think you showed me that), then let's indeed keep it as this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
```diff
@@ -137,16 +138,23 @@ public void constructCache(DefaultSolverScope solverScope) {
...
                     + ") which is higher than Integer.MAX_VALUE.");
         }
+        Stream<Object> entities = StreamSupport.stream(entitySelector.spliterator(), false);
+        /*
+         * sort the entities in case we need sequential pillars; this will result in all the pillars being sorted
+         * without having to sort them individually later.
+         */
+        Optional<Comparator<?>> comparator = subpillarConfigPolicy.getEntityComparator();
+        entities = comparator.isPresent() ? entities.sorted((Comparator) comparator.get()) : entities;
+        // create all the pillars from a stream of entities; if sorted, the pillars will be sequential
```
ge0ffrey: Comments are sentences, start with a capital

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/pillar/DefaultPillarSelector.java
```diff
@@ -259,27 +271,38 @@ public RandomSubPillarIterator() {
...
-                Object[] sandboxPillar = basePillar.toArray(); // Clone to avoid changing basePillar
-                List<Object> subPillar = new ArrayList<>(subPillarSize);
-                for (int i = 0; i < subPillarSize; i++) {
-                    int index = i + workingRandom.nextInt(basePillarSize - i);
-                    subPillar.add(sandboxPillar[index]);
-                    sandboxPillar[index] = sandboxPillar[i];
-                }
-                return subPillar;
             }
+            // when entity comparator present, it means we require sequential subpillars
```
ge0ffrey: => When an entity comparator is present, ...

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/domain/FlightAssignmentOrderComparator.java
triceo: @ge0ffrey This is going to be a massive change.

With `AbstractPersistable` no longer being `Comparable`, all uses of `CompareToBuilder` in examples need to be reviewed. `CompareToBuilder` doesn't require its arguments to be `Comparable`, and therefore - when comparing instances of `AbstractPersistable`, the new code could potentially fail at runtime.

I propose we still do this, but instead do it in a different PR. To not mix it with this work. (I will take the task.)

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/SubPillarConfigPolicy.java
triceo: I'm going to resist the urge to bikeshed now. But we will inevitably get back to this at some later point, since I'm not going to let you take `Optional` away from me so easily. :-)

optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/entity/pillar/SubPillarConfigPolicy.java
triceo: OK. In this case, I'll just remove the `toString()`.
The single quotes are auto-generated by IntelliJ. I assume it's because "}" is a single character and, as such, using a `String` for it is wasteful.

optaplanner-examples/src/main/resources/org/optaplanner/examples/flightcrewscheduling/solver/flightCrewSchedulingSolverConfig.xml
triceo: Your choice. The results are nearly identical - one is better here, the other is better there.
I suspect that, when there are additional constraints, the differences will become more pronounced.

triceo:
Here's a [definitive benchmark](https://github.com/kiegroup/optaplanner/files/3588912/fourth.tar.gz) on the latest code. Custom moves still perform best, and I no longer have any idea where that difference is coming from. Bad random seed?

PR #541, opened 2019-09-05
Planner 1593 deep cloning shadow variables map docs

ge0ffrey:


optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/clone/deepcloning/TestdataDeepCloningEntity.java
```diff
@@ -68,6 +72,8 @@ public void setValue(TestdataValue value) {
     }
 
     @DeepPlanningClone
+    @CustomShadowVariable(sources = {@PlanningVariableReference(variableName = "value")},
```
triceo: Why this update to `TestData`, when there is no update to any of the actual tests?

optaplanner-docs/src/main/asciidoc/LocalSearch/LocalSearch-chapter.adoc
```diff
@@ -283,15 +283,14 @@ Advanced configuration:
 [[tabuSearchAlgorithm]]
 === Algorithm Description
 
-Tabu Search works like Hill Climbing, but it maintains a tabu list to avoid getting stuck in local optima.
+Tabu Search is a Local Search that maintains a tabu list to avoid getting stuck in local optima.
```
triceo: This change is not related to PLANNER-1593.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/clone/deepcloning/TestdataDeepCloningEntity.java
ge0ffrey: the field is called shadowVariable - but it wasn't one already. Just makes more sense like this.

optaplanner-docs/src/main/asciidoc/LocalSearch/LocalSearch-chapter.adoc
ge0ffrey: Indeed, request from JustinG that I am lumping in this already small PR. For small PR's I think it's fine to combine such things (it's a separate commit for history of course).

ge0ffrey:
Issue renamed, good point. Thanks

PR #545, opened 2019-09-09
PLANNER-1598 Remove most custom moves from Nurse Rostering

triceo:
@ge0ffrey Please review.


ge0ffrey:
Of the 5 move selectors, we're only replacing 4 of them, but not "ShiftAssignmentPillarPartSwapMoveFactory". Are we sure that this isn't a regression, by removing that one too? Are we getting a better overall rank in the benchmarks?

kie-ci8:

triceo:
You were right, @ge0ffrey . I put it back.

PR #547, opened 2019-09-09
Detach JPA objects error message

ge0ffrey:
Relates to https://stackoverflow.com/questions/57806094/exception-after-version-upgrade/57848870?noredirect=1#comment102127089_57848870


ge0ffrey:
REJECTING THIS PR. See Frederic's answer here: https://stackoverflow.com/questions/57806094/exception-after-version-upgrade/57848870?noredirect=1#comment102127089_57848870
It was never JPA's fault.

PR #549, opened 2019-09-11
Add ConstraintProviders for Flight Crew Scheduling and Rock Tour

triceo:
I ran this with assertion score director and the results are 1:1 equivalent. I did not run any benchmarks.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
@@ -1,13 +1,15 @@
 package org.optaplanner.examples.flightcrewscheduling.optional.score;
 
-import java.util.Objects;
+import java.time.LocalDate;
 
 import org.optaplanner.core.api.score.buildin.hardsoftlong.HardSoftLongScore;
 import org.optaplanner.core.api.score.stream.Constraint;
 import org.optaplanner.core.api.score.stream.ConstraintFactory;
 import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.core.api.score.stream.Joiners;
 import org.optaplanner.examples.flightcrewscheduling.domain.Employee;
 import org.optaplanner.examples.flightcrewscheduling.domain.FlightAssignment;
+import org.optaplanner.examples.flightcrewscheduling.domain.Skill;
 
 public class FlightCrewSchedulingConstraintProvider implements ConstraintProvider {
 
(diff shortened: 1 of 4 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,63 @@
...
+                employeeUnavailability(constraintFactory),
+                transferBetweenTwoFlights(constraintFactory),
+                firstAssignmentDepartingFromHome(constraintFactory),
+                lastAssignmentArrivingAtHome(constraintFactory)
+        };
+    }
+
+    private Constraint flightConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUniquePair(FlightAssignment.class)
+                .filter((first, second) -> Objects.equals(first.getEmployee(), second.getEmployee()))
```
ge0ffrey: This should use a joiner, for performance:

constraintFactory.fromUniquePair(FlightAssignment.class, equals(FlightAssignment::getEmployee))

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,63 @@
...
+                .filter((first, second) -> second.getFlight().overlaps(first.getFlight()))
+                .filter((first, second) -> second.getId() > first.getId())
```
ge0ffrey: fromUniquePair automaticaly includes the "id > id" check, the "unique" signals that.
fromUniquePair is syntatic sugar for a join with an id comparison (see javadoc)

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private Constraint employeeUnavailability(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(FlightAssignment.class)
+                .filter(a -> !a.getEmployee().isAvailable(a.getFlight().getDepartureUTCDate()))
+                .penalize("Employee unavailable", HardSoftLongScore.ofHard(10));
+    }
+
+    private Constraint transferBetweenTwoFlights(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Employee.class)
+                .filter(e -> e.countInvalidConnections() > 0)
```
ge0ffrey: Can we use the variable name "employee" in such cases? (Same for a above constraints)

lamdba's don't have an explicit type written down. (It says "e", not "Employee e"), so for readability (most users start from one of these examples), let's use a full variable name.

In general, in optaplanner-core only the following short var names are used: int i, Object o, Exception e.

optaplanner-examples/src/main/java/org/optaplanner/examples/rocktour/optional/score/RockTourConstraintProvider.java
```diff
@@ -0,0 +1,82 @@
...
+                .filter(r -> r.getPreviousStandstill() != null)
+                .penalizeConfigurableLong(DRIVING_TIME_TO_SHOW_PER_SECOND,
+                        RockShow::getDrivingTimeFromPreviousStandstill);
+    }
+
+    private Constraint shortenDrivingTimePerMillisecondSquared(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(RockShow.class)
+                .filter(r -> r.getPreviousStandstill() != null)
+                .penalizeConfigurableLong(SHORTEN_DRIVING_TIME_PER_MILLISECOND_SQUARED,
+                        r -> (long) Math.pow(r.getDrivingTimeFromPreviousStandstill(), 2));
```
ge0ffrey: Math.pow does long to double conversation, which is inaccurate, try (double) 9007199254740993L
  https://ge0ffrey.github.io/ge0ffrey-presentations/cornerCaseCheatSheet/cheatSheetJava.html
then just to convert it back to long (again, inaccurate)

Lets simply do r.getDrivingTimeFromPreviousStandstill() * r.getDrivingTimeFromPreviousStandstill() ?

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
ge0ffrey: see also javadoc of join:

     * Important: {@link BiConstraintStream#filter(BiPredicate) Filtering} this is slower and less scalable
     * than a {@link #join(UniConstraintStream, BiJoiner)},
     * because it doesn't apply hashing and/or indexing on the properties,
     * so it creates and checks every combination of A and B.

optaplanner-examples/src/main/java/org/optaplanner/examples/rocktour/optional/score/RockTourConstraintProvider.java
triceo: This sounds more like an argument to avoid using `double`. What do we do when we want to ^50? We write a for loop?

Anyway... changing it.

ge0ffrey: thanks, in any case - at some point, we will write penalizeUnfairness or .rewardLoadBalance or something like that, replacing this code, making this discussion mute

PR #550, opened 2019-09-12
PLANNER-1607 Make it possible for entities to be Comparable

triceo:
@ge0ffrey This is the last bit of sequential subpillars we talked about.
If possible, please merge on Friday - it actually fixes a bug that we didn't notice until now.

FYI I ran all the examples by hand to make sure that there are no unexpected runtime failures.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintJustification.java
```diff
@@ -0,0 +1,13 @@
+package org.optaplanner.core.api.score.constraint;
+
+import java.util.Comparator;
+
+public interface ConstraintJustification {
+
+    /**
+     * Used by the GUI to sort the {@link ConstraintMatch} list by {@link ConstraintMatch#getJustificationList()}. Must
+     * be able to meaningfully compare classes of different types, if only by comparing their {@link Class#getName()}.
+     */
```
ge0ffrey: Please include @returns in the javadoc (if a javadoc is present, it should be valid)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -82,7 +83,12 @@ public int compareTo(ConstraintMatch other) {
             return constraintName.compareTo(other.constraintName);
         } else {
             for (int i = 0; i < justificationList.size() && i < other.justificationList.size(); i++) {
-                int comparison = ((Comparable) justificationList.get(i)).compareTo(other.justificationList.get(i));
+                Object left = justificationList.get(i);
+                Object right = other.justificationList.get(i);
+                Comparator comparator = left instanceof ConstraintJustification ?
```
ge0ffrey: This confuses me. "left" is a problem fact or a planning entity.
This code reads: "if a problem fact is an instance of ConstraintJustification, it has a method that returns a comperator?

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/domain/solver/TopicConflict.java
```diff
@@ -87,15 +93,11 @@ public int hashCode() {
 
     @Override
     public int compareTo(TopicConflict other) {
-        return new CompareToBuilder()
-                .append(leftTopic, other.leftTopic)
-                .append(rightTopic, other.rightTopic)
-                .toComparison();
+        return COMPARATOR.compare(this, other);
```
ge0ffrey: These kind of changes are a step forward, this should definitely be done.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintJustification.java
```diff
@@ -0,0 +1,13 @@
+package org.optaplanner.core.api.score.constraint;
+
+import java.util.Comparator;
+
+public interface ConstraintJustification {
```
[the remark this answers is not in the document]
ge0ffrey: It bugs me that problem facts implement an interface which forces them to either contain or build a Comperator.

If they implement this interface, can they not simply implement Comparable?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
triceo: Precisely. By implementing `ConstraintJustification`, you tell this code how to compare the instances without having the instances implement `Comparable<>`.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintJustification.java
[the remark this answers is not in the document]
triceo: They can, but then you can not have your own entities `Comparable` using your own `Comparator`. Which is the whole reason we started doing this.

triceo:
Closing as there is going to be a different implementation.

PR #551, opened 2019-09-13
Fix processing of Comparable on subpillar sequence comparator XML

triceo:


triceo:
@ge0ffrey This is just the bugfix from PLANNER-1607.

PR #553, opened 2019-09-17
PLANNER-1607 Allow entities to be Comparable on their own

triceo:
This is a second attempt. `AbstractPersistable` is no longer `Comparable`. Instead, constraint justifications have their own `Comparator` now.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintJustificationComparator.java
```diff
@@ -0,0 +1,37 @@
...
+import org.optaplanner.core.config.util.ConfigUtils;
+import org.optaplanner.core.impl.domain.common.accessor.MemberAccessor;
+
+final class ConstraintJustificationComparator implements Comparator<Object> {
+
+    private static Comparator<Object> CLASS_NAME_COMPARATOR = Comparator.comparing(o -> o.getClass().getCanonicalName());
+    private final Map<Class<?>, MemberAccessor> accessorCache = new HashMap<>(0);
+
+    @Override
+    public int compare(Object o1, Object o2) {
```
triceo: This is the meat of the PR. This is how all constraint justification objects would be compared.

triceo: This also probably deserves some tests. Will add.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/solver/TaskAssignmentDifficultyComparator.java
```diff
@@ -19,22 +19,15 @@
...
 
+    private static final Comparator<TaskAssignment> COMPARATOR = Comparator.comparing(TaskAssignment::getTask)
+            .thenComparingLong(TaskAssignment::getId);
+
     @Override
     public int compare(TaskAssignment a, TaskAssignment b) {
-        Task aTask = a.getTask();
-        Task bTask = b.getTask();
-        return new CompareToBuilder()
-                .append(aTask.getResourceUsageMultiplicand(), bTask.getResourceUsageMultiplicand())
```
ge0ffrey: Looks like this changes the behavior? The COMPERATOR does use getResourceUsageMultiplicand()

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/IdleCost.java
```diff
@@ -16,12 +16,21 @@
...
 import org.optaplanner.examples.cheaptime.domain.Machine;
 
-public class IdleCost {
+import static java.util.Comparator.comparing;
+import static java.util.Comparator.comparingLong;
+
+public class IdleCost implements Comparable<IdleCost> {
+
+    private static final Comparator<IdleCost> COMPARATOR =
+            comparing(IdleCost::getMachine, comparingLong(Machine::getId))
```
ge0ffrey: The original code did:
  .append(machine, other.machine)
Why not just compare the machines, as AbstractPersistable still implements Comparable?

optaplanner-examples/src/main/java/org/optaplanner/examples/cloudbalancing/optional/domain/CloudComputerStrengthComparator.java
```diff
@@ -19,18 +19,21 @@
...
 import org.optaplanner.examples.cloudbalancing.domain.CloudComputer;
 
 public class CloudComputerStrengthComparator implements Comparator<CloudComputer>, Serializable {
 
+    private static final Comparator<CloudComputer> DESCENDING_COST_COMPARATOR =
+            Comparator.comparing(CloudComputer::getCost).reversed();
+
+    private static final Comparator<CloudComputer> COMPARATOR =
+            Comparator.comparingInt(CloudComputer::getMultiplicand)
+            .thenComparing(DESCENDING_COST_COMPARATOR) // Descending (but this is debatable)
```
ge0ffrey: I'd inline the DESCENDING_COST_COMPARATOR here, it's not like we need it anywhere else

ge0ffrey: See also https://stackoverflow.com/a/30382620/472109

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintJustificationComparator.java
```diff
@@ -0,0 +1,37 @@
...
+        MemberAccessor memberAccessor = accessorCache.computeIfAbsent(o1.getClass(),
+                ConfigUtils::findPlanningIdMemberAccessor);
+        if (memberAccessor == null) {
+            // Same class, not comparable, does not have planning ID => don't sort. Not consistent with equals().
+            return 0;
+        }
+        // Compare planning IDs; Comparable guaranteed by MemberAccessor.
+        Comparable id1 = (Comparable)memberAccessor.executeGetter(o1);
+        Comparable id2 = (Comparable)memberAccessor.executeGetter(o2);
+        return id1.compareTo(id2);
```
ge0ffrey: +1 for the order of this method.

Style wise, we put a space after a cast :) So "(Comparable)o1" -> "(Comparable) o1"

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -81,8 +82,11 @@ public int compareTo(ConstraintMatch other) {
         } else if (!constraintName.equals(other.constraintName)) {
             return constraintName.compareTo(other.constraintName);
         } else {
+            Comparator<Object> comparator = new ConstraintJustificationComparator();
```
ge0ffrey: That ConstraintJustificationComparator caches the reflection results, but for every ConstriantMatch (= 1000's) we create a new one?
I am ok with merging this as is right now, but we need a new jira to improve performance of constraint match enabled by somehow having this comparator once on the scoreHolder and scoreInliner? Idunno

ge0ffrey: @triceo Can you add a TODO comment to point out the problem?

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/IdleCost.java
triceo: @ge0ffrey `AbstractPersistable` does not implement `Comparable`. That's the point of this PR.

ge0ffrey: Ah yes of course. Ignore this comment.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/domain/solver/TaskAssignmentDifficultyComparator.java
triceo: It does not change anything. It refers to the `Task`, which is now `Comparable` and does the rest of the comparison exactly as written here.

That said, I'll move that Comprator code here, to make it consistent with the other examples and to make it clearer what it actually does.

triceo:
Ready for merge, @ge0ffrey .

PR #555, opened 2019-09-19
PLANNER-1635 Termination should support java.time.Duration

ge0ffrey:
Needed so we can do this in quarkus/spring-boot application.properties:
optaplanner.spentLimit=PT2M30S


ge0ffrey:
I wanted to deprecate them, but the ISO format "PT1M30S" (= 1minute 30 seconds) is a bit too non-obvious for new users.
So, for now at least, we're supporting both ways, promoting the ISO format first in the docs and the quarkus/spring-boot example (but not yet the swing examples). We'll see how that goes - what people tend to use. Based on that data, we can revisit deprecation.

PR #556, opened 2019-09-19
PLANNER-1576 Add SolverManager api and default implementation

MusaTalluzi:
Issues to be fixed before ready to merge:
- Shutdown: see comment in DefaultSolverManager.shutdown()
- Test thread safety


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,188 @@
...
+            newSolverTask.addEventListener(
+                    bestSolutionChangedEvent ->
+                            eventHandlerExecutorService.submit(
+                                    () -> onBestSolutionChangedEvent.accept(bestSolutionChangedEvent.getNewBestSolution())));
+        }
+
+        CompletableFuture<Solution_> solverFuture = CompletableFuture.supplyAsync(newSolverTask::startSolving, solverExecutorService);
+        solverFuture.handle((solution_, throwable) -> {
+            if (throwable != null) {
+                logger.error("Exception while solving problem (" + problemId + ").", throwable.getCause());
```
yurloc: String concatenation in logging statements should be replaced with `{}` placeholders .

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManagerTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Test(timeout = 5000L)
+    public void basicUsageOfSolverManagerWithOneProblem() throws InterruptedException {
+        TestdataSolution problem = createTestProblem(tenantId);
+        solverManager.solve(tenantId, problem,
+                            taskAssigningSolution -> solutionChangedLatch.countDown(),
+                            taskAssigningSolution -> solvingEndedLatch.countDown());
+        solutionChangedLatch.await(5, TimeUnit.SECONDS);
+
+        assertEquals(solverManager.getSolverStatus(tenantId), SolverStatus.SOLVING);
```
yurloc: Wrong order of arguments. JUnit Assert uses `assertEquals(expected, actual)`.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManagerTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Test(timeout = 5000L)
+    public void shouldNotStartTwoSolverTasksWithSameProblemId() {
+        TestdataSolution problem = createTestProblem(tenantId);
+        solverManager.solve(tenantId, problem, null, null);
+        try {
+            solverManager.solve(tenantId, problem, null, null);
+        } catch (IllegalArgumentException e) {
+            assertEquals("Problem (" + tenantId + ") already exists.", e.getMessage());
+        }
```
yurloc: The test will pass if no exception is thrown. I recommend using AssertJ:
```java
assertThatIllegalArgumentException()
                .isThrownBy(() -> solverManager.solve(tenantId, problem, null, null))
                .withMessageContaining("Problem (" + tenantId + ") already exists.");

File: optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java [hunk 4]
```diff
@@ -0,0 +1,188 @@
...
+    @Override
+    public void shutdown() {
+        logger.info("Shutting down {}.", DefaultSolverManager.class.getName());
+        ThreadUtils.shutdownAwaitOrKill(solverExecutorService, "", "solverExecutorService");
+        ThreadUtils.shutdownAwaitOrKill(eventHandlerExecutorService, "", "eventHandlerExecutorService");
+        stopSolvers(); // without this: shutdownAwaitOrKill appears to only interrupt currently running tasks, those that are queued start solving
+    }
+
+    private void stopSolvers() {
+        for (SolverTask solverTask : problemIdToSolverTaskMap.values()) {
```
yurloc: raw type

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,188 @@
...
+        return solverTask.getBestSolution();
+    }
+
+    @Override
+    public Score<?> getBestScore(Object problemId) {
+        logger.debug("Getting best score of problemId ({}).", problemId);
+        SolverTask<Solution_> solverTask = problemIdToSolverTaskMap.get(problemId);
+        if (solverTask == null) {
+            logger.error("Problem (" + problemId + ") was not submitted.");
+            return null;
```
yurloc: Should throw an exception in my opinion. `SolverTask.getBestScore()` will return `null` if solution is uninitialized. So when you get `null` from `DefaultSolverManager.getBestScore()` how do you know your `problemId` is wrong or the solution is not yet initialized?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,188 @@
...
+
+    public DefaultSolverManager(String solverConfigResource, ClassLoader classLoader) {
+        this(solverConfigResource, classLoader, null);
+    }
+
+    public DefaultSolverManager(String solverConfigResource, ThreadFactory threadFactory) {
+        this(solverConfigResource, null, threadFactory);
+    }
+
+    public DefaultSolverManager(String solverConfigResource, ClassLoader classLoader, ThreadFactory threadFactory) {
```
yurloc: Should be private so that clients don't have access to a constructor that accepts `null` arguments.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,188 @@
...
+
+    public DefaultSolverManager(String solverConfigResource) {
+        this(solverConfigResource, null, null);
+    }
+
+    public DefaultSolverManager(String solverConfigResource, ClassLoader classLoader) {
+        this(solverConfigResource, classLoader, null);
+    }
+
+    public DefaultSolverManager(String solverConfigResource, ThreadFactory threadFactory) {
```
yurloc: And these public constructors should require non-null arguments. Tip: use `Objects.requireNonNull()`.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManagerTest.java
```diff
@@ -0,0 +1,197 @@
...
+        solverManager.stopSolver(tenantId);
+
+        solvingEndedLatch.await(5, TimeUnit.SECONDS);
+        assertTrue(bestSolutionChangedEventCount.get() > 0);
+        logger.info("Number of bestSolutionChangedEvents: {}.", bestSolutionChangedEventCount.get());
+    }
+
+    @Test
+    public void shutdownShouldStopAllSolvers() {
+        int[] problemIds = new int[Runtime.getRuntime().availableProcessors() * 3];
```
yurloc: Tip: use `IntStream` to get rid of the array and for loop:
```java
int problemCount = Runtime.getRuntime().availableProcessors() * 3;
IntStream.range(0, problemCount)
        .forEach(problemId -> {
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
MusaTalluzi: @ge0ffrey and I discussed this point previously and decided to only log an error (I can't find where we documented the discussion and can't remember the arguments, but I'll bring it up again in our next meeting)

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
```diff
@@ -0,0 +1,66 @@
...
+    void stopSolver(Object problemId);
+
+    boolean isProblemSubmitted(Object problemId);
+
+    Solution_ getBestSolution(Object problemId);
+
+    Score<?> getBestScore(Object problemId);
+
+    SolverStatus getSolverStatus(Object problemId);
+
```
cuijulian: There should be a `getScoreDirector()` method to get the IndictmentMap of a solution.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverStatus.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.api.solver.manager;
+
+public enum SolverStatus {
+    SOLVING,
+    NOT_SOLVING, // either hasn't started yet or finished solving
```
triceo: As this is a public API, comments such as these ought to be javadocs.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
```diff
@@ -0,0 +1,66 @@
...
+package org.optaplanner.core.api.solver.manager;
+
+import java.util.concurrent.ThreadFactory;
+import java.util.function.Consumer;
+
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.solver.manager.DefaultSolverManager;
+
+public interface SolverManager<Solution_> extends AutoCloseable {
```
triceo: This and the methods IMHO require Javadocs. Specifically, the fact that `problemId` is an `Object` begs for an explanation - what should I use for the id? What's the runtime behavior?

- What happens when I have two different instances of the id, but they `equals()`?)
- Is `null` accepted, where, and what is the behavior in that case?
- Can `null` be returned somewhere?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private DefaultSolverManager(String solverConfigResource, ClassLoader classLoader, ThreadFactory threadFactory) {
+        if (classLoader != null) {
+            solverFactory = SolverFactory.createFromXmlResource(solverConfigResource, classLoader);
+        } else {
+            solverFactory = SolverFactory.createFromXmlResource(solverConfigResource);
+        }
+
+        problemIdToSolverTaskMap = new ConcurrentHashMap<>();
```
triceo: Since you're not sizing the `Map`, you can keep the initialization at the declaration site.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,207 @@
...
+
+        // TODO implement throttling
+        if (onBestSolutionChangedEvent != null) {
+            newSolverTask.addEventListener(
+                    bestSolutionChangedEvent ->
+                            eventHandlerExecutorService.submit(
+                                    () -> onBestSolutionChangedEvent.accept(bestSolutionChangedEvent.getNewBestSolution())));
+        }
+
+        CompletableFuture<Solution_> solverFuture = CompletableFuture.supplyAsync(newSolverTask::startSolving, solverExecutorService);
```
triceo: What if the user decides to solve more tasks than they have CPU cores? Do we want to somehow prevent it or at least warn about it?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,207 @@
...
+    @Override
+    public void shutdown() {
+        logger.info("Shutting down {}.", DefaultSolverManager.class.getName());
+        ThreadUtils.shutdownAwaitOrKill(solverExecutorService, "", "solverExecutorService");
+        ThreadUtils.shutdownAwaitOrKill(eventHandlerExecutorService, "", "eventHandlerExecutorService");
+        stopSolvers(); // without this: shutdownAwaitOrKill appears to only interrupt currently running tasks, those that are queued start solving
+    }
+
+    private void stopSolvers() {
+        for (SolverTask<Solution_> solverTask : problemIdToSolverTaskMap.values()) {
```
triceo: Personally, I'd prefer `problemIdToSolverTaskMap.values().forEach(solverTask::stopSolver)`. Shorter, don't have to bother with typing.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
```diff
@@ -0,0 +1,66 @@
...
+    static <Solution_> SolverManager<Solution_> createFromXmlResource(String solverConfigResource, ClassLoader classLoader) {
+        return DefaultSolverManager.createFromXmlResource(solverConfigResource, classLoader);
+    }
+
+    static <Solution_> SolverManager<Solution_> createFromXmlResource(String solverConfigResource, ThreadFactory threadFactory) {
+        return DefaultSolverManager.createFromXmlResource(solverConfigResource, threadFactory);
+    }
+
+    static <Solution_> SolverManager<Solution_> createFromXmlResource(String solverConfigResource,
+                                                                      ClassLoader classLoader, ThreadFactory threadFactory) {
```
triceo: As far as I know, we don't align code on columns. Please configure your IDE code style properly.
(Yes, we do not respect the KIE code style in this particular regard.)

yurloc: My $0.02. Format it like this:
```java
    static <Solution_> SolverManager<Solution_> createFromXmlResource(
            String solverConfigResource,
            ClassLoader classLoader,
            ThreadFactory threadFactory
    ) {
```
- doesn't conflict with KIE conventions' alignment style
- contrary to ^^, this alignment style is robust in the sense that if you change the line with method name (e.g. add `public`) you don't have to move all the parameters, because they're simply indented, and not aligned with the opening `(`
- consistent (1 parameter per line)

triceo: We really should take the time to publish the code style we're using. It's mostly the KIE code style, with some notable exceptions.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
MusaTalluzi: The second point in the `Object problemId` is interesting and should have corresponding unit tests.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
MusaTalluzi: That's fine, and one of the reasons of using SolverManager is to be able to solve more problems than cpu cores OOTB.
In this case there are two policies:
- FIFO: solve as many problems as the number of available cores completely before starting queued up problems (current approach by setting the `FixedThreadPool` size in the `solverExectorService` to #cores).
- Round Robin: Solve all submitted tasks in parallel and let the JVM handle scheduling threads on available cores, this however, doesn't guarantee that each problem will get actual cpu time as configured in the Termination policy.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
yurloc: Or just `getIndictmentMap()` to avoid unnecessary object traversal. Do the users of this API need the ScoreDirector? If not, why should they bother getting it first in order to access indictment map?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -0,0 +1,242 @@
...
+        int numAvailableProcessors = Runtime.getRuntime().availableProcessors();
+        int activeThreadCount = Math.max(numAvailableProcessors - 2, 1);
+        logger.info("Number of available processors: {}. Active thread count: {}.", numAvailableProcessors, activeThreadCount);
+
+        if (threadFactory != null) {
+            this.solverFactory.getSolverConfig().setThreadFactoryClass(threadFactory.getClass());
+            solverExecutorService = Executors.newFixedThreadPool(activeThreadCount, threadFactory); // #threads < #cpus implies a FIFO scheduling policy
+            eventHandlerExecutorService = Executors.newSingleThreadExecutor(threadFactory);
+        } else {
+            solverExecutorService = Executors.newFixedThreadPool(numAvailableProcessors - 1);
```
yurloc: Is the `numAvailableProcessors - 1` intended? I don't see a reason why the number of threads in the pool should be different depending on whether you supply a thread factory or not.

yurloc: For example, if `#cpus == 1` then we'll have a thread pool of a zero size. So I'd really expect `activeThreadCount` here.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
MusaTalluzi: @yurloc  that's right we agreed that only the `indictmentMap` is necessary here and there's no reason to expose the ScoreDirector.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
MusaTalluzi: Nice catch, you're right it should be `activeThreadCount`, I updated it in one place and missed the other. Thanks!

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -142,43 +152,97 @@ public void solve(
             }
             newSolverTask = new SolverTask<>(problemId, solverFactory.buildSolver(), planningProblem);
             problemIdToSolverTaskMap.put(problemId, newSolverTask);
+            problemIdToEventHandlerFuturesMap.put(problemId, Collections.synchronizedList(new ArrayList<>()));
```
triceo: Is there any particular reason for using `Collections.synchronizedList()`? Java has concurrent collections whose performance is generally better.

If you insist on using synchronized collections, please make sure you comply [with its Javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedList(java.util.List)).

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -142,43 +152,97 @@ public void solve(
...
-        CompletableFuture<Solution_> solverFuture = CompletableFuture.supplyAsync(newSolverTask::startSolving, solverExecutorService);
+        submitSolverTask(newSolverTask);
+    }
+
+    private void submitSolverTask(SolverTask<Solution_> solverTask) {
+        final Object problemId = solverTask.getProblemId();
+        final Consumer<Solution_> onSolvingEnded = problemIdToOnSolvingEndedMap.get(problemId).orElse(solution_ -> {
+        });
+        final Consumer<Throwable> onException = problemIdToOnExceptionMap.get(problemId).orElse(throwable -> {
+        });
```
triceo: If you decide to get rid of the `Optional` (see above), these calls are easily replaced with `getOrDefault(problemId, ...)`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -142,43 +152,97 @@ public void solve(
...
+    public boolean terminateSolver(Object problemId) {
         Objects.requireNonNull(problemId);
-        logger.debug("Stopping solver of problemId ({}).", problemId);
-        SolverTask<Solution_> solverTask;
-        synchronized (this) {
+        logger.debug("Terminating solver of problemId ({}).", problemId);
+        synchronized (this) { // should be synchronized to avoid cleaning up resources in onSolvingEnded if it's executed after solverTask.terminateSolver() and before cleanUpResources()
             if (!isProblemSubmitted(problemId)) {
-                throw new IllegalArgumentException("Problem (" + problemId + ") was not submitted.");
+                throw new IllegalArgumentException("Problem (" + problemId + ") was not submitted or finished solving.");
```
triceo: `either was not submitted or already finished solving`?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -142,43 +152,97 @@ public void solve(
...
             }
-            solverTask = problemIdToSolverTaskMap.get(problemId);
+            SolverTask<Solution_> solverTask = problemIdToSolverTaskMap.get(problemId);
+            boolean willTerminate = solverTask.stopSolver();
+            if (!willTerminate) {
+                logger.error("Cannot terminate solver of problemId ({}).", problemId);
```
triceo: `solver with problemId`? Not sure.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -142,43 +152,97 @@ public void solve(
...
         // TODO implement throttling
         if (onBestSolutionChangedEvent != null) {
-            newSolverTask.addEventListener(
-                    bestSolutionChangedEvent ->
-                            eventHandlerExecutorService.submit(
-                                    () -> onBestSolutionChangedEvent.accept(bestSolutionChangedEvent.getNewBestSolution())));
+            newSolverTask.addEventListener(bestSolutionChangedEvent -> {
+                Future<?> bestSolutionChangedEventFuture = eventHandlerExecutorService.submit(
+                        () -> onBestSolutionChangedEvent.accept(bestSolutionChangedEvent.getNewBestSolution()));
+                problemIdToEventHandlerFuturesMap.get(problemId).add(bestSolutionChangedEventFuture);
```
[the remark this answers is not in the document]
MusaTalluzi: This'll definitely be re-factored into a separate method once I implement throttling.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
```diff
@@ -142,43 +152,97 @@ public void solve(
...
+    @Override
+    public synchronized boolean pauseSolver(Object problemId) {
+        logger.debug("Pausing solver of problemId ({}).", problemId);
+        SolverTask<Solution_> solverTask = getSolverTask(problemId);
+        return solverTask != null && solverTask.pauseSolver();
+    }
+
+    @Override
+    public synchronized boolean resumeSolver(Object problemId) {
+        logger.debug("Pausing solver of problemId ({}).", problemId);
```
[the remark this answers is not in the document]
MusaTalluzi: It should've hadn't `logger` been used throughout `optaplanner-core`, so it's better to follow the convention.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
```diff
@@ -0,0 +1,271 @@
...
+    boolean terminateSolver(Object problemId);
+
+    /**
+     * Notifies the solver that it should stop at its earliest convenience.
+     * To resume solving call {@link #resumeSolver(Object problemId)}
+     *
+     * @param problemId never null, a unique id for each planning problem
+     * @return true if successful, false if no problem with problemId has been submitted
+     */
+    boolean pauseSolver(Object problemId);
```
ge0ffrey: We can't pauze a solver, only terminateEarly() it

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
```diff
@@ -0,0 +1,271 @@
...
+     * <p>
+     * This method is useful in rare asynchronous situations (although {@code onBestSolutionChangedEvent} callback
+     * of {@link #solve} is often more appropriate).
+     *
+     * @param problemId never null, a unique id for each planning problem
+     * @return null if the problemId isn't associated with a previously submitted problem, or
+     * {@link PlanningSolution best solution} which might be uninitialized with a null {@link Score} if the solver job
+     * has not started yet
+     */
+    Solution_ getBestSolution(Object problemId);
```
ge0ffrey: This method causes a fundamental memory leak, by design.
It was part of the requirements, but we need to rethink this.
Let's discuss on Thursday.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/manager/SolverManager.java
ge0ffrey: Good suggestion that we need to support Indictments and CosntraintMatchTotals too (potentially through getScoreDirectorFactory). In phase one, this is out of scope.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/manager/DefaultSolverManager.java
MusaTalluzi: Thanks for pointing that out! No particular reason, actually `CopyOnWriteArrayList` is better than implementing the synchronization manually while iterating over the list.

ge0ffrey:
The other SolverManager PR has been merged, which copy-paste-adjusted much of the work in this PR. Closing this PR to avoid confusion. I will however still use this PR a bit more later: the consumer thread code - so please don't delete that branch yet (although I have it open locally).

PR #557, opened 2019-09-23
PLANNER-1666 Add min() and max() UniCollector

triceo:
Some questions to answer:

- Is this kind of `min` or `max` valid on bi-streams and higher? What would it do there?
- The collectors return null when the result makes no sense. (Such as "minimum of empty stream.") My assumption is that we would never get to see those situations, since an empty stream would never call the collector in the first place. Is the assumption correct?
- Does this help to address some of the feature gaps we have in the implementation currently? If so, what examples could we improve with this?

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -196,22 +196,38 @@ public final class ConstraintCollectors {
     }
 
     // ************************************************************************
-    // max
+    // min
     // ************************************************************************
 
-    public static <A> UniConstraintCollector<A, SortedMap<A, Long>, A> max(Comparator<A> comparator) {
+    private static <A> UniConstraintCollector<A, SortedMap<A, Long>, A> minOrMax(Comparator<A> comparator,
+            boolean min) {
+        Function<SortedMap<A, Long>, A> keySupplier = min ? SortedMap::firstKey : SortedMap::lastKey;
         return new DefaultUniConstraintCollector<>(
                 () -> new TreeMap<>(comparator),
                 (resultContainer, a) -> {
                     resultContainer.compute(a, (key, value) -> value == null ? 1 : value + 1);
-                    return (() -> {
-                        resultContainer.compute(a, (key, value) -> value == 1 ? null : value - 1);
-                    });
+                    return (() -> resultContainer.compute(a, (key, value) -> value == 1 ? null : value - 1));
                 },
-                (resultContainer) -> resultContainer.size() == 0 ? null : resultContainer.lastKey());
+                (resultContainer) -> resultContainer.size() == 0 ? null : keySupplier.apply(resultContainer));
     }
 
-    public static <A extends Comparable<A>> UniConstraintCollector<A, SortedMap<A, Long>, A> max() {
+    public static <A> UniConstraintCollector<A, ?, A> min(Comparator<A> comparator) {
+        return minOrMax(comparator, true);
+    }
+
+    public static <A extends Comparable<A>> UniConstraintCollector<A, ?, A> min() {
+        return min(Comparable::compareTo);
+    }
+
+    // ************************************************************************
+    // max
+    // ************************************************************************
+
+    public static <A> UniConstraintCollector<A, ?, A> max(Comparator<A> comparator) {
+        return minOrMax(comparator, false);
+    }
+
+    public static <A extends Comparable<A>> UniConstraintCollector<A, ?, A> max() {
         return max(Comparable::compareTo);
     }
 
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -199,18 +199,6 @@ public final class ConstraintCollectors {
     // min
     // ************************************************************************
 
-    private static <A> UniConstraintCollector<A, SortedMap<A, Long>, A> minOrMax(Comparator<A> comparator,
-            boolean min) {
-        Function<SortedMap<A, Long>, A> keySupplier = min ? SortedMap::firstKey : SortedMap::lastKey;
-        return new DefaultUniConstraintCollector<>(
-                () -> new TreeMap<>(comparator),
-                (resultContainer, a) -> {
-                    resultContainer.compute(a, (key, value) -> value == null ? 1 : value + 1);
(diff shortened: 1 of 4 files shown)


triceo:
Also @rsynek could be interested.

rsynek:
Discussed in person, the machine reassignment example would require `max()` on `BiStream`.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -192,6 +195,42 @@
                 resultContainer -> resultContainer[0]);
     }
 
+    // ************************************************************************
+    // min
+    // ************************************************************************
+
+    private static <A> UniConstraintCollector<A, SortedMap<A, Long>, A> minOrMax(Comparator<A> comparator,
```
ge0ffrey: code style: method ordening: normally we put the private method that are used by other methods to work just below the last method that uses it.
This way, reading from top to bottom, the external facing methods come first, but the private methods still stick close to where they are used (= vertical density)

Can we move this minOrMax method below the last max method?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintCollector.java
```diff
@@ -53,7 +53,7 @@
 
     /**
      * A lambda that converts the result container into the result.
-     * @return never null
+     * @return null when the result would be invalid, such as maximum value from an empty container.
```
ge0ffrey: Good point. Now I understand.
The user never calls this method - and the user can never run into an NPE with it due to the groupBy() way of using it, but this javadoc is indeed correct and it's an important fact.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Comparator;
+
+import org.junit.Test;
+import org.optaplanner.core.api.score.stream.uni.UniConstraintCollector;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConstraintCollectorsTest {
+
+    private <A, B, C> Runnable accumulate(UniConstraintCollector<A, B, C> collector, Object container, A value) {
```
ge0ffrey: code style, let's move these down too

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -0,0 +1,127 @@
...
+        return collector.accumulator().apply((B) container, value);
+    }
+
+    private <A, B, C> void assertResult(UniConstraintCollector<A, B, C> collector, Object container, C expectedResult) {
+        C actualResult = collector.finisher().apply((B) container);
+        assertEquals("Collector (" + collector + ") did not produced expected result.", expectedResult, actualResult);
+    }
+
+    @Test
+    public void maxComparable() {
```
ge0ffrey: in ConstraintCollectors the order is "min, max", here the order is "max, min".
Can we keep the same order?

Can we also add the
    // max

Can we open an issue to also test the others (count, countDistinct, sum, etc)?

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
```diff
@@ -387,6 +392,64 @@ public void groupBy_1Mapping1Collector_countDistinct() {
...
+                    .penalize(TEST_CONSTRAINT_NAME, SimpleScore.ONE, (entityGroup, min) -> min.getIntegerProperty());
+        });
+
+        // From scratch
+        scoreDirector.setWorkingSolution(solution);
+        assertScore(scoreDirector,
+                assertMatchWithScore(-1_000_000, entityGroup1, entity1),
+                assertMatchWithScore(-1, firstEntity.getEntityGroup(), firstEntity));
+
+        // Now remove all entities and see that the score is zero. (= all matches removed)
```
ge0ffrey: typo: dot should be behind braces

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -192,6 +195,42 @@
...
+            boolean min) {
+        Function<SortedMap<A, Long>, A> keySupplier = min ? SortedMap::firstKey : SortedMap::lastKey;
+        return new DefaultUniConstraintCollector<>(
+                () -> new TreeMap<>(comparator),
+                (resultContainer, a) -> {
+                    resultContainer.compute(a, (key, value) -> value == null ? 1 : value + 1);
```
ge0ffrey: This leaves null entries in the map, which is effectively a memory leak.

Example:
Map starts out as {}
insert "Apple": {"Apple":1}
insert "Orange": {"Apple":1, "Orange":1}
insert "Apple": {"Apple":2, "Orange":1}
remove "Orange": {"Apple":2, "Orange":null}
...
A problem fact change might have removed the Orange, but it will linger in memory forever.

ge0ffrey:
> The collectors return null when the result makes no sense. (Such as "minimum of empty stream.") My assumption is that we would never get to see those situations, since an empty stream would never call the collector in the first place. Is the assumption correct?
Yes, that is correct.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: I don't think that this is the case. If it were, the `finisher()` would not work. It checks for `size() == 0`, and the tests make sure that this is, in fact, the case.

Therefore I am assuming that returning `null` from `map.compute()` does, in fact, remove the entry. But I will play around with the `Map` to verify the assumption.

ge0ffrey: You're right, no need to check:
https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#compute-K-java.util.function.BiFunction-
"If the function returns null, the mapping is removed (or remains absent if initially absent)."

ge0ffrey: I suspect there is more code in optaplanner that does map.remove or removeKey but could benefit from this.

triceo:
@ge0ffrey Please squash when merging.

ge0ffrey:
Sorry, I was too quick to merge (I didn't squash)

PR #560, opened 2019-09-24
PLANNER-1667 Add code coverage for more ConstraintCollectors

triceo:
This PR introduces some code sharing and a lot of test coverage.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -21,12 +21,14 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Function;
 import java.util.function.ToIntBiFunction;
 import java.util.function.ToIntFunction;
 import java.util.function.ToLongBiFunction;
 import java.util.function.ToLongFunction;
 
+import com.google.common.base.Functions;
 import org.optaplanner.core.api.score.stream.bi.BiConstraintCollector;
 import org.optaplanner.core.api.score.stream.uni.UniConstraintCollector;
 import org.optaplanner.core.api.score.stream.uni.UniConstraintStream;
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -44,107 +46,82 @@
...
-                () -> new long[1],
+                LongAdder::new,
                 (resultContainer, a) -> {
-                    resultContainer[0]++;
-                    return (() -> resultContainer[0]--);
+                    resultContainer.increment();
+                    return (resultContainer::decrement);
                 },
-                resultContainer -> resultContainer[0]);
+                resultContainer -> countConverter.apply(resultContainer.longValue()));
```
triceo: The unifying theme of this PR is that the `int[1]` result container is replaced with `LongAdder` result container.
This way, it becomes immediately obvious what the code does - in my opinion, the single-element array did not do that.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -153,46 +130,46 @@
...
-                    resultContainer[0] += value;
-                    return (() -> resultContainer[0] -= value);
+                    resultContainer.add(value);
+                    return (() -> resultContainer.add(-value));
                 },
-                resultContainer -> resultContainer[0]);
+                LongAdder::intValue);
     }
 
     public static <A, B> BiConstraintCollector<A, B, ?, Long> sumLong(ToLongBiFunction<? super A, ? super B> groupValueMapping) {
```
triceo: Unfortunately, I cannot easily share code between `sum()` and `sumLong()` - there is no good way to unify `ToLongBiFunction` and `ToIntBiFunction` into some other generic function that would be able to take either `Long` or `Integer`.

Or rather there is a way - but the resulting code would become kinda cryptic. I decided to keep this clean, repetitious code.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -44,107 +46,82 @@
...
-                        }
-                        objectCount2[0]--;
-                        if (objectCount2[0] == 0) {
-                            resultContainer.objectCountMap.remove(value);
-                            resultContainer.count--;
-                        }
-                    });
-                },
-                resultContainer -> resultContainer.count);
+        return ConstraintCollectors.countDistinctLong(groupValueMapping, Long::intValue);
```
ge0ffrey: cast back and forth to longs might impact performance
but definitely impacts behavior when overflowing:
2 000 000 000 + 1 000 000 000 will result in a postive number after this change, instead of a negative number (which is easier to spot for overflow)

triceo:
I restored the original collector code. Therefore this PR is only about test coverage now.

PR #562, opened 2019-09-30
DroolsConstraintStream impl

ge0ffrey:
from, filter and scoring.
Known issue: 
This test still fails:
org.optaplanner.core.api.score.stream.uni.UniConstraintStreamTest#filter_problemFact
if AbstractConstraintStreamTest is set to include drools.


ge0ffrey:
We can use JUnit assumeTrue() to ignore tests that aren't supported yet in bavet or drools's impl, so the others can still run.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintStreamImplType.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.api.score.stream;
+
+public enum ConstraintStreamImplType {
```
triceo: We should add some Javadocs here, specifically:

- that both implementations are incomplete,
- and that for the foreseeable future, we would be spending our energy on the Drools impl.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -541,7 +558,8 @@ public ScoreDefinition buildDeprecatedScoreDefinition() {
                     "constraintProviderClass", this.constraintProviderClass);
             ConfigUtils.applyCustomProperties(constraintProvider, "constraintProviderClass",
                     constraintProviderCustomProperties, "constraintProviderCustomProperties");
-            return new ConstraintStreamScoreDirectorFactory<>(solutionDescriptor, constraintProvider);
+            ConstraintStreamImplType constraintStreamImplType_ = defaultIfNull(constraintStreamImplType, ConstraintStreamImplType.BAVET);
```
triceo: IMHO Drools should be the default as, in terms of features, it is expected to be more complete sooner.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSession.java
```diff
@@ -26,7 +26,7 @@
 /**
  * An instance of this class must be used in only one thread.
  */
-public interface ConstraintSession<Solution_>  {
+public interface ConstraintSession<Solution_> {
```
triceo: Shouldn't this rather implement `Closeable` or `AutoCloseable`? That would result in the IDE giving useful tips regarding forgetting to close it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -187,6 +186,9 @@ public void transitionTuple(BavetAbstractTuple tuple, BavetTupleState newState)
         return indictmentMap;
     }
 
+    @Override
+    public void close() {}
```
triceo: I'd add a comment saying that this is intentionally empty.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -137,6 +138,49 @@ private ConstraintMatchTotal findConstraintMatchTotal(RuleContext kcontext) {
...
     }
 
+    /**
+     * For internal use only, use penalize() or reward() instead.
+     * @param kcontext never null
+     */
+    public void impactScore(RuleContext kcontext) {
+        throw new UnsupportedOperationException("In the rule (" + kcontext.getRule().getName()
+                + "), the scoreHolder class (" + getClass()
+                + ") does not support an no weightMultiplier.");
```
triceo: `does not support an no`? How about `requires a` instead?

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
ge0ffrey: Let's discuss this when Drools surpasses Bavet in features.

triceo:
> We can use JUnit assumeTrue() to ignore tests that aren't supported yet in bavet or drools's impl, so the others can still run.
> 
> Or is there a similar ignoreIfTrue(boolean) construct?

`assumeTrue()` will work just fine. I'm not aware of anything else in JUnit 4. (JUnit 5 gives some other options.)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintStreamImplType.java
ge0ffrey: I'd leave that out of the javadocs, as it's too inaccurate for javadocs. We can mention it in the docs. (far more people read the docs than the javadocs anyway)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSession.java
ge0ffrey: agreed

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintFactory.java
```diff
@@ -0,0 +1,121 @@
...
+                throw new IllegalStateException("The constraint (" + constraint.getConstraintId()
+                        + ") must be created from the same constraintFactory.");
+            }
+            boolean added = constraintIdSet.add(constraint.getConstraintId());
+            if (!added) {
+                throw new IllegalStateException(
+                        "There are 2 constraints with the same constraintName (" + constraint.getConstraintName()
+                                + ") in the same constraintPackage (" + constraint.getConstraintPackage() + ").");
+            }
+            DroolsConstraint<Solution_> droolsConstraint = (DroolsConstraint) constraint;
```
rsynek: soft suggestion: instead of casting to `DroolsConstraint`, we could make the `Constraint[]` a type parameter in the abstract class and get `DroolsConstraint[]` directly here.

ge0ffrey:
Merged. @rsynek Lukas said he's process your comments in his work based on top of this merged PR.

PR #564, opened 2019-10-01
BAQE-1114 - Change revapi to check against 7.26.0.Final

MarianMacik:
depends on kiegroup/droolsjbpm-build-bootstrap#1070


MarianMacik:
jenkins retest this

PR #565, opened 2019-10-01
PLANNER-1668 Remove facts from the session properly

triceo:
@ge0ffrey Please squash when merging.

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
@@ -20,6 +20,7 @@ import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.function.Function;
 
+import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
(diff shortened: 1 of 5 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
```diff
@@ -53,6 +54,11 @@ public BiConstraintStreamTest(boolean constraintMatchEnabled, ConstraintStreamIm
     // Filter
     // ************************************************************************
 
+    @Before
+    public void assumeBavet() {
+        super.assumeBavet(); // Skip tests if we're not using Bavet
```
ge0ffrey: Nice to have:
the superclasses method and this method have the same name, but they are different:
- one is the check if it's bavet
- the other actives that check for all tests in this subclass

Maybe we can rename the second method for disableAllTestsForDrools (or a better name)?

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
```diff
@@ -60,6 +60,7 @@ public UniConstraintStreamTest(boolean constraintMatchEnabled, ConstraintStreamI
 
     @Test
     public void filter_problemFact() {
+        assumeBavet();
```
yurloc: `@Before ... setUp() { assumeBavet(); }` isn't suitable for this class?

triceo: Unfortunately, it is not. In this class, some tests already support both implementations.

PR #566, opened 2019-10-01
PLANNER-1669 Drools constraint streams now properly undo matches

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
```diff
@@ -45,6 +47,7 @@ public DroolsConstraintSessionFactory(SolutionDescriptor<Solution_> solutionDesc
     @Override
     public ConstraintSession<Solution_> buildSession(boolean constraintMatchEnabled, Solution_ workingSolution) {
         KieSession kieSession = kieBase.newKieSession();
+        ((RuleEventManager) kieSession).addEventListener(new OptaPlannerRuleEventListener()); // Enables undo in rules
```
triceo: @ge0ffrey I don't really like the fact that I'm referencing an implementation class from a different unrelated package. That said, I couldn't find a better place for it - suggestions?

triceo: (At the very least, I think it should be together with the actual undo listener. But that is in the public API, so I can not leave this listener there.)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -104,6 +107,14 @@ public void configureConstraintWeight(Rule rule, Score_ constraintWeight) {
         }
     }
 
+    /**
+     * Requires @{@link OptaPlannerRuleEventListener} to be added as event listener on {@link KieSession}, otherwise the
+     * score changes caused by the constraint matches would not be undone. See
+     * {@link DroolsScoreDirector#resetKieSession()} for an example.
+     * @param kcontext The rule for which to register the match.
```
ge0ffrey: nitpick (in according to sun's javadoc styleguides):
"The rule for which to register the match." is not a sentence - it doesn't have a verb in the main part - so it shouldn't  end with a dot. It's ok for javadocs parameters and return types not to be sentances - but a senstence-ending character (. or ? or !) can only used if it's a sentence.
=> @param kcontext The rule for which to register the match

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
ge0ffrey: This is a good start, so I wouldn't stop the merge, but I agree.
Both OptaPlannerRuleEventListener and GLOBAL_SCORE_HOLDER_KEY (in DroolsScoreDirector) are used by both drl and CS impl. We'll probably want to move them into for example the package org.optaplanner.core.impl.score.drools ? Definitively impl indeed.

ge0ffrey: ConstraintActivationUnMatchListener can also go into org.optaplanner.core.impl.score.drools, in 8.0. Add it to https://issues.jboss.org/browse/PLANNER-750 so we don't forget to do that.

PR #567, opened 2019-10-01
PLANNER-1638 Migrate TerminationTest to upstream

michaltomco:
The only class from the TerminationTest not yet covered in the upstream was the AndCompositeTeminationTest, which is covered in this commit.


ge0ffrey:
LGTM merging (improvements can come in separate PR)
As for the double's - the javadoc of org.optaplanner.core.impl.solver.termination.Termination#calculateSolverTimeGradient already covers what it is. So, my opinion:
- Comments to explain what the double's is or do, are duplication, so rather not (as such comments tend to rot)
- Comments to explain the math behind a combination of specific doubles in the test, are welcome, for example "// max(0.3, 0.5) = 0.5"

PR #568, opened 2019-10-02
PLANNER-1670 Implement Drools BiStreams

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintFactory.java
```diff
@@ -63,7 +68,15 @@ public DroolsConstraintFactory(SolutionDescriptor<Solution_> solutionDescriptor)
...
     @Override
     public <A> BiConstraintStream<A, A> fromUniquePair(Class<A> fromClass, BiJoiner<A, A> joiner) {
-        throw new UnsupportedOperationException();
+        MemberAccessor planningIdMemberAccessor = ConfigUtils.findPlanningIdMemberAccessor(fromClass);
+        if (planningIdMemberAccessor == null) {
+            throw new IllegalArgumentException("The fromClass (" + fromClass + ") has no member with a @"
+                    + PlanningId.class.getSimpleName() + " annotation,"
+                    + " so the pairs can be made unique ([A,B] vs [B,A]).");
+        }
+        // TODO Breaks node sharing + involves unneeded indirection
```
triceo: The TODO likely doesn't apply. I copy-pasted this code from Bavet impl.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsAbstractUniConstraintStream.java
```diff
@@ -38,12 +38,14 @@
...
 import org.optaplanner.core.impl.score.stream.drools.DroolsConstraintFactory;
+import org.optaplanner.core.impl.score.stream.drools.bi.DroolsAbstractBiConstraintStream;
+import org.optaplanner.core.impl.score.stream.drools.bi.DroolsJoinBiConstraintStream;
 import org.optaplanner.core.impl.score.stream.drools.common.DroolsAbstractConstraintStream;
 
 public abstract class DroolsAbstractUniConstraintStream<Solution_, A> extends DroolsAbstractConstraintStream<Solution_>
         implements UniConstraintStream<A> {
 
-    protected final List<DroolsAbstractUniConstraintStream<Solution_, A>> childStreamList = new ArrayList<>(2);
+    protected final List<DroolsAbstractConstraintStream<Solution_>> childStreamList = new ArrayList<>(2);
```
rsynek: Why do we initialize the ArrayList with the exact size of two? Is it guaranteed, that the size will never grow beyond this value?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsScoringBiConstraintStream.java
```diff
@@ -0,0 +1,161 @@
...
+        ruleItemBuilderList.add(getRightPattern());
+        Declaration<A> aVar = getLeftVariableDeclaration();
+        Declaration<B> bVar = getRightVariableDeclaration();
+        ConsequenceBuilder._3<? extends AbstractScoreHolder, A, B> consequence;
+        if (intMatchWeigher != null) {
+            consequence = PatternDSL.on(scoreHolderGlobal, aVar, bVar)
+                    .execute((drools, scoreHolder, a, b) -> {
+                        RuleContext kcontext = (RuleContext) drools;
+                        int weightMultiplier = intMatchWeigher.applyAsInt(a, b);
+                        scoreHolder.impactScore(kcontext, weightMultiplier);
```
rsynek: So in the end the decision was to use the `ScoreHolder` instead of `ScoreInliner`?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
```diff
@@ -0,0 +1,127 @@
...
+    public Declaration<B> getRightVariableDeclaration() {
+        return rightParentStream.getVariableDeclaration();
+    }
+
+    @Override
+    public PatternDSL.PatternDef<B> getRightPattern() {
+        return rightPattern;
+    }
+
+    private static boolean matches(Object left, Object right, JoinerType type) {
```
rsynek: Suggestion: how about  putting the matching logic into the `JoinerType` enum? Is it going to be reused for `TriStreams` too?

Another option is a hierarchy of matcher classes per JoinerType.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsAbstractUniConstraintStream.java
triceo: To be honest, I don't know. It's copy-pasted from the Bavet impl.

The size will often be 1. In less frequent cases, 2 or 0. 3, very unlikely. Seems to be a reasonable size.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsScoringBiConstraintStream.java
triceo: Yep, Geoffrey couldn't make it work otherwise. We may revisit in the future.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
triceo: I'm going to move the code if I need to reuse it. Right now, I didn't really think about `TriStream`s, but the code will definitely evolve as needed.

triceo:
I have now pushed a massive deduplication of all score impacts in all constraint streams.
Let me know if you prefer that I submit this change as a separate PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
```diff
@@ -144,133 +144,147 @@ public void addChildStream(BavetAbstractUniConstraintStream<Solution_, A> childS
     }
 
     // ************************************************************************
-    // Penalize/reward
+    // Penalize/impactScore
```
ge0ffrey: Penalize without reward is like ying without yang. So let's not remove "reward"

Users should only use penalize and reward, never impactScore.
Both penalize and reward only accept positive matchWeights.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
```diff
@@ -144,133 +144,147 @@ public void addChildStream(BavetAbstractUniConstraintStream<Solution_, A> childS
...
-    }
-
     @Override
     public Constraint penalize(String constraintPackage, String constraintName, Score<?> constraintWeight, ToIntFunction<A> matchWeigher) {
-        BavetConstraint<Solution_> constraint = buildConstraint(constraintPackage, constraintName, constraintWeight, false);
-        BavetScoringUniConstraintStream<Solution_, A> stream = new BavetScoringUniConstraintStream<>(constraintFactory, this, constraint, matchWeigher);
-        childStreamList.add(stream);
-        return constraint;
+        return impactScore(constraintPackage, constraintName, constraintWeight, matchWeigher, false);
+
```
ge0ffrey: code style: remove empty line

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
```diff
@@ -144,133 +144,147 @@ public void addChildStream(BavetAbstractUniConstraintStream<Solution_, A> childS
...
+    }
+
+    @Override
+    public Constraint rewardConfigurableBigDecimal(String constraintPackage, String constraintName, Function<A, BigDecimal> matchWeigher) {
+        return impactScoreConfigurableBigDecimal(constraintPackage, constraintName, matchWeigher, true);
+    }
+
+    @Override
+    protected Constraint impactScore(String constraintPackage, String constraintName, Score<?> constraintWeight,
+            boolean positively) {
```
ge0ffrey: Consistency: boolean "postively" has a different name than boolean "positive" in
   buildConstraintConfigurable(String constraintPackage, String constraintName, boolean positive) {

They should have the same name, because they are the same semantic.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintStream.java
```diff
@@ -84,4 +85,31 @@ protected void validateConstraintId(String constraintPackage, String constraintN
 
     public abstract InnerConstraintFactory<Solution_> getConstraintFactory();
 
+    @Override
+    public Constraint penalize(String constraintPackage, String constraintName, Score<?> constraintWeight) {
+        return impactScore(constraintPackage, constraintName, constraintWeight, false);
+    }
+
+
```
ge0ffrey: Code style: empty line

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintFactory.java
ge0ffrey: Sounds like a good unit test - which drools could win while bavet wouldn't.

Ignored tests can show how the impl differ in completeness :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsAbstractBiConstraintStream.java
```diff
@@ -0,0 +1,330 @@
...
+    public List<DroolsFromUniConstraintStream<Solution_, Object>> getFromStreamList() {
+        if (parent == null) {
+            DroolsJoinBiConstraintStream<Solution_, A, B> joinStream =
+                    (DroolsJoinBiConstraintStream<Solution_, A, B>) this;
+            List<DroolsFromUniConstraintStream<Solution_, Object>> leftParentFromStreamList =
+                    joinStream.getLeftParentStream().getFromStreamList();
+            List<DroolsFromUniConstraintStream<Solution_, Object>> rightParentFromStreamList =
+                    joinStream.getRightParentStream().getFromStreamList();
+            return Stream.concat(leftParentFromStreamList.stream(), rightParentFromStreamList.stream())
+                    .distinct()
```
ge0ffrey: 1) Should the distinct happen here? I doubt it:
I'd expect the first caller of getFromStreamList() to have to do the distinct too, so there it's a perf loss to do it in between too.

2) Is the distinct worth anything? Do the From's implement equals/hashcode? When in doubt, leave it out and put in a "TODO Should we distinct?".

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
```diff
@@ -0,0 +1,142 @@
...
+    private final PatternDSL.PatternDef<B> rightPattern;
+
+    public DroolsJoinBiConstraintStream(DroolsConstraintFactory<Solution_> constraintFactory,
+            DroolsAbstractUniConstraintStream<Solution_, A> parent,
+            DroolsAbstractUniConstraintStream<Solution_, B> otherStream, BiJoiner<A, B> biJoiner) {
+        super(constraintFactory, null);
+        this.leftParentStream = parent;
+        this.rightParentStream = otherStream;
+        this.biJoiner = (AbstractBiJoiner<A, B>) biJoiner;
+        this.rightPattern = otherStream.getPattern().expr(getLeftVariableDeclaration(), (b, a) -> matches(a, b));
```
ge0ffrey: This is not as performant as it could be.
See below

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
```diff
@@ -0,0 +1,142 @@
...
+            case DISJOINT:
+                return left.stream().noneMatch(right::contains) && right.stream().noneMatch(left::contains);
+            case INTERSECTING:
+                return left.stream().anyMatch(right::contains) || right.stream().anyMatch(left::contains);
+            default:
+                throw new IllegalStateException("Unsupported joiner type (" + type + ").");
+        }
+    }
+
+    private boolean matches(A left, B right) {
```
ge0ffrey: This is is not as performant as it could be.
It could use inlining:
Based on the joinerType(s), we can build a different lambda, to bring the switch(joinerType) cost to construction time.
Doesn't have to happen in this PR, but let's add a TODO in the constructor (at my other comment)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
```diff
@@ -0,0 +1,142 @@
...
+            case GREATER_THAN:
+                return comparison > 0;
+            case GREATER_THAN_OR_EQUAL:
+                return comparison >= 0;
+            default:
+                throw new IllegalStateException("Unsupported joiner type (" + type + ").");
+        }
+    }
+
+    private static boolean matchesCollection(Collection left, Collection right, JoinerType type) {
```
ge0ffrey: code style nitpick: would you mind if we name the JoinerType "joinerType" instead of just "type"?
And "left" as "leftCollection" etc?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsAbstractUniConstraintStream.java
```diff
@@ -103,139 +108,171 @@ public DroolsAbstractUniConstraintStream(DroolsConstraintFactory<Solution_> cons
...
+            ToIntFunction<A> matchWeigher) {
+        return impactScoreConfigurable(constraintPackage, constraintName, matchWeigher, true);
+
+    }
+
+    @Override
+    public Constraint rewardConfigurableLong(String constraintPackage, String constraintName,
+            ToLongFunction<A> matchWeigher) {
+        return impactScoreConfigurableLong(constraintPackage, constraintName, matchWeigher, true);
+    }
```
ge0ffrey: Should we create an interface InnerUniConstraintStream
and have rewire all penalize and reward methods to impactScore with default methods
so that the Drools/Bavet impl's don't have to do it themselves?

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AbstractConstraintStreamTest.java
```diff
@@ -119,14 +115,29 @@ protected static AssertableMatch assertMatchWithScore(int score, Object... justi
...
+            if (score != ((SimpleScore) constraintMatch.getScore()).getScore()) {
+                return false;
+            }
+            List<Object> actualJustificationList = constraintMatch.getJustificationList();
+            if (actualJustificationList.size() != justificationList.size()) {
+                return false;
+            }
+            /*
+             * We can not simply compare the lists, since the elements may be in different orders. The order of these
+             * lists is not relevant.
```
ge0ffrey: Code style nitpick: Would you mind if we turn this into a 2 line comment line this:
// We can not ...
// lists is not relevant.
 Or even
// We can not ...
/// The order of ...

4 lines lost for a 2 line comment breaks the visual structure for me a bit.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsScoringUniConstraintStream.java
```diff
@@ -98,8 +98,9 @@ private DroolsScoringUniConstraintStream(DroolsConstraintFactory<Solution_> cons
 
     @Override
     public void createRuleItemBuilders(List<RuleItemBuilder<?>> ruleItemBuilderList,
-            Global<? extends AbstractScoreHolder> scoreHolderGlobal,
-            Declaration<A> aVar, PatternDSL.PatternDef<A> parentPattern) {
+            Global<? extends AbstractScoreHolder> scoreHolderGlobal) {
+        PatternDSL.PatternDef<A> parentPattern = getPattern();
+        Declaration<A> aVar = getVariableDeclaration();
```
ge0ffrey: Ah, this refactor is a problem :)
I intentionally passed the variable declarations - instead of fetching them like this - despite it causing less readable code. The motivation:

from(A.class) // Uni<A>
.groupBy(A::getB) // Uni<B>
.penalize(); // The variable declaration should be only B. The one for A got burried.

from(A.class) // Uni<A>
.join(B.class) // Bi<A, B>
.groupBy((A a, B b) -> a.getCMap().get(b), sum(a.price * b.vat)) // Bi<C, int>
.penalize // The variable declaration should be only C I think?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
triceo: Yeah, this is a mistake. It's a side-effect of mass-rename, shouldn't have happened.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsScoringUniConstraintStream.java
triceo: @ge0ffrey This is what I mean when I say we have several fundamentally different `groupBy` methods. I don't see a way around that, other than implementing them with different `ConstraintStream` implementation classes. I'll cross that bridge when I get there, OK?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
triceo: I probably don't understand what you mean. Do you have an example of code somewhere, where you do that already?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintFactory.java
triceo: Actually, I don't think we can test for this in Drools. What Drools does to the RETE network is an implementation detail that we don't really want to deal with. Personally, I don't think we should worry about that - we should write clean good rules, and the rest is Drools internal business.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
triceo: On second thought, I get it now. Although to fix this will make the code much less readable.

triceo: On secound thoght, I get it now. Although fixing it will make the code much less readable.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsScoringUniConstraintStream.java
ge0ffrey: I think you 'll have to undo this method refactor, the one done in this case, but I am ok with leaving it like this until then.

PR #572, opened 2019-10-04
integration with BPMN ideas + change DPI to 96

ge0ffrey:


ge0ffrey:
Agreed on Kogito/Quarkus - that's the way we'll get it done.

I wanted the "calculate distance matrix" step in there, even though it's a bit of an impl detail, because it's a challenging part which shows why it will take longer to implement than the other cases.

PR #573, opened 2019-10-07
enable SORT_PROPERTIES_ALPHABETICALLY for a deterministic order

contextshuffling:
Test `polymorphicScore()` uses `ObjectMapper` from `Jackson` to serialize and de-serialize a object. However, the `ObjectMapper` relies on the order of fields returned by `java.lang.Class.getDeclaredFields`. 
But there is no guarantee in the order returned by that API and test can fail for failing to de-serialize this. 
e.g., `java.lang.AssertionError: expected:<-1hard/-20soft> but was:<null>`
 
We previously send a PR for fixing some of tests in `jackson` library and realize there is also related to `polymorphicScore` test here.

In this PR, we propose to enable the `SORT_PROPERTIES_ALPHABETICALLY` feature so that the order is deterministic.

### DIFF
--- a/optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/OptaPlannerJacksonModuleTest.java
+++ b/optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/OptaPlannerJacksonModuleTest.java
@@ -16,6 +16,7 @@
 
 package org.optaplanner.persistence.jackson.api;
 
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Test;
 import org.optaplanner.core.api.score.Score;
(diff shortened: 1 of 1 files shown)


kie-ci9:

optaplanner-persistence/optaplanner-persistence-jackson/src/test/java/org/optaplanner/persistence/jackson/api/OptaPlannerJacksonModuleTest.java
```diff
@@ -29,6 +30,7 @@
     @Test
     public void polymorphicScore() {
         ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY);
```
triceo: @contextshuffling Would you mind adding a comment here, describing why this line is necessary?

ge0ffrey:
ok to test

contextshuffling:
@ge0ffrey According to the official specification (https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredMethods--), "**The elements in the returned array are not sorted and are not in any particular order.**"  Enabling this feature will make sure the test work as expected.

triceo:
@contextshuffling I'm closing this PR in favor of #602, which I hope to merge later today. The authorship of your commit will be preserved.
Thank you for your contribution to OptaPlanner!

PR #574, opened 2019-10-08
PLANNER-1671 Implement Tri constraint streams

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/DroolsScoringTriConstraintStream.java
```diff
@@ -0,0 +1,171 @@
...
+import org.optaplanner.core.api.function.ToLongTriFunction;
+import org.optaplanner.core.api.function.TriFunction;
+import org.optaplanner.core.api.score.holder.AbstractScoreHolder;
+import org.optaplanner.core.impl.score.stream.drools.DroolsConstraint;
+import org.optaplanner.core.impl.score.stream.drools.DroolsConstraintFactory;
+
+public final class DroolsScoringTriConstraintStream<Solution_, A, B, C>
+        extends DroolsAbstractTriConstraintStream<Solution_, A, B, C> {
+
+    private final DroolsConstraint<Solution_> constraint;
```
triceo: Interesting observation: none of the Drools-based CS (not uni, not bi, not tri) actually use `constraint`. Could that potentially have some implications, or can we just remove it?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/JoinerType.java
```diff
@@ -41,4 +51,34 @@ public JoinerType flip() {
         }
     }
 
+    public boolean matches(Object left, Object right) {
+        return matcher.test(left, right);
+    }
+
+    private static boolean lessThan(Comparable left, Object right) {
+        return left.compareTo(right) < 0;
```
ge0ffrey: why not simply inline this method? I don't think it's used somewhere else?

```
LESS_THAN((a, b) -> ((Comparable) a).compareTo(b) < 0),
LESS_THAN_OR_EQUAL((a, b) -> ((Comparable) a).compareTo(b) <= 0),
GREATER_THAN((a, b) -> ((Comparable) a).compareTo(b) > 0),
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/JoinerType.java
```diff
@@ -41,4 +51,34 @@ public JoinerType flip() {
...
+
+    private static boolean greaterThan(Comparable left, Object right) {
+        return left.compareTo(right) > 0;
+    }
+
+    private static boolean greaterThanOrEqual(Comparable left, Object right) {
+        return left.compareTo(right) >= 0;
+    }
+
+    private static boolean disjointColections(Collection leftCollection, Collection rightCollection) {
```
ge0ffrey: if the enum is called "DISJOINT", this method should be called "disjointColections(...)" for consistency, not "disjointColections(...)".

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/DroolsAbstractTriConstraintStream.java
```diff
@@ -0,0 +1,190 @@
...
+public abstract class DroolsAbstractTriConstraintStream<Solution_, A, B, C>
+        extends DroolsAbstractConstraintStream<Solution_> implements InnerTriConstraintStream<A, B, C> {
+
+    protected final DroolsAbstractTriConstraintStream<Solution_, A, B, C> parent;
+
+    public DroolsAbstractTriConstraintStream(DroolsConstraintFactory<Solution_> constraintFactory,
+            DroolsAbstractTriConstraintStream<Solution_, A, B, C> parent) {
+        super(constraintFactory);
+        if (parent == null && !(this instanceof DroolsJoinTriConstraintStream)) {
+            throw new IllegalArgumentException("Parent of stream (" + this + ") must not be null, or not join stream.");
```
ge0ffrey: typo: "not a join stream"
Actually, this sentence is incorrect: "Parent ... must not be null, or not join stream." implies "parent is not a join stream", while it should be "this is not a join stream"

I'd split this up into 2 separate if checks.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/DroolsJoinTriConstraintStream.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public Declaration<A> getAVariableDeclaration() {
+        return leftParentStream.getLeftVariableDeclaration();
+    }
+
+    @Override
+    public PatternDSL.PatternDef<A> getAPattern() {
+        return leftParentStream.getLeftPattern();
```
ge0ffrey: Doesn't DroolsAbstractBiConstraintStream have getAPattern()?

If Tri's have getAPattern(), getBPattern() and getCPattern()
then Bi's must have getAPattern(), getBPattern() - but not getLeftPattern() and getRightPattern()
and Uni's getAPattern()
and one day Quads will have getAPattern(), getBPattern() getCPattern() and getDPattern()

ge0ffrey: Notice that on Tri's there is no real "left" and "right": it is wrong to presume that the left is A+B and right is C, because the left might be A and the right might be C.
As for Bi's: they might be created from something else than a join (think a Tri.groupBy(key1, key2)), so the calling A left and calling B right starts breaking down.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/DroolsScoringTriConstraintStream.java
ge0ffrey: That's the magic of the kconstraint I guess :)
+0 to remove it, as you see fit

triceo:
@ge0ffrey Comments resolved, ready to merge unless there are more.

PR #575, opened 2019-10-11
PLANNER-1676 Simple groupBy()

triceo:
Feature complete. Still a bit drafty, but ready for review.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/UniAnchor.java
```diff
@@ -0,0 +1,177 @@
...
+import org.optaplanner.core.impl.score.stream.common.JoinerType;
+import org.optaplanner.core.impl.score.stream.drools.bi.BiAnchor;
+import org.optaplanner.core.impl.score.stream.drools.common.LogicalRuleMetadata;
+import org.optaplanner.core.impl.score.stream.drools.common.LogicalTuple;
+import org.optaplanner.core.impl.score.stream.drools.common.OriginalRuleMetadata;
+import org.optaplanner.core.impl.score.stream.drools.common.RuleMetadata;
+
+import static org.drools.model.DSL.on;
+
+public final class UniAnchor {
```
triceo: Not sure about the name, `Anchor`.

The reason why this class exists is generics. `GroupingStream` operates on variables of `A`, but its children naturally need to operate on `LogicalTuple`. This was causing all kinds of trouble with inheritance, so I encapsulated all the generic mess into `Anchor` which - from the outside - is still as much type-safe as possible.

Now, the `Anchor`s do most of the work.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AbstractConstraintStreamTest.java
```diff
@@ -56,6 +60,11 @@ protected void assumeBavet() {
                 constraintStreamImplType == ConstraintStreamImplType.BAVET);
     }
 
+    protected void assumeDrools() {
+        Assume.assumeTrue("This functionality is not yet supported in Bavet constraint streams.",
+                constraintStreamImplType == ConstraintStreamImplType.DROOLS);
```
triceo: This PR marks the moment where Drools-based CS learned something that Bavet can not do yet. :-)

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
```diff
@@ -709,6 +768,11 @@ public void globalNodeOrder() {
 
     @Test
     public void zeroConstraintWeightDisabled() {
+        /*
+         * This may never be possible for Drools. If implemented, please remember that some rules may be shared by
+         * different scoring streams, and therefore the rules can only be disabled when all the relevant scoring streams
+         * have their weights set to zero.
+         */
```
triceo: This is an important note. That said, I may have an idea how to make the rule disabling work - but we'll have to discuss the impacts. It's possible it would not be reasonable, performance-wise.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+        return streams;
+    }
+
+    /**
+     * Retrieves Drools rules required to process this constraint.
+     *
+     * @param ruleLibrary never null. Cache of rules already generated by previous constraints. This method will use
+     * {@link Map#computeIfAbsent(Object, Function)} to add rules from {@link DroolsAbstractConstraintStream}s which we
+     * have not yet processed. This way, we will only generate one rule per stream, allowing Drools to reuse the
+     * computations.
```
triceo: This is a neat little optimization. Let me know if I need to explain more and/or better.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/RuleMetadata.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    static <A> OriginalRuleMetadata<A> of(Declaration<A> variableDeclaration, PatternDSL.PatternDef<A> pattern) {
+        return new OriginalRuleMetadata<>(variableDeclaration, pattern);
+    }
+
+    Declaration<FactType> getVariableDeclaration();
+
+    PatternDSL.PatternDef<FactType> getPattern();
+}
```
triceo: This interface (and its two implementors) exists so that we can enforce some type-safety within `Anchor`s - specifically, that the variable and its pattern are always of the same type. Code that keeps and treats these items separately has proven to be hard to understand.

triceo:
@ge0ffrey I have a much better impl (cleaner, more type-safe, and better maintainable going forward) of this code on my local branch. Unfortunately, I have a problem - it doesn't work unless I can get the real `Class` of `GroupKey_` generic type.

How much would you be against changing the `groupBy()` public API (currently tech preview) to force the users to provide the class?

 `stream.groupBy(Something::getSomethineElse)` would become `stream.groupBy(Something::getSomethineElse, SomethingElse.class)`.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +187,109 @@
     // Group by
     // ************************************************************************
 
-    <GroupKey_, ResultContainer_, Result_> UniConstraintStream<Result_> groupBy(
+    /**
+     * Runs all items in the stream through a given @{@link UniConstraintCollector} and converts them into a new
+     * {@link UniConstraintStream} which only has one item in it, the result of applying
+     * {@link UniConstraintCollector#finisher()}.
```
ge0ffrey: How we figure out a good example format in a bluejeans meeting and add an example in every javadoc?

I am thinking something like:

```
[expanation]
<p>
For example, a Uni<Person> with elements [Ann(age 100), Beth (age 20), Carl (age 3)] groupBy sum(Person::getAge) returns a Uni<Integer> stream of [123]. 
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +187,109 @@
...
+     * @param <GroupKey_> the type of a match fact in the destination {@link UniConstraintStream}.
+     * @return never null.
+     */
+    <GroupKey_> UniConstraintStream<GroupKey_> groupBy(Function<A, GroupKey_> groupKeyMapping);
+
+    /**
+     * Convert the {@link UniConstraintStream} to a {@link BiConstraintStream}, consisting of:
+     *
+     * <ul>
+     *     <li>on the left, the set of values resulting from applying the group key mapping function on all items in
```
ge0ffrey: Sentences start with capital

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
```diff
@@ -16,6 +16,50 @@
 
 /**
  * The {@link org.optaplanner.core.api.score.stream.ConstraintStream} API:
- * a way to define constraints for {@link org.optaplanner.core.api.score.Score} calculation.
+ * a way to define constraints for {@link org.optaplanner.core.api.score.Score} calculation. Throughout this
+ * documentation, we will be using the following terminology:
```
ge0ffrey: This is pointless. Nobody reads the package javadocs :(

Defining this terminology can happen in the user documentation, as they 'll read that diagonally first, before coding and seeing javadocs. Let's remove it here and move it to there, as it's a maintenance burden here.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
```diff
@@ -16,6 +16,50 @@
...
+ *
+ * <dl>
+ *     <dt>Stream</dt>
```
ge0ffrey: Let's never shorten ConstraintStream to Stream, to avoid confusion with java.util.Stream

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
```diff
@@ -16,6 +16,50 @@
...
+ *          <dd>Short for "constraint stream". A chain of different operations, originated by
+ *          {@link org.optaplanner.core.api.score.stream.ConstraintFactory#from(java.lang.Class)} (or similar methods)
+ *          and terminated by a penalization or reward operation.</dd>
+ *     <dt>Operation</dt>
+ *          <dd>Operations (implementations of
+ *          {@link org.optaplanner.core.api.score.stream.ConstraintStream}) are parts of Stream which mutate it. They
+ *          may remove tuples from further evaluation, expand or contract streams. Every Stream has a terminal
```
ge0ffrey: Did to simplify become to contract? (ok, just asking)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
```diff
@@ -16,6 +16,50 @@
...
+ *          Operation, which is either a penalization or a reward.</dd>
+ *     <dt>Fact</dt>
+ *          <dd>Object instance entering the Stream.</dd>
+ *     <dt>Genuine Fact</dt>
+ *          <dd>Fact that enters the Stream either through a from(...) call or through a join(...) call. Genuine Facts
+ *          are either planning entities (see {@link org.optaplanner.core.api.domain.entity.PlanningEntity}) or problem
+ *          facts (see @{@link org.optaplanner.core.api.domain.solution.drools.ProblemFactProperty} or
+ *          {@link org.optaplanner.core.api.domain.solution.drools.ProblemFactCollectionProperty}).</dd>
```
ge0ffrey: Ok, we do want this info in the javadoc. How about in ConstraintStream then?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
ge0ffrey: Update: Let's move it to ConstrainStream's javadoc.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
```diff
@@ -16,6 +16,50 @@
...
+ *     <dt>Inferred Fact</dt>
+ *          <dd>Fact that enters Stream through a computation. This would typically happen through an Operation such as
+ *          groupBy(...).</dd>
+ *     <dt>Tuple</dt>
+ *          <dd>A collection of Facts that Stream operates on. For example,
+ *          {@link org.optaplanner.core.api.score.stream.uni.UniConstraintStream} operates on single-Fact Tuples {A} and
+ *          {@link org.optaplanner.core.api.score.stream.bi.BiConstraintStream} operates on two-Fact Tuples {A, B}.
+ *          Putting Facts into a Tuple implies a relationship exists between these Facts.</dd>
```
ge0ffrey: Things starting with a capital must be classes. Tuple and Fact are not classes, so let's call them tuple and fact (no capital start).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/package-info.java
```diff
@@ -16,6 +16,50 @@
...
+ *     <dt>Match</dt>
+ *          <dd>Match is a Tuple that reached the terminal Operation of a Stream and is therefore either penalized or
+ *          rewarded.</dd>
+ *     <dt>Cardinality</dt>
+ *          <dd>The number of Facts in a Tuple. Uni Streams have a cardinality of 1, Bi Streams have a cardinality of 2,
+ *          etc.</dd>
+ *     <dt>Expansion</dt>
+ *          <dd>An operation that increases the cardinality of Stream. This typically happens through join(...) or
+ *          a groupBy(...) Operations. Uni Streams are expanded into Bi Streams, Bi Streams into Tri Streams etc.
+ *          Expanding Stream infers relationships between Facts where none previously existed.</dd>
```
ge0ffrey: "where not previously existed" is debatable.
In this case it makes sense:
  from(RoomPenality.class).join(Talk.class, equals(RoomPenality::getRoom, Talk::getRoom))

But in this case that relation is already there between RoomPenalty and room:
  from(RoomPenality.class).join(Room.class, equals(RoomPenality::getRoom, identity()))

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,109 @@
     // Group by
     // ************************************************************************
 
-    <GroupKey_, ResultContainer_, Result_> UniConstraintStream<Result_> groupBy(
+    /**
+     * Runs all items in the stream through a given @{@link UniConstraintCollector} and converts them into a new
+     * {@link UniConstraintStream} which only has one item in it, the result of applying
+     * {@link UniConstraintCollector#finisher()}.
```
ge0ffrey: the result of applying {@link UniConstraintCollector#finisher()}
-> 
the result of applying {@link UniConstraintCollector}

The finisher is a impl detail for 99% of our users (even though it's in the public api...)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,109 @@
...
+     *
+     * @param collector never null, the collector to perform the grouping operation with
+     * @param <ResultContainer_> the mutable accumulation type (often hidden as an implementation detail)
+     * @param <Result_> the type of a match fact in the destination {@link UniConstraintStream}
+     * @return never null.
```
ge0ffrey: no dot, it's not a sentence (no verb)
never null. -> never null

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,109 @@
...
+    <ResultContainer_, Result_> UniConstraintStream<Result_> groupBy(
             UniConstraintCollector<A, ResultContainer_, Result_> collector);
 
-    <GroupKey_> UniConstraintStream<GroupKey_> groupBy(
-            Function<A, GroupKey_> groupKeyMapping);
-
+    /**
+     * Convert the {@link UniConstraintStream} to a different {@link UniConstraintStream}, consisting of the set of
+     * values resulting from applying the group key mapping function on all items in the original stream. None of the
+     * items in the new stream will {@link Objects#equals(Object, Object)} any other.
```
ge0ffrey: Would you mind doing me a favor for maintenance and start new sensenance on a new line, both in javadoc and asciidoc?
In both cases, those newlines get replaced by a space in the output, so it becomes a paragraph again.
But for editing, at most one sentence per line, makes it really easy to change that line without having to adjust wrapping afterwards for the rest of that paragraph.

ge0ffrey: Also one sentance per line creates "ventilzated proze", see
https://asciidoctor.org/docs/asciidoc-recommended-practices/
https://vanemden.wordpress.com/2009/01/01/ventilated-prose/

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,109 @@
...
+     *     stream will {@link Objects#equals(Object, Object)} any other.</li>
+     * </ul>
+     *
+     * Here's an example. Assuming:
+     *
+     * <ul>
+     *     <li>The original {@link UniConstraintStream} contains items "X1", "X2", "Y1"</li>.
+     *     <li>The "A" group key mapping function returns "X" for "X1" and "X2", "Y" for "Y1".</li>
+     *     <li>The "B" group key mapping function returns "1" for "X1" and "Y1", "2" for "A2".</li>
+     * </ul>
```
ge0ffrey: So we remove this example?

If we keep it, let's use the syntax we agreed: {} for tuples, [] for elements

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,109 @@
...
+     */
     <GroupKey_, ResultContainer_, Result_> BiConstraintStream<GroupKey_, Result_> groupBy(
             Function<A, GroupKey_> groupKeyMapping,
             UniConstraintCollector<A, ResultContainer_, Result_> collector);
 
+    /**
+     * Convert the {@link UniConstraintStream} to a {@link BiConstraintStream}, consisting of:
+     *
+     * <ul>
+     *     <li>on the left, the set of values resulting from applying the "A" group key mapping function on all items in
```
ge0ffrey: We can't use "left" and "right" for BiStream here, because for TriStream due to a groupBy(getA, getB, getC), we can't use "left" and "right" either. (If the tristream is a join of A+B and C, then we can use "left" and "right").
We should use "A" or "first fact".

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+            streams.addAll(assembleAllStreams(fromStream));
         }
-        return PatternDSL.rule(constraintPackage, constraintName)
-                .build(ruleItemBuilderList.toArray(new RuleItemBuilder<?>[0]));
+        return streams;
+    }
+
+    private Collection<DroolsAbstractConstraintStream<Solution_>> assembleAllStreams(
+            DroolsAbstractConstraintStream<Solution_> parent) {
+        final Set<DroolsAbstractConstraintStream<Solution_>> streams = new LinkedHashSet<>();
```
ge0ffrey: Code consistency: OptaPlanner's code uses plural for arrays and collections, *List for lists and *Set for sets.
So "streams" -> "streamSet"

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+        for (DroolsAbstractConstraintStream<Solution_> child: parent.getChildStreams()) {
+            streams.addAll(assembleAllStreams(child));
+        }
+        return streams;
+    }
+
+    /**
+     * Retrieves Drools rules required to process this constraint.
+     *
+     * @param ruleLibrary never null. Cache of rules already generated by previous constraints. This method will use
```
ge0ffrey: will use -> use
(active tense)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+            Rule rule = ruleLibrary.computeIfAbsent(stream, key ->
+                    key.buildRule(this, scoreHolderGlobal).orElse(null));
+            if (rule != null) {
+                rules.add(rule);
+            }
+        });
+        int expectedRuleCount = rules.size();
+        if (expectedRuleCount == 0) {
+            throw new IllegalStateException("A constraint stream for constraint " + constraintPackage + " " +
+                    constraintName + " resulted in no rules.");
```
ge0ffrey: code consistency: () around the param

for constraint " + constraintPackage ... + " resulted
->
for constraint (" + constraintPackage ... + ") resulted

ge0ffrey: Use Constraint.getConstraintId(...) in there:
for constraint (" + Constraint.getConstraintId(...) + ") resulted

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -85,4 +127,9 @@ public String getConstraintName() {
         return constraintName;
     }
 
+    @Override
+    public String toString() {
+        return "DroolsConstraint(" + constraintPackage + " " + constraintName + ") in " + fromStreamList.size() +
```
ge0ffrey: Constraint.getConstraintId(...)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+     * @param ruleLibrary never null. Cache of rules already generated by previous constraints. This method will use
+     * {@link Map#computeIfAbsent(Object, Function)} to add rules from {@link DroolsAbstractConstraintStream}s which we
+     * have not yet processed. This way, we will only generate one rule per stream, allowing Drools to reuse the
+     * computations.
+     * @param scoreHolderGlobal never null. The Drools global used to track changes to score within rule consequences.
+     * @return never null. List of all Drools {@link Rule}s required by this constraint. May update ruleLibrary as a
+     * side effect.
+     */
+    public List<Rule> createRules(Map<DroolsAbstractConstraintStream<Solution_>, Rule> ruleLibrary,
+            Global<? extends AbstractScoreHolder> scoreHolderGlobal) {
```
ge0ffrey: I presume the ruleLibrary 's lifecycle scope is inclosed in the ScoreDirectorFactory creation?
AKA, building 2 ScoreDirectorFactory's means they both have their own ruleLibrary? (That's a good thing, there is only one ScoreDirectorFactory build anyway normally).

ge0ffrey: yes, just confirmed it

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+        Set<DroolsAbstractConstraintStream<Solution_>> streams = assembleAllStreams();
+        List<Rule> rules = new ArrayList<>();
+        streams.forEach(stream -> {
+            Rule rule = ruleLibrary.computeIfAbsent(stream, key ->
+                    key.buildRule(this, scoreHolderGlobal).orElse(null));
+            if (rule != null) {
+                rules.add(rule);
+            }
+        });
+        int expectedRuleCount = rules.size();
```
ge0ffrey: Why expected?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+        if (expectedRuleCount == 0) {
```
ge0ffrey: Should this really fail fast?
I don't think Bavet fails on having an empty Constraint array.
We should have a unit test for that for an empty constraint array.

I'd argue that the ruleList.size() <= streamSet.size()
(and both can be zero)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,13 +60,52 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+     * have not yet processed. This way, we will only generate one rule per stream, allowing Drools to reuse the
+     * computations.
+     * @param scoreHolderGlobal never null. The Drools global used to track changes to score within rule consequences.
+     * @return never null. List of all Drools {@link Rule}s required by this constraint. May update ruleLibrary as a
+     * side effect.
+     */
+    public List<Rule> createRules(Map<DroolsAbstractConstraintStream<Solution_>, Rule> ruleLibrary,
+            Global<? extends AbstractScoreHolder> scoreHolderGlobal) {
+        Set<DroolsAbstractConstraintStream<Solution_>> streams = assembleAllStreams();
+        List<Rule> rules = new ArrayList<>();
```
ge0ffrey: streamSet
ruleList

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/BiAnchor.java
```diff
@@ -0,0 +1,160 @@
...
+import org.optaplanner.core.impl.score.stream.drools.common.LogicalTuple;
+import org.optaplanner.core.impl.score.stream.drools.common.GenuineRuleMetadata;
+import org.optaplanner.core.impl.score.stream.drools.common.RuleMetadata;
+import org.optaplanner.core.impl.score.stream.drools.tri.TriAnchor;
+import org.optaplanner.core.impl.score.stream.drools.uni.UniAnchor;
+import org.optaplanner.core.impl.score.stream.tri.AbstractTriJoiner;
+
+import static org.drools.model.DSL.on;
+
+public final class BiAnchor {
```
ge0ffrey: Knee jerk reaction: Anchor is a name already used in the context of chained variables. Can we use anything else?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/GenuineRuleMetadata.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.drools.model.Declaration;
+import org.drools.model.PatternDSL;
+
+public final class GenuineRuleMetadata<A> implements RuleMetadata<A> {
+
+    private final Declaration<A> variableDeclaration;
+    private final PatternDSL.PatternDef<A> pattern;
+
+    GenuineRuleMetadata(Declaration<A> variableDeclaration, PatternDSL.PatternDef<A> pattern) {
```
ge0ffrey: code style: we don't use friendly constructors. either private, protected or public

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/TriAnchor.java
```diff
@@ -0,0 +1,142 @@
...
+                        (Declaration<C>) getCMetadata().getVariableDeclaration())
+                        .execute(consequenceImpl);
+        return Arrays.asList(getAMetadata().getPattern(), getBMetadata().getPattern(), getCMetadata().getPattern(),
+                consequence);
+    }
+
+    private static <A> A inline(Object item) {
+        if (item instanceof LogicalTuple) {
+            return ((LogicalTuple) item).getItem(0);
+        }
```
ge0ffrey: This isn't inlining.
The instanceof check happens for every match.

An inlining approach, would return a Supplier lamdba that returns the identity() if the metadata says it's a LocalTuple (note that there are no item instances yet at that time) and the Logicaltuple getItem() extraction lamdba otherwise.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/TriAnchor.java
```diff
@@ -0,0 +1,142 @@
...
+        ConsequenceBuilder._4<ScoreHolder, A, B, C> consequence =
+                on(scoreHolderGlobal, (Declaration<A>) getAMetadata().getVariableDeclaration(),
+                        (Declaration<B>) getBMetadata().getVariableDeclaration(),
+                        (Declaration<C>) getCMetadata().getVariableDeclaration())
+                        .execute(consequenceImpl);
+        return Arrays.asList(getAMetadata().getPattern(), getBMetadata().getPattern(), getCMetadata().getPattern(),
+                consequence);
+    }
+
+    private static <A> A inline(Object item) {
```
ge0ffrey: item -> fact?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsFilterUniConstraintStream.java
```diff
@@ -19,23 +19,16 @@
...
 
 public final class DroolsFilterUniConstraintStream<Solution_, A> extends DroolsAbstractUniConstraintStream<Solution_, A> {
 
     private final DroolsAbstractUniConstraintStream<Solution_, A> parent;
-    private final PatternDSL.PatternDef<A> aPattern;
 
     public DroolsFilterUniConstraintStream(DroolsConstraintFactory<Solution_> constraintFactory,
             DroolsAbstractUniConstraintStream<Solution_, A> parent, Predicate<A> predicate) {
-        super(constraintFactory);
+        super(constraintFactory, parent.getAnchor().filter(predicate));
```
ge0ffrey: ow, this could be dangerous. We need to bluejeans to discuss.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/UniAnchor.java
ge0ffrey: -1 for using the name "Anchor" as it's confuses with chained variable's anchor.
Let's discuss what it is

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
ge0ffrey: We need Mario to provide Rule.setEnabled(false).
You can't do this while building the rules (= building the CSSessionFactory), because the constraint weights of @ConstraintConfiguration are part of teh dataset, so they aren't know yet at that moment. However, when the CSSession, the dataset is known, so the weights are known and we can disable the zero weights for that KieSession.

triceo: @ge0ffrey There is API to remove rules from KieBase. The only question is how to make it not a performance problem to call it when the session is being started.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/TriAnchor.java
triceo: Yeah, that will be easy to fix.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
triceo: Leftovers from a previous version, where it was really expected. Will change.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AbstractConstraintStreamTest.java
```diff
@@ -85,15 +120,18 @@ protected void assertScore(InnerScoreDirector<TestdataLavishSolution> scoreDirec
             String constraintPackage = scoreDirector.getSolutionDescriptor().getSolutionClass().getPackage().getName();
             ConstraintMatchTotal constraintMatchTotal = scoreDirector.getConstraintMatchTotalMap()
                     .get(ConstraintMatchTotal.composeConstraintId(constraintPackage, TEST_CONSTRAINT_NAME));
+            System.out.println(constraintMatchTotal.getConstraintMatchSet());
```
ge0ffrey: sout

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/GenuineRuleMetadata.java
triceo: This is one of the least understandable rules around the codebase.

In this case, it's going to have one of the following results:

1. The class gets a public constructor, even though nobody outside of this package has any business creating instances.
1. The class gets hidden inside another class, emulating friendly access and making the code worse readable at the same time.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AbstractConstraintStreamTest.java
```diff
@@ -74,6 +83,32 @@ public AbstractConstraintStreamTest(boolean constraintMatchEnabled, ConstraintSt
...
     }
 
+    private static List<Object> removeIndirection(LogicalTuple logicalTuple) {
+        return IntStream.range(0, logicalTuple.getCardinality())
+                .mapToObj(logicalTuple::getItem)
+                .collect(Collectors.toList());
+    }
+
+    private static List<Object> removeIndirection(List<Object> justificationList) {
+        final List<Object> l = justificationList.stream()
```
rsynek: We could as well directly return the `collect()` result without storing it in `l`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/UniAnchor.java
```diff
@@ -0,0 +1,184 @@
...
+    private <A, ScoreHolder extends AbstractScoreHolder> List<RuleItemBuilder<?>> terminateWithScoring(
+            Global<ScoreHolder> scoreHolderGlobal, Block3<Drools, ScoreHolder, A> consequenceImpl) {
+        ConsequenceBuilder._2<ScoreHolder, A> consequence =
+                on(scoreHolderGlobal, (Declaration<A>) getAMetadata().getVariableDeclaration())
+                        .execute(consequenceImpl);
+        return Arrays.asList(getAMetadata().getPattern(), consequence);
+    }
+
+    public static String createContextId() {
+        return UUID.randomUUID().toString();
```
ge0ffrey: id must be reproducible between runs -> count++

ge0ffrey: rule name == the contextId of the inserted logical tuple of the same rule that inserts it

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/UniAnchor.java
```diff
@@ -0,0 +1,184 @@
...
+                .expr(getAMetadata().getVariableDeclaration(),
+                        (b, a) -> matches(biJoiner, inlineA(a), (B) bInliner.apply(b)));
+        if (bMetadata instanceof InferredRuleMetadata) {
+            return new BiAnchor(getAMetadata(), ((InferredRuleMetadata) bMetadata).substitute(newPattern));
+        } else {
+            return new BiAnchor(getAMetadata(), ((GenuineRuleMetadata) bMetadata).substitute(newPattern));
+        }
+    }
+
+    public <A, GroupKey_> List<RuleItemBuilder<?>> terminateWithLogical(final String currentContextId,
```
ge0ffrey: terminate -> finish or end or ... or better yet:

- terminateWithLogical -> insertLogical or logicalInsertion
- terminateWithScoring -> score or scoring or ...

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/UniAnchor.java
ge0ffrey: DroolsUniCondition?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsGroupingUniConstraintStream.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public List<DroolsFromUniConstraintStream<Solution_, Object>> getFromStreamList() {
+        return parent.getFromStreamList();
+    }
+
+    @Override
+    public Optional<Rule> buildRule(DroolsConstraint<Solution_> constraint, Global<? extends AbstractScoreHolder> scoreHolderGlobal) {
+        final String currentContextId = getAnchor().getContextId();
+        final String ruleName = UUID.randomUUID().toString(); // Generated == groupings can be shared by many rules.
```
rsynek: After seeing the rule name being generated from UUID on several places I wonder if we potentially need the rule name to be readable by a human being. Do we ever want to drill down to the rule and be able to connect it to the particular CS Operation?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/RuleMetadata.java
```diff
@@ -0,0 +1,36 @@
...
+import org.drools.model.PatternDSL;
+
+public interface RuleMetadata<FactType> {
+
+    static InferredRuleMetadata ofInferred(Declaration<LogicalTuple> variableDeclaration,
+            PatternDSL.PatternDef<LogicalTuple> pattern) {
+        return new InferredRuleMetadata(variableDeclaration, pattern);
+    }
+
+    static <A> GenuineRuleMetadata<A> of(Declaration<A> variableDeclaration, PatternDSL.PatternDef<A> pattern) {
```
rsynek: `of` -> `ofGenuine`?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/UniAnchor.java
```diff
@@ -0,0 +1,184 @@
...
+                patternProvider.apply(contextId, aVariableDeclaration));
+        this.aInliner = INFERRED_INLINER;
+    }
+
+    private <A> UniAnchor(RuleMetadata<A> aMetadata) {
+        this.aMetadata = aMetadata;
+        this.aInliner = getInliner(aMetadata);
+    }
+
+    private <A> A inlineA(Object a) {
```
ge0ffrey: rename this method, it's not inlining

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/LogicalTuple.java
```diff
@@ -0,0 +1,112 @@
...
+    private final String context;
+    private final Object[] items;
+
+    public LogicalTuple(final String context, final Object... items) {
+        Objects.requireNonNull(context, "Logical tuple rule context must not be null.");
+        this.context = context;
+        int itemCount = items.length;
+        if (itemCount == 0) { // such a tuple makes no sense
+            throw new IllegalArgumentException("Logical tuple must have at least one element.");
+        } else if (itemCount > 3) { // there are no quad-streams yet
```
rsynek: If we do such a check (and more importantly, if there are similar check in other classes), it might be good to extract the value into some global constant named e.g. `SUPPORTED_CARDINALITY`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/GenuineRuleMetadata.java
rsynek: I consider using the most restrictive access modifiers as possible a rule of thumb. In my opinion, the code style convention brings little value for the price of preventing us from using the package private access.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsGroupingUniConstraintStream.java
triceo: There will be some changes in this area, specifically the rules will be numbered and that numbering will be consistent across multiple JVM runs. That is probably as much as can be done here.

triceo:
@ge0ffrey The `Condition`s (formerly `Anchor`s) are now more generic. Rule names are as good as they're ever going to be. Plus I have fixed a couple bugs I identified through the new tests I added.

Please squash before merging.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -57,21 +61,81 @@ public DroolsConstraint(DroolsConstraintFactory<Solution_> constraintFactory,
...
+     * @param root scoring node for this constraint
+     * @return all streams which have the root as a child, transitively
+     */
+    private Set<DroolsAbstractConstraintStream<Solution_>> assembleScoringStreamTree(
+            DroolsAbstractConstraintStream<Solution_> root) {
+        Set<DroolsAbstractConstraintStream<Solution_>> allStreamSet = assembleAllStreams();
+        Set<DroolsAbstractConstraintStream<Solution_>> allPredecessorSet = predecessors(allStreamSet, root);
+        allPredecessorSet.add(root);
+        return allPredecessorSet;
+    }
```
triceo: I had to add this method to the rule construction to fix the bug I discovered when I wrote tests for sharing streams.
(Both scoring streams would originally end up coming from the same constraint, because each constraint would originally be applied to each scoring stream, and due to the caching, the second constraint would never get a rule.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsGroupingUniConstraintStream.java
rsynek: I was thinking if potentially want for example to use the operation's `toString()` as a prefix.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,87 @@
...
+     * @param <GroupKeyA_> the type of the first fact in the destination {@link BiConstraintStream}'s tuple
+     * @param <GroupKeyB_> the type of the second fact in the destination {@link BiConstraintStream}'s tuple
+     * @return never null
+     */
     <GroupKeyA_, GroupKeyB_> BiConstraintStream<GroupKeyA_, GroupKeyB_> groupBy(
             Function<A, GroupKeyA_> groupKeyAMapping,
             Function<A, GroupKeyB_> groupKeyBMapping);
 
+    /**
+     * Merges the semantics of {@link #groupBy(Function, Function)} and {@link #groupBy(UniConstraintCollector)}.
```
ge0ffrey: I'd avoid the use of the word "merge" here (it confuses to think that the streams are merged).
It could be "Combines the semantics ...", but maybe we can leave it out entirely:
"A groupBy with two group key and 1 collector that leads to a TriConstraintStream."

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -186,20 +186,87 @@
...
+     * @param <ResultContainer_> the mutable accumulation type (often hidden as an implementation detail)
+     * @param <Result_> the type of the second fact in the destination {@link BiConstraintStream}'s tuple
+     * @return never null
+     */
     <GroupKey_, ResultContainer_, Result_> BiConstraintStream<GroupKey_, Result_> groupBy(
             Function<A, GroupKey_> groupKeyMapping,
             UniConstraintCollector<A, ResultContainer_, Result_> collector);
 
+    /**
+     * Expand the {@link UniConstraintStream} to a {@link BiConstraintStream}, consisting of tuples which have:
```
ge0ffrey: I am not convinced about using "Expand" in this context.
Yes, the cardinality increases.
But the element count decreases.

When we turn a Uni<Apple> into a Bi<Airplane, Orange>, it doesn't feel as expanding.
Better example: We turn a Uni<Person> into a Bi<Age, Gender> and that person stream has [Ann(40, female), Beth(40, female), Carl (31, male)] that becomes [{40, female}, {31, male}]. Does that feel like an expansion?

ge0ffrey: Instead of "expand", I 'd just use "Convert".
I'd also mention that the new stream will have less or equal elements as the original.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintFactory.java
```diff
@@ -125,6 +133,20 @@ public DroolsConstraintFactory(SolutionDescriptor<Solution_> solutionDescriptor)
...
 
+    /**
+     * The constraint creating code needs to call this method to retrieve an ID of the rule.
+     * This ID then needs to be used in the rule name for easier debugging.
+     * It is imperative that this method only be called once per every rule created, as it maintains an internal
+     * counter of created rules.
+     * The sequence of IDs returned will start with 1 and continue towards {@link Integer#MAX_VALUE}.
+     * As long as rule creation calls this method in the same order, the rules will always receive the same ID.
+     *
+     * @return A unique numeric ID of the rule.
```
ge0ffrey: code style: we don't have an empty line in javadocs before params/return

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsLogicalTuple.java
```diff
@@ -0,0 +1,115 @@
...
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || !Objects.equals(getClass(), o.getClass())) {
+            return false;
+        }
+        final DroolsLogicalTuple that = (DroolsLogicalTuple) o;
+        return Objects.equals(ruleId, that.ruleId) &&
+                Arrays.equals(items, that.items); // TODO do we care about the order if items?
```
ge0ffrey: Yes, I think we do.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
```diff
@@ -788,6 +852,61 @@ public void nodeSharing() {
         assertEquals(1, monitorCount.get());
     }
 
-    // TODO add test for reusing streams: have 2 scorings from the same BiConstraintStream instance
+    @Test
+    public void filteredStreamReuse() {
```
ge0ffrey: Can we rename these test method to reuseFilter() and reuseGroupBy()? it's first and foremost about reuse I'd argue

PR #577, opened 2019-10-14
PLANNER-1630 Split up SolverConfig and SolverFactory

ge0ffrey:
Motivation:
- BEFORE: new SolverFactory, read solver config XML, adjust solver config, create solver
- AFTER: read solver config XML, adjust solver config, new SolverFactory, [1], create solver
At [1] optimization or quarkus magic can happen.

Still to do in separate PR's:
- Similar refactor for benchmark config
- Benchmark.createFromSolverFactory can probalby not survive (deprecate + remove in 8.0 too). We can add createFromXmlSolverResource and createFromSolverConfig instead.

Upgrade recipe coming on optaplanner-website repo.


optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfigs.java
```diff
@@ -0,0 +1,161 @@
...
+import java.io.Reader;
+import java.io.UnsupportedEncodingException;
+import java.util.Objects;
+
+import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.converters.ConversionException;
+import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.impl.solver.io.XStreamConfigReader;
+
+public class SolverConfigs {
```
ge0ffrey: Instead of throwing these create methods on SolverConfig, I decided to create a SolverConfigs class.
Similar to java.util.Collections, Streams, etc

However, *this is inconsistent* with SolverFactory.create*() and BenchmarkFactory.create*(). We can start deprecating those and moving them too, though...

Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/io/XStreamConfigReader.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.solver.io;
+
+import com.thoughtworks.xstream.XStream;
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.config.solver.SolverConfig;
+
+public class XStreamConfigReader {
```
ge0ffrey: New impl class, both solver and benchmark config reading use this class.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
```diff
@@ -26,68 +26,60 @@
...
  * <p>
- * To build an instance, use {@link #createFromXmlResource(String)} or any of the other creation methods.
- * <p>
- * Supports tweaking the configuration programmatically before a {@link Solver} instance is build.
+ * To build an instance, use {@link #createFromXmlResource(String)}.
+ * To change the configuration programmatically, create a {@link SolverConfig} with {@link SolverConfigs}
+ * and use {@link #create(SolverConfig)}.
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public abstract class SolverFactory<Solution_> {
```
ge0ffrey: I moved the KieContainer and Emtpy methods down in this class, making the diff a mess...

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
```diff
@@ -26,68 +26,60 @@
...
+     * @param solverConfig never null
      * @return never null
+     * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
      */
-    public static <Solution_> SolverFactory<Solution_> createFromKieContainerXmlResource(
-            String solverConfigResource) {
-        KieContainer kieContainer = KieServices.Factory.get().getKieClasspathContainer();
-        return new XStreamXmlSolverFactory<Solution_>(new SolverConfigContext(kieContainer))
-                .configure(solverConfigResource);
+    public static <Solution_> SolverFactory<Solution_> create(SolverConfig solverConfig) {
```
ge0ffrey: This new method "create(SolverConfig)" is important.
Originally I called it "createFromSolverConfig(SolverConfig)" to be consistent with the other create*() methods.
But I changed my mind later on, because it's inherently what happens...

Most users won't create a SolverConfig - they 'll directly create a SolverFactory from the XML.
WDYT?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/io/XStreamConfigReader.java
```diff
@@ -0,0 +1,57 @@
...
+        xStream.aliasSystemAttribute("xStreamRef", "reference");
+        xStream.processAnnotations(SolverConfig.class);
+        XStream.setupDefaultSecurity(xStream);
+        xStream.allowTypesByRegExp(new String[]{"org\\.optaplanner\\.\\w+\\.config\\..*"});
+        return xStream;
+    }
+
+    public static XStream buildXStream(ClassLoader classLoader) {
+        XStream xStream = buildXStream();
+        if (classLoader != null) {
```
triceo: And if `classloader == null`? Sounds like something we usually fail fast about.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/io/XStreamConfigReader.java
triceo: I find this name good enough. But since it's a utility class with a private constructor, perhaps it should be `final`?

optaplanner-docs/src/main/asciidoc/PlannerConfiguration/PlannerConfiguration-chapter.adoc
```diff
@@ -136,27 +135,29 @@ These `$$*$$Config` classes are the Java representation of the XML format.
...
 ====
-The `SolverFactory` is only thread-safe after it's configured.
-So the `getSolverConfig()` method is not thread-safe.
-To configure a `SolverFactory` dynamically for each user request, build a basic `SolverFactory` during initialization
-and clone it with the `cloneSolverFactory()` method for a user request:
+Don't modify a `SolverConfig` after creating a `SolverFactory`,
+because it currently breaks the thread-safety of the `SolverFactory`.
+To configure a `SolverFactory` dynamically for each user request,
+build a template `SolverConfig` during initialization
+and copy it with the copy constructor for each user request:
```
triceo: This is indeed surprising. Can we make the defensive copy ourselves and not bother the user with such details?

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/app/NQueensApp.java
```diff
@@ -78,14 +78,13 @@ public NQueensApp() {
     }
 
     /**
-     * Unused alternative. Abnormal way to create a {@link Solver}.
+     * Unused alternative. A way to create a {@link Solver} without using XML.
```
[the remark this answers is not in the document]
ge0ffrey: It's used, will remove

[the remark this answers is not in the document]
ge0ffrey: It's not used in production code, but it's tested. It's to have a hello world example of the programmatic API.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
```diff
@@ -199,31 +272,42 @@
...
+     * <p>
      * Build a {@link SolverFactory} quickly (without parsing XML) that builds the exact same {@link Solver}
      * with {@link #buildSolver()}, but can also be modified with {@link #getSolverConfig()} to build a different
      * {@link Solver} without affecting the original {@link SolverFactory}.
      * @return never null, often a different {@link SolverFactory} subclass implementation than this instance
+     * @deprecated in favor of {@link SolverConfig(SolverConfig)}
+     * and {@link SolverFactory#create(SolverConfig)}. Will be removed in 8.0.
      */
+    @Deprecated
     public abstract SolverFactory<Solution_> cloneSolverFactory();
```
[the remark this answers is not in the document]
ge0ffrey: No, people aren't extending SolverFactory, nor Solver.
SolverFactory should have been an interface, in retrospect. Maybe something to change for 8.0? But we can't prepare for it

In any case, This cloneSolverFactory() method, in our API, used by a portion of our users, explained in the docs, has been deprecated and replaced in the docs.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/io/XStreamConfigReader.java
ge0ffrey: This is intentional, to avoid DRY in code that calls this method.

I'll add a javadoc to point out that we accept nulls.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfigs.java
ge0ffrey: You make a good point. I was really in doubt on this. Internal consistency trumps JDK consistency?

@yurloc @rsynek Wdyt?

It's pretty easy to change for me btw before merging.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
ge0ffrey: Devil's advocate: we want most users not to use SolverConfig (it's the power user option), so is it ok to have create() be short and sweet when we want them to use createFromXmlResource() by default?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/io/XStreamConfigReader.java
ge0ffrey: done locally

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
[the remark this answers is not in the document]
ge0ffrey: Summary: this method that people use is now deprecated and will be removed in 8.0.

optaplanner-docs/src/main/asciidoc/PlannerConfiguration/PlannerConfiguration-chapter.adoc
triceo: Yeah, +1 on the defensive copy. I don't want to be thinking about when I can or can not modify the config.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
triceo: I could see that as an argument do shorten the `createFromXmlResource` method name instead. :-)

ge0ffrey:
@rsynek Test coverage increased where it makes sense: createFromXmlFile().
The deprecated methods on SolverFactory aren't worth it (they'd be removed anyway).

ge0ffrey:
Ok, if no one objects, and jenkins and sonar are happy (enough), I am merging this in the morning.

ge0ffrey:
@rsynek I went through the sonar report and added a few tests. Sonar is useful!
I am ignoring fails fast and deprecated code that hasn't really changed (the latter will be removed in 8.0). Due to this, the test coverage is still far from 80%, but I merge anyway.

PR #579, opened 2019-10-16
PLANNER-1676 groupBy() with a Collector

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsUniAccumulateFunctionAdapter.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public DroolsUniAccumulateFunctionAdapter(UniConstraintCollectorAdapter<A, ResultContainer_, NewA> collector) {
+        this.supplier = collector.get().supplier();
+        this.accumulator = collector.get().accumulator();
+        this.finisher = collector.get().finisher();
+    }
+
+    @Override
+    public ResultContainer_ createContext() {
+        return supplier.get();
```
triceo: This is a very nasty thing that I'm doing here. `ResultContainer_` here and the `ResultContainer_` used throughout the class are not, in fact, the same type.

Drools `AccumulateFunction` forces the `ResultContainer_` to be serializable and our `UniConstraintCollector` is not. Using this trick, which only works because of erasure, I don't need to have a `Serializable` wrapper for `UniConstraintCollector` which - I believe - would not be possible to do.

Alternatively, this hack can go away if we just say `UniConstraintCollector extends Serializable`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsUniAccumulateFunctionAdapter.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public void writeExternal(ObjectOutput out) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void readExternal(ObjectInput in) {
+        throw new UnsupportedOperationException();
+    }
```
triceo: These would be very hard to implement. The first one, impossible.
Thankfully, Drools seems to not require them.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsUniAccumulateFunctionAdapter.java
```diff
@@ -0,0 +1,131 @@
...
+import org.optaplanner.core.impl.score.stream.drools.uni.UniConstraintCollectorAdapter;
+
+/**
+ * Drools {@link AccumulateFunction} that calls {@link UniConstraintCollector} underneath.
+ * @param <A> input to accumulate
+ * @param <ResultContainer_> implementation detail
+ * @param <NewA> result of accumulation
+ */
+public final class DroolsUniAccumulateFunctionAdapter<A, ResultContainer_ extends Serializable, NewA>
+        implements AccumulateFunction<ResultContainer_> {
```
ge0ffrey: Design patterns: I 'd argue it's a Bridge, not an Adapter pattern.
https://stackoverflow.com/questions/1425171/difference-between-bridge-pattern-and-adapter-pattern
But that's like the difference between an ale and a lager: most people will happily interchange them.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsUniAccumulateFunctionAdapter.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public void accumulate(ResultContainer_ context, Object value) {
+        undoMap.compute(value, (k, v) -> {
+           if (v == null) {
+               Runnable runnable = accumulator.apply(context, (A) value);
+               return new Undo(runnable);
+           } else {
+               v.incrementUseCount();
```
ge0ffrey: Can you do me a favor and proof through debugging that this increment sometimes goes above 1 for some of the examples?

I find it strange that drools would do accumulate(context, Ann) twice without calling reverse(context, Ann) in between.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsUniAccumulateFunctionAdapter.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public void readExternal(ObjectInput in) {
+        throw new UnsupportedOperationException();
+    }
+
+    private static final class Undo {
+
+        private final Runnable runnable;
+        private final AtomicLong useCount = new AtomicLong(1);
```
ge0ffrey: AtomicLong do synchronization (so light locking). But this is used in a single-threaded context, so that not really needed, performance wise. How about adding a comment (maybe a TODO?) to point that out?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsGroupingUniConstraintStream.java
```diff
@@ -35,18 +36,29 @@
 import static org.drools.model.PatternDSL.declarationOf;
 import static org.drools.model.PatternDSL.pattern;
 
-public final class DroolsGroupingUniConstraintStream<Solution_, A, GroupKey_>
-        extends DroolsAbstractUniConstraintStream<Solution_, GroupKey_> {
+public final class DroolsGroupingUniConstraintStream<Solution_, A, NewA>
+        extends DroolsAbstractUniConstraintStream<Solution_, NewA> {
```
ge0ffrey: If we start using NewA here instead of GroupKey_, we should do it consistenly in all of ConstraintStreams (including the API and Bavet, not just Drools).
Let's discuss on a bluejeans. (I agree that GroupKey_ isn't perfect, but not sure yet if NewA is the best option). Good to merge this meanwhile - temporarly.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsUniAccumulateFunctionAdapter.java
triceo: I could go do that, and it may very well turn out that Drools never does that. But this code is still safer, and there's really no extra cost for it. (The map operations would have to be there anyway.) I'm inclined to keep it as is, WDYT?

PR #580, opened 2019-10-16
PLANNER-1630 Split up benchmark config and factory

ge0ffrey:


ge0ffrey:
@rsynek Same story here on test coverage. We went from very little testing (20-30% or so) to 40-50%. Once the deprecated methods are removed in 8.0, that test coverage will increase heavily.

optaplanner-benchmark/src/test/resources/org/optaplanner/benchmark/config/testdataBenchmarkConfig.xml
```diff
@@ -1,11 +1,17 @@
...
     <solver>
-      <solutionClass>org.optaplanner.examples.nqueens.domain.NQueens</solutionClass>
-      <entityClass>org.optaplanner.examples.nqueens.domain.Queen</entityClass>
+      <solutionClass>org.optaplanner.core.impl.testdata.domain.TestdataSolution</solutionClass>
+      <entityClass>org.optaplanner.core.impl.testdata.domain.TestdataEntity</entityClass>
+      <scoreDirectorFactory>
+        <scoreDrl>org/optaplanner/core/api/solver/testdataScoreRules.drl</scoreDrl>
+      </scoreDirectorFactory>
+      <termination>
+        <secondsSpentLimit>20</secondsSpentLimit>
```
rsynek: Do we use this benchmark config for tests that run the benchmark or just for verifying the configuration can be built? If it's the former we should use more deterministic termination.

ge0ffrey: Actually, the others use a termination of 0 seconds. I am going to set this one to 0 seconds too, even though it doesn't run (it just for comparison of before and after reading).

PR #581, opened 2019-10-17
More explanatory naming of generic parameters in Bavet group bridge

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetGroupBridgeUniConstraintStream.java
```diff
@@ -26,24 +26,24 @@
 import org.optaplanner.core.impl.score.stream.bavet.bi.BavetGroupBiNode;
 import org.optaplanner.core.impl.score.stream.bavet.common.BavetNodeBuildPolicy;
 
-public final class BavetGroupBridgeUniConstraintStream<Solution_, A, GroupKey_, ResultContainer_, Result_>
+public final class BavetGroupBridgeUniConstraintStream<Solution_, A, NewA, ResultContainer_, B>
```
ge0ffrey: A is coming from the parent, but B is not.
Shall we call B instead NewB?

PR #583, opened 2019-10-20
PLANNER-1676 Feature parity with Bavet

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
```diff
@@ -76,41 +81,69 @@ private DroolsUniCondition(DroolsMetadata<Object, A> aMetadata) {
...
+        ExprViewItem<Object> outerAccumulatePattern = DSL.accumulate(innerAccumulatePattern,
+                accFunction(() -> accumulateFunction, aVariable).as(accumulateResultVariable));
+        ConsequenceBuilder._2<?, ?> consequence = on(actualNewADeclaration, accumulateResultVariable)
+                .execute((drools, newA, newB) -> {
+                    RuleContext kcontext = (RuleContext) drools;
+                    kcontext.insertLogical(new DroolsLogicalTuple(ruleId, newA, newB));
+                });
+        return Arrays.asList(collectingPattern, pattern(setOfNewADeclaration), pattern(actualNewADeclaration),
+                outerAccumulatePattern, consequence);
     }
```
triceo: This is the meat of the pull request. A single Drools rule that both groups and accumulates.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniAccumulateFunctionBridge.java
```diff
@@ -18,29 +18,31 @@
...
-        implements AccumulateFunction<ResultContainer_> {
+final class DroolsUniAccumulateFunctionBridge<A, ResultContainer_, NewA>
+        implements AccumulateFunction<DroolsAccumulateContext<ResultContainer_>> {
 
     private final Supplier<ResultContainer_> supplier;
     private final BiFunction<ResultContainer_, A, Runnable> accumulator;
     private final Function<ResultContainer_, NewA> finisher;
-    private final Map<Object, Runnable> undoMap = new HashMap<>(0);
+    // Weak in order to prevent a memory leak, as the the context is kept by Drools outside of our control.
+    private final Map<DroolsAccumulateContext<ResultContainer_>, Map<Object, Runnable>> undoMaps = new WeakHashMap<>(0);
```
triceo: I needed to introduce a context carrier (`DroolsAccumulateContext`) with a stable `equals()`. (Therefore not the `ResultContainer_` itself.)

The reason why I didn't do this before was that until now, every accumulate was only operating on one set of data, and therefore there was really only one `context`. With the new, more complex rule, Drools started using different `context`s and the old code naturally broke.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
```diff
@@ -140,8 +139,9 @@ public void addChildStream(BavetAbstractUniConstraintStream<Solution_, A> childS
     }
 
     @Override
-    public <GroupKeyA_, GroupKeyB_, ResultContainer_, Result_> TriConstraintStream<GroupKeyA_, GroupKeyB_, Result_> groupBy(
-            Function<A, GroupKeyA_> groupKeyAMapping, Function<A, GroupKeyB_> groupKeyBMapping,
+    public <GroupKeyA_, GroupKeyB_, ResultContainer_, Result_>
+    TriConstraintStream<GroupKeyA_, GroupKeyB_, Result_> groupBy(Function<A, GroupKeyA_> groupKeyAMapping,
```
ge0ffrey: Maybe we should figure out a shorter name for GroupKeyA_ (that would make me and Radovan happy, I am sure Lukas loves any shorter name) so we don't go over the line char limit before hitting the first (

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsAbstractBiConstraintStream.java
```diff
@@ -190,6 +182,8 @@ public final Constraint impactScoreConfigurableBigDecimal(String constraintPacka
         }
     }
 
+    protected abstract DroolsAbstractConstraintStream<Solution_> getParent();
```
ge0ffrey: I would expect the parent of a DroolsBi to return a DroolsBi instead of a general CS. Can the return type be more specific? I haven't checked in depth, could be this is wrong.
Sometimes it could be null (for example the join CS with 2 bridge parents maybe?)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsBiCondition.java
```diff
@@ -53,6 +57,14 @@ public DroolsBiCondition(DroolsMetadata<Object, A> aMetadata, DroolsMetadata<Obj
         this.bMetadata = bMetadata;
     }
 
+    public DroolsBiCondition(Declaration<DroolsLogicalTuple> aVariableDeclaration,
+            Function<Declaration<DroolsLogicalTuple>, PatternDSL.PatternDef<DroolsLogicalTuple>> patternProvider) {
+        // We share both the declaration and the pattern, as the data is all coming from the same DroolsLogicalTuple.
```
ge0ffrey: style nitpick: use active voice (instead of "we") to be more compact:
=> // Share both the declaration and the pattern, ...

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsBiCondition.java
```diff
@@ -62,31 +74,21 @@ public DroolsBiCondition(DroolsMetadata<Object, A> aMetadata, DroolsMetadata<Obj
     }
 
     public DroolsBiCondition<A, B> andFilter(BiPredicate<A, B> predicate) {
-        PatternDSL.PatternDef<Object> newPattern = bMetadata.getPattern()
+        Supplier<PatternDSL.PatternDef<Object>> patternSupplier = () -> bMetadata.buildPattern()
```
ge0ffrey: Does the `.buildPattern()` and `aMetadata.getVariableDeclaration()` calls need to be called lazely?
I don't grok the entire code, but can you verify that this is doesn't work:

```
bBuildPattern = bMetadata.buildPattern();
Supplier<PatternDSL.PatternDef<Object>> patternSupplier = () -> bBuildPattern.expr(...);
```

If it does, it's more efficient. Some question for `aMetadata.getVariableDeclaration()`. Clearly, the predicate test must be called later and cannot be extracted from the supplier, because it uses a and b.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsBiCondition.java
```diff
@@ -62,31 +74,21 @@ public DroolsBiCondition(DroolsMetadata<Object, A> aMetadata, DroolsMetadata<Obj
...
-        }
+        return new DroolsBiCondition<>(aMetadata, bMetadata.substitute(patternSupplier));
     }
 
     public <C> DroolsTriCondition<A, B, C> andJoin(DroolsUniCondition<C> cCondition,
             AbstractTriJoiner<A, B, C> triJoiner) {
         DroolsMetadata<Object, C> cMetadata = cCondition.getAMetadata();
-        PatternDSL.PatternDef<Object> newPattern = cMetadata.getPattern()
+        Supplier<PatternDSL.PatternDef<Object>> newPattern = () -> cMetadata.buildPattern()
                 .expr(UUID.randomUUID().toString(), aMetadata.getVariableDeclaration(),
```
ge0ffrey: There are still usages of UUID - Is that intended?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsGroupingBiConstraintStream.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.drools.model.DSL.declarationOf;
+import static org.drools.model.PatternDSL.pattern;
+
+public class DroolsGroupingBiConstraintStream<Solution_, A, NewA, ResultContainer_, NewB>
+        extends DroolsAbstractBiConstraintStream<Solution_, NewA, NewB> {
+
+    private final DroolsAbstractUniConstraintStream<Solution_, A> parent;
+    private final Function<A, NewA> groupKeyMapping;
+    private final UniConstraintCollector<A, ResultContainer_, NewB> collector;
+    private final AtomicInteger ruleId = new AtomicInteger(-1);
```
ge0ffrey: AtomicInteger has the method `getAndIncrement()` too.
Usually other optaplanner-core code starts from 0 and uses `getAndIncrement()` and this would be called nextRuleId instead.

But there's more complexity in play here, let me grok that first...

ge0ffrey: Ok, I get it.
Instead of using a "final AtomicInteger", can we simply not use a non-final "int"?

AtomicInteger are needed for thread-safety, which will never be in play here (score directors are isolated per thread). AtomicIntegers are sometimes used as a unique id dispenser, which I believed what was going on here.
But it's not: it's actually just a field containing the ruleId. The only reason why the AtomicInteger wrapping is in play, as far as I can tell, is so that the field can be made final?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsAbstractBiConstraintStream.java
triceo: Sometimes the parent of a `Bi` is a `Uni`. (See `Uni.groupBy(groupKeyMapping, collector)`.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsBiCondition.java
triceo: The reason for all this laziness, which I don't like, is so that we can reuse patterns in two different places. Use case is in the new rule in this PR:

- I want to use the pattern in the first accumulate.
- And then I want to use the same pattern, with some modifications (a new filter) in the second accumulate.
- Drools does not let me to clone patterns.
- Therefore, I need to create a new pattern every time I use it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsGroupingBiConstraintStream.java
```diff
@@ -0,0 +1,104 @@
...
+    public DroolsBiCondition<NewA, NewB> createCondition() {
+        return new DroolsBiCondition<>(declarationOf(DroolsLogicalTuple.class), var -> {
+            Object createdRuleId = createRuleIdIfAbsent(getConstraintFactory());
+            return pattern(var).expr(logicalTuple ->
+                    Objects.equals(logicalTuple.getRuleId(), createdRuleId));
+        });
+    }
+
+    /**
+     * Exists in order to be able to bind the condition in the new rule to the {@link DroolsLogicalTuple}s generated by
```
ge0ffrey: nitpick style: Instead of "Exists in order to ..." I 'd use something like "Guarantees that ..."

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsBiCondition.java
triceo: The Drools API requires a textual ID for the expression. As far as I know, it doesn't make it into the rule and may be completely random. But I'll look around to see if I can somehow get rid of it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsGroupingBiConstraintStream.java
```diff
@@ -0,0 +1,104 @@
...
+     * just retrieve the value that already exists and query for all the {@link DroolsLogicalTuple}s having this value.
+     * Even if the order of operations were switched (condition created before rule built), the ID would still be
+     * correctly shared though.
+     * @param constraintFactory never null, used to generate rule IDs
+     * @return unique id for the rule
+     */
+    private int createRuleIdIfAbsent(DroolsConstraintFactory<Solution_> constraintFactory) {
+        return ruleId.updateAndGet(currentRuleId -> (currentRuleId < 0) ?
+                constraintFactory.getRuleIdAndIncrement() :
+                currentRuleId);
```
ge0ffrey: Took me a while to understand this.
See previous comment, it could be made:

```
if (ruleId < 0) {
   ruleId = constraintFactory.getRuleIdAndIncrement();
}
```

Btw, `constraintFactory.getRuleIdAndIncrement()` might be worth renaming to `constraintFactory.getNextRuleId()`. The incrementing is an impl detail.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/bi/DroolsJoinBiConstraintStream.java
```diff
@@ -48,6 +49,11 @@ public DroolsJoinBiConstraintStream(DroolsConstraintFactory<Solution_> constrain
         return rightParentStream;
     }
 
+    @Override
+    protected DroolsAbstractConstraintStream<Solution_> getParent() {
+        return null; // There is no one single parent for a join stream.
```
ge0ffrey: Ah yes, so a BiCS.getParent can return a BiCS instead of an AbstractCS.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsAccumulateContext.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Each context is uniquely identified by its ID.
+ * This is necessary so that the Drools accumulate function can properly undo in
+ * {@link DroolsUniAccumulateFunctionBridge} and its Bi, Tri, ... alternatives.
+ * @param <ResultContainer_> The same type from {@link UniConstraintCollector} and its Bi, Tri, ... alternatives.
+ */
+public final class DroolsAccumulateContext<ResultContainer_> implements Serializable {
+
+    private static final AtomicLong CONTEXT_COUNTER = new AtomicLong();
```
ge0ffrey: Can we get rid of making this static?

In multithreaded-solving, multiple score directors (each with it's own CS session) will be created in parallels. Normally they don't feel each other: each one is used on it's own thread, so they are single-threaded and don't share anything.
But this static CONTEXT_COUNTER will be shared. So the calls to it will be in a different order between runs, affecting the DroolsAccumulateContext's id (which is a pain for debugging and reproducible debugging). 

Furthermore, when we scale to 100 move threads, this means a 100 move threads are concurrently hitting that CONTEXT_COUNTER for every accumulate during session creation I presume. That's probably not the end of the world, as that congestion will last only a few milliseconds, but it's still nice to avoid.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsAccumulateContext.java
```diff
@@ -0,0 +1,61 @@
...
+    public ResultContainer_ getContainer() {
+        return container;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || !Objects.equals(getClass(), o.getClass())) {
```
ge0ffrey: DroolsAccumulateContext is a final class, can we simply do an instanceof check?

if (o instanceof DroolsAccumulateContext)

PS: null is not an instanceof DroolsAccumulateContext.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsAccumulateContext.java
triceo: I see your point. That said, it is `static` because the counter needs to be incremented. I'm not sure how to do this without having a single source of truth *somewhere.*

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/tri/DroolsTriCondition.java
```diff
@@ -123,7 +118,15 @@ public DroolsTriCondition(DroolsMetadata<Object, A> aMetadata, DroolsMetadata<Ob
...
                         .execute(consequenceImpl);
-        return Arrays.asList(aMetadata.getPattern(), bMetadata.getPattern(), cMetadata.getPattern(), consequence);
+        if (aMetadata instanceof DroolsInferredMetadata && bMetadata instanceof DroolsInferredMetadata &&
+                cMetadata instanceof DroolsInferredMetadata) {
+            // In case of logical tuples, all patterns will be the same logical tuple, and therefore we just add one.
+            return Stream.of(cMetadata.buildPattern(), consequence)
+                    .collect(Collectors.toList());
+        } else {
+            return Stream.of(aMetadata.buildPattern(), bMetadata.buildPattern(), cMetadata.buildPattern(), consequence)
+                    .collect(Collectors.toList());
```
ge0ffrey: I suspect that `Arrays.asList(aMetadata.buildPattern(), bMetadata.buildPattern(), cMetadata.buildPattern(), consequence)` is faster and less verbose?

triceo: Yeah. This is a leftover from a time when this code was doing something more complex. I'll change it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniAccumulateFunctionBridge.java
ge0ffrey: WeakHashMap might not do what you think it does (IIRC talks by either Goetz or Shipilev correctly) Let's discuss in bluejeans :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
```diff
@@ -76,41 +81,69 @@ private DroolsUniCondition(DroolsMetadata<Object, A> aMetadata) {
...
-        } else {
-            return new DroolsBiCondition<>(aMetadata, ((DroolsGenuineMetadata) bMetadata).substitute(newPattern));
-        }
+        return new DroolsBiCondition<>(aMetadata, bMetadata.substitute(patternSupplier));
+    }
+
+    public <ResultContainer, NewA, NewB> List<RuleItemBuilder<?>> completeWithLogicalInsert(
+            Object ruleId, Function<A, NewA> groupKeyMapping,
+            UniConstraintCollector<A, ResultContainer, NewB> collector) {
+        Function1<Object, A> extractor = o -> getAMetadata().extract(o);
```
ge0ffrey: Same question here: Can we extract the getAMetedata() call from the lamdba? So effectively cache the getAMetadata() call.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
```diff
@@ -76,41 +81,69 @@ private DroolsUniCondition(DroolsMetadata<Object, A> aMetadata) {
...
+        Function1<Object, NewA> grouper = a -> groupKeyMapping.apply(extractor.apply(a));
```
ge0ffrey: Can we use JDK's Function.andThen() here?
grouper = extractor.andThen(groupKeyMapping);

the JDK might do more magic with that internally.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
```diff
@@ -76,41 +81,69 @@ private DroolsUniCondition(DroolsMetadata<Object, A> aMetadata) {
...
+        // Accumulate all NewA into a set.
+        Declaration<NewA> innerNewADeclaration = (Declaration<NewA>) declarationOf(Object.class);
```
ge0ffrey: for bluejeans: why Object?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsAccumulateContext.java
triceo: Well, the IDs needn't be in sequence. They may also be random. But I'm not sure that this is better.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
triceo: The reason is the same in both cases - `aMetadata` may either be inferred or genuine. So the `extract(a)` either needs to return `a` for genuine or `a.getItem(...)` for inferred. And when we call the function, we do not know which is which.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsAccumulateContext.java
```diff
@@ -0,0 +1,61 @@
...
+            return false;
+        }
+        final DroolsAccumulateContext context = (DroolsAccumulateContext) o;
+        return id == context.id;
+    }
+
+    @Override
+    public int hashCode() {
```
ge0ffrey: return System.identityHashCode(this);

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsAccumulateContext.java
```diff
@@ -0,0 +1,61 @@
...
+    public DroolsAccumulateContext(ResultContainer_ container) {
+        this.container = container;
+    }
+
+    public ResultContainer_ getContainer() {
+        return container;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
ge0ffrey: return this == o;

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniAccumulateFunctionBridge.java
ge0ffrey: Put Map<Object, Runnable> into DroolsAccumulateContext, remove this Weak*Map.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
triceo: when
	accumulate(Person(), $set: collectSet(Person::getCity))
        $newA : City() from $set // grouping
        accumulate(Person(getCity() == $newA), $newB: collector(ConstraintCollectors.count()))
then
	insertLogical($newA, $newB);
end

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
```diff
@@ -454,7 +454,6 @@ public void groupBy_0Mapping1Collector_max() {
 
     @Test
     public void groupBy_1Mapping1Collector_count() {
```
ge0ffrey: Interesting extra test coverage:
- groupKeyMapping returns primitive type: something like: groupBy(Person::getAge) and age is an int
- a groupKey that has no instances has no match: groupBy(Person::getCity) but here is no Person in New York.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
triceo: I need to use Drools' `Function1` (not Java's `Function`), so I only have a limited number of options. This, unfortunately, is not one of them.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
```diff
@@ -50,21 +54,22 @@
...
-        this.aMetadata = (DroolsGenuineMetadata) DroolsMetadata.ofGenuine(aVariableDeclaration,
-                PatternDSL.pattern(aVariableDeclaration));
+        this.aMetadata = (DroolsGenuineMetadata) DroolsMetadata.ofGenuine(aVariableDeclaration);
     }
 
     public DroolsUniCondition(Declaration<DroolsLogicalTuple> aVariableDeclaration,
             Function<Declaration<DroolsLogicalTuple>, PatternDSL.PatternDef<DroolsLogicalTuple>> patternProvider) {
         this.aMetadata = (DroolsInferredMetadata) DroolsMetadata.ofInferred(aVariableDeclaration,
-                patternProvider.apply(aVariableDeclaration));
+                () -> patternProvider.apply(aVariableDeclaration));
```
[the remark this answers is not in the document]
triceo: In this case, it is not possible. (I usually do that automatically, since IDE usually points it out.)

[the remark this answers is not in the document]
ge0ffrey: Strange though, because EntityDescriptor.isInitialized(Object entity) has a parameter Object too, sure there must be a way to reference it.
But oth, I would expect IntelliJ to point it out indeed. I 'll merge and see if I can change it.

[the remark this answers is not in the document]
ge0ffrey: You're right.

PR #576, opened 2019-10-13
Remove dependency to jcommon and fix code due to JFreeChart upgrade 

FredPraca:
Requires: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1077

PR #585, opened 2019-10-21
PLANNER-1685: Benchmark Regression Fix

cuijulian:
Fixes regression caused by upgrading `JFreeChart` from `1.0.19` to `1.5.0` in #576

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
@@ -76,6 +76,8 @@ import freemarker.template.Configuration;
 import freemarker.template.Template;
 import freemarker.template.TemplateException;
 
+import static java.lang.Double.isFinite;
+
 public class BenchmarkReport {
 
     protected final transient Logger logger = LoggerFactory.getLogger(getClass());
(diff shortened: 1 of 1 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -353,10 +355,12 @@ private void writeBestScoreSummaryChart() {
...
                     for (int i = 0; i < levelValues.length && i < CHARTED_SCORE_LEVEL_SIZE; i++) {
-                        if (i >= datasetList.size()) {
-                            datasetList.add(new DefaultCategoryDataset());
+                        if (isFinite(levelValues[i])) {
+                            if (i >= datasetList.size()) {
+                                datasetList.add(new DefaultCategoryDataset());
+                            }
+                            datasetList.get(i).addValue(levelValues[i], solverLabel, problemLabel);
                         }
-                        datasetList.get(i).addValue(levelValues[i], solverLabel, problemLabel);
```
triceo: This code is apparently copy-pasted 3 times in this class. Any chance we could extract it into a separate method?

triceo: We could probably go even one block up, assuming we extract the `singleBenchmarkResult.get...()` into a lambda.

ge0ffrey: Not sure if it's worth the trouble:
If we do intent to do DRY on this code, it can go a lot further than just extracting these 3 lines, such as the for loop etc.
In any case, separate issue, separate PR. Nice to have, but not a blocker for me.

ge0ffrey:
Great work! Thank you @cuijulian !

PR #588, opened 2019-10-23
PLANNER-1691 Beta indexing bi joins in Drools-based Constraint Streams

triceo:
See [performance with this commit](https://docs.google.com/spreadsheets/d/1SP1QzVkGSiPah4WzLtlZD7Y3bk9FWcx5H0J3BY7-4ag/edit#gid=1008041303).

- taskassigning and flightcrewscheduling get very close to native DRL performance.
- vehiclerouting and cloudbalancing perform poorly due to groupBy().

This PR does not appear to affect CS-B performance in a measurable way.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
```diff
@@ -86,11 +90,59 @@ private DroolsUniCondition(DroolsMetadata<Object, A> aMetadata) {
...
+            case LESS_THAN:
+                return Index.ConstraintType.LESS_THAN;
+            case LESS_THAN_OR_EQUAL:
+                return Index.ConstraintType.LESS_OR_EQUAL;
+            case GREATER_THAN:
+                return Index.ConstraintType.GREATER_THAN;
+            case GREATER_THAN_OR_EQUAL:
+                return Index.ConstraintType.GREATER_OR_EQUAL;
+            default:
+                throw new IllegalStateException("Unsupported joiner type (" + type + ").");
```
triceo: Can we afford to do this?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
```diff
@@ -40,9 +50,23 @@ public CompositeBiJoiner(List<SingleBiJoiner<A, B>> joinerList) {
     // Builders
     // ************************************************************************
 
+    @Override
+    public Function<A, Object> getLeftMapping(int joinerId) {
+        final int maxId = leftMappings.length - 1;
+        if (joinerId > maxId) {
+            throw new IllegalArgumentException("Only joiners up to no. (" + maxId + ") are supported, was (" + joinerId + ").");
+        }
```
triceo: Performance would be marginally better if I removed these bounds checks.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsUniCondition.java
ge0ffrey: Yes, we always have a similar default for all switch statements, because you never know someone adds a enum value without adjusting this code.

That being said, should we switch case all cases? Yes, we should, if this exception pops up, it's a priority issue. Unit testing should cover all enum cases.
I don't see another way of dealing with this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/NoneBiJoiner.java
```diff
@@ -22,23 +22,35 @@
...
+    private static final Object[] EMPTY_OBJECT_ARRAY = new Object[0];
+    private static final JoinerType[] EMPTY_JOINER_ARRAY = new JoinerType[0];
+
     // ************************************************************************
     // Builders
     // ************************************************************************
 
+    @Override
+    public Function<A, Object> getLeftMapping(int joinerIndex) {
+        throw new UnsupportedOperationException();
```
ge0ffrey: Can we change this to:
throw new UnsupportedOperationException("Impossible state: getLeftMapping is never called on a NoneBiJoiner.");

The "Impossible state" is our way of saying: we're theoretically a 100% sure that our code will never do this, so if you do see this, it's a bug and it's not your fault, user.

PR #589, opened 2019-10-23
No need to provide the same ClassLoader twice as an argument

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
```diff
@@ -286,7 +271,9 @@
      * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
      */
     public static <Solution_> SolverFactory<Solution_> createEmptyFromKieContainer(KieContainer kieContainer) {
-        return new DefaultSolverFactory<>(new SolverConfig(), new SolverConfigContext(kieContainer));
+        SolverConfig solverConfig = new SolverConfig(kieContainer.getClassLoader());
+        return new DefaultSolverFactory<>(solverConfig,
+                new SolverConfigContext(kieContainer));
```
yurloc: Strange formatting. In my opinion this:
```java
return new DefaultSolverFactory<>(solverConfig, new SolverConfigContext(kieContainer));
```
or this
```java
return new DefaultSolverFactory<>(
        solverConfig,
        new SolverConfigContext(kieContainer)
);
```
are more clear.

yurloc:
Revapi fails on this change.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/api/PlannerBenchmarkFactory.java
```diff
@@ -471,17 +471,6 @@ public static PlannerBenchmarkFactory create(PlannerBenchmarkConfig benchmarkCon
...
-    /**
-     * As defined by {@link #create(PlannerBenchmarkConfig)}.
-     * @param benchmarkConfig never null
-     * @param classLoader sometimes null, the {@link ClassLoader} to use for loading all resources and {@link Class}es,
-     * null to use the default {@link ClassLoader}
-     * @return never null
-     */
-    public static PlannerBenchmarkFactory create(PlannerBenchmarkConfig benchmarkConfig, ClassLoader classLoader) {
-        return new DefaultPlannerBenchmarkFactory(benchmarkConfig, new SolverConfigContext(classLoader));
-    }
```
triceo: Can we remove public API at this point?

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
```diff
@@ -177,20 +177,6 @@
...
-     * @param classLoader sometimes null, the {@link ClassLoader} to use for loading all resources and {@link Class}es,
-     *      null to use the default {@link ClassLoader}
-     * @return never null
-     * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
-     */
-    public static <Solution_> SolverFactory<Solution_> create(SolverConfig solverConfig, ClassLoader classLoader) {
-        // Defensive copy of solverConfig, because the DefaultSolverFactory doesn't internalize it yet
-        solverConfig = new SolverConfig(solverConfig);
-        return new DefaultSolverFactory<>(solverConfig, new SolverConfigContext(classLoader));
-    }
```
triceo: Again, removing public API.

triceo: If it wasn't released yet, then Revapi wouldn't catch it and therefore wouldn't fail. I'm not in VPN right now, so cannot check why specifically the build failed.

rsynek: @yurloc is right, revapi complains about removal of org.optaplanner.core.config.SolverConfigContext::getClassLoader()

ge0ffrey: SolverConfigContext is in the config package. So it's special :)
It's not in the api package, nor in the impl package.
It's in the package for which we're guarantee BC only from the XML point of view not from the methods point of view. The lack of a clean line is hurting us and should be dealt with in the 8.0 timeframe.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/api/PlannerBenchmarkFactory.java
ge0ffrey: This wasn't released yet - so if we hurry, yes :)

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
ge0ffrey: Adjusted it here.
There are many more cases where this happens (even though we do try to avoid it). None any more in SolverFactory though.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigMultiThreadedTest.java
```diff
@@ -22,22 +22,14 @@
...
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.when;
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
ge0ffrey: I don't mind adhering to this one (I rarely look at the import statements). Any opinion on this @triceo and @rsynek ?

Do note that the KIE conventions had an exception for org.junit.Assert and org.mockito.Mockito:
  https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/1d35e1062cc735557a2be7177361e988ef728218/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml#L8

But the infamous "minimum common denominator" for Eclipse interoperability changed that too: https://github.com/kiegroup/droolsjbpm-build-bootstrap/commit/76aa61384524100db57968e674df9c140f7fa21f#diff-9d663bfb476dd8a86f4f213da123bfa5
It's a real shame that that PR got merged, not just because of this. (Ironically I was to one to actually pus the merge button...). Now that Eclipse is effectively dead (long live VSCode and Che), we might want to rethink that "minimum common denominator" approach.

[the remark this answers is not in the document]
triceo: No opinion. I will use whatever we agree on and what can be configured in the IDE.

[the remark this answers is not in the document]
rsynek: Generally I try to avoid star imports, but I don't mind these two exceptions. If we configure our IDEs in this way, it's fine with me.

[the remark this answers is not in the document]
ge0ffrey: I've add the code style discussion into this JIRA:
  https://issues.jboss.org/browse/PLANNER-1696
Please edit that jira to add any others we're missing.

PR #590, opened 2019-10-23
Add missing finite check for JFreeChart regression

cuijulian:
Add missing case in #585

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
@@ -564,7 +564,9 @@ public class BenchmarkReport {
                 if (singleBenchmarkResult.hasAllSuccess()) {
                     double worstScoreCalculationSpeedDifferencePercentage
                             = singleBenchmarkResult.getWorstScoreCalculationSpeedDifferencePercentage();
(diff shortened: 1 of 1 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -564,7 +564,9 @@ private void writeWorstScoreCalculationSpeedDifferencePercentageSummaryChart() {
                 if (singleBenchmarkResult.hasAllSuccess()) {
                     double worstScoreCalculationSpeedDifferencePercentage
                             = singleBenchmarkResult.getWorstScoreCalculationSpeedDifferencePercentage();
-                    dataset.addValue(worstScoreCalculationSpeedDifferencePercentage, solverLabel, problemLabel);
+                    if (isFinite(worstScoreCalculationSpeedDifferencePercentage)) {
+                        dataset.addValue(worstScoreCalculationSpeedDifferencePercentage, solverLabel, problemLabel);
+                    }
                 }
```
triceo: One more reason, I think, to extract that logic to a new method, and reuse that throughout the class.

ge0ffrey: Yes, but I believe using lamdba's can help us extract the solverLabel creation, DefaultCategoryDataset line's etc too. Maybe not.
I still argue that is not blocking this PR - it's a separate issue.

@triceo Do you mind if I merge?

PR #591, opened 2019-10-24
Fix datasetList index out of bounds issue

cuijulian:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -356,7 +356,7 @@ private void writeBestScoreSummaryChart() {
                     double[] levelValues = ScoreUtils.extractLevelDoubles(singleBenchmarkResult.getAverageScore());
                     for (int i = 0; i < levelValues.length && i < CHARTED_SCORE_LEVEL_SIZE; i++) {
                         if (isFinite(levelValues[i])) {
-                            if (i >= datasetList.size()) {
+                            while (i >= datasetList.size()) {
```
ge0ffrey: This doesn't entirely fix the bug:

Presume levelValues.length = 5
So i goes 0, 1, 2, 3, 4
Now if isFinite(levelValues[2]) returns false, but the others return true, then this will work properly.
However if isFinite(levelValues[4]) returns false (the last index), then this won't work properly.

ge0ffrey:
I'd like to add a unit test to the requirements now :)
Try to follow "Test Driven Design": First write a unit test that demonstrates the bug.
(This should be doable with a unit test, instead of an integration test). Ask @rsynek or me for guidance if you have trouble creating the unit test.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
cuijulian: If the last value `isFinite(levelValues[4])` returns false, then the loop will exit without adding the value. I don't see an out of bounds access here. Maybe I'm not getting you, is there a consequence in this corner case I'm not seeing?

triceo:
@ge0ffrey @cuijulian Guys, I've been doing some benchmarking and without this fix, the benchmarker is pretty much useless. What's worse, it only fails at the very end, after the user has already wasted possibly hours of time.

In my opinion, we absolutely need to get this merged during this sprint, as this regression must not make it into the product. WDYT?

ge0ffrey:
I understand, it completly ruins your benchmark currently, so it's a show stopper, not just for the release.
I 'll merge this, and take a quick look at it myself how I want the code to be and assign that improvement for Julian to review.
@cuijulian Can you write a unit test in a separate PR for this, later? If it's not straightforward how to do approach this unit test, set up a meeting with Lukas to discuss a good approach.

ge0ffrey:
Followed up by: https://github.com/kiegroup/optaplanner/pull/594

cuijulian:
> Lukas, I understand, it completely ruins your benchmark currently, so it's a show stopper, not just for the release.
> I 'll merge this, and take a quick look at it myself how I want the code to be and assign that improvement for Julian to review.
> @cuijulian Can you write a unit test in a separate PR for this, later? If it's not straightforward how to do approach this unit test, set up a meeting with Lukas to discuss a good approach.

Absolutely, thank you for the fix Geoffrey, I'll look into unit tests for the benchmarker report this sprint. @triceo Sorry for the blocker, I realize it's probably eaten up many hours of your time. Hopefully some unit tests will improve things :)

PR #592, opened 2019-10-25
Add Collectors to sum durations. 

ge0ffrey:
Breaks compilation due to ambigous use...

TODO: testing + docs

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -16,11 +16,13 @@
 
 package org.optaplanner.core.api.score.stream;
 
+import java.time.Duration;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.ToIntBiFunction;
 import java.util.function.ToIntFunction;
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -194,7 +194,7 @@ public void sum() {
         // Add second value, we have two now.
         int secondValue = 1;
         Runnable secondRetractor = accumulate(collector, container, secondValue);
-        assertResult(collector, container,firstValue + secondValue);
+        assertResult(collector, container, firstValue + secondValue);
```
ge0ffrey: In other tests, I have the tendency:
1) not to make them too smart, by actually doing math. For example:
int a = Integer.MAX +1;
int b = Integer.MAX +1;
assertEquals(calculator.sum(a, b), a + b); // green if buggy

2) Use numbers such as 1000, 200, 30, 4, 0.5, ... etc that are very easy to sum up.

So I'd argue that we can improve readability and safety by doing this:
int firstValue = 100;
int secondValue = 20;
assertResult(collector, container, 120);

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -197,6 +199,29 @@
                 resultContainer -> resultContainer[0]);
     }
 
+    public static <A> UniConstraintCollector<A, ?, Duration> sum(Function<? super A, Duration> groupValueMapping) {
```
ge0ffrey: By adding these methods, previously compiling code no longer does due to ambiguity.

triceo: I really don't think that this is a way forward. Adding a new function `sumSomething()` for every `Something` somebody may want to sum, that simply does not scale. 

We can approach it from a different angle: 

`<X> sum(Function<A, X>, Function<X, X> adder)`

And if you want to be extra nice to people:

`apply(Sum.duration())`, where `Sum.duration()` is a convenience implementation we offer to people, and we could also provide averages etc. in a similar manner.

Christopher-Chianelli: In  Java Stream API (and nearly all like it), this operation is called `reduce()`:  (https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#Reduction). Although in streams, a map is done before the reduce (so it would look like `stream.map(toDuration).reduce(sumBiOperation)`)

ge0ffrey:
Closed in favor of Lukas's work

PR #594, opened 2019-10-28
Benchmark: provisioning of new DefaultCategoryDataset is independent of the isFinite check

ge0ffrey:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -72,11 +74,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import freemarker.template.Configuration;
-import freemarker.template.Template;
-import freemarker.template.TemplateException;
-
-import static java.lang.Double.isFinite;
+import static java.lang.Double.*;
```
cuijulian: Should be `import static java.lang.Double.isFinite`. Is there a reason wild card import is used here?

ge0ffrey: Agreed, this be import static java.lang.Double.isFinite;

triceo:
@cuijulian Waiting for your approval, after which I can merge.

PR #595, opened 2019-10-28
SolverManager, Spring Boot starter and autoconfigure

ge0ffrey:
Related: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1125


optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverManager.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.UUID;
+
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.config.solver.SolverConfig;
+import org.optaplanner.core.impl.solver.DefaultSolverManager;
+
+/**
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ */
+public interface SolverManager<Solution_> extends AutoCloseable {
```
ge0ffrey: Ignore this file, this is a dummy implementation (that works but is not up to our code quality). SolverManager itself is being worked on in a separate PR that will be merged before this one.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverManager.java
```diff
@@ -0,0 +1,58 @@
...
+package org.optaplanner.core.impl.solver;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import org.optaplanner.core.api.solver.Solver;
+import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.api.solver.SolverManager;
+
+public class DefaultSolverManager<Solution_> implements SolverManager<Solution_> {
```
ge0ffrey: Ignore this file, this is a dummy implementation (that works but is not up to our code quality). SolverManager itself is being worked on in a separate PR that will be merged before this one.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfigurationTest.java
```diff
@@ -0,0 +1,55 @@
...
+    private final ApplicationContextRunner contextRunner;
+
+    public OptaPlannerAutoConfigurationTest() {
+        this.contextRunner = new ApplicationContextRunner()
+                .withUserConfiguration(TestConfiguration.class)
+                .withConfiguration(AutoConfigurations.of(OptaPlannerAutoConfiguration.class));
+    }
+
+    @Test @Ignore("The @EntityScan annotation is ignored in tests") // TODO FIXME
+    public void emptyProperties() {
```
ge0ffrey: I want to add the following tests:
- the properties are empty, there is no solverConfig.xml
- the properties are empty, and there is a solverConfig.xml
- the properties state the solverConfig is foo.xml, but it doesn't exist
- the properties state the solverConfig is foo.xml and it exist
- ...

nealeu:
Thanks for working on this @ge0ffrey.  I'm coming at Optaplanner from a Spring angle today, and it's great to see a clean out of the box approach with autoconfigure, starter and example modules for Spring Boot.

ge0ffrey:
jenkins retest this

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private void validateSolverFactory() {
+        solverFactory.buildSolver();
+    }
+
+    @Override
+    public DefaultSolverFuture solve(Solution_ planningProblem,
+            Consumer<Solution_> bestSolutionConsumer) {
+        Solver<Solution_> solver = solverFactory.buildSolver();
+        // TODO consumption should happen on different thread than solver thread, doing skipAhead and throttling
```
triceo: Do we want to release it to people without that?

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
```diff
@@ -0,0 +1,70 @@
...
+        solver.addEventListener(event -> bestSolutionConsumer.accept(event.getNewBestSolution()));
+        DefaultSolverFuture solverFuture = new DefaultSolverFuture<>(solver);
+        executorService.submit(() -> {
+            try {
+                solver.solve(planningProblem);
+            } catch (Exception e) {
+                e.printStackTrace(); // TODO generated
```
triceo: This definitely isn't production-ready.

triceo:
Looking at Sonar, there's some things we probably want to fix.
Also, ~45 % code coverage is less than ideal for something we expect to bring us so much new audience.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/TerminationProperties.java
```diff
@@ -0,0 +1,70 @@
...
+
+package org.optaplanner.spring.boot.autoconfigure;
+
+import java.time.Duration;
+
+public class TerminationProperties {
+
+    /**
+     * How long the solver can run. Uses ISO 8601 format, see {@link Duration}.
+     * For example: "30S" is 30 seconds. "5M" is 5 minutes. "2H" is 2 hours.
```
yurloc: The conversion from string to a `Duration` is handled by [Spring Boot](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration). ISO 8601 is one of the supported formats. I recommend using the more readable format (30s, 5m, 2h30m etc.).

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfigurationTest.java
```diff
@@ -0,0 +1,157 @@
...
+                    SolverConfig solverConfig = context.getBean(SolverConfig.class);
+                    assertEquals("2", solverConfig.getMoveThreadCount());
+                    assertNotNull(context.getBean(SolverFactory.class));
+                });
+    }
+
+    @Test
+    public void terminationProperties() {
+        contextRunner
+                .withPropertyValues("optaplanner.solver.termination.spent-limit=PT4H")
```
yurloc: 4h

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfigurationTest.java
```diff
@@ -0,0 +1,157 @@
...
+                .run(context -> {
+                    TerminationConfig terminationConfig = context.getBean(SolverConfig.class).getTerminationConfig();
+                    assertEquals(Duration.ofHours(4), terminationConfig.getSpentLimit());
+                    assertNotNull(context.getBean(SolverFactory.class));
+                });
+        contextRunner
+                .withPropertyValues("optaplanner.solver.termination.unimproved-spent-limit=PT5H")
```
yurloc: 5h

optaplanner-integration/optaplanner-spring-boot-example/src/main/resources/application.properties
```diff
@@ -0,0 +1,10 @@
+# The solver runs for 30 seconds. To run for 5 minutes use "5M" and for 2 hours use "2H".
+optaplanner.solver.termination.spent-limit=30S
```
yurloc: We're showing users how to do it. Let's not confuse them and let them believe they should use capital letters for time units. While ISO 8601 is one of the formats supported by Spring Boot, I'd show them the more natural way of denoting duration (`30s`) and also show them they can combine units: `2h45m`.

rsynek:
@ge0ffrey every time I refresh the page in a browser (NOT by using the Refresh button), I get:
```
java.lang.NullPointerException: null
        at org.webjars.WebJarAssetLocator.getFullPathExact(WebJarAssetLocator.java:254) ~[webjars-locator-core-0.39.jar:na]
```
in the log. But the page shows up.

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/PocSolverManagerProducer.java
```diff
@@ -0,0 +1,38 @@
...
+package org.optaplanner.spring.boot.example.poc;
+
+import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.spring.boot.example.poc.api.solver.SolverManager;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+// THIS IS JUST A PROOF OF CONCEPT IN THE EXAMPLE FOR 7.30.0.Final. THIS CLASS WILL BECOME OBSOLETE VERY SOON.
+// TODO Move this functionality into optaplanner-spring-boot-autoconfigure
```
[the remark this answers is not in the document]
ge0ffrey: This is in the example code, not in the autoconfigure or starter. The reason I pulled it out of the starter is because it's not officially supported until it's part of that. The reason I am marking it as poc in the example is so it sticks out like a soar thumb. Finishing the SolverManager API is still a priority - we're almost there - but I didn't want it to prevent the autoconfigure and starter to be merged, as those do a lot of good work already (SolverFactory injection for example)

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
ge0ffrey: The SolverFactory generation is a giant step forward. No more XML. Termination time settable in application.properties.
Throttling etc are nice to have's, that will come as soon as we finalize the SolverManager API and refactor the example to use it.

rsynek:
@ge0ffrey Have you considered putting the example under optaplanner-examples? It would require moving the existing examples into a sub-module, but it makes sense to place this new example there. I would definitely look for it in the optaplanner-examples module.

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
ge0ffrey: Hmm - good point. We can't release example code like this. Maybe we should mask out the example?

rsynek:
A readme showing how to run the example would be nice.

ge0ffrey:
> Looking at Sonar, there's some things we probably want to fix.
> Also, ~45 % code coverage is less than ideal for something we expect to bring us so much new audience.

Good point. The code coverage of the example is deplorable. The autoconfigure and starter files are 100% though, except for one file that's 77% (not all fail fasts are tested):
https://sonarcloud.io/component_measures?id=org.optaplanner%3Aoptaplanner&metric=new_coverage&pullRequest=595&selected=org.optaplanner%3Aoptaplanner%3Aoptaplanner-integration%2Foptaplanner-spring-boot-autoconfigure%2Fsrc%2Fmain%2Fjava%2Forg%2Foptaplanner%2Fspring%2Fboot%2Fautoconfigure%2FOptaPlannerAutoConfiguration.java&view=list

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/service/TimeTableService.java
```diff
@@ -0,0 +1,68 @@
...
+import org.optaplanner.spring.boot.example.poc.api.solver.SolverFuture;
+import org.optaplanner.spring.boot.example.poc.api.solver.SolverManager;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/timeTable")
+public class TimeTableService {
```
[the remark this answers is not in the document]
ge0ffrey: Should be easy to do, this will fix the low coverage too. Will do.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/TerminationProperties.java
ge0ffrey: Yes, absolutely true for UX.

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
triceo: I'm not sure what you're suggesting. How would masking out look?

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/PocSolverManagerProducer.java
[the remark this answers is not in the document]
triceo: Honestly, I've been thinking for a while that the SolverManager API is taking way too long. There's scope creep (Reactive etc.) and there's poor time management (meetings every fortnight).

We should finalize SolverManager. If we instead allow this PoC code in, people will start copying this code. For us, it's PoC. For them, it'll live forever.

optaplanner-integration/optaplanner-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+    <groupId>org.optaplanner</groupId>
+    <artifactId>optaplanner-integration</artifactId>
+    <version>7.31.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>optaplanner-spring-boot-starter</artifactId>
+
+  <name>OptaPlanner Spring Boot starter</name>
+  <description>
+    OptaPlanner solves planning problems.
```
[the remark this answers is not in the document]
ge0ffrey: Agreed, but I 'll handle this in a separate issue as it applies to all pom.xml files, not just this one.

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
```diff
@@ -56,12 +56,17 @@ public DefaultSolverFuture solve(Solution_ planningProblem,
             try {
                 solver.solve(planningProblem);
             } catch (Exception e) {
-                e.printStackTrace(); // TODO generated
+                e.printStackTrace();
```
triceo: You really mean to keep this in? If so, at the very least let's log it.

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/solver/TimeTableConstraintProvider.java
```diff
@@ -0,0 +1,78 @@
...
+                        Joiners.equal(Lesson::getRoom))
+                .reward("Teacher room stability", HardSoftScore.ONE_SOFT);
+    }
+
+//    private Constraint minimizeTeacherWorkDays(ConstraintFactory constraintFactory) {
+//        return constraintFactory
+//                .from(Lesson.class)
+//                .groupBy(Lesson::getTeacher, countDistinct((Lesson lesson) -> lesson.getTimeslot().getDayOfWeek()))
+//                .penalize("Minimize teacher work days", HardSoftScore.ONE_SOFT);
+//    }
```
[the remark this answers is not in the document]
ge0ffrey: Fixed, reanmed to teacherTimeEfficiency

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/PocSolverManagerProducer.java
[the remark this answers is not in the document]
ge0ffrey: > We should finalize SolverManager. If we instead allow this PoC code in, people will start copying this code. For us, it's PoC. For them, it'll live forever.

This PR now includes the SolverManager entirely :)

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
ge0ffrey: Right now, the SolverManager impl still does it on the solver thread. I 'll fix that soon. Not sure if it should block this PR from merging though.
(the lack of testing should, fixing that first)

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/poc/impl/solver/DefaultSolverManager.java
ge0ffrey: Fixed, now an exceptionHandler is supported

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/Solver.java
```diff
@@ -35,19 +35,50 @@
...
 
+    /**
+     * Solves the planning problem and returns the best solution encountered
+     * (which might or might not be optimal, feasible or even initialized).
+     * <p>
+     * It can take seconds, minutes, even hours or days before this method returns,
+     * depending on the {@link Termination} configuration.
+     * To terminate a {@link Solver} early, call {@link #terminateEarly()}.
+     * @param problem never null, a {@link PlanningSolution}, usually its planning variables are uninitialized
+     * @return never null, but it can return the original, uninitialized {@link PlanningSolution} with a {@link Score} null.
```
ron-murhammer: I think this should say "uninitialized PlanningSolution with a null Score" instead of "uninitialized PlanningSolution with a Score null".

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/Solver.java
```diff
@@ -35,19 +35,50 @@
...
+     * Notifies the solver that it should stop at its earliest convenience.
+     * This method returns immediately, but it takes an undetermined time
+     * for the {@link #solve} to actually return.
+     * <p>
+     * If the solver is running in daemon mode, this is the only way to terminate it normally.
+     * <p>
+     * This method is thread-safe.
+     * It can only be called from a different thread
+     * because the original thread is still calling {@link #solve(Object)}.
+     * @return true if successful
```
ron-murhammer: It might be useful to mention when this could return false?

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
```diff
@@ -36,9 +37,11 @@
  * Creates {@link Solver} instances.
  * Most applications only need one SolverFactory.
  * <p>
- * To build an instance, use {@link #createFromXmlResource(String)}.
- * To change the configuration programmatically, create a {@link SolverConfig}
- * and use {@link #create(SolverConfig)}.
+ * To create an SolverFactory, use {@link #createFromXmlResource(String)}.
```
ron-murhammer: Minor: should be "a SolverFactory" instead of "an SolverFactory"

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverManagerConfig.java
```diff
@@ -0,0 +1,139 @@
...
+        if (resolvedParallelSolverCount > availableProcessorCount) {
+            logger.warn("The resolvedParallelSolverCount ({}) is higher "
+                            + "than the availableProcessorCount ({}), which is counter-efficient.",
+                    resolvedParallelSolverCount, availableProcessorCount);
+            // Still allow it, to reproduce issues of a high-end server machine on a low-end developer machine
+        }
+        return resolvedParallelSolverCount;
+    }
+
+    protected int getAvailableProcessors() {
```
ron-murhammer: Should this method be made static?

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of a submitted problem, such as {@link Long} or {@link UUID}.
+ */
+public interface SolverJob<Solution_, ProblemId_> {
+
+    /**
+     * @return never null, a value given to {@link SolverManager#solveBatch(Object, Function, Consumer)}
+     *or {@link SolverManager#solveObserving(Object, Function, Consumer)}
```
triceo:
```suggestion
     * or {@link SolverManager#solveObserving(Object, Function, Consumer)}
```

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    ProblemId_ getProblemId();
+
+    /**
+     * Returns if the {@link Solver} is scheduled to solve, actively solving or not.
```
triceo:
```suggestion
     * Returns whether the {@link Solver} is scheduled to solve, actively solving or not.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    void terminateEarly();
+
+    /**
+     * Waits if necessary for the solver to complete and then returns the final best {@link PlanningSolution}.
+     * @return never null, but it could be the original uninitialized problem
+     * @throws InterruptedException if the current thread was interrupted while waiting
+     * @throws ExecutionException if the computation threw an exception
+     */
+    Solution_ getFinalBestSolution() throws InterruptedException, ExecutionException;
```
triceo: No `getCurrentBestSolution()`? Why do I need to wait?

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverManager.java
```diff
@@ -0,0 +1,203 @@
...
+            Function<ProblemId_, Solution_> problemFinder, Consumer<Solution_> finalBestSolutionConsumer,
+            BiConsumer<ProblemId_, Throwable> exceptionHandler);
+
+    /**
+     * Submits a planning problem to solve and returns immediately.
+     * The planning problem is solved on a solver {@link Thread}, as soon as one is available.
+     * <p>
+     * When the solver finds a new best solution, the {@code bestSolutionConsumer} is called every time,
+     * on a consumer {@link Thread}, as soon as one is available (taking into account any throttling waiting time),
+     * unless a newer best solution is already available by then (in which case skip ahead discards it).
```
triceo: "throttling time" and "skip ahead" are terms I have not seen described in the Javadocs.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverManager.java
```diff
@@ -0,0 +1,203 @@
...
+     * @param problemId never null, a value given to {@link #solveBatch(Object, Function, Consumer)}
+     * or {@link #solveObserving(Object, Function, Consumer)}
+     */
+    void terminateEarly(ProblemId_ problemId);
+
+    /**
+     * Calculates the {@link Score} of a {@link PlanningSolution} and updates its {@link PlanningScore} member.
+     * @param solution never null
+     */
+    void updateScore(Solution_ solution);
```
triceo: This is surprising to me. Why give users the power to do this? This looks like a call that would be internal to the solver. It completely breaks the abstraction of this class.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverStatus.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.api.solver;
+
+public enum SolverStatus {
+    SOLVING_SCHEDULED,
+    SOLVING_ACTIVE,
+    NOT_SOLVING;
+}
```
triceo: Do we want to describe the possible transitions between these?

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverManagerConfig.java
```diff
@@ -0,0 +1,139 @@
...
+
+    // ************************************************************************
+    // Builder methods
+    // ************************************************************************
+
+    public Integer resolveParallelSolverCount() {
+        int availableProcessorCount = getAvailableProcessors();
+        Integer resolvedParallelSolverCount;
+        if (parallelSolverCount == null || parallelSolverCount.equals(PARALLEL_SOLVER_COUNT_AUTO)) {
+            // Leave one for the Operating System and 1 for the solver thread, take the rest
```
triceo: Not sure why leave just 1 for solver thread. Shouldn't solver threads get most of the CPU unless the user wants otherwise?

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverManagerConfig.java
triceo: Also, it's a simple wrapper for another method. Why not remove it altogether?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -0,0 +1,119 @@
...
+                finalBestSolutionConsumer.accept(finalBestSolution);
+            }
+            return finalBestSolution;
+        } catch (Exception e) {
+            exceptionHandler.accept(problemId, e);
+            throw new IllegalStateException("Solving failed for problemId (" + problemId + ").", e);
+        } finally {
+            // TODO What happens if this code throws an exception? Is it eaten?
+            solverManager.getProblemIdToSolverJobMap().remove(problemId);
+            solverStatus = SolverStatus.NOT_SOLVING; // TODO FIXME race condition, lock on problemId
```
triceo: Sounds like an important `FIXME`. :-)

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/Solver.java
ge0ffrey: Yes, fixing

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/Solver.java
ge0ffrey: Yes, changed to: "@return true if successful, false if was already terminating or terminated"

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverManagerConfig.java
ge0ffrey: No, because some unit tests override it (at least for SolverConfig, but they'll do so here too) to mimic computers with a different number of cpu's than the random jenkins job that runs these tests too.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverManager.java
ge0ffrey: Users need it when loading a timetable from the database and showing it in the UI. It's already solved, but the UI needs to show the score (and indictments later).
Let's discuss alternative proposals for this tomorrow .

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverManagerConfig.java
ron-murhammer: Ah, makes sense.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverManagerConfig.java
ge0ffrey: This comment is actually false, so I am removing it.
By default, we only take half of the CPU threads. The reason that taking too many introduces freezes (including on the consumer threads, the REST threads, the OS, etc). That gives a very poor user experience. People don't like it. If they want to live closer to the border, they can configure a number manually (or use the availableProcesses variable).

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
ge0ffrey: Yea, I was in doubt about it. It could make a good future feature, but to track current best solutions, we already have solveObserving(). So I went with "When in doubt, leave it out." We can easily add it later if more people ask for it.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
ge0ffrey: Thanks for the review, Ron :)
What do you think of the naming? SolverManager, SolverJob, ...
and the example's code size?

optaplanner-integration/pom.xml
```diff
@@ -23,7 +23,76 @@
...
   </modules>
 
+
+  <properties>
+    <!-- Remove once https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1086 is merged -->
+    <version.org.springframework.boot>2.2.0.RELEASE</version.org.springframework.boot>
+    <version.org.springframework>5.2.0.RELEASE</version.org.springframework>
+    <!-- Remove once https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1075 is merged-->
+    <version.javax.validation>2.0.1.Final</version.javax.validation>
+    <version.org.hibernate.validator>6.0.17.Final</version.org.hibernate.validator>
```
ge0ffrey: Several of these can be removed, todo

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
ron-murhammer: Overall, I think the class naming is pretty good and makes sense. Only alternative I could think of is something like SolverPoolManager or something that indicates its a "pool" of solver threads. A few minor thoughts:
- Could use a bit more class javadoc on SolverJob
- I'm not sure about the `solveBatch` method name as it kind of makes me think you are sending in a 'batch' of problems to solve but that doesn't appear to be the case
- I don't fully understand why there are 3 different `solveBatch` methods with varying parameters and whether they are all useful (I see they just chain together so not a huge deal) but reading through I don't really understand when I would use one vs another
- I do agree that `updateScore` seems a little out of place so might be interesting in considering alternatives to avoid exposing that there

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverStatus.java
ge0ffrey: Fixed, thanks for reporting :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
ge0ffrey: Agreed to postpone from this PR into the next PR for this work that will include the consumer thread refactor (which will affect those 2 lines too). Also note that the race condition is not proven, but it looks fishy.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverManager.java
ge0ffrey: Agreed to move into ScoreService (a new class)

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverFactory.java
ge0ffrey: We went to rename solveBatch to solve(), more overloaded methods, clearer javadocs to explain which to use when.
We also renamed solveObserving() to solveAndListen()

optaplanner-integration/optaplanner-spring-boot-example/src/main/java/org/optaplanner/spring/boot/example/service/TimeTableService.java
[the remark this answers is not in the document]
ge0ffrey: Done. Excluding the REST methods that Spring automatically exposes for CrudRepository, as they already test that.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/ScoreManager.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.api.score;
+
+import java.util.UUID;
```
yurloc: Several unused imports in this class.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverManager.java
```diff
@@ -0,0 +1,256 @@
...
+ */
+
+package org.optaplanner.core.api.solver;
+
+import java.util.UUID;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.optaplanner.core.api.domain.solution.PlanningScore;
```
yurloc: Unused imports.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/DefaultScoreManager.java
```diff
@@ -0,0 +1,47 @@
...
+import org.optaplanner.core.api.score.ScoreManager;
+import org.optaplanner.core.impl.score.director.ScoreDirector;
+import org.optaplanner.core.impl.score.director.ScoreDirectorFactory;
+
+/**
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ */
+public class DefaultScoreManager<Solution_> implements ScoreManager<Solution_> {
+
+    private ScoreDirectorFactory<Solution_> scoreDirectorFactory;
```
yurloc: May be final.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -0,0 +1,119 @@
...
+    private final Consumer<Solution_> finalBestSolutionConsumer;
+    private final BiConsumer<ProblemId_, Throwable> exceptionHandler;
+
+    private volatile SolverStatus solverStatus;
+    private Future<Solution_> future;
+
+    public DefaultSolverJob(DefaultSolverManager<Solution_, ProblemId_> solverManager, ProblemId_ problemId,
+            Solver<Solution_> solver,
+            Function<ProblemId_, Solution_> problemFinder,
+            Consumer<Solution_> finalBestSolutionConsumer,
```
yurloc: Can generalize to `<? super Solution_>`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverManager.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of a submitted problem, such as {@link Long} or {@link UUID}.
+ */
+public class DefaultSolverManager<Solution_, ProblemId_> implements SolverManager<Solution_, ProblemId_> {
+
+    protected final transient Logger logger = LoggerFactory.getLogger(getClass());
+
+    private BiConsumer<ProblemId_, Throwable> defaultExceptionHandler;
```
yurloc: May be final. I find this keyword helpful even when just reading the code. It's immediately clear that this field will never change.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerProperties.java
```diff
@@ -0,0 +1,64 @@
...
+public class OptaPlannerProperties {
+
+    public static final String DEFAULT_SOLVER_CONFIG_URL = "solverConfig.xml";
+
+    /**
+     * A classpath resource to read the solver configuration XML.
+     * Defaults to {@value DEFAULT_SOLVER_CONFIG_URL}.
+     * If this property isn't specified, that solverConfig.xml is optional.
+     */
+    private String solverConfigXML;
```
yurloc: Should be `solverConfigXml`.

When it's `solverConfigXML`, IDEA auto-suggests a property called `optaplanner.solver-config-x-m-l` which is not what we want.

optaplanner-integration/optaplanner-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+  <dependencies>
+    <!-- Internal dependencies -->
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-spring-boot-autoconfigure</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-persistence-jackson</artifactId>
+    </dependency>
```
yurloc: Missing dependency on `optaplanner-core`.

https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-custom-starter:
> The starter module that provides a dependency to the autoconfigure module as well as the library and any additional dependencies that are typically useful. In a nutshell, adding the starter should provide everything needed to start using that library.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
```
yurloc: Missing dependency:
```xml


<dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-autoconfigure-processor</artifactId>
    <optional>true</optional>
</dependency>
```
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-custom-starter:
> 

Spring Boot uses an annotation processor to collect the conditions on auto-configurations in a metadata file (META-INF/spring-autoconfigure-metadata.properties). If that file is present, it is used to eagerly filter auto-configurations that do not match, which will improve startup time. It is recommended to add the following dependency in a module that contains auto-configurations:

optaplanner-integration/optaplanner-spring-boot-autoconfigure/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    This module contains the Spring Boot autoconfigure.
+  </description>
+  <url>https://www.optaplanner.org</url>
+
+  <dependencies>
+    <!-- Internal dependencies -->
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-core</artifactId>
+    </dependency>
```
yurloc: Should this be optional, I believe. IIUC, in theory, you can have optaplanner-autoconfiguration and unless you also have optaplanner-core on classpath, this autoconfiguration can be skipped. Conversely, if you hard-depend on optaplanner-core here, it will always be on classpath and so this autoconfiguration can never be skipped.

https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-custom-starter-module-autoconfigure:
> You should mark the dependencies to the library as optional so that you can include the autoconfigure module in your projects more easily. If you do it that way, the library is not provided and, by default, Spring Boot backs off.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverManagerProperties.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.spring.boot.autoconfigure;
+
+import org.optaplanner.core.config.solver.SolverManagerConfig;
+
+public class SolverManagerProperties {
```
yurloc: Should be an inner class of `OptaPlannerProperties` to enable metadata generation.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
```diff
@@ -0,0 +1,67 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.spring.boot.autoconfigure;
+
+import org.optaplanner.core.config.solver.EnvironmentMode;
+import org.optaplanner.core.config.solver.SolverConfig;
+
+public class SolverProperties {
```
yurloc: Should be an inner class of `OptaPlannerProperties` to enable metadata generation.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/TerminationProperties.java
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.spring.boot.autoconfigure;
+
+import java.time.Duration;
+
+public class TerminationProperties {
```
yurloc: Should be an inner class of `SolverProperties` to enable metadata generation.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerProperties.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private SolverProperties solver;
```
yurloc: Should be annotated with `@NestedConfigurationProperty`:
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#configuration-metadata-nested-properties

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerProperties.java
```diff
@@ -0,0 +1,64 @@
...
+    private SolverManagerProperties solverManager;
```
yurloc: Should be annotated with `@NestedConfigurationProperty`:
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#configuration-metadata-nested-properties

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
```diff
@@ -0,0 +1,67 @@
...
+    private EnvironmentMode environmentMode;
+    /**
+     * Enable multithreaded solving for a single problem, which increases CPU consumption.
+     * Defaults to {@value SolverConfig#MOVE_THREAD_COUNT_NONE}.
+     * Other options include {@value SolverConfig#MOVE_THREAD_COUNT_AUTO}, a number
+     * or formula based on the available processor count.
+     */
+    private String moveThreadCount = null;
+
+    private TerminationProperties termination = new TerminationProperties();
```
yurloc: Should be annotated with `@NestedConfigurationProperty`:
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#configuration-metadata-nested-properties

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/TerminationProperties.java
yurloc: Resolving. There's a better solution with `@NestedConfigurationProperty`.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
yurloc: Resolving. There's a better solution with `@NestedConfigurationProperty`.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverManagerProperties.java
yurloc: Resolving. There's a better solution with `@NestedConfigurationProperty`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
ge0ffrey: Inetresting. PECS I presume? (Producer Extends Consumer Super)

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerProperties.java
ge0ffrey: Very good catch. (fixing too)

optaplanner-integration/optaplanner-spring-boot-autoconfigure/pom.xml
ge0ffrey: Is it not in there on line 46?

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerProperties.java
ge0ffrey: That explains why IntelliJ didn't pick them up

optaplanner-integration/optaplanner-spring-boot-starter/pom.xml
ge0ffrey: optaplanner-core is a non-optional dependency of the autoconfigure, so it's not an "additional" dependency I'd argue? wdyt?

optaplanner-integration/optaplanner-spring-boot-autoconfigure/pom.xml
ge0ffrey: Agreed, fixed

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
yurloc: Yes. Maybe a separate PR but worth thinking about, I'd say.

ge0ffrey: Definitely, on my loose ends list for separate PR's.

PR #598, opened 2019-10-30
PLANNER-1693 Make comparator null-safe in cheaptime example

triceo:


optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/swingui/CheapTimePanel.java
```diff
@@ -56,6 +53,10 @@
 import org.optaplanner.examples.common.swingui.SolutionPanel;
 import org.optaplanner.swing.impl.TangoColorFactory;
 
+import static com.google.common.base.Functions.identity;
```
ge0ffrey: Theres's a identity function in JDK too:
https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#identity--

In time, we want to get rid of guava (= com.google.common) because it's not backward compatible between minor versions.

PR #605, opened 2019-11-04
Constraint stream images: join

ge0ffrey:


ge0ffrey:
@triceo Can you review? We can add the text around it soon, it's a good start I think.

PR #609, opened 2019-11-12
PLANNER-1711: No more insertLogical()

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -43,8 +46,12 @@ public ConstraintMatch(String constraintPackage, String constraintName,
             List<Object> justificationList, Score score) {
         this.constraintPackage = constraintPackage;
         this.constraintName = constraintName;
-        this.justificationList = justificationList;
+        this.justificationList = new ArrayList<>(justificationList);
         this.score = score;
+        this.hashCode = (((17 * 37)
+                + constraintPackage.hashCode()) * 37
+                + constraintName.hashCode()) * 37
+                + justificationList.hashCode();
```
triceo: `justificationList` contains results of `accumulate`, which could be `Set`s with varying contents at varying times. Therefore, `justificationList` *changes from the outside* and we must enforce stable `hashCode` - otherwise all hell breaks loose later.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AbstractConstraintStreamTest.java
```diff
@@ -84,17 +84,22 @@ public AbstractConstraintStreamTest(boolean constraintMatchEnabled, ConstraintSt
...
                 .flatMap(item -> {
-                    if (item instanceof DroolsLogicalTuple) {
-                        return removeIndirection((DroolsLogicalTuple) item).stream();
+                    if (item instanceof Collection) {
+                        /*
+                         * Removes results of intermediate accumulate operations from constraint matches. In CS-D, those are
+                         * typically by-products of other operations and including them would mean that constraint
+                         * matches of CS-D would be incomparable to CS-B.
+                         */
+                        return Stream.empty();
```
triceo: There's likely corner cases that this will fail on.

triceo:
Jenkins please retest this.

triceo:
@ge0ffrey First piece of 7.7 CS-D ready for review.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
ge0ffrey: -1 until discussed.
This affects the ConstraintMatch DRL usage and Bavet's ConstraintMatch usage. Do we want that?
Or can we isolate this into a flag on ScoreHolder to active CS-D special stuff?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -161,6 +163,9 @@
                 () -> new int[1],
                 (resultContainer, a) -> {
                     int value = groupValueMapping.applyAsInt(a);
+                    if (value == 0) {
+                        return NOOP;
```
ge0ffrey: Does this give a noticable performance boost? Is it worth the added cost in code readability?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+        UnaryOperator<PatternDSL.PatternDef<T>> expanderWithLogging = p -> {
+            LOGGER.trace("Expanding {} pattern with '{}'.", baseVariable, description);
+            return expander.apply(p);
+        };
+        return expand(expanderWithLogging);
+    }
+
+    private PatternDSL.PatternDef<T> build(PatternDSL.PatternDef<T> basePattern) {
+        PatternDSL.PatternDef<T> result = builder.apply(basePattern);
+        LOGGER.trace("Pattern for {} finished: {}.", baseVariable, result);
```
ge0ffrey: We encourage users to turn on "TRACE" logging to see the actual moves that optaplanner is taking.
Therefore, even the trace log must be _readable_ and contain useful information only.
One trick that often works well, is to ident trace logging with 2 spaces, so it's easier to see "solving started" and "step ended" log lines etc.

If these trace loggings are purely here for our debugging purposes, I recommend to use "debugger log messages" instead.
I think it would be useful to see which drools constraints are build from the CS, in trace logging, at the end of creating a constraint. I am not sure it's useful to see both "start" and "end" messages - I am not sure if those are worth their weight.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsRuleStructure.java
```diff
@@ -0,0 +1,131 @@
...
+import org.drools.model.DeclarationSource;
+import org.drools.model.RuleItemBuilder;
+import org.drools.model.Variable;
+
+/**
+ * Represents the left-hand side of a Drools rule.
+ */
+public abstract class DroolsRuleStructure {
+
+    private static final AtomicInteger COUNTER = new AtomicInteger();
```
ge0ffrey: This "static" bugs me.
If I build a SolverFactory twice (or build another one first or in parallel concurrently), will I get different ID's for my CS-D implementation and therefore make reproduction harder?
Think optaplanner-benchmark AUTO.

This counter must not be a static, but a shared stated, starting for zero for one CS-D build.

Furthermore. ScoreDirector is single threaded. No need for AtomicInteger. But what about ScoreDirectorFactory? I 'd still say no need, but let's discuss :)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: There are two different situations here:

1. The constraint weight is zero, and therefore the rule should not be executed at all. I agree, that should be handled somewhere else. (It isn't handled at all.)
1. The input value to the collector is zero, and therefore there is no change to the collector. I would argue that this is still a valid situation - if you're collecting min/max of something, you can reasonably expect zeros to be there. And in case of a zero, there is no change to the collector - and therefore this is a valid shortcut.

Whether or not this brings actual value in terms of performance, that is hard to say. This PR does so much with performance that I no longer have separate data just for this code change. I can remove this, if you insist.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsPatternBuilder.java
triceo: Wrt. indentation - can do.

Wrt."debugger log messages" - I'm probably not familiar with that pattern. Can you point me someplace?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsRuleStructure.java
triceo: The IDs here are used plainly to guarantee variable name uniqueness - they have no other functional use, and therefore I would argue that they're irrelevant for the purposes of reproducibility.

However, I see your point. I'll see if I can somehow guarantee uniqueness without the IDs.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
triceo: Let's discuss. But it's hard to argue that having inconsistent `hashCode()` is anything else but a bug.

triceo:
All comments resolved, ready for another round of review, @ge0ffrey.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -224,4 +229,18 @@ public final void run() {
...
+     * This is necessary because CS-D uses some advanced Drools constructions leveraging various metadata objects.
+     * Had we not called this adapter, these metadata objects would have been present in the justification list,
+     * defeating its purpose.
+     * Code outside of CS-D has no need to call this method.
+     *
+     * @param justificationListAdapter function to clean up the justification list
+     */
+    public void setJustificationListAdapter(BiFunction<List<Object>, Rule, List<Object>> justificationListAdapter) {
+        this.justificationListAdapter = justificationListAdapter;
+    }
```
triceo: I don't really like that I had to put this on the public API.
I considered to put it in the constructor, but that would require me to update every single `ScoreHolder` impl.
I'm listening for other ideas.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -56,7 +59,7 @@ public String getConstraintName() {
     }
 
     public List<Object> getJustificationList() {
-        return justificationList;
+        return Collections.unmodifiableList(justificationList);
```
ge0ffrey: We're duplicating it in the constructor and wrapping it here in an unmodifiable List on every call?
Can we do that in the constructor then too?

Still not sure if it's worth it though. This costs performance wise.
2 proposals:
Proposal A): in constructor: Collections.unmodifiableList(new ArrayList<>(justificationList))
Proposal B:  in constructor: Collections.unmodifiableList(justificationList)

I suspect that for DRL and CS-B, proposal B is bugfree, so the duplication cost of proposal A is a perf loss.
For CS-D we can duplicate before giving it to that constructor?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
ge0ffrey: Yea :(
Let's discuss on bluejeans.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
```diff
@@ -164,6 +92,19 @@ public String getConstraintName() {
         return constraintName;
     }
 
+    public int getConstraintStreamCardinality() {
+        if (scoringStream instanceof UniConstraintStream) {
+            return 1;
+        } else if (scoringStream instanceof BiConstraintStream) {
+            return 2;
+        } else if (scoringStream instanceof TriConstraintStream) {
+            return 3;
```
ge0ffrey: Any idea what PolyConstraintStream will return? Int.MAX?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
triceo: The copy is there so that it can not be modified from the outside by people having the reference.
The `unmodifiableList` is there, so that the copy can not be modified after the user requests it.
Both are necessary. But I agree that it can be done in a single place.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraint.java
triceo: This is a solution to the justification list problem - I think we'll have to solve that before poly streams.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -122,10 +125,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return (((17 * 37)
-                + constraintPackage.hashCode()) * 37
-                + constraintName.hashCode()) * 37
-                + justificationList.hashCode();
+        return Objects.hash(constraintPackage, constraintName, justificationList);
```
ge0ffrey: +1 great idea.
Can you create a jira to look for other places that magically use the number 37 to create a hashcode. I am pretty sure I did it a lot before java 8 come in.

triceo: https://issues.jboss.org/browse/PLANNER-1725

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsGroupByInvoker.java
```diff
@@ -0,0 +1,101 @@
...
+        Class<?> accumulateClass = DroolsGroupBy.class;
+        String classFileName = accumulateClass.getCanonicalName().replace('.', '/') + ".class";
+        try (InputStream is = getClass().getClassLoader().getResourceAsStream(classFileName)) {
+            final byte[] data = new byte[1024];
+            int byteCount;
+            ByteArrayOutputStream bos = new ByteArrayOutputStream();
+            while ((byteCount = is.read(data, 0, 1024)) > -1) {
+                bos.write(data, 0, byteCount);
+            }
+            return bos.toString();
```
rsynek: This `toString()` is platform-dependent; maybe we want to pass a charset there.

rsynek:
@triceo btw., nice job with getting coverage over 92%. I am sure that after seeing this @ge0ffrey will agree with keeping the default 80% threshold :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsGroupByInvoker.java
triceo: After talking to Radovan, we think this is not an issue. We're reading a classfile, and classfile encoding must be platform independent. (Otherwise it would defeat the purpose of bytecode, to be the same on all platforms.)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -194,7 +198,8 @@ public void impactScore(RuleContext kcontext, BigDecimal weightMultiplier) {
 
     protected List<Object> extractJustificationList(RuleContext kcontext) {
         // Unlike kcontext.getMatch().getObjects(), this includes the matches of accumulate and exists
-        return ((org.drools.core.spi.Activation) kcontext.getMatch()).getObjectsDeep();
+        Activation activation = (Activation) kcontext.getMatch();
+        return justificationListAdapter.apply(activation.getObjectsDeep(), kcontext.getRule());
```
triceo: Add a null check?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
triceo: We keep the method. We rename it to *Converter.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/AbstractScoreHolder.java
```diff
@@ -224,4 +229,18 @@ public final void run() {
...
         }
     }
+
+    /**
+     * Used for Drools-based constraint streams (CS-D) to convert justification list to the same format as the one used
+     * by Bavet constraint streams.
+     * This is necessary because CS-D uses some advanced Drools constructions leveraging various metadata objects.
+     * Had we not called this adapter, these metadata objects would have been present in the justification list,
+     * defeating its purpose.
+     * Code outside of CS-D has no need to call this method.
```
triceo: Remove from Javadoc, move inside the method.

triceo:
@rsynek Actually I'm more proud of the near-zero code duplication. That's hard to do in CS.

PR #613, opened 2019-11-18
Planner1641 Migrate EnvironmentModeTest to upstream

michaltomco:


optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
```diff
@@ -0,0 +1,237 @@
...
+        // for full assert modes it should throw exception about corrupted score
+        setSolverConfigCalculatorClass(TestdataCorruptedDifferentValuesCalculator.class);
+
+        switch (environmentMode) {
+            case FULL_ASSERT:
+            case NON_INTRUSIVE_FULL_ASSERT:
+                assertIllegalStateExceptionWhileSolving("not the uncorruptedScore");
+                break;
+            case FAST_ASSERT:
+                assertIllegalStateExceptionWhileSolving("Score corruption analysis could not be generated ");
```
triceo: There is a weird inconsistency. Sometimes you start the `String` with a capital letter, sometimes not.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
```diff
@@ -0,0 +1,237 @@
...
+        ((DefaultSolver<TestdataSolution>) solver2).addPhaseLifecycleListener(listener2);
+
+        for (int i = 0; i < NUMBER_OF_TIMES_RUN && areSame; i++) {
+            solver.solve(inputProblem);
+            solver2.solve(inputProblem);
+
+            List<Integer> scoreTimeSeries = listener.getScores();
+            List<Integer> scoreTimeSeries2 = listener2.getScores();
+
+            areSame = scoreTimeSeries.equals(scoreTimeSeries2);
```
triceo: Personally, I'd add the `Assertion` here. This way, the failure message will be more descriptive than ("false was not true").

triceo: Since you want to check multiple scores in a loop, you may want to give `SoftAssertions` a try.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
```diff
@@ -0,0 +1,237 @@
...
+        return areSame;
+    }
+
+    private boolean areGeneratingSameNumbers(RandomFactory f1, RandomFactory f2) {
+        boolean areSame = true;
+        Random random = f1.createRandom();
+        Random random2 = f2.createRandom();
+        for (int i = 0; i < EnvironmentModeTest.NUMBER_OF_RANDOM_NUMBERS_GENERATED; i++) {
+            if (random.nextInt() != random2.nextInt()) {
+                areSame = false;
```
triceo: Ditto.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/custom/TestdataFirstValueInitializer.java
```diff
@@ -0,0 +1,33 @@
...
+            entity.setValue(firstValue);
+            scoreDirector.afterVariableChanged(entity, "value");
+        }
+
+        scoreDirector.triggerVariableListeners();
+        Score<?> score = scoreDirector.calculateScore();
+
+        if (!score.isSolutionInitialized()) {
+            throw new IllegalStateException(
+                    "The solution [" + TestdataEntity.class.getSimpleName()
```
triceo: We use `(` and `)` to quote parameters in exception messages.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/custom/TestdataFirstValueInitializer.java
```diff
@@ -0,0 +1,33 @@
...
+                            + "] was not fully initialized by CustomSolverPhase: "
+                            + this.getClass().getCanonicalName());
```
triceo: And since you're quoting, be consistent and quote the phase here as well.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/event/TestdataStepScoreListener.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    public void stepEnded(AbstractStepScope<TestdataSolution> stepScope) {
+        TestdataSolution solution = stepScope.getWorkingSolution();
+
+        if (solution.getScore() != null) {
+            scores.add(solution.getScore().getScore());
+        }
+    }
+
+    public List<Integer> getScores() {
```
triceo: Any reason for this `Integer` extraction from `Score`? You can compare scores directly, I couldn't think of a need for this.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/TestdataDifferentValuesCalculator.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.score.director.easy.EasyScoreCalculator;
+import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
+import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
+import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+
+public class TestdataDifferentValuesCalculator implements EasyScoreCalculator<TestdataSolution> {
```
triceo: Personally, since the change from the corrupted to this one is just in the one line at the end, I'd merge both impls in one class and differentiate them by a boolean constructor argument or something. That way, the difference between the two calculators will be more obvious.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
```diff
@@ -0,0 +1,237 @@
...
+                .isTrue();
+    }
+
+    private void assertNonReproducibility(Solver solver1, Solver solver2) {
+        assertThat(areGeneratingSameNumbers(
+                ((DefaultSolver) solver1).getRandomFactory(),
+                ((DefaultSolver) solver2).getRandomFactory())
+        )
+                .as("Random factories generate exactly same results. "
+                            + "It can happen but the probability is very low. Run test again")
```
triceo: This test is intentionally flaky? Not sure we should include it.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
michaltomco: The String is a part of an exception message (not the whole message), sometimes it's mid-sentence and needs to be lower-case.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
michaltomco: I made the problem more complex and, since it being rather fast, had it running in the background while working on this PR today for a while. There were no failed tests out of about 120,000 runs. The probability of it failing is very low and would suggest keeping it as it is. What do you think?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/TestdataDifferentValuesCalculator.java
michaltomco: Since ScoreDirectorFactoryConfig does not take an instance of ScoreCalculator, but a Class object of a Calculator instead, I am not sure this would be a plausible solution. Do you think this can be done in a different way or it should be kept as it is?

michaltomco:
Thank you Lukáš for your suggestions! They helped a lot. Hopefully everything is fine now. Sorry for tinkering with the repository's history by the way. In addition to the commit I wanted to push, I've managed to push a few ugly one with it as well. I decided to clean up the history and force-push it back neat and clean.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/TestdataDifferentValuesCalculator.java
triceo: @michaltomco You can have two classes sharing a super-class, which could have this boolean switch in the constructor.

(I'm kinda sensitive to code duplication. We can probably live without this.)

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
```diff
@@ -0,0 +1,251 @@
...
+                    .isEqualTo(random2.nextInt());
+        }
+    }
+
+    private void assertGeneratingDifferentNumbers(RandomFactory factory1, RandomFactory factory2) {
+        Random random = factory1.createRandom();
+        Random random2 = factory2.createRandom();
+
+        for (int i = 0; i < EnvironmentModeTest.NUMBER_OF_RANDOM_NUMBERS_GENERATED; i++) {
+            Assertions.assertThat(random.nextInt()).as("Random factories should not generate exactly the same results "
```
triceo: Check the `SoftAssertions` pattern. You will be able to get the entire loop covered by one assertion.

optaplanner-core/src/test/java/org/optaplanner/core/config/solver/EnvironmentModeTest.java
triceo: OK. But if I see it failing a build, I'll get back to you. :-)

michaltomco: Deal! :)

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/AbstractTestdataDifferentValuesCalculator.java
```diff
@@ -0,0 +1,43 @@
...
+                    alreadyUsedValues.add(value);
+                }
+            }
+        }
+        if(isCorrupted) {
+            numOfCalls += 1;
+            return SimpleScore.of(score - numOfCalls);
+        }else{
+            return SimpleScore.of(score);
+        }
```
triceo: Code style. (Spaces before `(` and on each side of `else`.)

PR #615, opened 2019-11-19
PLANNER-1716 Switch OptaPlanner to junit-vintage-engine

rsynek:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1116

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,13 @@
     </dependencies>
   </dependencyManagement>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+    </dependency>
+  </dependencies>
+
   <profiles>
     <profile>
       <id>fullProfile</id>


rsynek:
jenkins retest this

ge0ffrey:
@rsynek optaplanner-test currently grabs in junit 4 at comile time because it imports org.junit.Assert.* . That's should remain as-is indeed: but in 8.0, we'll want to change that to junit 5. I 'll add it to the 8.0 jira todo list.

rsynek:
@yurloc yes, it should be possible to start using JUnit 5 features in the code base - of course first we need to add some other dependencies, e.g. junit-jupiter-api.

triceo:
Correct. `vintage` is just the v4 APIs implemented with v5 backend. Jupiter is the v5 API over the v5 backend.

mareknovotny:
Shouldn't you start open discussion for other teams to suggest other teams to start use this one step ahead?
As now with this PR we will have in kiegroup different junit and only Optaplanner will use vintage module

rsynek:
@mareknovotny +1 for discussing it with the entire group. I tried to do the switch in kie-parent, but all tests based on https://github.com/kiegroup/lienzo-tests/blob/master/src/main/java/com/ait/lienzo/test/LienzoMockitoTestRunner.java fail due to a classloading issue.

ge0ffrey:
Merged these PR's, but that discussion with the entire group should start in parallel @rsynek - Ideally the drools and droolsjbpm-integration repository also switch to the new engine sooner rather than later.

PR #618, opened 2019-11-25
PLANNER-1678 Basic CS-D support for QuadStreams

triceo:
No groupBy yet.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -136,6 +139,17 @@
...
     // TODO other TriJoiner methods
 
+    // ************************************************************************
+    // QuadJoiner
+    // ************************************************************************
+
+    public static <A, B, C, D, Property_> QuadJoiner<A, B, C, D> equal(
+            TriFunction<A, B, C, Property_> leftMapping, Function <D, Property_> rightMapping) {
+        return new SingleQuadJoiner<>(leftMapping, JoinerType.EQUAL, rightMapping);
+    }
```
ge0ffrey: Yes, we need this equals(TriFunction, Function).

Open question: Do we _need_ equals(BiFunction, BiFunction) too?
I can think of use cases where this would give a performance benefit:
- Give me every A, B, C, D for which A.foo == B.foo and C.bar == D.bar and A.number * B.number * C.number * D.number == 42

But when in doubt, leave it out?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintCollector.java
```diff
@@ -0,0 +1,64 @@
...
+ * Loosely based on JDK's {@link Collector}, but it returns an undo operation for each accumulation
+ * to enable incremental score calculation in {@link ConstraintStream constraint streams}.
+ * @param <A> the type of the first matched fact in the source {@link BiConstraintStream}
+ * @param <B> the type of the second matched fact in the source {@link BiConstraintStream}
+ * @param <C> the type of the third matched fact in the source {@link BiConstraintStream}
+ * @param <ResultContainer_> the mutable accumulation type (often hidden as an implementation detail)
+ * @param <Result_> the type of a match fact in the destination {@link ConstraintStream}
+ * @see ConstraintCollectors
+ */
+public interface TriConstraintCollector<A, B, C, ResultContainer_, Result_> {
```
ge0ffrey: I don't mind this being here, good to merge, but this is unrelated to quad steams, right?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/quad/AbstractQuadJoiner.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        if (joinerList.isEmpty()) {
+            return new NoneQuadJoiner<>();
+        } else if (joinerList.size() == 1) {
+            return joinerList.get(0);
+        }
+        return new CompositeQuadJoiner<>(joinerList);
+    }
+
+    public abstract TriFunction<A, B, C, Object> getLeftMapping(int joinerId);
```
ge0ffrey: Here's where Bi.join(Bi) would make our life painful. As we'd need a different QuadJoiner for Tri.join(Uni) and Bi.join(Bi).
As this method is defined in an impl class, we can figure this out later. So when in doubt, leave Bi.join(Bi) out.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/quad/QuadConstraintStreamTest.java
```diff
@@ -0,0 +1,311 @@
...
+                = new ConstraintStreamScoreDirectorFactory<>(
+                TestdataSimpleBigDecimalScoreSolution.buildSolutionDescriptor(), (factory) -> new Constraint[]{
+                factory.from(TestdataEntity.class)
+                        .join(TestdataValue.class, equal(TestdataEntity::getValue, identity()))
+                        .join(TestdataValue.class)
+                        .reward("myConstraint1", SimpleBigDecimalScore.ONE),
+                factory.from(TestdataEntity.class)
+                        .join(TestdataValue.class, equal(TestdataEntity::getValue, identity()))
+                        .join(TestdataValue.class)
+                        .rewardBigDecimal("myConstraint2", SimpleBigDecimalScore.ONE, (entity, value, extra) -> new BigDecimal("0.2"))
```
ge0ffrey: This looks to me as if it creates a Tri stream, not a quad stream.

ge0ffrey: the matcher confirms that, clearly a tri stream matcher.

ge0ffrey: Other cases above too - the quad streams tests is really good at testing tri streams :)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -52,7 +53,7 @@
      * Exhaustively test each tuple of facts against the {@link BiPredicate}
      * and match if {@link BiPredicate#test(Object, Object)} returns true.
      * <p>
-     * Important: This is slower and less scalable than {@link UniConstraintStream#join(UniConstraintStream, BiJoiner[])}
+     * Important: This is slower and less scalable than {@link UniConstraintStream#join(UniConstraintStream, BiJoiner)}
```
ge0ffrey: See comment on TriConstraintStream

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintStream.java
```diff
@@ -50,14 +55,147 @@
      * Exhaustively test each tuple of facts against the {@link TriPredicate}
      * and match if {@link TriPredicate#test(Object, Object, Object)} returns true.
      * <p>
-     * Important: This is slower and less scalable than {@link BiConstraintStream#join(UniConstraintStream, TriJoiner[])}
+     * Important: This is slower and less scalable than {@link BiConstraintStream#join(Class, TriJoiner[])}
```
ge0ffrey: I am ok with doing this change from stream to class,
but all javadoc constraint must be consistent in all Javadocs for constraintstreams.
BiStream's version doesn't change this into Class as far as I can tell.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintStream.java
```diff
@@ -50,14 +55,147 @@
...
+     * @param joiner never null
+     * @param <D> the type of the fourth matched fact
+     * @return a stream that matches every combination of [A, B, C] and D for which the {@link QuadJoiner} is true
+     */
+    <D> QuadConstraintStream<A, B, C, D> join(UniConstraintStream<D> otherStream, QuadJoiner<A, B, C, D> joiner);
+
+    /**
+     * Create a new {@link QuadConstraintStream} for every combination of [A, B, C] and D.
+     * <p>
+     * Important: {@link QuadConstraintStream#filter(QuadPredicate)}  Filtering} this is slower and less scalable
```
ge0ffrey: Double space before Filtering.

Is it consistent otherwise with the Tri/Bi/Uni versions?
Yes, this javadoc and API duplication is a pita (but we take the pain, to avoid pain on the user): as long as we have the api duplication, we can't change anything in the api or javadoc without also adjusting the other stream types consistently in the same commit.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintCollector.java
triceo: Well spotted, I'll take it out. Let's make this PR clean.

PR #624, opened 2019-11-29
Deprecate forceUpdateBestSolution

ge0ffrey:
Motivation: forceUpdateBestSolution was a hack, introduced before PFC probably, to use in a Santa compo once. I am not aware of anyone else actually using it.
This is not compatible with https://issues.jboss.org/browse/PLANNER-241 which allows for skipping phases if all entities are immovable.
Furthermore, it is fundamentally wrong for phases to worsen the best score.


ge0ffrey:
SonarCloud says "Do not forget to remove this deprecated code someday." so I am ignoring it.

PR #625, opened 2019-11-29
PLANNER-241 Solve with an empty entity list or an all-pinned entity list

ge0ffrey:


ge0ffrey:
@wmedvede @yurloc This PR should fix (some of) your workarounds: if the entity list is empty or has only pinned entities (for example only pinned tasks), it will no longer fail, but it will skip solving instead:

```
15:37:25.526 [main        ] INFO  Solving started: time spent (13), best score (0), environment mode (REPRODUCIBLE), random (JDK with seed 0).
15:37:25.535 [main        ] INFO  Skipped all phases (1): out of 0 planning entities, none are movable (non-pinned).
15:37:25.535 [main        ] INFO  Solving ended: time spent (24), best score (0), score calculation speed (41/sec), phase total (1), environment mode (REPRODUCIBLE).
```

If the value list (for example employee list) is empty, it still fails.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1166,6 +1155,22 @@ public int countReinitializableVariables(ScoreDirector<Solution_> scoreDirector,
...
+            Object entity = extractMemberObject(memberAccessor, solution);
+            if (entity != null) {
+                streamList.add(Stream.of(entity));
+            }
+        }
+        for (MemberAccessor memberAccessor : entityCollectionMemberAccessorMap.values()) {
+            Collection<Object> entityCollection = extractMemberCollectionOrArray(memberAccessor, solution, false);
+            streamList.add(entityCollection.stream());
+        }
+        return streamList.stream().flatMap(Function.identity());
```
triceo: I think that you can do the same by using `Stream.concat()`. That will allow you to get rid of the `List`.

ge0ffrey: It behaves like concat (= lazy), but concat() needs an array as input. We don't know how long the array will be in advance, so we need to create a list anyway :)

triceo: The `List` isn't lazy either. :-) But you can replace `list.add(Stream.of(x))` by `stream = Stream.concat(stream, x)` and save yourself quite some code.

ge0ffrey: ah, I didn't know, let me check that.

ge0ffrey: yes, it works :)

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/DefaultSolverTest.java
```diff
@@ -77,12 +82,122 @@ public void solveLegacy() {
...
+        solution.setChainedAnchorList(Arrays.asList(new TestdataChainedAnchor("v1"), new TestdataChainedAnchor("v2")));
+        solution.setChainedEntityList(Collections.emptyList());
+
+        solution = solver.solve(solution);
+        assertNotNull(solution);
+        assertEquals(true, solution.getScore().isSolutionInitialized());
+        assertSame(solution, solver.getBestSolution());
+    }
+
+    @Test @Ignore("We currently don't support an empty value list yet if the entity list is not empty.")
```
rsynek: Is there a follow-up Jira ticket for this? If so, let's link it here, please.

ge0ffrey: Done, added https://issues.jboss.org/browse/PLANNER-1738

PR #629, opened 2019-12-04
PLANNER-1650 Write tests for NearbySelectionConfig

michaltomco:


rsynek:
@michaltomco thanks, will take a look tomorrow, if you're fine with that.

PR #630, opened 2019-12-06
issues.jboss.org renamed to issues.redhat.com

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/PlanningEntity.java
```diff
@@ -26,8 +26,8 @@
 import org.optaplanner.core.impl.heuristic.selector.common.decorator.SelectionSorterWeightFactory;
 import org.optaplanner.core.impl.score.director.ScoreDirector;
 
-import static java.lang.annotation.ElementType.*;
-import static java.lang.annotation.RetentionPolicy.*;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
```
ge0ffrey:
```suggestion
import static java.lang.annotation.RetentionPolicy.RUNTIMEggg;
```

PR #631, opened 2019-12-07
Reduce dependence on commons-lang3

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendablebigdecimal/BendableBigDecimalScore.java
```diff
@@ -402,20 +403,15 @@ public boolean equals(Object o) {
...
-        }
-        for (BigDecimal softScore : softScores) {
-            hashCode = (37 * hashCode) + softScore.stripTrailingZeros().hashCode();
-        }
-        return hashCode;
+        int[] scoreHashCodes = Stream.concat(Arrays.stream(hardScores), Arrays.stream(softScores))
+                .map(BigDecimal::stripTrailingZeros)
+                .mapToInt(BigDecimal::hashCode)
+                .toArray();
+        return Objects.hash(initScore, Arrays.hashCode(scoreHashCodes));
```
triceo: In my opinion, this is too much to do in `hashCode()`.
In this particular case, I would argue the trailing zeros could be stripped as they come into the `Score`. Precision loss doesn't matter here, as the end value is always going to be the same.

ge0ffrey: The stripTrailingZeros is a requirement of the equals() method, which leaks into the hashcode method due to the Java equal contract.
IIRC, the equals method required that due to this case:
BendableScore.ZERO (= "0") must equal BendableScore.ZERO.add("1.00").subtract("1.00") (= "0.00") or it causes score corruption.

ge0ffrey: That being said, that stripTrailingZeros does look weird, so the equals() method should really comment that example as a motivation so it doesn't get removed by accident.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendablebigdecimal/BendableBigDecimalScore.java
```diff
@@ -402,20 +403,15 @@ public boolean equals(Object o) {
...
     }
 
     @Override
     public int compareTo(BendableBigDecimalScore other) {
-        // A direct implementation (instead of CompareToBuilder) to avoid dependencies
         validateCompatible(other);
         if (initScore != other.getInitScore()) {
             return initScore < other.getInitScore() ? -1 : 1;
```
ge0ffrey: (Ok to leave to another PR): this can use Integer.compare(int, int)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/anchor/AnchorVariableDemand.java
```diff
@@ -68,7 +68,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return CLASS_NAME_HASH_CODE + sourceVariableDescriptor.hashCode();
+        return Objects.hash(this.getClass().getName(), sourceVariableDescriptor);
```
ge0ffrey: Change breaks equals contract, because equality doesn't take into account subclass names.
To avoid that, use Objects.hash(AnchorVariableDemand.class.getName(), ...)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/custom/CustomShadowVariableDemand.java
```diff
@@ -61,7 +61,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return CLASS_NAME_HASH_CODE + targetShadowVariableDescriptor.hashCode();
+        return Objects.hash(this.getClass().getName(), targetShadowVariableDescriptor);
```
ge0ffrey: same

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/CollectionInverseVariableDemand.java
```diff
@@ -67,7 +67,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return CLASS_NAME_HASH_CODE + sourceVariableDescriptor.hashCode();
+        return Objects.hash(this.getClass().getName(), sourceVariableDescriptor);
```
ge0ffrey: same

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/SingletonInverseVariableDemand.java
```diff
@@ -64,7 +64,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return CLASS_NAME_HASH_CODE + sourceVariableDescriptor.hashCode();
+        return Objects.hash(this.getClass().getName(), sourceVariableDescriptor);
```
ge0ffrey: same

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/support/VariableListenerNotification.java
```diff
@@ -58,8 +59,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        // A direct implementation (instead of HashCodeBuilder) to avoid dependencies
-        return (((17 * 37) + System.identityHashCode(entity))) * 37 + type.hashCode();
+        return Objects.hash(entity, type);
```
ge0ffrey: This is not a pure refactor: now the entity's hashcode is used instead of the entity's identityHashCode.
*Breaks equals contract* (and this one unlike the ones above this one will cause trouble): the equals method uses entity same semantics, but the hashcode uses entity equals semantics.
For the record, it needs to remain entity same semantics (the more we force our users to implement equals/hashcode the way we want it, the less compatible we are with other Java frameworks)
PS: we probably want a unit test proving that this breaks the equals contract before we fix it - clearly the original code is missing test coverage :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/ChangeMove.java
```diff
@@ -105,28 +103,22 @@ public String getSimpleMoveTypeDescription() {
...
 
     @Override
-    public boolean equals(Object o) {
+    public boolean equals(final Object o) {
         if (this == o) {
             return true;
-        } else if (o instanceof ChangeMove) {
-            ChangeMove<?> other = (ChangeMove) o;
-            return new EqualsBuilder()
-                    .append(entity, other.entity)
```
ge0ffrey: Yea, yea, I know...
This already uses equals() instead of same semantics on entities, but in practice Move equality is only used for Move Tabu Search and nobody uses that. So this hasn't posed anywhere the trouble that VariableListenerNotification would have if you make that equals() semantics.
That being said, move equality should probably use same semantics. Can you create a separate JIRA for that? It will need to be thoroughly tested.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/PillarChangeMove.java
```diff
@@ -125,28 +123,22 @@ public String getSimpleMoveTypeDescription() {
     }
 
     @Override
-    public boolean equals(Object o) {
+    public boolean equals(final Object o) {
```
ge0ffrey: (suggestion, non-blocking) Code style: normally we don't put "final" keywords in method parameters.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/KOptMove.java
```diff
@@ -172,26 +172,23 @@ public String getSimpleMoveTypeDescription() {
...
     }
 
     @Override
     public int hashCode() {
-        return new HashCodeBuilder()
-                .append(entity)
-                .append(values)
-                .toHashCode();
+        int result = Objects.hash(entity);
+        result = 31 * result + Arrays.hashCode(values);
```
ge0ffrey: Any reason why this doesn't use Objects.hash()?

I am ok for changing the hashcode here as long as it desn't violate the equals contract.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/PillarChangeMove.java
```diff
@@ -125,28 +123,22 @@ public String getSimpleMoveTypeDescription() {
...
-        } else if (o instanceof PillarChangeMove) {
-            PillarChangeMove<?> other = (PillarChangeMove) o;
-            return new EqualsBuilder()
-                    .append(variableDescriptor, other.variableDescriptor)
-                    .append(pillar, other.pillar)
-                    .append(toPlanningValue, other.toPlanningValue)
-                    .isEquals();
-        } else {
+        }
+        if (o == null || getClass() != o.getClass()) {
```
ge0ffrey: +1 to doing the getClass check instead of instanceof. Especially with ChainedChangeMove and ChangeMove it might matter some day.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFilterUniNode.java
```diff
@@ -48,9 +49,7 @@ public void addChildNode(BavetAbstractUniNode<A> childNode) {
 
     @Override
     public int hashCode() {
-        // Similar to Object.hash() without autoboxing
-        return 31 * System.identityHashCode(parentNode)
-                + System.identityHashCode(predicate);
+        return Objects.hash(parentNode, predicate);
```
ge0ffrey: IMPORTANT (see above), break equals contract + changes implementation

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetFilterTriNode.java
```diff
@@ -48,9 +49,7 @@ public void addChildNode(BavetAbstractTriNode<A, B, C> childNode) {
 
     @Override
     public int hashCode() {
-        // Similar to Object.hash() without autoboxing
-        return 31 * System.identityHashCode(parentNode)
-                + System.identityHashCode(predicate);
+        return Objects.hash(parentNode, predicate);
```
ge0ffrey: IMPORTANT (see above), break equals contract + changes implementation

ge0ffrey: There a bunch more of these, I won't point them all out.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetFilterBiNode.java
```diff
@@ -48,9 +49,7 @@ public void addChildNode(BavetAbstractBiNode<A, B> childNode) {
 
     @Override
     public int hashCode() {
-        // Similar to Object.hash() without autoboxing
-        return 31 * System.identityHashCode(parentNode)
-                + System.identityHashCode(predicate);
+        return Objects.hash(parentNode, predicate);
```
ge0ffrey: IMPORTANT (see above), break equals contract + changes implementation

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendablebigdecimal/BendableBigDecimalScore.java
triceo: @ge0ffrey My point is... if we strip all trailing zeros always, we don't need to worry about it in `equals()`.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/MachinePeriodPart.java
```diff
@@ -97,40 +104,30 @@ public int getPeriod() {
...
-                    .isEquals();
-        } else {
+        }
+        if (o == null || getClass() != o.getClass()) {
             return false;
         }
+        final MachinePeriodPart that = (MachinePeriodPart) o;
+        return period == that.period &&
+                active == that.active &&
+                Objects.equals(machine, that.machine) &&
```
ge0ffrey: Code style: please keep original order (machine, period, active), which is the order of the field declarations.

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/MachinePeriodPart.java
```diff
@@ -97,40 +104,30 @@ public int getPeriod() {
...
-                    .append(period, other.period)
-                    .append(active, other.active)
-                    .append(resourceAvailableList, other.resourceAvailableList)
-                    .isEquals();
-        } else {
+        }
+        if (o == null || getClass() != o.getClass()) {
             return false;
         }
+        final MachinePeriodPart that = (MachinePeriodPart) o;
```
ge0ffrey: Code nitpick: This renames the variable "other" to "that".
All other equals() code in optaplanner uses "o" for Object and "other" for casted instance.

optaplanner-examples/src/main/java/org/optaplanner/examples/travelingtournament/domain/TravelingTournament.java
```diff
@@ -114,12 +114,8 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        HashCodeBuilder hashCodeBuilder = new HashCodeBuilder();
-        for (Match match : matchList) {
-            // Notice: we don't use hashCode()
-            hashCodeBuilder.append(match.solutionHashCode());
-        }
-        return hashCodeBuilder.toHashCode();
+        int[] hashes = matchList.stream().mapToInt(Match::solutionHashCode).toArray();
```
ge0ffrey: "Notice: we don't use hashCode()" - why not? 
Ow... On Match.java
```
    /**
     * The normal methods {@link #equals(Object)} and {@link #hashCode()} cannot be used
     * because the rule engine already requires them (for performance in their original state).
     * @see #solutionHashCode()
     */
    public boolean solutionEquals(Object o) {
```
Ok, that didn't age well... (the code was written in 2007 or even 2006, because it's part of the first commit that came from SourceForge)
That javadocs is probably not true any more. Drools doesn't use Java equality since Drools 4 or so.
I've created a jira: https://issues.redhat.com/browse/PLANNER-1746

triceo:
@ge0ffrey Please re-review, the changes were significant.
Also, if merging, please squash.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/simple/SimpleScore.java
```diff
@@ -169,15 +170,11 @@ public boolean equals(Object o) {
...
-                + initScore) * 37
-                + score;
+        return Objects.hash(initScore, score);
     }
 
     @Override
     public int compareTo(SimpleScore other) {
-        // A direct implementation (instead of CompareToBuilder) to avoid dependencies
         if (initScore != other.getInitScore()) {
             return initScore < other.getInitScore() ? -1 : 1;
```
ge0ffrey: Nitpick: we can use return Integer.compare(initScore, other.getInitScore()); here too

optaplanner-examples/src/main/java/org/optaplanner/examples/travelingtournament/domain/Match.java
```diff
@@ -69,17 +69,15 @@ public void setDay(Day day) {
...
-                    .append(homeTeam, other.homeTeam)
-                    .append(awayTeam, other.awayTeam)
-                    .append(day, other.day)
-                    .isEquals();
-        } else {
+        }
+        if (o == null || getClass() != o.getClass()) {
             return false;
         }
+        final Match match = (Match) o;
```
ge0ffrey: Nitpick: variable name other?

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/solver/drools/IdleCost.java
```diff
@@ -58,25 +57,19 @@ public long getCost() {
...
-                    .append(machine, other.machine)
-                    .append(activePeriodAfterIdle, other.activePeriodAfterIdle)
-                    .append(cost, other.cost)
-                    .isEquals();
-        } else {
+        }
+        if (o == null || getClass() != o.getClass()) {
             return false;
         }
+        final IdleCost idleCost = (IdleCost) o;
```
ge0ffrey: nitpick: other variable name

PR #635, opened 2019-12-11
PLANNER-1749: Replace non-free fonts used in SVGs with Liberation

yurloc:
- Replaced Arial and Courier New with Liberation Sans and Liberation Mono respectively using `find` and `sed`.
- Removed a few empty text elements referencing other exotic fonts.
- Reexported all SVGs to PNGs using `find` and `inkscape` command.

#### Commands used
```bash
find -name *.svg -not -path '*/target/*' -exec sed \
-e "s/font-family:Arial/font-family:'Liberation Sans'/g" \
-e "s/-inkscape-font-specification:Arial/-inkscape-font-specification:'Liberation Sans, Normal'/g" -i {} \;
```
and variants of that to replace fonts. Then:
```bash
find -name *.svg -not -path '*/target/*' -execdir bash -c 'PNG=${0/svg/png}; inkscape -e $PNG $0' {} \;
```
to reexport SVGs to PNGs using Inkscape without GUI (`-z`).

I have changed a few pages in the beginning of the documentation and it looks good.

EDIT: Regenerated only modified SVGs with:
```bash
for svg in $(git diff master --name-only); do inkscape -z -e ${svg/svg/png} $svg; done
```


ge0ffrey:
Great that you're making this happen.

I noticed some images are changing size:
image

yurloc:
Good catch, thanks! I'll fix that.

ge0ffrey:
I recommend reverting all png images who's svg hasn't changed (such as those swingui icon images)

ge0ffrey:
Once this is merged, I wonder if we can do the dpi change in the same way. Let me check

PR #638, opened 2019-12-12
Replace serif fonts with Liberation Sans

yurloc:
Notice that most of the changes are on empty text elements - should we remove them?


ge0ffrey:
This features far more SVG's than PNG files?

ge0ffrey:
Good catch ! we indeed shouldn't use serif, only sans-serif

ge0ffrey:
This file switchTree.png changed size

ge0ffrey:
integrationOverview.svg/png features some source code text. Some other images do too (like ConstraintStreams)
Source code text can be in serif, but it can also be in monospace. It can't be in sans-serif.
Especially if it's 3+ lines, source code in monospace is much better.

yurloc:
Comments solved.
1. Reverted SVGs containing empty text elements. Changing font on en empty text element didn't affect the matching PNG when exported. I'll remove these empty text elements in a separate PR.
2. Excluded two images that are used in Swing UI, where it either actually makes sense to use a serif font or it simply doesn't matter.

If you go through PNG changes in this PR, there are three types of changes:
1. Ellipses (...) on algorithm diagrams (difficult to spot).
2. Circled numbers designating procedure steps (medium difficult to spot).
3. Actual text on score comparison diagrams (easy to spot).

ge0ffrey:
Thanks Jiri, great work of handling loose ends :)

PR #646, opened 2019-12-23
SolverManager PECS (Producer Extends, Consumer Super)

ge0ffrey:
As Jiri pointed out. This solves this problem:
```
// Cat extends Animal
public Cat read(Long id) {...}
public void save(Object o) {...}

SolverManager<Long, Animal> solverManager;
solverManager.solve(1L, this::read, this::save); // Now compiles
```

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
@@ -38,8 +38,8 @@ public final class DefaultSolverJob<Solution_, ProblemId_> implements SolverJob<
     private final DefaultSolverManager<Solution_, ProblemId_> solverManager;
     private final ProblemId_ problemId;
     private final Solver<Solution_> solver;
-    private final Function<ProblemId_, Solution_> problemFinder;
-    private final Consumer<Solution_> finalBestSolutionConsumer;
+    private final Function<ProblemId_, ? extends Solution_> problemFinder;
+    private final Consumer<? super Solution_> finalBestSolutionConsumer;
     private final BiConsumer<ProblemId_, Throwable> exceptionHandler;
 
     private volatile SolverStatus solverStatus;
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -38,17 +38,17 @@
     private final DefaultSolverManager<Solution_, ProblemId_> solverManager;
     private final ProblemId_ problemId;
     private final Solver<Solution_> solver;
-    private final Function<ProblemId_, Solution_> problemFinder;
-    private final Consumer<Solution_> finalBestSolutionConsumer;
+    private final Function<ProblemId_, ? extends Solution_> problemFinder;
```
yurloc: Can be `? super ProblemId_`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -38,17 +38,17 @@
...
 
     private volatile SolverStatus solverStatus;
     private Future<Solution_> future;
 
     public DefaultSolverJob(DefaultSolverManager<Solution_, ProblemId_> solverManager, ProblemId_ problemId,
             Solver<Solution_> solver,
-            Function<ProblemId_, Solution_> problemFinder,
-            Consumer<Solution_> finalBestSolutionConsumer,
+            Function<ProblemId_, ? extends Solution_> problemFinder,
+            Consumer<? super Solution_> finalBestSolutionConsumer,
```
yurloc: The next line can be `BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverManager.java
```diff
@@ -76,24 +76,24 @@ private void validateSolverFactory() {
 
     @Override
     public SolverJob<Solution_, ProblemId_> solve(ProblemId_ problemId,
-            Function<ProblemId_, Solution_> problemFinder,
-            Consumer<Solution_> finalBestSolutionConsumer,
+            Function<ProblemId_, ? extends Solution_> problemFinder,
```
yurloc: Can be `? super ProblemId_`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
ge0ffrey: Really good catches, Jiri. Those methods should indeed support accepting a Function<Object, ...> as a problemFinder for a SolverManager<Long, ...>.

PR #649, opened 2020-01-07
PLANNER-1684 First BiGroup operation

triceo:
Also includes:

- A lot of refactoring, hopefully leading to a code that is easier to understand.
- A ban on repeated groupBy() operations.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
@@ -195,12 +195,10 @@ public interface BiConstraintStream<A, B> extends ConstraintStream {
     }
 
     // ************************************************************************
-    // Group by (TODO implement these)
+    // Group by
     // ************************************************************************
 
-    /*
-    <GroupKey_> UniConstraintStream<GroupKey_> groupBy(
-            BiFunction<A, B, GroupKey_> groupKeyMapping);
+    <GroupKey_> UniConstraintStream<GroupKey_> groupBy(BiFunction<A, B, GroupKey_> groupKeyMapping);
 
     <GroupKey_, ResultContainer_, Result_> BiConstraintStream<GroupKey_, Result_> groupBy(
             BiFunction<A, B, GroupKey_> groupKeyMapping,
@@ -210,6 +208,8 @@ public interface BiConstraintStream<A, B> extends ConstraintStream {
             BiFunction<A, B, GroupKeyA_> groupKeyAMapping,
             BiFunction<A, B, GroupKeyB_> groupKeyBMapping);
 
+    /*
+    // TODO implement this
     <GroupKeyA_, GroupKeyB_, ResultContainer_, Result_>
     TriConstraintStream<GroupKeyA_, GroupKeyB_, Result_> groupBy(
             BiFunction<A, B, GroupKeyA_> groupKeyAMapping,
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
@@ -200,13 +200,19 @@ public interface BiConstraintStream<A, B> extends ConstraintStream {
 
     <GroupKey_> UniConstraintStream<GroupKey_> groupBy(BiFunction<A, B, GroupKey_> groupKeyMapping);
 
+    /*
+    // TODO implement this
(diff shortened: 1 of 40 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -195,21 +195,27 @@
...
 
     // ************************************************************************
-    // Group by (TODO implement these)
+    // Group by
     // ************************************************************************
 
-    /*
-    <GroupKey_> UniConstraintStream<GroupKey_> groupBy(
-            BiFunction<A, B, GroupKey_> groupKeyMapping);
+    <GroupKey_> UniConstraintStream<GroupKey_> groupBy(BiFunction<A, B, GroupKey_> groupKeyMapping);
```
ge0ffrey: Shall we immediately add the javadoc if we uncomment it?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/uni/DroolsValuePair.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return hashCode;
+    }
+
+    @Override
+    public String toString() {
+        return "Pair[" + key + ", " + value + "]";
```
ge0ffrey: In toString() and log messages,
we use "[" to denote lists, collections, arrays, etc.
We use "(" to denote fields.
So let's change this into: return "Pair(" + key + ", " + value + ")";

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
triceo: Good point, will add.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AssortedConstraintStreamTest.java
```diff
@@ -0,0 +1,183 @@
...
+import org.junit.Test;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.score.director.InnerScoreDirector;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishEntity;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishEntityGroup;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishSolution;
+
+import static org.optaplanner.core.api.score.stream.Joiners.equal;
+
+public class AssortedConstraintStreamTest extends AbstractConstraintStreamTest {
```
ge0ffrey: Assorted?
I don't like "Assorted"Test. It's like Drools's notorious "Misc"Test (and Misc2Test and Misc3Test). It's like a "lost+find" directory - tests that haven't been sorted yet.

So far, the "where can I find it" logic in the tests has been:
if the final stream is a Uni stream, in the UniTest, if it's a Bi then in the BiTest, etc.
But if you think it makes more sense to put some of these grouping tests into a (Advanced)GroupingConstraintStreamTest, etc to avoid bloating the BiTest etc, I am ok with that.

Veto on naming a test "Assorted" or "Misc" Test. It's too broad.

triceo: Yeah, this was one of the things I was going to "clean up later before I submit the PR." :-)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -197,8 +197,8 @@
             UniConstraintCollector<A, ResultContainer_, Result_> collector);
 
     /**
-     * Convert the {@link UniConstraintStream} to a different {@link UniConstraintStream}, containing of the set of
-     * tuples resulting from applying the group key mapping function on all tuples of the original stream.
+     * Convert the {@link UniConstraintStream} to a different {@link UniConstraintStream}, containing the set of tuples
+     * resulting from applying the group key mapping function on all tuples of the original stream.
      * Neither tuple of the new stream will {@link Objects#equals(Object, Object)} any other.
```
ge0ffrey: I am not sure if these javadocs can't be made more "simple & clear". Also notice the big out-of-sync between the javadoc of groupBy(Function) and groupBy(Function, Collector)

ge0ffrey: We should improve these (if possible), before copy-pasting it into the other streams cardinalities.

PR #653, opened 2020-01-09
PLANNER-1652 Migrate ExhaustiveSearchTest

michaltomco:
I've constructed a custom domain for this test, which tries to find the best solution using the least amount of value duplicates. The data in ExhaustiveSearchTest has been checked manually and matches the expected results of different configurations tested.


optaplanner-core/src/test/java/org/optaplanner/core/impl/exhaustivesearch/ExhaustiveSearchTest.java
```diff
@@ -0,0 +1,363 @@
...
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableEntity;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableSolution;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableValue;
+import org.optaplanner.core.impl.testdata.score.director.TestdataComparableDifferentValuesCalculator;
+import org.optaplanner.core.impl.testdata.util.PlannerTestUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+public class ExhaustiveSearchTest {
```
triceo: This test deserves a Javadoc. Something explaining how the test was put together, what the magic values mean (`13--`, ...) and that they are, indeed, painstakingly manually crafter.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/comparable/TestdataCodeComparator.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.apache.commons.lang3.builder.CompareToBuilder;
+import org.optaplanner.core.impl.testdata.domain.TestdataObject;
+
+public class TestdataCodeComparator implements Comparator<TestdataObject> {
+
+    public int compare(TestdataObject a, TestdataObject b) {
+        return new CompareToBuilder()
+                .append(a.getCode(), b.getCode())
+                .toComparison();
```
ge0ffrey: Please Comparator.comparing().thenComparing() instead and avoid having to create the class to begin with.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/comparable/TestdataComparableEntity.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+import org.optaplanner.core.impl.testdata.domain.TestdataObject;
+
+@PlanningEntity(difficultyComparatorClass = TestdataCodeComparator.class)
+public class TestdataComparableEntity extends TestdataObject {
```
ge0ffrey: It's named ComparableEntity but doesn't implement Comparable?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/event/TestdataStepValueListener.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListenerAdapter;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
+import org.optaplanner.core.impl.solver.scope.DefaultSolverScope;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableEntity;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableSolution;
+
+public class TestdataStepValueListener extends PhaseLifecycleListenerAdapter<TestdataComparableSolution> {
+
+    private final List<String> dataConfigurations = new ArrayList<>();
```
ge0ffrey: What's a dataConfiguration? This names doesn't seem to fit the payload.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/event/TestdataStepValueListener.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListenerAdapter;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
+import org.optaplanner.core.impl.solver.scope.DefaultSolverScope;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableEntity;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableSolution;
+
+public class TestdataStepValueListener extends PhaseLifecycleListenerAdapter<TestdataComparableSolution> {
```
ge0ffrey: Looks like this is a solution state recorder.
But while listening to steps, you can also have a step (= winning move) recorder.
That only has the delta's.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/comparable/TestdataCodeComparator.java
michaltomco: I've created the comparator in order to have a difficulty comparator class which I can put into the @PlanningEntity(difficultyComparatorClass = TestdataCodeComparator.class) annotation above the comparable entity. Having a comparable entities by difficulty is mandatory when using EntitySorterManner.DECREASING_DIFFICULTY in the ExhaustiveSearchPhaseConfig, that's why it's there. Is there a simpler way how to achieve this without needing to create a Comparator class that you explicitly put into the annotation?

rsynek: @ge0ffrey I agree with @michaltomco regarding the necessity of this `Comparator` class. The usage of the `CompareToBuilder` can be replaced by standard `java.util.Comparator`, though. See e.g. https://github.com/kiegroup/optaplanner/blob/master/optaplanner-examples/src/main/java/org/optaplanner/examples/cloudbalancing/optional/domain/CloudProcessDifficultyComparator.java#L24

optaplanner-core/src/test/java/org/optaplanner/core/impl/exhaustivesearch/BlackBoxExhaustiveSearchPhaseTest.java
```diff
@@ -0,0 +1,433 @@
...
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableEntity;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataComparableSolution;
+import org.optaplanner.core.impl.testdata.phase.event.TestdataSolutionSateRecorder;
+import org.optaplanner.core.impl.testdata.score.director.TestdataComparableDifferentValuesCalculator;
+import org.optaplanner.core.impl.testdata.util.PlannerTestUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * The tests run through all available configuration combinations related to ExhaustiveSearch and compares the results
```
rsynek: test, not tests - there is a single one, as far as I can see

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/comparable/TestdataComparableEntity.java
michaltomco: Yes, it does sound a bit confusing. I've found it fitting for the use-case at that time, but it does break the Comparable naming convention. I've changed the entity to TestdataEntityWithDifficultyComparator and the solution to TestdataSolutionWithDifficultyComparatorEntity. I've wanted to fit in "Class" and make them into TestdataEntityWithDifficultyComparatorClass and TestdataSolutionWithDifficultyComparatorClassEntity, but decided not to, because I found them to be very lenghty even withouth it.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/event/TestdataStepValueListener.java
michaltomco: Changed it to workingSolutions. Does it fit the payload properly?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/event/TestdataStepValueListener.java
michaltomco: The way it is implemented now seems to me being similarly complex as having the winning move recorder with deltas, while being a bit more comprehensible to a human and achieving the same outcome than having the deltas instead, or am I missing something?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/comparable/TestdataSolutionWithDifficultyComparatorEntity.java
```diff
@@ -0,0 +1,65 @@
...
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.solution.drools.ProblemFactCollectionProperty;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.testdata.domain.TestdataObject;
+import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+
+@PlanningSolution
+public class TestdataSolutionWithDifficultyComparatorEntity extends TestdataObject {
```
ge0ffrey: This naming is very confusing: it ends with "Entity", but it's a planning solution.
This naming should be consistent with it's entity.

Proposal A) for names:
- the solution: TestdataDifficultyComparingSolution
- the entity: TestdataDifficultyComparingEntity

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/TestdataComparableDifferentValuesCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataEntityWithDifficultyComparator;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataSolutionWithDifficultyComparatorEntity;
+
+public class TestdataComparableDifferentValuesCalculator implements EasyScoreCalculator<TestdataSolutionWithDifficultyComparatorEntity> {
+
+    @Override
+    public SimpleScore calculateScore(TestdataSolutionWithDifficultyComparatorEntity solution) {
+        int score = 0;
+        Map<TestdataValue, Integer> alreadyUsedValues = new HashMap<>();
```
ge0ffrey: Why is this a Map when a Set suffices?
The value side is never read. There is never a "get(key)", only a "contains(key").

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/TestdataComparableDifferentValuesCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.impl.score.director.easy.EasyScoreCalculator;
+import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataEntityWithDifficultyComparator;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataSolutionWithDifficultyComparatorEntity;
+
+public class TestdataComparableDifferentValuesCalculator implements EasyScoreCalculator<TestdataSolutionWithDifficultyComparatorEntity> {
```
ge0ffrey: Does not belong in package "testdata", that package is reserved for domain objects.

Put it near the test, either as a static class inside the test, or next to it.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/phase/event/TestdataSolutionSateRecorder.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListenerAdapter;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
+import org.optaplanner.core.impl.solver.scope.DefaultSolverScope;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataEntityWithDifficultyComparator;
+import org.optaplanner.core.impl.testdata.domain.comparable.TestdataSolutionWithDifficultyComparatorEntity;
+
+public class TestdataSolutionSateRecorder extends PhaseLifecycleListenerAdapter<TestdataSolutionWithDifficultyComparatorEntity> {
```
ge0ffrey: This doesn't belong in the "testdata" package because it's not a domain object.
Put it near the test (see later comment).

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/comparable/TestdataEntityWithDifficultyComparator.java
```diff
@@ -0,0 +1,55 @@
...
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+import org.optaplanner.core.impl.testdata.domain.TestdataObject;
+import org.optaplanner.core.impl.testdata.domain.TestdataValue;
+
+@PlanningEntity(difficultyComparatorClass = TestdataCodeComparator.class)
+public class TestdataEntityWithDifficultyComparator extends TestdataObject {
```
ge0ffrey: This naming is confusing. It ends with "Comparator", so is it a Comparator?
No, it's a entity. See other comment for rename suggestion.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/score/director/TestdataComparableDifferentValuesCalculator.java
michaltomco: This one was a leftover from when I was trying to make a more complex Calculator in order to make the solver do more different steps, but after trying out a simple Set-based one and not seeing any benefit in having the previous one, I've changed it to Set as you've recommended, thank you.

optaplanner-core/src/test/java/org/optaplanner/core/impl/exhaustivesearch/BlackBoxExhaustiveSearchPhaseTest.java
```diff
@@ -0,0 +1,490 @@
...
+
+    /**
+     * This class calculates the score of a solution by penalizing repeated value occurrences held by entities.
+     */
+    public static class TestdataComparableDifferentValuesCalculator implements EasyScoreCalculator<TestdataDifficultyComparingSolution> {
+
+        @Override
+        public SimpleScore calculateScore(TestdataDifficultyComparingSolution solution) {
+            int score = 0;
+            Set<TestdataValue> alreadyUsedValues = new HashSet<>();
```
ge0ffrey: Rename to valueSet (or alreadyUsedValueSet)

PR #659, opened 2020-01-20
PLANNER-1787 Constraint match justification list must be modifiable

triceo:


ge0ffrey:
I believe there used to be a policy that the justification list is the order of the matches of DRL

For example, presume:
when
   ShiftAssignment($e : employee)
   ShiftAssignment(employee != $e)
then
Now we insert 2 ShiftAssignments. SA-1 with employee "Ann" and SA-2 with employee "Beth". what are the matches?
Well, in the past, the matches where:
- match SA-1, SA-2
- match SA-2, SA-1
but with this policy of sorting them by id or based on comparable, it becomes:
- match SA-1, SA-2
 - match SA-1, SA-2
That's information loss.

PR #660, opened 2020-01-22
PLANNER-1615 ifExists() for UniStream

triceo:
This is a PoC of the ifExists() implementation. If accepted, I will add ifNotExists() similarly and move on to bi/tri/quad.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -218,6 +218,88 @@
...
+     * <p>
+     * This method causes <i>Unchecked generics array creation for varargs parameter</i> warnings,
+     * but we can't fix it with a {@link SafeVarargs} annotation because it's an interface method.
+     * Therefore, there are overloaded methods with up to 4 {@link BiJoiner} parameters.
+     * @param otherClass never null
+     * @param joiners never null
+     * @param <B> the type of the second matched fact
+     * @return a stream that matches every A where B exists for which all the {@link BiJoiner}s are true
+     */
+    <B> UniConstraintStream<A> ifExists(Class<B> otherClass, BiJoiner<A, B>... joiners);
```
Christopher-Chianelli: Suggested solution: add a default `ifExists(Class<B> otherClass, BiJoiner<A,B>... joiners); that calls`ifExists(Class<B> otherClass, List<BiJoiner> joiners)`

```
default <B> UniConstraintStream<A> ifExists(Class<B> otherClass, BiJoiner<A, B>... joiners) {
    ifExists(otherClass, Arrays.asList(joiners));
}

<B>  UniConstraintStream<A> ifExists(Class<B> otherClass, List<BiJoiner<A, B>> joiners);

```

triceo: Chris, I don't have anything against this change, but if we do it, we need to do the same for joins and perhaps in couple other places. All in all, that's a large amount of work, and scope creep in this context.

Would you mind filing a JIRA for this? You could even tackle it yourself, if you want to feel like diving into the CS code.

Christopher-Chianelli: I will first try it out in an example project to make sure my suggested change actually works as intended; if so, I will file a JIRA.

Christopher-Chianelli: Cannot make an interface method final, which is a blocker for `@SafeVarargs`. If you want to use this suggestion, we have two options:
1. Turn the interfaces into abstract classes (which allow us to use final).
2. Create a static wrapper.
I think both of those options are bad.

If you allow `@SuppressWarnings`, there a third option:
```
public interface TestInterface<A> {
    
    default <B> List<A> myMethodVarargs(Class<B> otherClass, @SuppressWarnings("unchecked") A... data) {
        return myMethodList(otherClass, Arrays.asList(data));
    }
    
    <B> List<A> myMethodList(Class<B> otherClass, List<A> data);
}
```

triceo: Actually, thinking about this again, could you perhaps explain why you would need a method like that?

Personally I consider this:

   ifExists(Something.class, joiner1, joiner2, joiner3, joiner4)

To be much better than this:

   ifExists(Something.class, Arrays.asList(joiner1, joiner2, joiner3, joiner4);

But perhaps there is some aspect that I'm missing.

Christopher-Chianelli: It to make implementing the interface not inherit the varargs warnings (and working with Lists is easier than working with arrays). The varargs option is still available (under the same name; above was from testing trying to find a way to remove the warning). `@SafeVarargs` might work on implementing classes, as long as the methods are marked as `final`.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -41,6 +44,10 @@
     // BiJoiner
     // ************************************************************************
 
+    public static <A> BiJoiner<A, A> equal() {
```
ge0ffrey: This is cool!

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -50,6 +57,19 @@
...
+        return notEqual(Function.identity());
+    }
+
+    public static <A, Property_> BiJoiner<A, A> notEqual(Function<A, Property_> mapping) {
+        return notEqual(mapping, mapping);
+    }
+
+    public static <A, B, Property_> BiJoiner<A, B> notEqual(Function<A, Property_> leftMapping,
+            Function <B, Property_> rightMapping) {
+        return new SingleBiJoiner<>(leftMapping, JoinerType.NOT_EQUAL, rightMapping);
```
ge0ffrey: Will notEqual be indexed?
If it's not indexed, there must be fail-fast if any indexed joiner is behind it (so only notEquals and filtering() can come after it).

ge0ffrey: All in all, the notEquals raises a red flag of "when in doubt, leave it out" for me, especially due to that last question, until we discuss it properly and/or know more about how users use ConstraintStreams. Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -89,6 +109,18 @@
         return new SingleBiJoiner<>(leftMapping, JoinerType.GREATER_THAN_OR_EQUAL, rightMapping);
     }
 
+    /**
+     * Applies a filter to the joined tuple, with the semantics of {@link BiConstraintStream#filter(BiPredicate)}.
```
ge0ffrey: Similar to the javadoc of that filter() method, this javadoc should mention the lack of indexing.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -89,6 +109,18 @@
...
+     *
+     * @param filter filter to apply
```
ge0ffrey: always start with "never null" or "sometimes null". (in this case never null)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -89,6 +109,18 @@
...
+     * @param <A> type of the first fact in the tuple
+     * @param <B> type of the second fact in the tuple
+     * @return the joiner
```
ge0ffrey: never null or sometimes null?

ge0ffrey: I'd not even mention "the joiner" as the return type already signals that information.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
triceo: In Drools, NOT_EQUAL is an implemented index type, so CS-D behaves the same for EQUAL as for NOT_EQUAL.

There is no Bavet support for NOT_EQUAL and it will throw an exception if you try to use it.

triceo: There is no way to leave this out, as otherwise I can't have `ifExistsOther()`. That requires `notEqual()`. (Alternatively, a filter with poor performance.)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
triceo: In fact, the reason why I reference that other Javadoc is to avoid having to say this everywhere. That Javadoc says it, and this Javadoc says "same semantics".

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
      * @return a stream that matches every combination of A and B for which all the {@link BiJoiner joiners} are true
      */
     default <B> BiConstraintStream<A, B> join(Class<B> otherClass, BiJoiner<A, B>... joiners) {
-        return join(otherClass, AbstractBiJoiner.merge(joiners));
```
ge0ffrey: There is a `join(Class<B>, joiners)` and a `join(Uni<B>, joiners)`.
Do both support filtering() joiners now? Isn't this new code below here duplicated between them?

It feels to me like we want to extract these new lines below into a separate non-public method if possible.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+            if (joiner instanceof FilteringBiJoiner) {
+                if (!hasAFilter) { // From now on, we only allow filtering joiners.
+                    indexOfFirstFilter = i;
+                }
+            } else if (hasAFilter) {
+                throw new IllegalStateException("Indexing joiner (" + joiner + ") must not follow a filtering joiner ("
+                        + joiners[indexOfFirstFilter] + ").");
+            }
+        }
+        if (indexOfFirstFilter < 0) { // Only found indexing joiners.
```
ge0ffrey: "Only found indexing joiners." is not a sentence, don't end with "."

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
triceo: This is an interface and Java 8, so there are no non-public methods.
Yes, all the `join()` methods support `filtering()`. In fact, `filtering()` is converted to `join().filter()`, so the JoinStream never even gets to see that particular joiner.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+            BiJoiner<A, B> joiner = joiners[i];
+            boolean hasAFilter = indexOfFirstFilter >= 0;
+            if (joiner instanceof FilteringBiJoiner) {
+                if (!hasAFilter) { // From now on, we only allow filtering joiners.
+                    indexOfFirstFilter = i;
+                }
+            } else if (hasAFilter) {
+                throw new IllegalStateException("Indexing joiner (" + joiner + ") must not follow a filtering joiner ("
+                        + joiners[indexOfFirstFilter] + ").");
+            }
```
ge0ffrey: For readability I'd turn these if's order and avoid the boolean hasAFilter:

```
if (indexOfFirstFilter >= 0) {
      if (joiner instanceof FilteringBiJoiner) {
        throw ...
      }
} else if (joiner instanceof FilteringBiJoiner) {
    indexOfFirstFilter = i;
}
```

Most of OptaPlanner's code has the style to first check the fail fasts, before doing the code. I've found this is more readable, because when reading you can skip the fail-fasts, reducing the lines to grok (fail-fast's are usually easy to recognize based on their "throw" keyword).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+        }
+        if (indexOfFirstFilter < 0) { // Only found indexing joiners.
+            return join(otherClass, AbstractBiJoiner.merge(joiners));
+        }
+        // Assemble the join stream.
+        BiConstraintStream<A, B> joined = indexOfFirstFilter == 0 ?
+                join(otherClass) :
+                join(otherClass, Arrays.copyOf(joiners, indexOfFirstFilter));
+        // We merge all filters into one, so that we don't pay the penalty for lack of indexing more than once.
+        BiPredicate<A, B> resultingFilter = (a, b) -> true;
```
ge0ffrey: Don't create this "always true" predicate:

- If there is no filtering, don't call filter().
- If there is only 1 filter, let's say `Foo::isReady`, don't create a wrapper predicate of an always true predicate combined with foo.isReady(), just use `Foo::isReady``directly.
- Same reasoning for 2 filters (don't make it 3)
- ...

ge0ffrey: Motivation:
- performance (especially with the extra filter() call for those that don't use it, it currently will cause non-trivial perf degredation in Bavet that creates tuples for filters)
- node sharing (same lambda recognition)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
rsynek: I partially agree - with the order, not with avoiding the `hasAFilter` variable, which in my opinion rather adds to readability.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+    }
+
+    /**
+     * Create a new {@link BiConstraintStream} for every combination of A and a different A for which the
+     * {@link BiJoiner} is true (for the properties it extracts from both facts).
+     * <p>
+     * Important: This is faster and more scalable than a {@link #joinOther(UniConstraintStream) joinOther}
+     * followed by a {@link BiConstraintStream#filter(BiPredicate) filter},
+     * because it applies hashing and/or indexing on the properties,
+     * so it doesn't create nor checks every combination of A and B.
```
ge0ffrey: join(Class) and join(Class, joiner) are very commonly used, so they don't use "As defined by" in the javadocs. They are an exception, because using "as defined by" reduces maintenance.

I'd argue that "joinOther(Class)" can use "As defined by join(Class)" and then something like "With the additional joiner than they are not equal" (but better written).
Same for "joinOther(Class, Joiner)" => "As defined by join(Class, Joiner) with the additional joiner than they are not equal"
But indeed for joinOther(Class, Joiner, Joiner) and more, I 'd keep using "As defined by joinOther(Class, Joiner, Joiner)" (so no need of "with the additional ..." because it references joinOther, not join).

Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+     * @param otherStream never null
+     * @param joiner never null
+     * @return a stream that matches every combination of A and a different A for which the {@link BiJoiner} is true
+     */
+    default BiConstraintStream<A, A> joinOther(UniConstraintStream<A> otherStream, BiJoiner<A, A> joiner) {
+        if (joiner instanceof NoneBiJoiner) {
+            return join(otherStream, Joiners.notEqual());
+        } else {
+            return join(otherStream, AbstractBiJoiner.merge(Joiners.notEqual(), joiner));
```
ge0ffrey: If nonEquals is indexed, does this merge make sure that if we do:

`joinOther(Foo.class, equal(getType), filtering(a, b -> a.getAge * b.getAge == 42))`

it is syntactic sugar for:

`join(Foo.class, equal(getType), notEqual(), filtering(a, b -> a.getAge * b.getAge == 42))`

So the first joiner remained the first joiner, but the second joiner became third joiner, because it's the first non-indexed joiner.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+     * Therefore, there are overloaded methods with up to 4 {@link BiJoiner} parameters.
+     * @param otherClass never null
+     * @param joiners never null
+     * @return a stream that matches every combination of A and a different A for which all the
+     * {@link BiJoiner joiners} are true
+     */
+    default BiConstraintStream<A, A> joinOther(Class<A> otherClass, BiJoiner<A, A>... joiners) {
+        BiJoiner<A, A> otherness = Joiners.notEqual();
+        BiJoiner[] allJoiners = Stream.concat(Stream.of(otherness), Arrays.stream(joiners))
+                .toArray(BiJoiner[]::new);
```
ge0ffrey: The order of joiners gets changed here:
`joinOther(Foo.class, x, y z)` should be `join(Foo.class, x, y, z, notEquals())`
That's currently not the case.

ge0ffrey: Motivation: The order of the joiners can heavily affect performance, so we must respect the user's order such that they can tweak it.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+    // If Exists
+    // ************************************************************************
+
+    /**
+     * Create a new {@link UniConstraintStream} for every A, if B exists.
+     * @param otherClass never null
+     * @param <B> the type of the second matched fact
+     * @return a stream that matches every A where a B exists
+     */
+    default <B> UniConstraintStream<A> ifExists(Class<B> otherClass) {
```
ge0ffrey: Is this method every going to be useful?
What is the specification? If we have A1, A2, A3, ... they will match (= survive) if in these cases:
- there is a B1
- there is a B1, B2
but not if there not a single B.

Theory: "Nobody's every going to need this method." When in doubt, leave it out. Let the users tell us that they need and we 'll add it. Or let us feel the need for it in an example and let's then add it. But meanwhile, let's leave it out.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+
+    /**
+     * Create a new {@link UniConstraintStream} for every A where B exists for which the {@link BiJoiner} is true
+     * (for the properties it extracts from both facts).
+     * <p>
+     * This method has overloaded methods with multiple {@link BiJoiner} parameters.
+     * @param otherClass never null
+     * @param joiner never null
+     * @param <B> the type of the second matched fact
+     * @return a stream that matches every A where B exists for which the {@link BiJoiner} is true
```
ge0ffrey: never null?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
triceo: Ordering of joiners is never changed. When `notEqual()` is added by the `joinOther()` statement, it as added as *the first* joiner.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
triceo: I strongly disagree.
When using `joinOther()`, your implicit assumption is that you'll be getting only those that are not equal. Therefore, the `notEqual()` joiner *must* come first.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
triceo: I don't mind if we leave it out. I just remember having "meta-objects" in Drools working memory to signify some sort of a state change - do something if the state is X, but don't do it when it isn't.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
triceo: (I agree on the performance point. But the fact that you'll get this joiner with `joinOther` is just assumed - if you don't want this joiner, call `join()` and do whatever you wish.)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
ge0ffrey: I don't want to introduce methods that accepts List or Collection in this API.
Motivation:
- Don't proliferate more combinations of each method (join(List) vs join(vararg) and that for join, ifExists, joinOther, ... and then for Uni, Bi, Tri, ...). Not worth the maintenance burden I'd argue.
-  We're not really accepting an array - we're accepting a vararg. I don't see use cases at this point in time where users would actually use an array. I do see cases where they would provide 7+ parameters, hence the need for a vararg.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
```diff
@@ -90,8 +91,23 @@ public void addChildStream(BavetAbstractUniConstraintStream<Solution_, A> childS
         }
         if (!(joiner instanceof AbstractBiJoiner)) {
             throw new IllegalArgumentException("The joiner class (" + joiner.getClass() + ") is not supported.");
+        } else if (joiner instanceof FilteringBiJoiner) {
+            return join(otherStream)
+                    .filter(((FilteringBiJoiner<A, B>) joiner).getFilter());
```
ge0ffrey: Who's responsible for creaming off the FilteringBiJoiner? Is it UniConstraintStream or the Drools/BavetAbstractUniConstraintStream? currently I see code doing this in both sides.

Can we extra some of this into seperate method?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/FilteringBiJoiner.java
```diff
@@ -0,0 +1,55 @@
...
+
+public final class FilteringBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
+
+    public FilteringBiJoiner(BiPredicate<A, B> filter) {
+        super(filter);
+    }
+
+    @Override
+    public Function<A, Object> getLeftMapping(int joinerId) {
+        throw new UnsupportedOperationException();
```
ge0ffrey: Please add a Exceptions message that starts with "Impossible state:" (which indicates, you should never see this, this is an OptaPlanner bug, this is not your fault). For example "Impossible state: this joiner should never be treated like a normal joiner internally."

ge0ffrey:
How confident are we that the unit test covers different orderings of joiners? I haven't checked the unit tests in details, but are cases like this covered for join (or ifExists)?

- 2 filtering joiners
- an equal, a lessThan and a filtering joiner
- ...Other() with an equal, a lessThan and a filtering joiner // If the notEquals in indexed, it must end up before lessThan, if it's not indexed, it must end up after it
- fail fasts, such as an equal, a filtering and an equal joiner

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -180,7 +190,332 @@
...
+        int indexOfFirstFilter = -1;
+        for (int i = 0; i < joinerCount; i++) {
+            BiJoiner<A, B> joiner = joiners[i];
+            boolean hasAFilter = indexOfFirstFilter >= 0;
+            if (joiner instanceof FilteringBiJoiner) {
+                if (!hasAFilter) { // From now on, we only allow filtering joiners.
+                    indexOfFirstFilter = i;
+                }
+            } else if (hasAFilter) {
+                throw new IllegalStateException("Indexing joiner (" + joiner + ") must not follow a filtering joiner ("
```
rsynek: This exception and the expected order of joiners should be probably described in a javadoc.

This fail-fast is not a super trivial check of inputs - it establishes a contract regarding in which order users are supposed to register joiners of different kinds. There should be a test for it.

triceo:
> How confident are we that the unit test covers different orderings of joiners?

Pretty sure the coverage for that sort of thing is poor. The combinations are endless.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
triceo: We've just agreed during our call that we're removing `joinOther` altogether. :-)

triceo:
Now that we've spoken to Mario, we know that NOT_EQUALS index is ignored by Drools. (Not implemented.) Therefore, we really do need to put the `notEquals()` joiner as much to the end as we can.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
triceo: Unfortunately, this needs to happen here. This method could be called directly (and not through the other `join()` methods), and therefore this joiner could be a `FilteringBiJoiner`. There is no other case where that could happen.

triceo:
Jenkins retest this.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -121,66 +125,268 @@
...
+     * @return never null, a stream that matches every combination of A and B for which the {@link BiJoiner} is true
      */
     default <B> BiConstraintStream<A, B> join(Class<B> otherClass, BiJoiner<A, B> joiner) {
-        return join(getConstraintFactory().from(otherClass), joiner);
+        if (joiner instanceof FilteringBiJoiner) {
+            FilteringBiJoiner<A, B> filteringJoiner = (FilteringBiJoiner<A, B>) joiner;
+            return join(otherClass)
+                    .filter(filteringJoiner.getFilter());
+        } else {
+            return join(getConstraintFactory().from(otherClass), joiner);
```
ge0ffrey: Why do we need to handle the FilteringJoiner here already?
The join(UniStream, BiJoiner) method must also deal with wether or not that BiJoiner is a FilteringBiJoiner?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -121,66 +125,268 @@
...
+        }
     }
 
     /**
-     * As defined by {@link #join(Class, BiJoiner)}.
+     * As defined by {@link #join(Class, BiJoiner...)}.
```
ge0ffrey: This puts the vararg method front-and-center. This has 2 issues:

1)The vararg method suffers from "Unchecked generics array creation for varargs parameter" JDK warnings, which we can only fix by making the method non-abstract (which is obviously impossible). Therefore, I see the method as a necessary evil: not something we want to put front-and-center.
2) This javadoc delegates to the vararg method javadoc, which in turn delegates to the single joiner method. Every extra click through looses users. Yet we do delegates due to reduce maintenance burden - and to make users that already know the base method, be able to "start from there". For varargs that latest approach doesn't work.
3) vararg's javadoc is NOT a subset of this javadocs. For example, that warning doesn't apply here at all.

ge0ffrey: I recommend going back to delegating to the single joiner param, wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
ge0ffrey: Style issue is still here: remove the "." at the end in the comment because it's not a sentance

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
ge0ffrey: Good that the JoinStream never sees that partical joiner.

I am just wondering - if both join(Class) and join(Stream) handle the FilteringJoiner-extraction, do they duplicate the code below? If so, can we de-duplicate it by letting this one delegate to the Stream one?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -121,66 +125,268 @@
...
+
+    /**
+     * Create a new {@link UniConstraintStream} for every A, if another A exists that does not {@link #equals(Object)}
+     * the first. If multiple {@link BiJoiner}s are provided, for performance reasons, the indexing joiners must be
+     * placed before filtering joiners.
+     * @param otherClass never null
+     * @return never null, a stream that matches every A where a different A exists
+     */
+    default UniConstraintStream<A> ifExistsOther(Class<A> otherClass) {
+        return ifExists(otherClass, Joiners.filtering((a, b) -> !a.equals(b)));
```
ge0ffrey: Due to Lambda identity, I'd implement this slight different. Let me explain:

These 2 statements use the same lambda instance:
- from(A.class).join(B.class).filter(Objects::equal)
- from(A.class).join(B.class).filter(Objects::equal)

These 2 statements do NOT use the lambda instance:
- from(C.class).join(D.class).filter((c, d) -> c.equals(d))
- from(C.class).join(D.class).filter((c, d) -> c.equals(d))

You can try this out by printing System.systemHashCode() on those BiPredicates.

Now, why do we care about that? Because of node sharing in Drools AND Bavet.
If its the same lambda instance, node sharing work guaranteed.
If it isn't, Bavet won't see it and Drools has some dirty ASM magic that might or might not recognize they are the same code.

Unfortunately, there is no Objects.notEquals in java.util.Objects.
Apache commons-lang3 does have ObjectUtils.notEqual, presuming we want to use that, I would write:

return ifExists(otherClass, Joiners.filtering(ObjectUtils::notEqual);

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
ge0ffrey: good point. Maybe we need to introduce a AbstractUniConstraintStream to capture code that is common between Drools and Bavet to avoid duplicating that code, without exposing it in the interface? Similar to how AbstractScoreDirector sits between ScoreDirector and DroolsScoreDirector/etc?
We could make an innerJoin() method that the implementations need to implement.

Not sure if this is worth it's weight though?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/FilteringBiJoiner.java
```diff
@@ -0,0 +1,60 @@
...
+import org.optaplanner.core.impl.score.stream.common.JoinerType;
+
+public final class FilteringBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
+
+    public FilteringBiJoiner(BiPredicate<A, B> filter) {
+        super(filter);
+    }
+
+    @Override
+    public Function<A, Object> getLeftMapping(int joinerId) {
```
ge0ffrey: Just noticed (not a problem in this PR). This `joinerId` is not an ID, it's an index, for CombinedBiJoiners. That variable should be renamed to `joinerIndex` or just `index` (I 'd say the latter). Again, doesn't need to happen in this PR.

What the different between an ID and and index?
ID's have gaps, as elements get deleted. Examples of ID's: 0, 1, 2, 4, 7
Indexes don't have gaps, they would naturally change if elements get deleted (ussually they won't): Examples of index: 0, 1, 2, 3, 4, 5

PR #661, opened 2020-01-23
PLANNER-1802 Add early termination test for SolverManager

michaltomco:
There has been a potential bug found: not being able to terminate a job with terminateEarly() and close() having status SOLVING_SCHEDULED

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
@@ -250,7 +250,7 @@ public class SolverManagerTest {
      * status change in a deterministic way while solving jobs on separate threads. The test waits for a status change in
      * a loop and time-outs when it doesn't reach it's expected status.
      */
-    @Ignore("SOLVING_SCHEDULED tasks are not terminated successfully neither by terminateEarly(), nor close(). Will be enabled after bug-fix.")
+    @Ignore("https://issues.redhat.com/browse/PLANNER-1836")
     @Test
     public void terminateEarlyAndClose() throws InterruptedException {
         final SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
@@ -316,19 +316,15 @@ public class SolverManagerTest {
     private void waitForJobStatusChangeOrTimeout(SolverManager<TestdataSolution, Long> solverManager, Long problemId, SolverStatus expectedStatusChange) throws InterruptedException {
         long t = System.currentTimeMillis();
         long end = t + 5000;
-        // Checks every 5 milliseconds for solverJob status change for 5 seconds at maximum, otherwise time-outs and fails.
-        while (true) {
-            t = System.currentTimeMillis();
-
-            if (t < end) {
-                if (solverManager.getSolverStatus(problemId) == expectedStatusChange) {
-                    break;
-                }
-                TimeUnit.MILLISECONDS.sleep(5);
-            } else {
-                fail("Job with id " + problemId + " took too long and timed-out not changing its status from "
-                             + solverManager.getSolverStatus(problemId) + " to " + expectedStatusChange + ".");
+        // Check every 5 milliseconds for solverJob status change for 5 seconds at maximum, otherwise time-out and fail.
+        while (t < end) {
+            if (solverManager.getSolverStatus(problemId) == expectedStatusChange) {
+                return;
             }
+            TimeUnit.MILLISECONDS.sleep(5);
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +245,86 @@ public void skipAhead() throws ExecutionException, InterruptedException {
         assertTrue(eventCount.get() < 4);
     }
 
+    /**
+     * In order to effectively test the terminateEarly() and close() methods there had to be a way how to make the job
+     * status change deterministic while solving jobs on a separate thread. The test waits for a status change in
+     * a loop and time-outs when it doesn't reach it's expected status.
+     */
+    @Test
+    public void terminateEarlyAndClose() throws InterruptedException {
```
rsynek: If this test fails (as it reproduces an issue), let's mark it as ignored with a link to that issue and get this merged. Once the issue is fixed, it will be enabled.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +245,87 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+
+    private void waitForJobStatusChangeOrTimeout(SolverManager<TestdataSolution, Long> solverManager, Long problemId, SolverStatus expectedStatusChange) throws InterruptedException {
+        long t = System.currentTimeMillis();
+        long end = t + 5000;
+        // Check every 5 milliseconds for solverJob status change for 5 seconds at maximum, otherwise time-out and fail.
+        while (t < end) {
+            if (solverManager.getSolverStatus(problemId) == expectedStatusChange) {
+                return;
+            }
+            TimeUnit.MILLISECONDS.sleep(5);
```
ge0ffrey: Above the sleep, can you add these comments?

```
// Quick polling (not a Test Thread Sleep anti-pattern)
// Test is still fast on fast machines and doesn't randomly fail on slow machines.
```

Motivation: the use of Thread.sleep() is a red flag in tests, because it's often used in a catch 22 situation: wait long enough to succeed on the slow machine, but don't wait long so the tests run quickly. This comment makes it clear that this code is doing it correctly.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +245,87 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        assertSame(SolverStatus.SOLVING_SCHEDULED, solverManager.getSolverStatus(secondProblemId));
+
+        solverManager.close();
+        waitForJobStatusChangeOrTimeout(solverManager, firstProblemId, SolverStatus.NOT_SOLVING);
+        waitForJobStatusChangeOrTimeout(solverManager, secondProblemId, SolverStatus.NOT_SOLVING);
+    }
+
+    private void waitForJobStatusChangeOrTimeout(SolverManager<TestdataSolution, Long> solverManager, Long problemId, SolverStatus expectedStatusChange) throws InterruptedException {
+        long t = System.currentTimeMillis();
+        long end = t + 5000;
```
ge0ffrey: 5000 is risky on our slow jenkins machines (or a community user compiling while something else is running on his/her machine).

I recommend just adding a `@Timeout(600_000)` on the test itself and do

```
while (solverManager.getSolverStatus(problemId) == expectedStatusChange) {
   Thread.sleep(100L);
}
```

Calling currentTimeMillis() every 5 milliseconds might not be as cheap as one would expect. And questionable on Windows that has a 17 millisecond clock granularity :)

ge0ffrey: @rsynek @triceo  That 100 milliseconds, I pulled out of thin air. But I presume michal did the same with the 5 milliseconds. We should figure out a good standard for that number for for such "quick polling" patterns used in tests.

 I wouldn't make it less than 20 milliseconds due to clock granularity on common machines. But 100ms is probably too long, now that I think about it. Having 200 tests that do quick polling with 100ms, means they run at least for 20 seconds.

How about we use 20ms as the quick polling standard? Wdyt?

ge0ffrey: How long does calling a REST method take? That polling should not be able to DDOS a slow rest service.

ge0ffrey: @rsynek These comments didn't make it in (nor have a comment to reject them), I 'll provide a Pr for it together with the issue itself.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
ge0ffrey: This timeout approach is not race condition free. If we activate this test, it will cause randomly failing unit tests (which is never acceptable). I am going to refactor this to use Latches.

PR #663, opened 2020-01-24
[DROOLS-4092] Avoid using Drools' executable model

lucamolteni:
https://github.com/kiegroup/droolsjbpm-integration/pull/1864
https://github.com/kiegroup/drools/pull/2636
https://github.com/kiegroup/openshift-drools-hacep/pull/68
https://github.com/kiegroup/kie-wb-playground/pull/78
https://github.com/kiegroup/droolsjbpm-knowledge/pull/397

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
@@ -30,6 +30,7 @@ import com.thoughtworks.xstream.annotations.XStreamConverter;
 import com.thoughtworks.xstream.annotations.XStreamImplicit;
 import com.thoughtworks.xstream.annotations.XStreamOmitField;
 import org.apache.commons.lang3.BooleanUtils;
+import org.drools.compiler.kie.builder.impl.DrlProject;
 import org.kie.api.KieBase;
 import org.kie.api.KieBaseConfiguration;
 import org.kie.api.KieServices;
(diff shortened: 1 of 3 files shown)


triceo:
@lucamolteni Would you mind explaining the rationale behind this change?
I am a little confused - the other PRs actually make executable model the default, while only this one disables it?

lucamolteni:
@triceo I spoke with @ge0ffrey on Friday about this and, even if we want the executable model as a default in Drools we don't necessarily want it in every project that uses Drools.
This is the safest choice so far, when you want to switch to executable model you simply change that switch.

ge0ffrey:
@triceo I still had to update you on this.
Basically, Drools until recently used a DRL parser implementation that doesn't use the exe model. Luca's work replaces that. They're shipping it as the default now, but it still has a known bug for one of OptaPlanner's examples (and it hasn't been tested by the notorious turtle tests yet, which will probably discover more). We've agreed that OptaPlanner temporarily doesn't use it yet for scoreDRL (not for ConstraintStreams obviously, as that doesn't use DRL) and that once Luca fixes that known issue, he reverts this change so the turtle tests start grinding at it.

ge0ffrey:
@lucamolteni What's the jira key for the known issue that makes optaplanner's test fail with the new DRL parser?

lucamolteni:
@ge0ffrey https://issues.redhat.com/browse/DROOLS-4979

It was fixed by Mario today, so after the merge we'll start investigating enabling it.

This the issue for reenabling it 

https://issues.redhat.com/browse/DROOLS-4984

@mariofusco ^

lucamolteni:
Please merge this now as 4092 was merged

PR #665, opened 2020-01-29
PLANNER-1791 Add SolverManager tests

michaltomco:
Tests the following requirements:
- Test how multiple solver manager threads work together with multi-threaded solving
- Test submitting more problems than number of cpu cores.
- Submitting same problem twice

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
@@ -283,14 +283,14 @@ public class SolverManagerTest {
         // Use twice the amount of problems than available processors.
         int problemCount = Runtime.getRuntime().availableProcessors() * 2;
 
-        SolverManager<TestdataSolution, Integer> solverManager = createSolverManagerTestableByDifferentConsumers(problemCount);
+        SolverManager<TestdataSolution, Integer> solverManager = createSolverManagerTestableByDifferentConsumers();
 
         assertSolveWithoutConsumer(problemCount, solverManager);
         assertSolveWithBestSolutionConsumer(problemCount, solverManager);
         assertSolveWithFinalBestSolutionConsumer(problemCount, solverManager);
     }
 
-    private SolverManager<TestdataSolution, Integer> createSolverManagerTestableByDifferentConsumers(int processCount) {
+    private SolverManager<TestdataSolution, Integer> createSolverManagerTestableByDifferentConsumers() {
         final SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
                 .withPhases(new CustomPhaseConfig().withCustomPhaseCommands(
                         (ScoreDirector<TestdataSolution> scoreDirector) -> {
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
@@ -16,6 +16,8 @@
 
 package org.optaplanner.core.api.solver;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.CyclicBarrier;
@@ -37,8 +39,12 @@ import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
 import org.optaplanner.core.impl.testdata.domain.extended.TestdataUnannotatedExtendedSolution;
 import org.optaplanner.core.impl.testdata.util.PlannerTestUtils;
 
-import static org.junit.Assert.*;
-import static org.optaplanner.core.impl.testdata.util.PlannerAssert.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.optaplanner.core.impl.testdata.util.PlannerAssert.assertSolutionInitialized;
 
 public class SolverManagerTest {
 
@@ -240,4 +246,153 @@ public class SolverManagerTest {
         assertTrue(eventCount.get() < 4);
     }
 
+    /**
+     * Tests whether SolverManager can solve on multiple threads problems that use multiple thread counts.
+     */
+    @Test(timeout = 600_000)
+    public void multipleThreadSolvingWithSolverManager() throws ExecutionException, InterruptedException {
+        int processCount = Runtime.getRuntime().availableProcessors();
+        CyclicBarrier barrier = new CyclicBarrier(processCount);
+        final SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
+                .withPhases(new ConstructionHeuristicPhaseConfig(),
+                            new CustomPhaseConfig().withCustomPhaseCommands(
+                                    scoreDirector -> {
+                                        try {
+                                            barrier.await();
+                                        } catch (InterruptedException | BrokenBarrierException e) {
+                                            fail("Cyclic barrier failed.");
+                                        }
+                                    }))
+                // Adds moveThreadCount to the solver config.
+                .withMoveThreadCount(String.valueOf(processCount));
+        // Creates solverManagerConfig with multiple threads.
+        SolverManager<TestdataSolution, Long> solverManager = SolverManager.create(
+                solverConfig, new SolverManagerConfig().withParallelSolverCount(String.valueOf(processCount)));
+
+        List<SolverJob<TestdataSolution, Long>> jobs = new ArrayList<>();
+        for (long i = 0; i < processCount; i++) {
+            jobs.add(solverManager.solve(i, PlannerTestUtils.generateTestdataSolution("s" + i)));
+        }
+
+        for (SolverJob<TestdataSolution, Long> job : jobs) {
+            assertSolutionInitialized(job.getFinalBestSolution());
+        }
+    }
+
+    /**
+     * Tests whether SolverManager can cope with the situation when more problems than processors are submitted.
+     * In addition, the test verifies whether solving with consumers is affected by this.
+     */
+    @Test(timeout = 600_000)
+    public void moreProblemsThanCpus() throws InterruptedException, ExecutionException {
+        // Use twice the amount of processes than available processors.
+        int problemCount = Runtime.getRuntime().availableProcessors() * 2;
+
+        SolverManager<TestdataSolution, Integer> solverManager = createSolverManagerTestableByDifferentConsumers(problemCount);
+
+        assertSolveWithoutConsumer(problemCount, solverManager);
+        assertSolveWithBestSolutionConsumer(problemCount, solverManager);
+        assertSolveWithFinalBestSolutionConsumer(problemCount, solverManager);
+    }
+
+    private SolverManager<TestdataSolution, Integer> createSolverManagerTestableByDifferentConsumers(int processCount) {
+        final SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
+                .withPhases(new CustomPhaseConfig().withCustomPhaseCommands(
+                        (ScoreDirector<TestdataSolution> scoreDirector) -> {
+                            TestdataSolution solution = scoreDirector.getWorkingSolution();
+                            TestdataEntity entity = solution.getEntityList().get(0);
+                            scoreDirector.beforeVariableChanged(entity, "value");
+                            entity.setValue(solution.getValueList().get(0));
+                            scoreDirector.afterVariableChanged(entity, "value");
+                            scoreDirector.triggerVariableListeners();
+                        }), new CustomPhaseConfig().withCustomPhaseCommands(
+                        (ScoreDirector<TestdataSolution> scoreDirector) -> {
+                            TestdataSolution solution = scoreDirector.getWorkingSolution();
+                            TestdataEntity entity = solution.getEntityList().get(1);
+                            scoreDirector.beforeVariableChanged(entity, "value");
+                            entity.setValue(solution.getValueList().get(1));
+                            scoreDirector.afterVariableChanged(entity, "value");
+                            scoreDirector.triggerVariableListeners();
+                        }));
+
+        SolverManagerConfig solverManagerConfig = new SolverManagerConfig().withParallelSolverCount(String.valueOf(processCount));
+
+        return SolverManager.create(solverConfig, solverManagerConfig);
+    }
+
+    private void assertSolveWithoutConsumer(int processCount, SolverManager<TestdataSolution, Integer> solverManager) throws InterruptedException, ExecutionException {
+        List<SolverJob<TestdataSolution, Integer>> jobs = new ArrayList<>(processCount);
+
+        for (int id = 0; id < processCount; id++) {
+            jobs.add(solverManager.solve(id, PlannerTestUtils.generateTestdataSolution("s" + id, 2)));
+        }
+
+        for (SolverJob<TestdataSolution, Integer> job : jobs) {
+            assertSolutionInitialized(job.getFinalBestSolution());
+        }
+    }
+
+    private void assertSolveWithBestSolutionConsumer(int processCount, SolverManager<TestdataSolution, Integer> solverManager)
+            throws InterruptedException {
+        int bestSolutionEventTriggeredCountExpected = processCount * 2;
+        CountDownLatch bestSolutionConsumerLatch = new CountDownLatch(bestSolutionEventTriggeredCountExpected);
+
+        AtomicInteger bestSolutionEventTriggerCount = new AtomicInteger();
+        Consumer<TestdataSolution> bestSolutionConsumer = bestSolution -> {
+            bestSolutionEventTriggerCount.incrementAndGet();
+            bestSolutionConsumerLatch.countDown();
+        };
+        for (int id = 0; id < processCount; id++) {
+            solverManager.solveAndListen(
+                    id,
+                    problemId -> PlannerTestUtils.generateTestdataSolution("s" + problemId, 2),
+                    bestSolutionConsumer, null);
+        }
+        bestSolutionConsumerLatch.await();
+
+        assertThat(bestSolutionEventTriggerCount.intValue()).isSameAs(bestSolutionEventTriggeredCountExpected);
+    }
+
+    private void assertSolveWithFinalBestSolutionConsumer(int processCount, SolverManager<TestdataSolution, Integer> solverManager)
+            throws InterruptedException {
+        int finalBestSolutionEventTriggeredCountExpected = processCount;
+        CountDownLatch finalBestSolutionConsumerLatch = new CountDownLatch(finalBestSolutionEventTriggeredCountExpected);
+
+        AtomicInteger finalBestSolutionEventTriggerCount = new AtomicInteger();
+        Consumer<TestdataSolution> finalBestSolutionConsumer = finalBestSolution -> {
+            finalBestSolutionEventTriggerCount.incrementAndGet();
+            finalBestSolutionConsumerLatch.countDown();
+        };
+        for (int id = 0; id < processCount; id++) {
+            solverManager.solve(
+                    id,
+                    problemId -> PlannerTestUtils.generateTestdataSolution("s" + problemId, 2),
+                    finalBestSolutionConsumer, null);
+        }
+        finalBestSolutionConsumerLatch.await();
+
+        assertThat(finalBestSolutionEventTriggerCount.intValue()).isSameAs(finalBestSolutionEventTriggeredCountExpected);
+    }
+
+    @Test(timeout = 600_000)
+    public void runSameIdProcesses() {
+        SolverManagerConfig config = new SolverManagerConfig().withParallelSolverCount("2");
+
+        CyclicBarrier barrier = new CyclicBarrier(2);
+        final SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
+                .withPhases(new CustomPhaseConfig().withCustomPhaseCommands(
+                        scoreDirector -> {
+                            try {
+                                barrier.await();
+                            } catch (InterruptedException | BrokenBarrierException e) {
+                                fail("Cyclic barrier failed.");
+                            }
+                        }));
+
+        SolverManager<TestdataSolution, Long> solverManager = SolverManager.create(solverConfig, config);
+
+        solverManager.solve(1L, PlannerTestUtils.generateTestdataSolution("solver"));
+        assertThatThrownBy(() -> solverManager.solve(1L, PlannerTestUtils.generateTestdataSolution("solver")))
+                .isInstanceOf(IllegalStateException.class).hasMessageContaining("already solving");
+    }
 }
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
@@ -17,6 +17,7 @@
 package org.optaplanner.core.api.solver;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +246,153 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+            assertSolutionInitialized(job.getFinalBestSolution());
+        }
+    }
+
+    /**
+     * Tests whether SolverManager can cope with the situation when more problems than processors are submitted.
+     * In addition, the test verifies whether solving with consumers is affected by this.
+     */
+    @Test(timeout = 600_000)
+    public void moreProblemsThanCpus() throws InterruptedException, ExecutionException {
```
rsynek: The comment might not be needed, if the test is called e.g.:
submitMoreProblemsThanCpus_allGetSolved()

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +246,153 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        // Use twice the amount of processes than available processors.
+        int problemCount = Runtime.getRuntime().availableProcessors() * 2;
+
+        SolverManager<TestdataSolution, Integer> solverManager = createSolverManagerTestableByDifferentConsumers(problemCount);
```
rsynek: I am not entirely sure I understand what this test aims to do. The `problemCount` != CPU count used by a solver manager.

One can submit 1000 problems while having only 1 CPU used by a solver manager.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +246,153 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+                            scoreDirector.afterVariableChanged(entity, "value");
+                            scoreDirector.triggerVariableListeners();
+                        }));
+
+        SolverManagerConfig solverManagerConfig = new SolverManagerConfig().withParallelSolverCount(String.valueOf(processCount));
+
+        return SolverManager.create(solverConfig, solverManagerConfig);
+    }
+
+    private void assertSolveWithoutConsumer(int processCount, SolverManager<TestdataSolution, Integer> solverManager) throws InterruptedException, ExecutionException {
```
rsynek: Opposed to the other `assertSolver*` methods, this one lacks any synchronization between submitting the problems to solver manager and checking results. It's a risk in terms of test stability.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +246,153 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+                    id,
+                    problemId -> PlannerTestUtils.generateTestdataSolution("s" + problemId, 2),
+                    bestSolutionConsumer, null);
+        }
+        bestSolutionConsumerLatch.await();
+
+        assertThat(bestSolutionEventTriggerCount.intValue()).isSameAs(bestSolutionEventTriggeredCountExpected);
+    }
+
+    private void assertSolveWithFinalBestSolutionConsumer(int processCount, SolverManager<TestdataSolution, Integer> solverManager)
```
rsynek: It might be interesting to also check if the best solution publish to the consumer is the same as the best solution available via the `SolverJob`.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
michaltomco: Yes, the count of problems added (e.g. on line 321) exceeds the cpu count. The idea of the test is to check whether SolverManager can manage to solve more problems than it has cpus. ( https://docs.google.com/document/d/1W_lgu7Aliv1XVX133yUrHG1V7-b97dMNGsbAOc-DGRg/edit# under Testing, the 3rd point )

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
michaltomco: Done for the consumer that gets the finalBestSolution. As for the solveAndListen portion, I am not entirely sure how to obtain the bestSolution that is not final from a job. Currently I check whether the final ones are contained in the list of consumers and that the list is of size 2x the problem count. Any ideas how to deal with this issue?

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
rsynek: That's fine. By why passing the problemCount to the configuration of SolverManager as the number of threads it should create?

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
rsynek: You can have (and probably should have) a separate consumer for each submitter problem. That's how the API most likely will be used. In such a case, knowing there are two best solutions per problem, just take the second one.

michaltomco:
I've reworked the solveMultipleThreadedMovesWithSolverManager_allGetSolved() test, which still needs a bit refurbishing. I still have to find a good way how to check whether solverManager does not eat all the cores in a consistent manner. I've ran it on my pc and Jenkins with different configurations, from which none of them froze and all of them came to a satisfying result. I still have to find a good way how to check this in a consistent manner that doesn't take too much time.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
michaltomco: Yes, I agree it doesn't make sense, I've removed it.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +249,159 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        for (long i = 0; i < processCount; i++) {
+            jobs.add(solverManager.solve(i, PlannerTestUtils.generateTestdataSolution("s" + i, 10)));
+        }
+
+        for (SolverJob<TestdataSolution, Long> job : jobs) {
+            assertSolutionInitialized(job.getFinalBestSolution());
+        }
+    }
+
+    @Test(timeout = 600_000)
```
triceo: Surely this is excessive.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +249,159 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+                            scoreDirector.triggerVariableListeners();
+                        }), new CustomPhaseConfig().withCustomPhaseCommands(
+                        (ScoreDirector<TestdataSolution> scoreDirector) -> {
+                            TestdataSolution solution = scoreDirector.getWorkingSolution();
+                            TestdataEntity entity = solution.getEntityList().get(1);
+                            scoreDirector.beforeVariableChanged(entity, "value");
+                            entity.setValue(solution.getValueList().get(1));
+                            scoreDirector.afterVariableChanged(entity, "value");
+                            scoreDirector.triggerVariableListeners();
+                        }));
```
triceo: These two phase commands are nearly identical. Reuse?

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +249,159 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+                    problemId -> PlannerTestUtils.generateTestdataSolution("s" + problemId, 2),
+                    finalBestSolutionConsumer, null).getFinalBestSolution());
+        }
+        finalBestSolutionConsumerLatch.await();
+
+        assertThat(finalBestSolutionEventTriggerCount.intValue()).isSameAs(finalBestSolutionEventTriggeredCountExpected);
+        assertThat(consumedFinalBestSolutions).containsAll(jobFinalBestSolutions);
+    }
+
+    @Test(timeout = 600_000)
```
triceo: 10 minutes is just way too much. Are these turtle tests instead?

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+    private void assertConsumedFirstBestSolution(TestdataSolution solution) {
+        TestdataEntity entity = solution.getEntityList().get(0);
+        assertThat(entity.getCode()).isEqualTo("e1");
+        assertThat(entity.getValue().getCode()).isEqualTo("v1");
+        entity = solution.getEntityList().get(1);
+        assertThat(entity.getCode()).isEqualTo("e2");
+        assertThat(entity.getValue()).isNull();
+    }
+
+    @Test(timeout = 100)
```
rsynek: I wouldn't take decreasing timeouts to extreme. 100 ms could be too little. Units to tens of seconds is the most usual choice.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
         assertTrue(eventCount.get() < 4);
     }
 
+    @Ignore("A test that runs forever and is here to investigate a faulty thread-creating behavior.")
+    @Test
+    public void solveMultipleThreadedMovesWithSolverManager_allGetSolved() throws ExecutionException, InterruptedException {
```
rsynek: @michaltomco would you please create a Jira for making it a runnable test?

michaltomco: Done: https://issues.redhat.com/browse/PLANNER-1833

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -37,23 +44,19 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.optaplanner.core.impl.testdata.util.PlannerAssert.assertSolutionInitialized;
 
 public class SolverManagerTest {
 
-    @Test(timeout = 600_000)
+    @Test(timeout = 1_000)
```
ge0ffrey: Some of our jenkins machines are really slow. If the machine is too slow, this test will randomly fail if the timeout is too short. Always put it on 600_000 even if it runs in less than a second on your machine, just for safety.
The timeout is just there to prevent a forever run.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+    }
+
+    private void assertInitializedJobs(List<SolverJob<TestdataSolution, Integer>> jobs) throws InterruptedException, ExecutionException {
+        for (SolverJob<TestdataSolution, Integer> job : jobs) {
+            // Method getFinalBestSolution() waits for the solving to finish, therefore it ensures synchronization.
+            assertSolutionInitialized(job.getFinalBestSolution());
+        }
+    }
+
+    @Test(timeout = 2_000)
```
ge0ffrey: 600_000

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+    private void assertConsumedFirstBestSolution(TestdataSolution solution) {
+        TestdataEntity entity = solution.getEntityList().get(0);
+        assertThat(entity.getCode()).isEqualTo("e1");
+        assertThat(entity.getValue().getCode()).isEqualTo("v1");
+        entity = solution.getEntityList().get(1);
+        assertThat(entity.getCode()).isEqualTo("e2");
+        assertThat(entity.getValue()).isNull();
+    }
+
+    @Test(timeout = 1_000)
```
ge0ffrey: 600_000

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+    public void runSameIdProcesses_throwsIllegalStateException() {
+        SolverManagerConfig solverManagerConfig = new SolverManagerConfig();
+
+        SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
+                .withPhases(createPhaseWithConcurrentSolvingStart(2));
+
+        SolverManager<TestdataSolution, Long> solverManager =
+                SolverManager.create(solverConfig, solverManagerConfig);
+
+        solverManager.solve(1L, PlannerTestUtils.generateTestdataSolution("solver"));
```
ge0ffrey: using a solution code ID of "solver" is confusing. Please use "s1" which is short for solution 1.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        assertThatThrownBy(() -> solverManager.solve(1L, PlannerTestUtils.generateTestdataSolution("solver")))
```
ge0ffrey: also "s1"

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+
+    private void assertConsumedBestSolutions(int problemCount, Map<Integer, List<TestdataSolution>> consumedSolutions) {
+        for (int i = 0; i < problemCount; i++) {
+            assertThat(consumedSolutions.get(i)).hasSize(2);
+            assertConsumedFirstBestSolution(consumedSolutions.get(i).get(0));
+            assertConsumedFinalBestSolution(consumedSolutions.get(i).get(1));
+        }
+    }
+
+    private void assertConsumedFirstBestSolution(TestdataSolution solution) {
```
ge0ffrey: I wouldn't extract this into a seperate method, because:
- that code isn't used twice. No DRY justifying this
- this extraction doesn't make the code easier to read (assertConsumedFirstBestSolution isn't an obvious abstraction). It makes it harder to follow the flow.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+            assertConsumedBestSolutions(problemCount, consumedSolutions);
+        }
+    }
+
+    private void assertConsumedFinalBestSolutions(int problemCount, Map<Integer, List<TestdataSolution>> consumedSolutions) {
+        for (int i = 0; i < problemCount; i++) {
+            assertThat(consumedSolutions.get(i)).hasSize(1);
+            TestdataSolution solution = consumedSolutions.get(i).get(0);
+            assertConsumedFinalBestSolution(solution);
+        }
```
ge0ffrey: Sorry, I can't follow how the code flows here. It's not readable.
Try unextracting all methods starting with assertConsumed and then figuring out which methods make sense to extract. Avoid using both "assertConsumedFinalBestSolutions" and "assertConsumedFinalBestSolution"

ge0ffrey: There is SOFT guideline to avoid having more than 20 lines in a single method. Don't hesitate to violate that guideline to keep the code readable, it's not a hard guideline. Especially for tests, 20 lines is awfully low IMO.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+    private void assertConsumedSolutions(int problemCount, boolean onlyFinalBestSolutions, Map<Integer, List<TestdataSolution>> consumedSolutions) {
+        if (onlyFinalBestSolutions) {
+            assertConsumedFinalBestSolutions(problemCount, consumedSolutions);
+        } else {
+            assertConsumedBestSolutions(problemCount, consumedSolutions);
+        }
+    }
+
+    private void assertConsumedFinalBestSolutions(int problemCount, Map<Integer, List<TestdataSolution>> consumedSolutions) {
+        for (int i = 0; i < problemCount; i++) {
```
ge0ffrey: Use for each whenever possible.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        TestdataEntity entity = solution.getEntityList().get(0);
+        assertThat(entity.getCode()).isEqualTo("e1");
+        assertThat(entity.getValue().getCode()).isEqualTo("v1");
+        entity = solution.getEntityList().get(1);
+        assertThat(entity.getCode()).isEqualTo("e2");
+        assertThat(entity.getValue().getCode()).isEqualTo("v2");
+    }
+
+    private void assertConsumedBestSolutions(int problemCount, Map<Integer, List<TestdataSolution>> consumedSolutions) {
+        for (int i = 0; i < problemCount; i++) {
```
ge0ffrey: Use for each, no need to have an i variable, dimishes readability

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+            jobs.add(solverManager.solve(id, PlannerTestUtils.generateTestdataSolution(Integer.toString(id))));
+        }
+        assertInitializedJobs(jobs);
+    }
+
+    private void assertSolveWithConsumer(
+            int problemCount, SolverManager<TestdataSolution, Integer> solverManager, boolean onlyFinalBestSolutions)
+            throws ExecutionException, InterruptedException {
+
+        // Two solutions for every problem.
```
ge0ffrey: Not a sentance, remove the dot at the end, xor use a verb in it

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        Map<Integer, List<TestdataSolution>> consumedSolutions = new HashMap<>(problemCount * 2);
```
ge0ffrey: For consistency with other optaplanner code, the variable name of:
- Array and Collection uses plurals, for example "solutions"
- List uses a List suffix: for example  "solutionList"
- Map **definitly** uses a Map suffix, for example "solutionMap" (or "countToSolutionMap" in case the key is not dead obvious)

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+
+        List<SolverJob<TestdataSolution, Integer>> jobs = new ArrayList<>(problemCount);
+
+        for (int id = 0; id < problemCount; id++) {
+            List<TestdataSolution> consumedBestSolutions = Collections.synchronizedList(new ArrayList<>());
+            String solutionName = Integer.toString(id);
```
ge0ffrey: Other testcases typically do solutionName = "s" + i

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+                        problemId -> PlannerTestUtils.generateTestdataSolution(solutionName, 2),
+                        consumedBestSolutions::add, null));
+            }
+            consumedSolutions.put(id, consumedBestSolutions);
+        }
+        assertInitializedJobs(jobs);
+        assertConsumedSolutions(problemCount, onlyFinalBestSolutions, consumedSolutions);
+    }
+
+    private void assertConsumedSolutions(int problemCount, boolean onlyFinalBestSolutions, Map<Integer, List<TestdataSolution>> consumedSolutions) {
```
ge0ffrey: Can we rename "onlyFinalBestSolutions" to "listenWhileSolving" so it uses the same terminology as solveAndListen() vs solve(), to indicate that it's basically making that difference?

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+
+    private PhaseConfig[] createGraduallyInitializedPhaseConfigs() {
+        PhaseConfig[] configs = new PhaseConfig[2];
+        for (int x = 0; x < 2; x++) {
+            configs[x] = createSetXEntityToXValuePhaseConfig(x);
+        }
+        return configs;
+    }
+
+    private CustomPhaseConfig createSetXEntityToXValuePhaseConfig(int x) {
```
ge0ffrey: I wouldn't extract this method (see other case below). No DRY and no readability improvement.  createSetXEntityToXValuePhaseConfig isn't an obvious or near-obvious abstraction

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -240,4 +255,170 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
+        for (SolverJob<TestdataSolution, Integer> job : jobs) {
+            // Method getFinalBestSolution() waits for the solving to finish, therefore it ensures synchronization.
+            assertSolutionInitialized(job.getFinalBestSolution());
+        }
+    }
+
+    @Test(timeout = 2_000)
+    public void submitMoreProblemsThanCpus_allGetSolved() throws InterruptedException, ExecutionException {
+        // Use twice the amount of problems than available processors.
+        int problemCount = Runtime.getRuntime().availableProcessors() * 2;
```
ge0ffrey: Intresting - we shoud indeed test this. Good idea.
@rsynek Is there any risk for random failing tests? (not acceptable)
Do we accept the impact for test reproducibility? (acceptable for me)

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -66,7 +69,19 @@ public void solveBatch_2InParallel() throws ExecutionException, InterruptedExcep
         assertSolutionInitialized(solverJob2.getFinalBestSolution());
     }
 
-    @Test(timeout = 600_000)
+    private CustomPhaseConfig createPhaseWithConcurrentSolvingStart(int barrierPartiesCount) {
```
ge0ffrey: Is this method reused anywhere else? Otherwise, same arguments against extraction as below

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
rsynek: I will unify all the SolverManager tests to 60_000. 1 minute is more than enough, as the test usually finishes below 1 second. If some of the tests takes more than a minute in CI, I would argue we should be notified about it.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
rsynek: It is used by two different tests, I would rather keep it.

rsynek:
Can be closed in favour of #676 (original commits are included there).

PR #666, opened 2020-01-29
PLANNER-1788 ifExists()/ifNotExists() for BiStream

triceo:


triceo:
@ge0ffrey Ready for review.

One thing I just realized and that we didn't think about: so, how does `ifExists()` behave with regards to uninitialized entities?

For `from()`, you have two different methods, so that the user can chose.
If we do this here, the API will become just insane, essentially the existing 10 methods per stream will  double to 20.
But I'm not sure that the current API is reasonable either - `from()` will give the user initialized entities, but `ifExists()` also operates with uninitialized ones.

So, three options I see:

- Ignore the issue. The current API is fine.
- We double the number of methods on the API (`ifExistsOtherUnfiltered()`?).
- If `from()` is called, `ifExists()` is smart and knows that. If `fromUnfiltered()` is called, `ifExists()` also adapts.

WDYT?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -198,6 +198,166 @@
...
+     * This method causes <i>Unchecked generics array creation for varargs parameter</i> warnings,
+     * but we can't fix it with a {@link SafeVarargs} annotation because it's an interface method.
+     * Therefore, there are overloaded methods with up to 4 {@link TriJoiner} parameters.
+     * @param otherClass never null
+     * @param joiners never null
+     * @param <C> the type of the third matched fact
+     * @return never null, a stream that matches every pair of A and B where C does not exist for which the
+     * {@link TriJoiner}s are true
+     */
+    <C> BiConstraintStream<A, B> ifNotExists(Class<C> otherClass, TriJoiner<A, B, C>... joiners);
```
triceo: I just experimented with, this, and if called like `ifNotExists(Visit.class)` - without the joiners - this gives varargs warnings in IDE. Not sure what to do about this method.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -198,6 +198,166 @@
         return join(otherClass, AbstractTriJoiner.merge(joiners));
     }
 
+    // ************************************************************************
+    // If (Not) Exists
```
ge0ffrey: style: Not and Exists without capital. only first word has capital in comments

ge0ffrey: BTW, in general for titles (including docs), we're now following scientific titles, not newspaper titles. So basically only the first word has a capital.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
ge0ffrey: In the javadoc it says:
     * This method causes <i>Unchecked generics array creation for varargs parameter</i> warnings,
     * but we can't fix it with a {@link SafeVarargs} annotation because it's an interface method.
     * Therefore, there are overloaded methods with up to 4 {@link TriJoiner} parameters.
I've seen that issue on all other varargs methods before - is this the same issue as what you are seeing?

If it is, ignore it. If it's not, let's discuss it.

triceo: Yeah, it is the same with joins etc. Ok, leaving it as it is.

PR #667, opened 2020-01-30
PLANNER-1832 Add generic ConstraintCollectors.sum()

triceo:
A tiny little PR adding a useful feature.
Many of our collectors can actually be reimplemented using it, getting rid of some code duplication.


triceo:
@Christopher-Chianelli You are right. Although in the case of primitive types (`int`, `long`), we keep the method overloads. Otherwise we would lose performance from all the auto-boxing.

PR #674, opened 2020-02-05
PLANNER-1825 Add constraint provider for TSP

triceo:
Example verified with assertion score director.
CS-D performance on par with DRL.


optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/app/TspApp.java
```diff
@@ -55,7 +55,7 @@ protected TspPanel createSolutionPanel() {
 
     @Override
     public SolutionFileIO<TspSolution> createSolutionFileIO() {
-        return new XStreamSolutionFileIO<>(TspSolution.class);
+        return new TspXStreamSolutionFileIO();
```
ge0ffrey: Why extract that in a separate file? The more files in an example, the harder it is to understand for our users.

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/optional/score/TspConstraintProvider.java
```diff
@@ -0,0 +1,53 @@
...
+        return new Constraint[]{
+                distanceToPreviousStandstill(constraintFactory),
+                distanceFromLastVisitToDomicile(constraintFactory)
+        };
+    }
+
+    private Constraint distanceToPreviousStandstill(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Visit.class)
+                .penalizeLong("Distance to previous standstill",
+                        SimpleLongScore.of(1),
```
ge0ffrey: Use SimpleLongScore.ONE

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/app/TspApp.java
triceo: I thought you were going to ask. :-) I have a harness that allows me to run all examples in a single benchmark - essentially an abstraction to easily perform score director benchmarks over all examples and all score directors, without having a benchmark template of a million lines.

The downside of that is that it relies pretty heavily on generics. And I had to do this to standardize it with the other examples. (This was the only one that didn't have its own solution file IO.)

I will submit the harness as a PR later and we can discuss it.

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/optional/score/TspConstraintProvider.java
```diff
@@ -0,0 +1,53 @@
...
+                        Visit::getDistanceFromPreviousStandstill);
+    }
+
+    private Constraint distanceFromLastVisitToDomicile(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Visit.class)
+                .ifNotExists(Visit.class, Joiners.equal(visit -> visit, Visit::getPreviousStandstill))
```
ge0ffrey: I guess "visit -> visit" is better than "Function.identity()"?

Should we experiment with thisEqual() or equalToThat() ? I guess that's overkill...

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/app/TspApp.java
ge0ffrey: If all other examples do it, there is very little point indeed in this example not doing it to be easier to understand. +1 for this change (consistency)

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/optional/score/TspConstraintProvider.java
ge0ffrey: (just a thought, no changes needed)

triceo: In this particular case, `Function.identity()` didn't compile.`identity()` returns a `Visit` while `getPreviousStandstill()` returns a `Standstill`. The compiler didn't like that.

PR #676, opened 2020-02-07
PLANNER-1791 Add SolverManager tests

rsynek:
This PR incorporates #665 + addresses a majority of review comments.

@michaltomco please go over the original comments from #665 and open a new PR to address the remaining ones of them.

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
@@ -25,10 +25,13 @@ import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -47,31 +57,35 @@
...
+                scoreDirector -> {
+                    try {
+                        barrier.await();
+                    } catch (InterruptedException | BrokenBarrierException e) {
+                        fail("Cyclic barrier failed.");
+                    }
+                });
+    }
+
+    @Test(timeout = 60_000)
```
ge0ffrey: Are you sure you want to bring the timeouts down to 1 minute isntead of 10 minutes? I remember we set it to 10 minutes for good reasons, even though the slowest test only took 10 seconds on our machines back then.

rsynek: Each test runs below 1 sec, I don't believe the CI environment will be > 60 times slower. If some test runs longer than 1 minute, I prefer to know about it (by seeing it failing on timeout).

PR #678, opened 2020-02-10
PLANNER-1840 Extra test coverage for joiners

triceo:
Besides adding test coverage, I also refactor parts of the joiners to make sure code is reused as much as reasonable.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/AbstractBiJoiner.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/AbstractBiJoiner.java
@@ -17,15 +17,17 @@
 package org.optaplanner.core.impl.score.stream.bi;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
+import java.util.stream.IntStream;
 
 import org.optaplanner.core.api.score.stream.bi.BiJoiner;
 import org.optaplanner.core.impl.score.stream.common.AbstractJoiner;
 import org.optaplanner.core.impl.score.stream.common.JoinerType;
 
-public abstract class AbstractBiJoiner<A, B> extends AbstractJoiner implements BiJoiner<A, B> {
+public abstract class AbstractBiJoiner<A, B> extends AbstractJoiner<B> implements BiJoiner<A, B> {
 
     private final BiPredicate<A, B> filter;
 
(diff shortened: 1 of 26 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
```diff
@@ -0,0 +1,349 @@
...
+
+import org.junit.Test;
+import org.optaplanner.core.api.function.QuadFunction;
+import org.optaplanner.core.api.function.TriFunction;
+import org.optaplanner.core.impl.score.stream.bi.AbstractBiJoiner;
+import org.optaplanner.core.impl.score.stream.penta.AbstractPentaJoiner;
+import org.optaplanner.core.impl.score.stream.quad.AbstractQuadJoiner;
+import org.optaplanner.core.impl.score.stream.tri.AbstractTriJoiner;
+
+import static java.math.BigInteger.*;
```
rsynek: Code style: I think that `*` imports are generally less preferable. Is it intentional here or is it an accident caused by IDE?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractJoiner.java
```diff
@@ -16,8 +16,30 @@
 
 package org.optaplanner.core.impl.score.stream.common;
 
-public abstract class AbstractJoiner {
+import java.util.Arrays;
+import java.util.function.Function;
+import java.util.stream.IntStream;
+
+public abstract class AbstractJoiner<Right> {
```
ge0ffrey: This presumes that that we'll never do this:
`Bi<A,B>.join(Bi<C,D>)`

-1 to abstract the Right generatic type here...

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/AbstractBiJoiner.java
```diff
@@ -75,11 +77,14 @@ public boolean matches(A a, B b) {
...
-
-    public abstract Function<B, Object> getRightMapping(int index);
-
-    public abstract Function<B, Object[]> getRightCombinedMapping();
+    public Function<A, Object[]> getLeftCombinedMapping() {
+        Function<A, Object>[] mappings = IntStream.range(0, getJoinerTypes().length)
+                .mapToObj(this::getLeftMapping)
+                .toArray(Function[]::new);
+        return (A a) -> Arrays.stream(mappings)
+                .map(f -> f.apply(a))
```
ge0ffrey: You can probably use f::apply (one less wrapper lambda)

triceo: Not in this case. It would work if I was trying to get rid of `f`, but here I need to pass `a` to `f` and therefore I need to do it like this.

triceo:
Jenkins please retest this.

PR #679, opened 2020-02-10
PLANNER-1829: Add a ConstraintProvider for Scrabble

Christopher-Chianelli:
Note: FAST_ASSERT passes, but FULL_ASSERT fails

```
[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 7.996 s <<< FAILURE! - in org.optaplanner.examples.scrabble.app.ScrabblePerformanceTest
[ERROR] solveModel[0: NONE]  Time elapsed: 7.994 s  <<< ERROR!
java.lang.IllegalStateException: 
Score corruption (4hard): the workingScore (-130init/2hard/0medium/-13soft) is not the uncorruptedScore (-130init/-2hard/0medium/-13soft) after completedAction ([ OPTAPLANNER  {HORIZONTAL -> HORIZONTAL},  OPTAPLANNER  {(1,20) -> (1,20)}]):
Score corruption analysis:
  The corrupted scoreDirector has 4 ConstraintMatch(s) which are in excess (and should not be there):
    org.optaplanner.examples.scrabble.domain/Character confict/[(12,20)]=1hard/0medium/0soft
    org.optaplanner.examples.scrabble.domain/Character confict/[(13,20)]=1hard/0medium/0soft
    org.optaplanner.examples.scrabble.domain/Pull to the center/[ MICROCONTAINER ]=0hard/0medium/0soft
    org.optaplanner.examples.scrabble.domain/Pull to the center/[ OPTAPLANNER ]=0hard/0medium/-13soft
  The corrupted scoreDirector has 136 ConstraintMatch(s) which are missing:
    org.optaplanner.examples.scrabble.solver/Character confict/[(12,20)]=-1hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Character confict/[(13,20)]=-1hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Maximize merges per word/[0,  AEROGEAR ]=0hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Maximize merges per word/[0,  ACTIVEMQ ]=0hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Maximize merges per word/[0,  CAMEL ]=0hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Maximize merges per word/[0,  CXF ]=0hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Maximize merges per word/[0,  KARAF ]=0hard/0medium/0soft
    org.optaplanner.examples.scrabble.solver/Maximize merges per word/[0,  SERVICEMIX ]=0hard/0medium/0soft
    ... 128 more
  Maybe there is a bug in the score constraints of those ConstraintMatch(s).
  Maybe a score constraint doesn't select all the entities it depends on, but finds some through a reference in a selected entity. This corrupts incremental score calculation, because the constraint is not re-evaluated if such a non-selected entity changes.
Shadow variable corruption in the corrupted scoreDirector:
  None
	at org.optaplanner.core.impl.score.director.AbstractScoreDirector.assertScoreFromScratch(AbstractScoreDirector.java:663)
	at org.optaplanner.core.impl.score.director.AbstractScoreDirector.assertWorkingScoreFromScratch(AbstractScoreDirector.java:641)
	at org.optaplanner.core.impl.score.director.AbstractScoreDirector.doAndProcessMove(AbstractScoreDirector.java:191)
	at org.optaplanner.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider.doMove(ConstructionHeuristicDecider.java:126)
	at org.optaplanner.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider.decideNextStep(ConstructionHeuristicDecider.java:100)
	at org.optaplanner.core.impl.constructionheuristic.DefaultConstructionHeuristicPhase.solve(DefaultConstructionHeuristicPhase.java:74)
	at org.optaplanner.core.impl.solver.AbstractSolver.runPhases(AbstractSolver.java:98)
	at org.optaplanner.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:189)
	at org.optaplanner.examples.common.app.SolverPerformanceTest.runSpeedTest(SolverPerformanceTest.java:98)
	at org.optaplanner.examples.scrabble.app.ScrabblePerformanceTest.solveModel(ScrabblePerformanceTest.java:44)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
@@ -63,7 +63,7 @@ public class ScrabbleConstraintProvider implements ConstraintProvider {
 
     private Constraint pullToCenter(ConstraintFactory cf) {
         return cf.from(ScrabbleWordAssignment.class)
-                 .penalize("Pull to the center", HardMediumSoftScore.ONE_SOFT, (swa) -> swa.getDistanceToCenter());
+                 .penalize("Pull to the center", HardMediumSoftScore.ONE_SOFT, ScrabbleWordAssignment::getDistanceToCenter);
     }
 
 }
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
@@ -32,19 +32,17 @@ public class ScrabbleConstraintProvider implements ConstraintProvider {
 
     private Constraint noParallelHorizontalNeighbours(ConstraintFactory cf) {
         return cf.from(ScrabbleCell.class).filter(sc -> sc.hasWordSet(ScrabbleWordDirection.HORIZONTAL))
-                 .join(ScrabbleCell.class,
-                       Joiners.equal(ScrabbleCell::getX), Joiners.lessThan(ScrabbleCell::getId),
-                       Joiners.filtering((first, other) -> other.hasWordSet(ScrabbleWordDirection.HORIZONTAL)))
-                 .filter((c1, c2) -> Math.abs(c1.getY() - c2.getY()) == 1)
+                 .ifExists(ScrabbleCell.class,
+                           Joiners.equal(ScrabbleCell::getX), Joiners.equal(ScrabbleCell::getY, c -> c.getY() + 1),
+                           Joiners.filtering((first, second) -> second.hasWordSet(ScrabbleWordDirection.HORIZONTAL)))
                  .penalize("No parallel horizontal neighbours", HardMediumSoftScore.ONE_HARD);
     }
 
     private Constraint noParallelVerticalNeighbours(ConstraintFactory cf) {
         return cf.from(ScrabbleCell.class).filter(sc -> sc.hasWordSet(ScrabbleWordDirection.VERTICAL))
-                 .join(ScrabbleCell.class,
-                       Joiners.equal(ScrabbleCell::getY), Joiners.lessThan(ScrabbleCell::getId),
-                       Joiners.filtering((first, other) -> other.hasWordSet(ScrabbleWordDirection.VERTICAL)))
-                 .filter((c1, c2) -> Math.abs(c1.getX() - c2.getX()) == 1)
+                 .ifExists(ScrabbleCell.class,
+                           Joiners.equal(ScrabbleCell::getY), Joiners.equal(ScrabbleCell::getX, c -> c.getX() + 1),
+                           Joiners.filtering((first, second) -> second.hasWordSet(ScrabbleWordDirection.VERTICAL)))
                  .penalize("No parallel vertical neighbours", HardMediumSoftScore.ONE_HARD);
     }
 
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
@@ -15,53 +15,52 @@ public class ScrabbleConstraintProvider implements ConstraintProvider {
     @Override
     public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
         return new Constraint[]{
-                                characterConflict(constraintFactory),
-                                noParallelHorizontalNeighbours(constraintFactory),
-                                noParallelVerticalNeighbours(constraintFactory),
-                                outOfGrid(constraintFactory),
-                                maximizeMergesPerWord(constraintFactory),
-                                pullToCenter(constraintFactory)
+                characterConflict(constraintFactory),
+                noParallelHorizontalNeighbours(constraintFactory),
+                noParallelVerticalNeighbours(constraintFactory),
+                outOfGrid(constraintFactory),
+                maximizeMergesPerWord(constraintFactory),
+                pullToCenter(constraintFactory)
         };
     }
 
     private Constraint characterConflict(ConstraintFactory cf) {
         return cf.from(ScrabbleCell.class)
-                 .filter(sc -> sc.getCharacterSet().size() >= 2)
-                 .penalize("Character confict", HardMediumSoftScore.ONE_HARD, sc -> sc.getCharacterSet().size() - 1);
+                .filter(sc -> sc.getCharacterSet().size() >= 2)
+                .penalize("Character confict", HardMediumSoftScore.ONE_HARD, sc -> sc.getCharacterSet().size() - 1);
     }
 
     private Constraint noParallelHorizontalNeighbours(ConstraintFactory cf) {
         return cf.from(ScrabbleCell.class).filter(sc -> sc.hasWordSet(ScrabbleWordDirection.HORIZONTAL))
-                 .ifExists(ScrabbleCell.class,
-                           Joiners.equal(ScrabbleCell::getX), Joiners.equal(ScrabbleCell::getY, c -> c.getY() + 1),
-                           Joiners.filtering((first, second) -> second.hasWordSet(ScrabbleWordDirection.HORIZONTAL)))
-                 .penalize("No parallel horizontal neighbours", HardMediumSoftScore.ONE_HARD);
+                .ifExists(ScrabbleCell.class,
+                          Joiners.equal(ScrabbleCell::getX), Joiners.equal(ScrabbleCell::getY, c -> c.getY() + 1),
+                          Joiners.filtering((first, second) -> second.hasWordSet(ScrabbleWordDirection.HORIZONTAL)))
+                .penalize("No parallel horizontal neighbours", HardMediumSoftScore.ONE_HARD);
     }
 
     private Constraint noParallelVerticalNeighbours(ConstraintFactory cf) {
         return cf.from(ScrabbleCell.class).filter(sc -> sc.hasWordSet(ScrabbleWordDirection.VERTICAL))
-                 .ifExists(ScrabbleCell.class,
-                           Joiners.equal(ScrabbleCell::getY), Joiners.equal(ScrabbleCell::getX, c -> c.getX() + 1),
-                           Joiners.filtering((first, second) -> second.hasWordSet(ScrabbleWordDirection.VERTICAL)))
-                 .penalize("No parallel vertical neighbours", HardMediumSoftScore.ONE_HARD);
+                .ifExists(ScrabbleCell.class,
+                          Joiners.equal(ScrabbleCell::getY), Joiners.equal(ScrabbleCell::getX, c -> c.getX() + 1),
+                          Joiners.filtering((first, second) -> second.hasWordSet(ScrabbleWordDirection.VERTICAL)))
+                .penalize("No parallel vertical neighbours", HardMediumSoftScore.ONE_HARD);
     }
 
     private Constraint outOfGrid(ConstraintFactory cf) {
         return cf.from(ScrabbleWordAssignment.class)
-                 .filter(ScrabbleWordAssignment::isOutOfGrid)
-                 .penalize("Out of grid", HardMediumSoftScore.ONE_HARD, swa -> swa.getWord().length());
+                .filter(ScrabbleWordAssignment::isOutOfGrid)
+                .penalize("Out of grid", HardMediumSoftScore.ONE_HARD, swa -> swa.getWord().length());
     }
 
     private Constraint maximizeMergesPerWord(ConstraintFactory cf) {
         return cf.from(ScrabbleWordAssignment.class)
-                 .join(ScrabbleCell.class, Joiners.filtering((swa, sc) -> sc.getWordSet().contains(swa) && sc.hasMerge()))
-                 .groupBy((swa, sc) -> swa.getId(), ConstraintCollectors.countBi())
-                 .reward("Maximize merges per word", HardMediumSoftScore.ONE_MEDIUM, (id, count) -> count * count);
+                .join(ScrabbleCell.class, Joiners.filtering((swa, sc) -> sc.getWordSet().contains(swa) && sc.hasMerge()))
+                .groupBy((swa, sc) -> swa.getId(), ConstraintCollectors.countBi())
+                .reward("Maximize merges per word", HardMediumSoftScore.ONE_MEDIUM, (id, count) -> count * count);
     }
 
     private Constraint pullToCenter(ConstraintFactory cf) {
         return cf.from(ScrabbleWordAssignment.class)
-                 .penalize("Pull to the center", HardMediumSoftScore.ONE_SOFT, ScrabbleWordAssignment::getDistanceToCenter);
+                .penalize("Pull to the center", HardMediumSoftScore.ONE_SOFT, ScrabbleWordAssignment::getDistanceToCenter);
     }
-
 }
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
@@ -1,0 +1,69 @@
+package org.optaplanner.examples.scrabble.optional.score;
+
+import org.optaplanner.core.api.score.buildin.hardmediumsoft.HardMediumSoftScore;
+import org.optaplanner.core.api.score.stream.Constraint;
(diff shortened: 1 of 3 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private Constraint noParallelVerticalNeighbours(ConstraintFactory cf) {
+        return cf.fromUniquePair(ScrabbleCell.class)
+                 .filter((c1, c2) -> c1.getY() == c2.getY() &&
+                                     Math.abs(c1.getX() - c2.getX()) == 1 &&
+                                     c1.hasWordSet(ScrabbleWordDirection.VERTICAL) &&
+                                     c2.hasWordSet(ScrabbleWordDirection.VERTICAL))
+                 .penalize("No parallel vertical neighbours", HardMediumSoftScore.ONE_HARD);
+    }
```
triceo: Some suggestions, for reasons of performance:

- The `getY()` comparisons should be done using `Joiners.equal(...)`.
- I would consider first filtering our all the cells where `c1.hasWordSet(...)` and only then joining with `c2`.

optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private Constraint outOfGrid(ConstraintFactory cf) {
+        return cf.from(ScrabbleWordAssignment.class)
+                 .filter(ScrabbleWordAssignment::isOutOfGrid)
+                 .penalize("Out of grid", HardMediumSoftScore.ONE_HARD, swa -> swa.getWord().length());
+    }
+
+    private Constraint maximizeMergesPerWord(ConstraintFactory cf) {
+        return cf.from(ScrabbleWordAssignment.class)
+                 .join(ScrabbleCell.class, new FilteringBiJoiner<>((swa, sc) -> sc.getWordSet().contains(swa) && sc.hasMerge()))
```
triceo: Please replace `FilteringBiJoiner` with `Joiners.filtering(...)`.

optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
```diff
@@ -0,0 +1,69 @@
...
+                                outOfGrid(constraintFactory),
+                                maximizeMergesPerWord(constraintFactory),
+                                pullToCenter(constraintFactory)
+        };
+    }
+
+    private Constraint characterConflict(ConstraintFactory cf) {
+        return cf.from(ScrabbleCell.class)
+                 .filter(sc -> sc.getCharacterSet().size() >= 2)
+                 .penalize("Character confict", HardMediumSoftScore.ONE_HARD, sc -> 1 - sc.getCharacterSet().size());
```
triceo: This is where the score corruption is going to be. DRL does this:

    scoreHolder.addHardConstraintMatch(kcontext, 1 - $size);

So the number will be 0 or less. But that is because in DRL, you need to input *a negative number* to penalize.
In CS, you have the `penalize()` call, and that expects *a positive number*. 

Therefore I think that if you invert the logic of the match weigher here, this score corruption will go away.

Christopher-Chianelli: Thanks for that; (although I still find it odd FAST_ASSERT didn't catch this) Suggestion for Constrant Stream: output a warning to the console whenever it see a penalize/reward with a negative number (if possible)

Christopher-Chianelli: There was another one: `maximizeMergesPerWord` was suppose to be a reward, not a penalty. But I got it working in FULL_ASSERT mode now.

triceo: Please make it run like that for a couple minutes, sometimes the corruptions are hiding in less likely corners of the state space.

Wrt. the warning - I'll think about it. We generally don't do warnings in Optaplanner, but I'll consider if fail-fast would be a good idea here.

Christopher-Chianelli: Error is probably better than a warning so users will immediately know something wrong instead of having to look at the logs to see why it not working correctly. Although that raises the question if someone has a constraint that can be positive or negative (Drools can do that since its reward function is also its penalize function).

Christopher-Chianelli: I ran it for 30 minutes; no Score Corruption was detected on FULL_ASSERT (manually terminated on Local Search step 1677)

optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private Constraint noParallelVerticalNeighbours(ConstraintFactory cf) {
+        return cf.from(ScrabbleCell.class).filter(sc -> sc.hasWordSet(ScrabbleWordDirection.VERTICAL))
+                 .join(ScrabbleCell.class,
+                       Joiners.equal(ScrabbleCell::getY), Joiners.lessThan(ScrabbleCell::getId),
+                       Joiners.filtering((first, other) -> other.hasWordSet(ScrabbleWordDirection.VERTICAL)))
+                 .filter((c1, c2) -> Math.abs(c1.getX() - c2.getX()) == 1)
+                 .penalize("No parallel vertical neighbours", HardMediumSoftScore.ONE_HARD);
+    }
```
triceo: Looking at the code now, the `join().filter()` could be replace by `ifExists(...)` with the appropriate joiners. The performance of that would likely be the best - only not sure if it would match 1:1 with what the DRL does. If FULL_ASSERT passes, I say let's do it.

optaplanner-examples/src/main/resources/org/optaplanner/examples/scrabble/solver/scrabbleSolverConfig.xml
```diff
@@ -6,10 +6,15 @@
...
 
   <!-- Score configuration -->
   <scoreDirectorFactory>
-    <scoreDrl>org/optaplanner/examples/scrabble/solver/scrabbleScoreRules.drl</scoreDrl>
+    <constraintProviderClass>org.optaplanner.examples.scrabble.optional.score.ScrabbleConstraintProvider</constraintProviderClass>
     <initializingScoreTrend>ONLY_DOWN</initializingScoreTrend>
+    <assertionScoreDirectorFactory>
+      <scoreDrl>org/optaplanner/examples/scrabble/solver/scrabbleScoreRules.drl</scoreDrl>
+      <initializingScoreTrend>ONLY_DOWN</initializingScoreTrend>
+    </assertionScoreDirectorFactory>
```
triceo: When you're done with the constraint provider, please bring this in sync with the rest of the examples. (Only `scoreDrl` is enabled, the rest is commented out.)

optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
Christopher-Chianelli: I don't think they are the same. Consider this scenario:

```
WORD1
WORD2
WORD3
```

Where `id(WORD2) < id(WORD1) < id(WORD3)`.
Using the current method, the constraint matches are `[(WORD2, WORD1),(WORD2,WORD3)]`.
Using the `ifExists` method, the constraint matches are `[(WORD2,WORD1|WORD2)]` (there is only one constraint match, with the second element being either WORD1 or WORD2). This situation is extremely rare, and probably take a long time to be explored. I ran 20 minutes on FULL_ASSERT using `ifExists` and got no errors so far, but I still think it not one-to-one.

Christopher-Chianelli: If (`cell1.x < cell2.x -> cell1.id < cell2.id` and `cell1.y < cell2.y -> cell1.id < cell2.id`) or (`cell1.x > cell2.x -> cell1.id > cell2.id` and `cell1.y > cell2.y -> cell1.id > cell2.id`), this could never happen (it requires an id discontinuity)

Christopher-Chianelli: From the importer:
```
        private void readCellList() throws IOException {
            int gridWidth = readIntegerValue("gridWidth:");
            solution.setGridWidth(gridWidth);
            int gridHeight = readIntegerValue("gridHeight:");
            solution.setGridHeight(gridHeight);
            List<ScrabbleCell> cellList = new ArrayList<>(gridWidth * gridHeight);
            for (int x = 0; x < gridWidth; x++) {
                for (int y = 0; y < gridHeight; y++) {
                    ScrabbleCell cell = new ScrabbleCell();
                    cell.setId((long) (y * gridWidth + x));
                    cell.setX(x);
                    cell.setY(y);
                    cell.setWordSet(new LinkedHashSet<>());
                    cell.setCharacterCountMap(new LinkedHashMap<>());
                    cellList.add(cell);
                }
            }
            solution.setCellList(cellList);
        }
```
In particular: `cell.setId((long) (y * gridWidth + x))`
Due to this, the corner case will never be hit, and the constraints are "equivalent".
"Proof":
Assume, for contradiction, there exists cells c1, c2, c3 such that `c1.x = c2.x = c3.x` and `c1.hasWordSet(HORIZONTAL) and c2.hasWordSet(HORIZONTAL) and c3.hasWordSet(HORIZONTAL)`, with `c1.y < c2.y < c3.y`. From the id formula we can derive:
```
c1.id 
= c1.y * W + x
< c2.y * W + x
= c2.id
```
Thus `c1.id < c2.id`. Similarly we have:
```
c2.id
= c2.y * W + x
< c3.y * W + x
= c3.id
```
Thus `c2.id < c3.id` and we have `c1.id < c2.id < c3.id`. This make "double" matches impossible, as they can only occur if a cell is between two cells with an id less than both of them, which we proved is not possible. QED.

Christopher-Chianelli: Side note: This proof also proves the filter can be simplified to `c1.y == c2.y - 1`.

Christopher-Chianelli: @triceo if Id generating code won't change and no one using Scrabble in production (with their own id generating code that might not have this property that cause the constraints to be equivalent), then I am okay changing it to `ifExists`.

Christopher-Chianelli: Looking at the `Drools`, they indeed use `c1.x = c2.x and c2.y = c1.y + 1`, meaning we can safety change this to an `ifExists` and remove the id joiner (which turn out was only changing `Math.abs(c1.y - c2.y) = 1` to `c2.y = c1.y + 1`).

Christopher-Chianelli: (To bad there isn't a `Joiners.fixed((first,second) -> boolean)`, since the cells `x` and `y` don't ever change, meaning we already know what cell we need to check for a particular cell and don't need to check all the cells in the same column)

Christopher-Chianelli: (`fixed` is basically a filtering that is only run once)

Christopher-Chianelli: (wait, `equals` have an override that use two different functions; I can use that)

triceo: Well proven!

optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
+        return new Constraint[]{
+                                characterConflict(constraintFactory),
+                                noParallelHorizontalNeighbours(constraintFactory),
+                                noParallelVerticalNeighbours(constraintFactory),
+                                outOfGrid(constraintFactory),
+                                maximizeMergesPerWord(constraintFactory),
+                                pullToCenter(constraintFactory)
+        };
```
triceo: One last thing, @Christopher-Chianelli, and then I'll merge: this isn't the proper code style, we don't do alignment on brackets. Please check the rest of the file as well, it's not the only place I see it.

optaplanner-examples/src/main/resources/org/optaplanner/examples/scrabble/solver/scrabbleSolverConfig.xml
```diff
@@ -6,10 +6,15 @@
   <solutionClass>org.optaplanner.examples.scrabble.domain.ScrabbleSolution</solutionClass>
   <entityClass>org.optaplanner.examples.scrabble.domain.ScrabbleWordAssignment</entityClass>
   <entityClass>org.optaplanner.examples.scrabble.domain.ScrabbleCell</entityClass>
+  <environmentMode>FULL_ASSERT</environmentMode>
```
triceo: This needs to be commented out before merging.

triceo: Also, we usually keep it at the top of the file. (Check other examples for reference.)

optaplanner-examples/src/main/java/org/optaplanner/examples/scrabble/optional/score/ScrabbleConstraintProvider.java
Christopher-Chianelli: I use `KIE Java Conventions` in Eclipse. Fun fact: the IDEA `KIE Java Formatter` and Eclipse `KIE Java Formatter` disagrees in some places, and this is one of them. I remember I created a script once that ran Idea formatter for me from command line. Might be harder to do that now that I have IDEA in a flatpak....

triceo: Dammit. :-( We should somehow get them in sync. I'll bring it up.
In the meantime - any chance you could configure yours to be the same as in IDEA? As far as I know, most of the team uses IntelliJ.

Christopher-Chianelli: They are fundamentally incompatible (IDEA's have more features than Eclipse). I will try running the IDEA's formatter in a container and package it as a program.

Christopher-Chianelli: Done; managed to find out where `format.sh` is in the flatpak.

PR #681, opened 2020-02-11
[PLANNER-1851] Add MOXy for JAXB JSON binding in optaplanner-persistence-jaxb

cuijulian:
Merge with: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1185
Jira: https://issues.redhat.com/browse/PLANNER-1851

- Adds the MOXy JAXB provider dependency for JSON marshalling support.
- Adds JSON score marshalling tests

### DIFF
--- a/optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
+++ b/optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
@@ -55,21 +55,26 @@
       <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-core</artifactId>
+      <groupId>org.glassfish.jaxb</groupId>
+      <artifactId>jaxb-runtime</artifactId>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>jakarta.xml.bind</groupId>
+          <artifactId>jakarta.xml.bind-api</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-impl</artifactId>
+      <groupId>org.glassfish</groupId>
+      <artifactId>jakarta.json</artifactId>
+      <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>javax.activation</groupId>
-      <artifactId>activation</artifactId>
-    </dependency>
-    <!-- MOXy needed for JSON -->
+    <!-- MOXy needed as JAXB JSON provider -->
     <dependency>
       <groupId>org.eclipse.persistence</groupId>
       <artifactId>org.eclipse.persistence.moxy</artifactId>
+      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
+++ b/optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
@@ -66,6 +66,11 @@
       <groupId>javax.activation</groupId>
       <artifactId>activation</artifactId>
     </dependency>
+    <!-- MOXy needed for JSON -->
+    <dependency>
+      <groupId>org.eclipse.persistence</groupId>
+      <artifactId>org.eclipse.persistence.moxy</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
+++ b/optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,11 +19,14 @@ package org.optaplanner.persistence.jaxb.api.score;
 import java.io.Serializable;
 import java.io.StringReader;
 import java.io.StringWriter;
+
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Marshaller;
 import javax.xml.bind.Unmarshaller;
 
+import org.eclipse.persistence.jaxb.MarshallerProperties;
+import org.eclipse.persistence.jaxb.UnmarshallerProperties;
 import org.optaplanner.core.api.score.Score;
 
 import static org.junit.Assert.*;
@@ -35,6 +38,11 @@ public abstract class AbstractScoreJaxbXmlAdapterTest {
     // ************************************************************************
 
     protected <S extends Score, W extends TestScoreWrapper<S>> void assertSerializeAndDeserialize(S expectedScore, W input) {
+        assertSerializeAndDeserializeXML(expectedScore, input);
+        assertSerializeAndDeserializeJson(expectedScore, input);
+    }
+
+    protected <S extends Score, W extends TestScoreWrapper<S>> void assertSerializeAndDeserializeXML(S expectedScore, W input) {
         String xmlString;
         W output;
         try {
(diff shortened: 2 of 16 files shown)


optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
```diff
@@ -63,8 +63,21 @@
...
+      <artifactId>jakarta.json-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>jakarta.json</artifactId>
+    </dependency>
+    <!-- MOXy needed for JSON -->
+    <dependency>
+      <groupId>org.eclipse.persistence</groupId>
+      <artifactId>org.eclipse.persistence.moxy</artifactId>
```
triceo: I do not understand the need for this. We're a JEE company, we have our own JSON/JAXB stuff. :-) 
See RESTEasy Jackson 2 Provider, which should also come with Quarkus by default.

cuijulian: The purpose of this PR is to add JSON marshalling support for JAXB. Jackson already gives us XML and JSON support, but we're trying to give users the choice of using JAXB as well.

triceo: RESTEasy can be used for that as well, [see the docs](https://docs.jboss.org/resteasy/docs/4.4.2.Final/userguide/html_single/index.html#JAXB_+_JSON_provider). I know, because I use it for the exact same purpose in my own projects.

Either that, or I don't understand how your use case differs. I've been wrong before.

cuijulian: From my understanding, this `optaplanner-persistence-jaxb` module is responsible for implementing JSON marshalling using the JAXB standard, so I'm looking for a JAXB provider (not Jackson, a module for that already exists in `optaplanner-persistence-jackson`) that supports JSON marshalling (MOXy is the provider used in this PR).

RESTEasy is a JAX-RS implementation. It does support JSON marshalling, but it does so by using its own JAXB provider. Since this module is for JAXB only, I don't think we should introduce RESTEasy dependencies into it. We just need a JAXB provider.

After some thought, I think I approached this the wrong way. @triceo I see what you're saying now, I could just use Quarkus' RESTEasy module to do this in the OptaPlanner extension, rather than adding MOXy to handle JAXB JSON.

triceo: Yes, for the extension, I definitely think that reusing what is already there in Quarkus is the way to go. 

FYI, plain Jackson can be used with JAXB as well, without the use of RESTEasy:
https://github.com/FasterXML/jackson-modules-base/tree/master/jaxb

cuijulian: Yup, that does it for Quarkus.

Hmm, this JAXB Jackson dependency looks interesting, it seems like an alternative to adding JAXB annotated JSON support in `optaplanner-spring-boot-autoconfigure`. Not sure if it belongs in `optaplanner-persistence-jaxb`, since we already have a module for Jackson. @ge0ffrey thoughts on these alternatives?

optaplanner-persistence/optaplanner-persistence-jaxb/pom.xml
```diff
@@ -63,8 +63,17 @@
...
     </dependency>
     <dependency>
-      <groupId>javax.activation</groupId>
-      <artifactId>activation</artifactId>
+      <groupId>jakarta.activation</groupId>
+      <artifactId>jakarta.activation-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>jakarta.json</artifactId>
```
ge0ffrey: Can we make these new dependencies test dependencies?
In fact, some of the old dependencies should probably be test dependencies.
We want to provide XmlAdaptors for score, implemented against the JAXB API.
We don't want to choose a JAXB implementation for the users, only for our testing.

ge0ffrey: And we certainly don't want to bloat the classpath.
Having both com.sun.xml.bind and org.glassfish and org.eclipse.persistence dependencies, is suspicious.
A lot of com.sun things became org.glassfish things at some point. Nowadays everything should be in jakarta.
To settle these questions, look at the master of the resteasy project. Or the quarkus-resteasy-jaxb project. What is in their classpath?

cuijulian:
>     * the more is test scope, the better
> 
>     * use the canonical jars that resteasy does?

@ge0ffrey I got rid of the `com.sun.xml.bind` dependencies and added the `jaxb-runtime` dependency (which `quarkus-jaxb` uses) to provide the necessary Jakarta spec jars. Everything except the `jboss-jaxb-api_2.3_spec` jar is test scoped, which should clean up our classpath a little.

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/json/AbstractScoreJaxbJsonAdapterTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.junit.Assert.fail;
+
+public abstract class AbstractScoreJaxbJsonAdapterTest {
+
+    // ************************************************************************
+    // Helper methods
+    // ************************************************************************
+
+    protected <S extends Score, W extends TestScoreWrapper<S>> void assertSerializeAndDeserialize(S expectedScore, W input) {
+        System.setProperty("javax.xml.bind.context.factory", "org.eclipse.persistence.jaxb.JAXBContextFactory");
```
ge0ffrey: This is dangerous, especially because it doesn't unset it.
The order in which the unit tests are run can break other unit tests that do XML checking.

ge0ffrey: Later it sets
jaxbMarshaller.setProperty(MarshallerProperties.MEDIA_TYPE, "application/json");

So is that system property still needed?

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/json/buildin/bendablebigdecimal/BendableBigDecimalScoreJaxbJsonAdapterTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+
+import org.junit.Test;
+import org.optaplanner.core.api.score.buildin.bendablebigdecimal.BendableBigDecimalScore;
+import org.optaplanner.persistence.jaxb.api.score.buildin.bendablebigdecimal.BendableBigDecimalScoreJaxbXmlAdapter;
+import org.optaplanner.persistence.jaxb.api.score.json.AbstractScoreJaxbJsonAdapterTest;
+
+public class BendableBigDecimalScoreJaxbJsonAdapterTest extends AbstractScoreJaxbJsonAdapterTest {
```
ge0ffrey: The other test, `SimpleScoreXMLJsonAdapterTest` is called like that because it is `SimpleScoreXMLJsonAdapter` + `Test`.

This json stuff, is also a test for the `SimpleScoreXMLJsonAdapter`, so it too, can go into the `SimpleScoreXMLJsonAdapterTest`... (yes, sounds weird). And it can reuse the wrapper.

ge0ffrey: Doing that, will heavily reduce the number of code lines, making the code easier to maintain.

"Good programmers add code to add features. Great programmers remove code (without removing features)." (there's a better version of this quoth somewhere).

I wouldn't even mind that the we just adjust the superclass's serializeAndDeserialize() to call serializeAndDeserializeXML() and serializeAndDeserializeJson(), avoiding any changes to specific score tests

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/json/AbstractScoreJaxbJsonAdapterTest.java
cuijulian: Yes, unfortunately the system property is still needed to use MOXy as the JAXB provider according to the [MOXy docs](https://www.eclipse.org/eclipselink/documentation/2.7/concepts/blocks003.htm#CHDBGGJD). But I see what you're saying about affecting other unit tests when a specific order is chosen.

A quick fix would be to set the system property in the `@Before` method of the MOXy test classes. Maybe there's a better alternative.

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/json/buildin/bendablebigdecimal/BendableBigDecimalScoreJaxbJsonAdapterTest.java
cuijulian: Point noted; good quote! My thought process about the `SimpleScoreJaxbXMLAdapterTest` was that it tests the `SimpleScoreJaxbXMLAdapter`, so the test is specific to JAXB standard XML. I would argue that the JSON tests I added is a test for the `SimpleScoreJaxbJsonAdapter`, so the `SimpleScoreJaxbJsonAdapterTest` is responsible for JAXB standard JSON (not XML). That's why I put them in separate classes.

But I absolutely agree with you that there's a lot of code duplication here. The problem is that the `Marshaller` created for XML serialization is different from the `Marshaller` created for JSON serialization (because of the MEDIATYPE property that JSON needs). I like your `serializeAndDeserializeXML()` and `serializeAndDeserializeJson()` separation idea, I should be able to use this to avoid duplicating all the test classes.

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/buildin/bendable/BendableScoreJaxbXmlJsonAdapterTest.java
```diff
@@ -21,17 +21,30 @@
 
 import org.junit.Test;
 import org.optaplanner.core.api.score.buildin.bendable.BendableScore;
-import org.optaplanner.persistence.jaxb.api.score.AbstractScoreJaxbXmlAdapterTest;
+import org.optaplanner.persistence.jaxb.api.score.AbstractScoreJaxbXmlJsonAdapterTest;
 
-public class BendableScoreJaxbXmlAdapterTest extends AbstractScoreJaxbXmlAdapterTest {
+public class BendableScoreJaxbXmlJsonAdapterTest extends AbstractScoreJaxbXmlJsonAdapterTest {
```
ge0ffrey: By default, a test's name should be <the class it tests> + "Test". By keeping this predictable it's easier to:
1) Find the test for FooBar (it will be called FooBarTest)
2) Avoid that there are 2 test classes for FooBar, which helps to avoid that there are duplicate tests (maintenance issues, less is more, if the coverage is the same)
This test tests BendableScoreJaxbXmlAdapter, so it's name must remain BendableScoreJaxbXmlAdapterTest

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/buildin/bendable/BendableScoreJaxbXmlJsonAdapterTest.java
```diff
@@ -21,17 +21,30 @@
...
+    public void serializeAndDeserializeJson() {
+        assertSerializeAndDeserializeJson(null, new TestBendableScoreWrapper(null));
+
+        BendableScore score = BendableScore.of(new int[]{1000, 200}, new int[]{34});
+        assertSerializeAndDeserializeJson(score, new TestBendableScoreWrapper(score));
+
         score = BendableScore.ofUninitialized(-7, new int[]{1000, 200}, new int[]{34});
-        assertSerializeAndDeserialize(score, new TestBendableScoreWrapper(score));
+        assertSerializeAndDeserializeJson(score, new TestBendableScoreWrapper(score));
     }
```
ge0ffrey: By splitting up serializeAndDeserialize() into serializeAndDeserializeXML() and serializeAndDeserializeJson() for every Test class, we're create a lot of duplication, for all of the 10+ score classes. Note how both methods are indentical except for the assertSerializeAndDeserialize*() call.

Let's be practical. Let's keep it as 1 method and keep using only assertSerializeAndDeserialize(), but let's have that - in the superclass - call assertSerializeAndDeserializeXML() and assertSerializeAndDeserializeJson().
That will heavily reduce maintenance once we mess with scores again (like the time I added initScore, all these tests had to be refactored).

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/buildin/bendable/BendableScoreJaxbXmlJsonAdapterTest.java
cuijulian: With the addition of `assertSerializeAndDeserializeJson()` in `AbstractScoreJaxbXmlJsonAdapterTest` the test subclasses now include JSON.

ge0ffrey: I still think the class name shouldn't change because it's a unit test and the unit test class hasn't changed. Let's discuss, I 'll send an invite.

ge0ffrey:
Thanks @cuijulian !

PR #682, opened 2020-02-12
PLANNER-1854 Fail fast on negative constraint match weight

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetAbstractConstraintStream.java
```diff
@@ -68,4 +69,12 @@ public BavetAbstractConstraintStream(BavetConstraintFactory<Solution_> constrain
         return constraintFactory;
     }
 
+    protected static void assertPositiveImpact(Constraint constraint, Number impact) {
+        if (impact.doubleValue() < 0) {
```
ge0ffrey: -1 Avoid number.doubleValue(), intValue(), ... like the plague. Why?
- rounding errors, try `(double) 9007199254740993L`, see https://ge0ffrey.github.io/ge0ffrey-presentations/cornerCaseCheatSheet/cheatSheetJava.html
- doubleValue() on some large or extremely accurate BigDecimals are questionable

Also, keep the operands of the same type, to avoid java-puzzlers-bugs. So keep the second operand specifically a double: `if (impact.doubleValue() < 0.0) {`. (Same goes for longs: in LongScore etc, we use `0L` to send a zero long, not `0`.)

This implies we'll have an assertPositiveImpact(, int), assertPositiveImpact(, long), etc

triceo:
Merged, having addressed the comment.

PR #683, opened 2020-02-12
docs: Address constraint streams doc issues

michaltomco:
I found two minor issues while going through the documentation and amended them.


triceo:
Thank you, nice catch!

PR #684, opened 2020-02-12
PLANNER-1613 toList() and toSet() constraint collectors

triceo:
This is a simple "last resort" collector for when our other collectors are insufficient. People can always retrieve a collection of the facts, and perform whatever operation they want on them afterwards.

WIthout providing this functionality, people would have to resort to submitting feature requests and be blocked before we implement them.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -21,14 +21,21 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.time.Duration;
 import java.time.Period;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.function.BiFunction;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
+import java.util.function.IntFunction;
+import java.util.function.Supplier;
 import java.util.function.ToIntBiFunction;
 import java.util.function.ToIntFunction;
 import java.util.function.ToLongBiFunction;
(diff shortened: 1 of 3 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -554,4 +561,171 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
     private ConstraintCollectors() {
     }
 
+    // ************************************************************************
+    // toCollection
+    // ************************************************************************
+
+    /**
+     * Creates constraint collector that will return matched facts as a collection.
```
ge0ffrey: avoid future tense.
-> "that returns"

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -554,4 +561,171 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+     *
+     * @param collectionFunction never null, provides new instance of the target collection with given initial capacity
+     * @param <A> type of the matched fact, matching the type of elements in the resulting collection
+     * @param <Result> type of the target collection
+     * @return never null
+     */
+    public static <A, Result extends Collection<A>> UniConstraintCollector<A, ?, Result> toCollection(
```
ge0ffrey: Do we need toCollection? When in doubt leave it out. I 'd rather start with just toSet() and toList() until someone complains of the lack of toCollection(). Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -554,4 +561,171 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+            IntFunction<Result> collectionFunction) {
+        return toCollection(Function.identity(), collectionFunction);
+    }
+
+    public static <A> UniConstraintCollector<A, ?, Set<A>> toSet() {
+        return toSet(Function.identity());
+    }
+
+    public static <A> UniConstraintCollector<A, ?, List<A>> toList() {
```
ge0ffrey: javadocs :)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: I consider it useful for when you want a stable iteration order, so you request `toCollection(TreeSet::new)`.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -554,4 +561,171 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+                (resultContainer, a) -> {
+                    Mapped mapped = mappingFunction.apply(a);
+                    resultContainer.add(mapped);
+                    return () -> resultContainer.remove(mapped);
+                },
+                resultContainer -> finishToCollection(collectionFunction, resultContainer));
+    }
+
+    private static <Mapped, Container extends List<Mapped>, Result extends Collection<Mapped>> Result
+    finishToCollection(IntFunction<Result> collectionFunction, Container resultContainer) {
```
ge0ffrey: Thinking aloud: The method name doesn't indicate it's used as the undo operation.
How about calling it to undoToCollection?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: Yeah, the Javadoc is on `toCollection()` now.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: This is not the undo. This is the finisher, which takes the incremental container and turns it into a final immutable product.

ge0ffrey: Agreed, then the name is correct. I interpreted "finish" as last part that is common, not as finisher.

Maybe `toCollectionFinisher`? Although name is fine now.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
ge0ffrey: Add at least a "As defined by ..."

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
ge0ffrey: Agreed, or to pick your collection type. Keep it indeed.

This is the special case though. The toList() is the normal case. So toList() gets a full proper javadoc (they others can use "As defined by"). Wdyt?

ge0ffrey: So, we want `toCollection(constructor)`, but not `toCollection()`?

PR #685, opened 2020-02-12
Clean up the code coverage configuration

rsynek:
- getting rid of some obsolete pieces of configuration
- using the jacoco maven plugin to define the surefire arg line instead of preparing it manually


pom.xml
```diff
@@ -274,6 +241,36 @@
...
+              <goals>
+                <goal>prepare-agent</goal>
+              </goals>
+              <configuration>
+                <append>true</append>
+                <destFile>${jacoco.exec.file}</destFile>
+                <excludes>
+                  <exclude>*Lexer</exclude>
+                  <exclude>org.optaplanner.core.impl.testdata.domain.*</exclude>
+                  <exclude>*TestdataAbstractSolutionBasedSolution</exclude>
```
triceo: Aren't these `src/test` classes? Why would they be included in code coverage in the first place?

rsynek: Added comments explaining why these classes need to be excluded.

pom.xml
```diff
@@ -274,6 +241,40 @@
...
+          </executions>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <configuration>
+            <!-- Append jacoco.agent.argLine property populated by JaCoCo's prepare-agent goal. -->
+            <argLine>@{jacoco.agent.argLine}</argLine>
+          </configuration>
+        </plugin>
```
ge0ffrey: @rsynek Maybe we can keep this surefire plugin override in the run-code-coverage profile, so we don't need the new jacoco.agent.argLine property in <properties/>?

rsynek: Meanwhile, I ran into a reincarnation of [1], which makes it impossible to run/debug individual tests from IntelliJ Idea unless you change its maven settings to avoid including argLine to Surefire.

[1] https://youtrack.jetbrains.com/issue/IDEA-143187?_ga=2.118432521.1268541826.1581583492-754822794.1502997082

PR #686, opened 2020-02-12
[DROOLS-5030] improve support for different resource types in KieFile…

mariofusco:
…System


ge0ffrey:
@mariofusco Thanks for taking this in the optaplanner repo too :)

PR #688, opened 2020-02-14
Move surefire argLine configuration to the run-code-coverage profile

rsynek:


rsynek:
Even if the configuration is in a separate (inactive) profile, the issue with IntelliJ Idea persists.

PR #689, opened 2020-02-14
PLANNER-1194 Include License file in the distribution

yurloc:


optaplanner-distribution/src/main/assembly/assembly-optaplanner.xml
```diff
@@ -112,12 +113,9 @@
       <includes>
         <include>org.optaplanner:optaplanner-docs:zip</include>
       </includes>
-      <useProjectArtifact>false</useProjectArtifact>
       <outputDirectory>reference_manual</outputDirectory>
-      <!-- Workaround for https://jira.codehaus.org/browse/MASSEMBLY-557 -->
-      <fileMode>644</fileMode>
```
ge0ffrey: Are you sure that https://jira.codehaus.org/browse/MASSEMBLY-557 is fixed?
Ok, that's a dead link. But it's to make sure the manual gets actually copied.

ge0ffrey: Ok to remove it, this is very old, it's proably fixed and we're otherwise never going to remove it.
@yurloc We do need to check the zip for the next 3 released to confirm that the entire manual is in there.

yurloc:
I've added the current MASSEBMBLY-557 link to the description.

I've built the distribution module locally and checked the expected content is there.

PR #692, opened 2020-02-17
PLANNER-1850 Add min/max collectors for streams of higher cardinality

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -526,36 +526,382 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+     *     Further assume that Person compares by age.
+     *     The resulting collector returns 20, as that is the minimum age of all the elements.
+     *
+     * @param <A> type of the first matched fact
+     * @param <B> type of the second matched fact
+     * @param <Mapped> type of the result
+     * @param groupValueMapping never null, maps matched facts to the result type
+     * @return never null
+     */
+    public static <A, B, Mapped extends Comparable<Mapped>> BiConstraintCollector<A, B, ?, Mapped> min(
```
triceo: So, fun problem...

`ConstraintCollectors.min(BiFunction)`
`ConstraintCollectors.min(Comparator)`

These are ambiguous, for example:
`ConstraintCollectors.min((Integer a, Integer b) -> a + b);`
IDE doesn't know which one to pick.

What now?

- We can't remove the Comparator one, as that's old API.
- People could just cast to whichever they prefer, but that's ugly.

This is only going to be a problem with bi-stream. tri-stream and quad-stream are fine.
It's also only going to be a problem for `BiFunction<A, A, Integer>`, as that is ambiguous with `Comparator<A>`.

Considering all of that, I don't see it as a big enough problem to fix it.
Essentially the only solution is to rename `min(BiFunction)` to something like `minBi(...)`, and that's too draconian.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -526,36 +526,380 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
     // min
     // ************************************************************************
 
+    /**
+     * Returns a collector that finds a minimum value in a group of {@link Comparable} elements.
+     * <p>
+     *     It is expected that the {@link Comparable}'s {@link Comparable#compareTo(Object)} is consistent with
```
ge0ffrey: Consistency: other javadocs don't ident after the `<p>`. Note that we also never do `</p>` (that didn't work well). All in all, I am happy to re-standarize all code on whatever openJDK is doing, but that's for a separate PR. Let's have this PR follow the rest of the code.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -526,36 +526,380 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+     *     {@link #equals(Object)}.
+     *     In other words, if two elements compare to equal, any of them may be returned by the collector as they are
```
ge0ffrey: "compare to equal" -> "compare to zero" (it's more specific)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -526,36 +526,380 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+     *     considered to be {@link #equals(Object)}.
```
ge0ffrey: I would change this paragraph into something like this:

```
Important: The {@link Comparable}'s {@link Comparable#compareTo(Object)} must be <i>consistent with equals</i>, such that <tt>e1.compareTo(e2) == 0</tt> has the same boolean value as <tt>e1.equals(e2)</tt>.
In other words, if two elements compare to zero, any of them can be returned by the collector. It can even differ between 2 score calculations on the exact same {@link PlanningSolution} state, due to incremental score calculation.
```

Motivation:
- start with "important" to stand out (we do this in a few other javadocs too).
-  Use "must" instead of "It is expected that ..." because that's more direct, stronger (which is appropriate here) and clearer.
- Mimic javadocs of Comparable on OpenJDK, use same terminology and phrasing.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -526,36 +526,380 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+     * <p>
+     *     Example: Assume the following elements of type Person: Ann(age = 20), Beth(age = 25), Cathy(age = 30),
+     *     David(age = 25), Eric(age = 20).
+     *     Further assume that the Person type compares by age.
+     *     Such collector returns either Ann or Eric as the minimum element.
```
ge0ffrey: Use `{@code Ann(age = 20)}` to format example input and output.
Wrap a collection of elements as input or output in `[]`.
Use formal (pseudo-)code-snippet `.groupBy(min(Person::getAge))` instead of assume that the Person type compares by age.
Find a shorter template for "input + constraintstream = output". The "Such collector" breaks this flow (too much ceremony).

ge0ffrey: Proposal:

```
For example, given `[Ann(age = 20), Beth(age = 25), Cathy(age = 30), David(age = 25), Eric(age = 20)]` with `.groupBy(min(Comparator.comparing(Person::getAge)))`, this returns `Ann` or `Eric` arbitrary.
```

ge0ffrey: If this template proposal works well, let's apply it to all of constraint streams for all example:

"For example, {@code <input>} with {@code <cs snippet>} returns {@code <output>}."

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -526,36 +526,380 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
+     *     Using Person::getAge as the groupValueMapping, the resulting collector returns 20, as that is the minimum
+     *     age of all the elements.
+     *
+     * @param <A> type of the matched fact
+     * @param <Mapped> type of the result
+     * @param groupValueMapping never null, maps facts from the matched type to the result type
+     * @return never null
+     */
+    public static <A, Mapped extends Comparable<Mapped>> UniConstraintCollector<A, ?, Mapped> min(
+            Function<A, Mapped> groupValueMapping) {
```
ge0ffrey: Soft suggestion: I'd order this method (min(Function)) before min(Comparator), because we want to stimulate users to use this one (I know, it doesn't change the order in javadocs or code completion, but people do look at the sources too and for us it's a signal too that we consider this one more important).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
ge0ffrey: Also,
in ConstraintStream.join(Class) we have a hint that they probably want to use join(Class, Joiners) instead if they can. Let's do something similar here, that links to min(Function).

ge0ffrey: And before the hint, let's recommend the good practice of ending with an identity comparison, by adding a 2th example with `.groupBy(min(Comparator.comparing(Person::getAge).comparing(Person::getId)))`

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
ge0ffrey: -1 to minBi. (proposal C) The comparator one takes a back seat to the function one, not the other way around.
I am ok with casting for now. (proposal B)

Proposal A2: Deprecate the Comparator ones, add `minWithComparator()` and `maxWithComparator()`. Then we ditch the comparator for 8.0 (whenever that may be, but it could be sooner than later due to kogito), adding it in https://issues.redhat.com/browse/PLANNER-750 so we don't forget.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -1381,6 +1598,224 @@ public void maxNotComparable() {
         assertResult(collector, container, null);
     }
 
+    @Test
+    public void maxComparableMapped() {
+        UniConstraintCollector<Class, ?, String> collector = max(Class::getCanonicalName);
+        Object container = collector.supplier().get();
+        // add first value, which becomes the max
+        Class firstValue = ConstraintCollectorsTest.class;
```
ge0ffrey: A Class as a firstValue?
Wouldn't it make more sense to either use:
Object o1 = new Object();
or use mockito's "mock()").

Or does it need to be a class instance? In that case, let's call it firstClass?

triceo: The reason I use class is that it doesn't implement `Comparable`, yet it has an easily accessible`String` field that does, so it's easy to write a `Comparator` for it.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -1324,13 +1325,229 @@ public void minNotComparable() {
         assertResult(collector, container, null);
     }
 
+    @Test
+    public void minComparableMapped() {
```
ge0ffrey: This test smells like its "abstracted". (I am not saying it is an abstracted test)
I personally favor concrete tests a lot of abstracted tests.

For example:

```
// Concrete test
assertEquals(7, calculator.sum(2, 5));
assertEquals(7, calculator.sum(4, 3));
assertEquals(9, calculator.sum(4, 5));

// Abstracted test
for (i = 0; i < 7; i++) {
  for (j = 0; j < 7; i++) {
   assertEquals(i + j, calculator.sum(i, j));
  }
}
```

Abstracted tests do more tests, but often don't increase test coverage. They make it hard to read what happens in corner cases - which can lead to invalid tests. And they often copy-paste the mistakes from the unit into the unit test code (especially corner cases).

PR #695, opened 2020-02-18
PLANNER-1836 SolverManager.terminateEarly() must cancel an un-started SolverJob

ge0ffrey:


optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -238,110 +237,70 @@ public void skipAhead() throws ExecutionException, InterruptedException {
...
-    /**
-     * In order to effectively test the terminateEarly() and close() methods there had to be a way how to make the job
-     * status change in a deterministic way while solving jobs on separate threads. The test waits for a status change in
-     * a loop and time-outs when it doesn't reach it's expected status.
-     */
-    @Ignore("https://issues.redhat.com/browse/PLANNER-1836")
-    @Test
-    public void terminateEarlyAndClose() throws InterruptedException {
+    @Test(timeout = 600_000L)
+    public void terminateEarly() throws InterruptedException {
```
ge0ffrey: @michaltomco An attempt at trying to explain how I changed this test:
- The original test uses waitForJobStatusChangeOrTimeout() - but these are race condition sensitive. For example, presume we wait for SOLVING_ACTIVE. In theory, it's possible that the solver runs so quick, that between polling it goes from SOLVING_SCHEDULED to SOLVING_ACTIVE to NOT_SOLVING.
- Furthermore - the TimeUnit.MILLISECONDS.sleep(5) could be both too quick (see "live lock" in Brian Goetz's book) and too slow (increases total unit testing time, which is a cost for CI and all developers for every commit). Whenever possible, avoiding it is nice.

**The latches approach follows a differently philosophy.** Instead of trying to "poll" at the right time (retrying if needed) to confirm the sequence of events happens in the order we except them to happen, we pause the sequence of events with latches, at which point we can guarantee the solving status.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -109,7 +116,20 @@ public Solution_ call() {
...
+            solverStatus = SolverStatus.NOT_SOLVING;
+        } else {
+            // The solver is either actively solving or has already terminated
+            solver.terminateEarly();
+            try {
+                // Don't return until bestSolutionConsumer won't be called any more
+                terminatedLatch.await();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                logger.warn("The terminateEarly() call is interrupted.", e);
```
ge0ffrey: Yes, I used a warn here. Usually, that's not a good idea.
Motivation:
A) When things go wrong during a normal shutdown, I want it to stick out, so we can figure out what's going on. Doing logger.info or lower means we'll never know.
B) Fail fasting by throwing an exception, might cause more havoc than it's worth? During shutdown, it's fairly normal that ALL threads get interrupted. I 'd presume that the "shutdown thread" isn't getting interrupted (that would be stupid) and that that one will call terminateEarly() (actually, it will call solverManager.shutdown(), so this discussion is void!).

I am fairly convinced of not doing A) (an info or lower log). I am less convinced for B)
 Wdyt? Should we fail fast instead?

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
michaltomco: Thanks for the explanation Geoffrey, I agree it is safer this way after reading your arguments.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -49,11 +48,11 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-import static org.optaplanner.core.impl.testdata.util.PlannerAssert.assertSolutionInitialized;
+import static org.junit.Assert.*;
```
michaltomco: Wouldn't it be better to import only the asserts used instead of using `org.junit.Assert.*`?

ge0ffrey: We more-or-less have an exception for static imports for Assert and Mockito.
I'd love for us to just follow the quarkus code style and enforce it in the build just like they do.

But trying to change or synchronize **Code Style is a religious discussion**. Such discussions always spiral out of control, take a lot of energy and kill productivity.
So lets not discuss whether or not star static imports are a good thing. Let's discuss whether we "just want to follow quarkus's style" (or spring's) and be done with it. And when we want to investigate that effort.

PR #697, opened 2020-02-19
PLANNER-1864 Introduce neutral impact() scoring to Constraint Streams

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Constraint.java
```diff
@@ -64,4 +64,10 @@ default String getConstraintId() {
         return ConstraintMatchTotal.composeConstraintId(getConstraintPackage(), getConstraintName());
     }
 
+    /**
+     * Whether this constraint penalizes, rewards or mixes both.
+     * @return never null
+     */
+    ScoreImpactType getScoreImpactType();
```
triceo: FYI Addition to public API.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraint.java
```diff
@@ -21,42 +21,27 @@
...
 import org.optaplanner.core.api.score.Score;
-import org.optaplanner.core.api.score.stream.Constraint;
+import org.optaplanner.core.api.score.stream.ScoreImpactType;
 import org.optaplanner.core.impl.score.stream.bavet.common.BavetNodeBuildPolicy;
 import org.optaplanner.core.impl.score.stream.bavet.uni.BavetFromUniConstraintStream;
 import org.optaplanner.core.impl.score.stream.bavet.uni.BavetFromUniNode;
+import org.optaplanner.core.impl.score.stream.common.AbstractConstraint;
 
-public final class BavetConstraint<Solution_> implements Constraint {
+public final class BavetConstraint<Solution_> extends AbstractConstraint<Solution_, BavetConstraintFactory<Solution_>> {
```
triceo: Both `BavetConstraint` and `DroolsConstraint` get a shared superclass, to get some code reuse in there.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Constraint.java
ge0ffrey: Who will use this?
- If it's an end-user, what's the use case.
- If it's for us, to do InitializedTrend magic later, I 'd rather not expose it in the public API (especially because there's no POC of the InitializedTrend magic (for which there is no bandwidth now)).

In the second case, we can put it in an AbstractConstraint or InnerConstraint class (a new class between Constraint and Drools/BavetConstraint). Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintStream.java
```diff
@@ -124,7 +124,7 @@ default Constraint penalize(String constraintName, Score<?> constraintWeight) {
      * <p>
      * The constraintWeight comes from an {@link ConstraintWeight} annotated member on the {@link ConstraintConfiguration},
      * so end users can change the constraint weights dynamically.
-     * This constraint is deactivated if the {@link ConstraintWeight} is zero.
+     * This constraint may be deactivated if the {@link ConstraintWeight} is zero.
```
ge0ffrey: This is changing the javadocs to the implementation reality (of CS-D), today.

We should investigate the work for the reverse: change the implementation reality to adhere to the javadocs :)
In theory, it's matter of deactiving rules if the constraintWeight is zero. We should reopen that discussion with Mario for `Rule.setEnabled(false)`.

ge0ffrey: This is changing the javadocs to the implementation reality (of CS-D), today.

We should investigate the work for the reverse: change the implementation reality to adhere to the javadocs :)
In theory, it's matter of deactiving rules if the constraintWeight is zero. We should reopen that discussion with Mario for `Rule.setEnabled(false)`.

triceo: I'm OK with talking to Mario. Yet, the Javadocs - in reality - are lying to users now. :-)

triceo: There is another option. We could rebuild the KieBase every time we create a new score director.
Since there isn't any compilation (we're doing straight executable model), it shouldn't be too slow.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -947,4 +947,163 @@ default Constraint rewardConfigurableBigDecimal(String constraintName, Function<
...
+     * @param constraintWeight never null
+     * @param matchWeigher never null
+     * @return never null
+     */
+    Constraint impactBigDecimal(String constraintPackage, String constraintName, Score<?> constraintWeight,
+            Function<A, BigDecimal> matchWeigher);
+
+    /**
+     * Positively or negatively impact the {@link Score} by the {@link ConstraintWeight} multiplied by the match weight.
+     * Otherwise as defined by {@link #impactConfigurable(String)}.
```
rsynek: This link (and other links to `impactConfigurable` below) is broken; I haven't found such a method.

The same for Bi,Tri and QuadConstraintStream classes.

triceo: Thanks, good catch.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ScoreImpactType.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.api.score.stream;
+
+public enum ScoreImpactType {
```
ge0ffrey: api -> impl

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/quad/QuadConstraintStream.java
```diff
@@ -624,4 +625,184 @@ default Constraint rewardConfigurableBigDecimal(String constraintName,
...
+     * <p>
+     * The constraintWeight comes from an {@link ConstraintWeight} annotated member on the
+     * {@link ConstraintConfiguration}, so end users can change the constraint weights dynamically.
+     * This constraint may be deactivated if the {@link ConstraintWeight} is zero.
+     * If there is no {@link ConstraintConfiguration}, use {@link #impact(String, Score)} instead.
+     * <p>
+     * The {@link Constraint#getConstraintPackage()} defaults to {@link ConstraintConfiguration#constraintPackage()}.
+     * <p>
+     * Use {@code penalizeConfigurableLong(...)} or {@code rewardConfigurableLong(...)} unless you intend to mix
+     * positive and negative weights.
```
ge0ffrey: Soft suggestion:
I'd use the word "instead" in these sentences for clarity.
Also "you intend to mix positive and negative weights" could be interpreted in general, not just for this constraint. We really want to discourage them using "impact".

Suggestion:
 "... unless you intend to mix positive and negative weights"
 -> "... instead, unless this constraint can both have positive and negative weights."

Also, I 'd always put it as the second paragraph, as I think this _nudging_ (yes that's what it is :)), is more important what getConstraintPackage() etc default to.

PR #699, opened 2020-02-20
PLANNER-1863: toMap() collector

triceo:
This, I think, completes our collection of collectors. I can't think of anything else we'd need out of the box.

Introduces six kinds of methods, the end result of which is:

1. `Map<X, Set<Y>>`.
1. `Map<X, C<Y>>`, where `C` is a `Set` provided by the user. (So that they can use `TreeSet`.)
1. `Map<X, Y>` with a reducer from `Set<Y>` to `<Y>`. 
1. `SortedMap<X, Set<Y>>`.
1. `SortedMap<X, C<Y>>`, where `C` is a `Set` provided by the user. (So that they can use `TreeSet`.)
1. `SortedMap<X, Y>` with a reducer from `Set<Y>` to `<Y>`. 

Also, for consistency, I'm adding similar `toSortedSet()` methods to accompany the current `toSet()` methods.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -73,9 +73,6 @@ public final class ConstraintCollectors {
     // count
     // ************************************************************************
 
-    private ConstraintCollectors() {
-    }
-
     public static <A> UniConstraintCollector<A, ?, Integer> count() {
         return new DefaultUniConstraintCollector<>(
                 () -> new int[1],
@@ -146,10 +143,6 @@ public final class ConstraintCollectors {
                 resultContainer -> resultContainer[0]);
     }
 
-    // ************************************************************************
-    // countDistinct
-    // ************************************************************************
-
     public static <A, B, C, D> QuadConstraintCollector<A, B, C, D, ?, Long> countLongQuad() {
         return new DefaultQuadConstraintCollector<>(
                 () -> new long[1],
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -68,6 +73,9 @@
     // count
     // ************************************************************************
 
+    private ConstraintCollectors() {
```
ge0ffrey: soft. Code style. Right now it's after the count comment header and before the method. Let's put this constructor after all static methods

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -138,6 +146,10 @@
                 resultContainer -> resultContainer[0]);
     }
 
+    // ************************************************************************
+    // countDistinct
+    // ************************************************************************
+
     public static <A, B, C, D> QuadConstraintCollector<A, B, C, D, ?, Long> countLongQuad() {
```
ge0ffrey: countDistinct header above countLongQuad? That latter isn't distinct?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -304,6 +307,10 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
                 resultContainer -> resultContainer[0]);
     }
 
+    // ************************************************************************
+    // sum
+    // ************************************************************************
```
ge0ffrey: sum header below sum method?

PR #702, opened 2020-02-24
PLANNER-1849 Constraint Stream support for regrouping

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -196,7 +198,42 @@
      * {@link TriJoiner joiners} are true
      */
     default <C> TriConstraintStream<A, B, C> join(Class<C> otherClass, TriJoiner<A, B, C>... joiners) {
```
triceo: **I do not know how we missed this**, but until now, filtering joiners were not supported for tri/quad joins. (The API allowed it, but threw a runtime exception.)

This PR added test coverage that identified the problem, and so I also add the fix. The fix is to bring the code to function equivalence with bi joins, which always supported filtering joiners.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintStream.java
```diff
@@ -197,7 +199,42 @@
...
+            return joined;
+        }
+        // We merge all filters into one, so that we don't pay the penalty for lack of indexing more than once.
+        FilteringQuadJoiner<A, B, C, D> filteringJoiner = (FilteringQuadJoiner<A, B, C, D>) joiners[indexOfFirstFilter];
+        QuadPredicate<A, B, C, D> resultingFilter = filteringJoiner.getFilter();
+        for (int i = indexOfFirstFilter + 1; i < joinerCount; i++) {
+            FilteringQuadJoiner<A, B, C, D> anoterFilteringJoiner = (FilteringQuadJoiner<A, B, C, D>) joiners[i];
+            resultingFilter = resultingFilter.and(anoterFilteringJoiner.getFilter());
+        }
+        return joined.filter(resultingFilter);
```
triceo: Ditto.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
ge0ffrey: Would they Sonar test coverage report allowed us to detect this test coverage gap? If so it might be worth checking if all api classes have any uncovered lines that matter enough to warren test coverage.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AdvancedGroupByConstraintStreamTest.java
```diff
@@ -251,4 +341,21 @@ public void groupByAfterExists() {
                 assertMatchWithScore(-2, solution.getFirstEntityGroup(), 2));
     }
 
+    private static <X> Set<X> asSet(X... x) {
```
ge0ffrey: These 3 methods belong in PlannerTestUtils

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/AdvancedGroupByConstraintStreamTest.java
```diff
@@ -71,80 +86,155 @@ public void collectedAndFiltered() {
...
+        solution.getEntityList().remove(entity);
+        scoreDirector.afterEntityRemoved(entity);
+        assertScore(scoreDirector,
+                assertMatchWithScore(-1,
+                        asMap(solution.getEntityGroupList().get(1), 12)));
     }
 
     @Test
-    @Ignore("Regrouping not yet supported.")
+    @Ignore
```
ge0ffrey: Still ignored? If so, why?

triceo: The PR is not yet complete. This will be fixed soon.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/ConstraintStreamHelper.java
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.score.stream.common;
+
+interface ConstraintStreamHelper<Right, JoinedStream, Joiner, Predicate> {
```
triceo: I expect a lively discussion about this. :-)

This exists so that I can reuse `JoinerUtils.join(...)` across all the streams. It may be a lot of code for such a simple purpose, but it's just a level of abstraction. The code that is being shared thanks to this is complicated and needs to be kept identical across all the streams, so I think it's worth it.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
triceo: Yeah, this would have been caught if we had that test. We do now.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/BiConstraintStreamDescriptor.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.function.BiPredicate;
+
+import org.optaplanner.core.api.score.stream.bi.BiConstraintStream;
+import org.optaplanner.core.api.score.stream.bi.BiJoiner;
+import org.optaplanner.core.api.score.stream.uni.UniConstraintStream;
+import org.optaplanner.core.impl.score.stream.bi.AbstractBiJoiner;
+import org.optaplanner.core.impl.score.stream.bi.FilteringBiJoiner;
+
+final class BiConstraintStreamDescriptor<A, B>
+        implements ConstraintStreamHelper<B, BiConstraintStream<A, B>, BiJoiner<A, B>, BiPredicate<A, B>> {
```
ge0ffrey: Need the same name, except for Bi

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/JoinerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        return join(new TriConstraintStreamDescriptor<>(stream), otherClass, joiners);
+    }
+
+    public static <A, B, C, D> QuadConstraintStream<A, B, C, D> join(TriConstraintStream<A, B, C> stream,
+            Class<D> otherClass, QuadJoiner<A, B, C, D>... joiners) {
+        return join(new QuadConstraintStreamDescriptor<>(stream), otherClass, joiners);
+    }
+
+    private static <Right, JoinedStream extends ConstraintStream, Joiner, Predicate> JoinedStream join(
+            ConstraintStreamHelper<Right, JoinedStream, Joiner, Predicate> constraintStreamDescriptor,
```
ge0ffrey: variable name

PR #704, opened 2020-02-25
PLANNER-1819 Add Examination CSProvider

michaltomco:
Add a constraint stream provider for the Examination example.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
@@ -1,0 +1,224 @@
+package org.optaplanner.examples.examination.solver.score;
+
+import static org.optaplanner.core.api.score.stream.Joiners.equal;
+import static org.optaplanner.core.api.score.stream.Joiners.filtering;
+import static org.optaplanner.core.api.score.stream.Joiners.greaterThan;
+
+import java.util.function.Function;
+
+import org.optaplanner.core.api.score.stream.Constraint;
+import org.optaplanner.core.api.score.stream.ConstraintCollectors;
+import org.optaplanner.core.api.score.stream.ConstraintFactory;
+import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.examples.examination.domain.Exam;
+import org.optaplanner.examples.examination.domain.ExaminationConstraintConfiguration;
+import org.optaplanner.examples.examination.domain.Period;
+import org.optaplanner.examples.examination.domain.PeriodPenalty;
+import org.optaplanner.examples.examination.domain.PeriodPenaltyType;
+import org.optaplanner.examples.examination.domain.Room;
+import org.optaplanner.examples.examination.domain.RoomPenalty;
+import org.optaplanner.examples.examination.domain.RoomPenaltyType;
+import org.optaplanner.examples.examination.domain.solver.TopicConflict;
+
+// TODO The ExaminationConstraintProvider is unusually slow. For more information, see: https://issues.redhat.com/browse/PLANNER-2011
+public class ExaminationConstraintProvider implements ConstraintProvider {
+
+    @Override
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
+        return new Constraint[] {
+                // Hard constraints
+                conflictingExamsInSamePeriod(constraintFactory),
+                periodDurationTooShort(constraintFactory),
+                roomCapacityTooSmall(constraintFactory),
+                periodPenaltyExamCoincidence(constraintFactory),
+                periodPenaltyExclusion(constraintFactory),
+                periodPenaltyAfter(constraintFactory),
+                roomPenaltyExclusive(constraintFactory),
+
+                // Soft constraints
+                twoExamsInARow(constraintFactory),
+                twoExamsInADay(constraintFactory),
+                periodSpread(constraintFactory),
+                mixedDurations(constraintFactory),
+                frontLoad(constraintFactory),
+                periodPenalty(constraintFactory),
+                roomPenalty(constraintFactory)
+        };
+    }
+
+    protected Constraint conflictingExamsInSamePeriod(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TopicConflict.class)
+                .join(Exam.class,
+                        equal(TopicConflict::getLeftTopic, Exam::getTopic),
+                        filtering((topicConflict, leftExam) -> leftExam.getPeriod() != null))
+                .ifExists(Exam.class,
+                        equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                        equal((topicConflict, leftExam) -> leftExam.getPeriod(), Exam::getPeriod))
+                .penalizeConfigurable("conflictingExamsInSamePeriod",
+                        (topicConflict, leftExam) -> topicConflict.getStudentSize());
+    }
+
+    protected Constraint periodDurationTooShort(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .filter(exam -> exam.getTopicDuration() > exam.getPeriodDuration())
+                .penalizeConfigurable("periodDurationTooShort", Exam::getTopicStudentSize);
+    }
+
+    protected Constraint roomCapacityTooSmall(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Room.class)
+                .join(Exam.class, equal(Function.identity(), Exam::getRoom))
+                .join(Period.class, equal((room, exam) -> exam.getPeriod(), Function.identity()))
+                .groupBy((room, exam, period) -> room,
+                        (room, exam, period) -> period,
+                        ConstraintCollectors.sum((period, exam, room) -> exam.getTopicStudentSize()))
+                .filter((room, period, totalStudentSize) -> totalStudentSize > room.getCapacity())
+                .penalizeConfigurable("roomCapacityTooSmall",
+                        (room, period, totalStudentSize) -> totalStudentSize - room.getCapacity());
+    }
+
+    protected Constraint periodPenaltyExamCoincidence(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PeriodPenalty.class)
+                .filter(periodPenalty -> periodPenalty.getPeriodPenaltyType() == PeriodPenaltyType.EXAM_COINCIDENCE)
+                .join(Exam.class,
+                        equal(PeriodPenalty::getLeftTopic, Exam::getTopic),
+                        filtering((periodPenalty, leftExam) -> leftExam.getPeriod() != null))
+                .join(Exam.class,
+                        equal((periodPenalty, leftExam) -> periodPenalty.getRightTopic(), Exam::getTopic),
+                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null),
+                        filtering((periodPenalty, leftExam, rightExam) -> leftExam.getPeriod() != rightExam.getPeriod()))
+                .penalizeConfigurable("periodPenaltyExamCoincidence",
+                        (periodPenalty, leftExam, rightExam) -> leftExam.getTopic().getStudentSize()
+                                + rightExam.getTopic().getStudentSize());
+    }
+
+    protected Constraint periodPenaltyExclusion(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PeriodPenalty.class)
+                .filter(periodPenalty -> periodPenalty.getPeriodPenaltyType() == PeriodPenaltyType.EXCLUSION)
+                .join(Exam.class,
+                        equal(PeriodPenalty::getLeftTopic, Exam::getTopic),
+                        filtering((periodPenalty, leftExam) -> leftExam.getPeriod() != null))
+                .join(Exam.class,
+                        equal((periodPenalty, leftExam) -> periodPenalty.getRightTopic(), Exam::getTopic),
+                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null),
+                        filtering((periodPenalty, leftExam, rightExam) -> leftExam.getPeriod().equals(rightExam.getPeriod())))
+                .penalizeConfigurable("periodPenaltyExclusion",
+                        (periodPenalty, leftExam, rightExam) -> leftExam.getTopic().getStudentSize()
+                                + rightExam.getTopic().getStudentSize());
+    }
+
+    protected Constraint periodPenaltyAfter(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PeriodPenalty.class)
+                .filter(periodPenalty -> periodPenalty.getPeriodPenaltyType() == PeriodPenaltyType.AFTER)
+                .join(Exam.class,
+                        equal(PeriodPenalty::getLeftTopic, Exam::getTopic),
+                        filtering((periodPenalty, leftExam) -> leftExam.getPeriod() != null))
+                .join(Exam.class,
+                        equal((periodPenalty, leftExam) -> periodPenalty.getRightTopic(), Exam::getTopic),
+                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null),
+                        filtering((periodPenalty, leftExam,
+                                rightExam) -> leftExam.getPeriod().getPeriodIndex() <= rightExam.getPeriod().getPeriodIndex()))
+                .penalizeConfigurable("periodPenaltyAfter",
+                        (periodPenalty, leftExam, rightExam) -> leftExam.getTopic().getStudentSize()
+                                + rightExam.getTopic().getStudentSize());
+    }
+
+    protected Constraint roomPenaltyExclusive(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(RoomPenalty.class)
+                .filter(roomPenalty -> roomPenalty.getRoomPenaltyType() == RoomPenaltyType.ROOM_EXCLUSIVE)
+                .join(Exam.class,
+                        equal(RoomPenalty::getTopic, Exam::getTopic),
+                        filtering((roomPenalty, leftExam) -> leftExam.getPeriod() != null && leftExam.getRoom() != null))
+                .join(Exam.class,
+                        equal((roomPenalty, leftExam) -> leftExam.getRoom(), Exam::getRoom),
+                        equal((roomPenalty, leftExam) -> leftExam.getPeriod(), Exam::getPeriod),
+                        filtering((roomPenalty, leftExam, rightExam) -> leftExam.getTopic() != rightExam.getTopic()))
+                .penalizeConfigurable("roomPenaltyExclusive",
+                        (periodPenalty, leftExam, rightExam) -> leftExam.getTopic().getStudentSize()
+                                + rightExam.getTopic().getStudentSize());
+    }
+
+    protected Constraint twoExamsInARow(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TopicConflict.class)
+                .join(Exam.class,
+                        equal(TopicConflict::getLeftTopic, Exam::getTopic),
+                        filtering((topicConflict, leftExam) -> leftExam.getPeriod() != null))
+                .ifExists(Exam.class,
+                        equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                        equal((topicConflict, leftExam) -> leftExam.getDayIndex(), Exam::getDayIndex),
+                        filtering((topicConflict, leftExam,
+                                rightExam) -> getPeriodIndexDifferenceBetweenExams(leftExam, rightExam) == 1))
+                .penalizeConfigurable("twoExamsInARow", (topicConflict, leftExam) -> topicConflict.getStudentSize());
+    }
+
+    protected Constraint twoExamsInADay(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TopicConflict.class)
+                .join(Exam.class,
+                        equal(TopicConflict::getLeftTopic, Exam::getTopic),
+                        filtering((topicConflict, leftExam) -> leftExam.getPeriod() != null))
+                .ifExists(Exam.class,
+                        equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                        equal((topicConflict, leftExam) -> leftExam.getDayIndex(), Exam::getDayIndex),
+                        // Find exams in a day, but not being held right after each other. That case is handled in the twoExamsInARow constraint.
+                        filtering((topicConflict, leftExam,
+                                rightExam) -> getPeriodIndexDifferenceBetweenExams(leftExam, rightExam) > 1))
+                .penalizeConfigurable("twoExamsInADay", (topicConflict, leftExam) -> topicConflict.getStudentSize());
+    }
+
+    protected Constraint periodSpread(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(ExaminationConstraintConfiguration.class)
+                .join(TopicConflict.class)
+                .join(Exam.class,
+                        equal((config, topicConflict) -> topicConflict.getLeftTopic(), Exam::getTopic),
+                        filtering((config, topicConflict, leftExam) -> leftExam.getPeriod() != null))
+                .ifExists(Exam.class,
+                        equal((config, topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                        filtering((config, topicConflict, leftExam, rightExam) -> rightExam.getPeriod() != null),
+                        filtering((config, topicConflict, leftExam,
+                                rightExam) -> getPeriodIndexDifferenceBetweenExams(leftExam,
+                                        rightExam) < (config.getPeriodSpreadLength() + 1)))
+                .penalizeConfigurable("periodSpread",
+                        (config, topicConflict, leftExam) -> topicConflict.getStudentSize());
+    }
+
+    protected Constraint mixedDurations(ConstraintFactory constraintFactory) {
+        // 4 mixed durations of 100, 150, 200 and 200 should only result in 2 penalties (for 100&150 and 100&200).
+        return constraintFactory.fromUniquePair(Exam.class,
+                equal(Exam::getPeriod),
+                equal(Exam::getRoom))
+                // Keep only those that have different topic durations in the same room-period pair.
+                .filter((leftExam, rightExam) -> leftExam.getTopicDuration() != rightExam.getTopicDuration())
+                // Keep only those where the left exam's id is the lowest one in the room-period pair.
+                .ifNotExists(Exam.class,
+                        equal((leftExam, rightExam) -> leftExam.getPeriod(), Exam::getPeriod),
+                        equal((leftExam, rightExam) -> leftExam.getRoom(), Exam::getRoom),
+                        greaterThan((leftExam, rightExam) -> leftExam.getId(), Exam::getId))
+                // Keep only those where the left exam's id is the lowest one in the room-period pair.
+                .ifNotExists(Exam.class, equal((leftExam, rightExam) -> leftExam.getPeriod(), Exam::getPeriod),
+                        equal((leftExam, rightExam) -> leftExam.getRoom(), Exam::getRoom),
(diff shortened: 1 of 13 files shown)


michaltomco:
The constraints at the moment are rather slow, not sure whether they are properly converted from the Drools language. I've used a lot of filtering after noticing it is not supported yet, so I've put .filter() instead.

A few questions:
- What is the ideal way how to model the != null? Two options come to mind, either use .ifNotExists() or .filter(!=null). What is the ideal one performance-wise?
- Regarding roomCapacityTooSmall() constraint, is the drools accumulate function properly converted? If not what is the ideal way to do this?

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+    private Constraint periodPenaltyExamCoincidence(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PeriodPenalty.class)
+                .filter(periodPenalty -> periodPenalty.getPeriodPenaltyType() == PeriodPenaltyType.EXAM_COINCIDENCE)
+                .join(Exam.class,
+                      equal(PeriodPenalty::getLeftTopic, Exam::getTopic))
+                .filter((periodPenalty, leftExam) -> leftExam.getPeriod() != null)
+                .join(Exam.class,
+                      equal((periodPenalty, exam) -> periodPenalty.getRightTopic(), Exam::getTopic),
+                      equal((periodPenalty, exam) -> exam.getPeriod(), Exam::getPeriod))
+                .filter((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null)
```
triceo: This is no longer necessary, right. `leftExam.getPeriod()` already isn't `null`, and `rightExam.getPeriod()` is joined where equal.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+    private Constraint periodPenaltyAfter(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PeriodPenalty.class)
+                .filter(periodPenalty -> periodPenalty.getPeriodPenaltyType() == PeriodPenaltyType.AFTER)
+                .join(Exam.class,
+                      equal(PeriodPenalty::getLeftTopic, Exam::getTopic))
+                .filter((periodPenalty, leftExam) -> leftExam.getPeriod() != null)
+                .join(Exam.class,
+                      equal((periodPenalty, leftExam) -> periodPenalty.getRightTopic(), Exam::getTopic),
+                      lessThanOrEqual((periodPenalty, leftExam) -> leftExam.getPeriodIndex(), Exam::getPeriodIndex))
+                .filter((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null)
```
triceo: Can  `getPeriod()` be `null` if `getPeriodIndex()` already worked fine?

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+    private Constraint roomPenaltyExclusive(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(RoomPenalty.class)
+                .filter(roomPenalty -> roomPenalty.getRoomPenaltyType() == RoomPenaltyType.ROOM_EXCLUSIVE)
+                .join(Exam.class,
+                      equal(RoomPenalty::getTopic, Exam::getTopic))
+                .filter((roomPenalty, leftExam) -> leftExam.getPeriod() != null && leftExam.getRoom() != null)
+                .join(Exam.class,
+                      equal((roomPenalty, leftExam) -> leftExam.getRoom(), Exam::getRoom),
+                      equal((roomPenalty, leftExam) -> leftExam.getPeriod(), Exam::getPeriod))
+                .filter((roomPenalty, leftExam, rightExam) -> leftExam.getTopic().equals(rightExam.getTopic()))
```
triceo: Can't you do this using `Joiners.equal()` within the preceding `join()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//    end
+    private Constraint periodSpread(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TopicConflict.class)
+                .join(Exam.class,
+                      equal(TopicConflict::getLeftTopic, Exam::getTopic))
+                .filter((topicConflict, leftExam) -> leftExam.getPeriod() != null)
+                .join(Exam.class,
+                      equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                      equal((topicConflict, leftExam) -> leftExam.getDayIndex(), Exam::getDayIndex))
+                .filter((topicConflict, leftExam, rightExam) -> Math.abs(leftExam.getPeriodIndex() - rightExam.getPeriodIndex()) > 1)
```
triceo: You use this `Math.abs(...) >/= 1` construct frequently. I'd extract it to a separate method and reuse.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+                                  lessThan(Exam::getId, Exam::getId))
+                .join(Exam.class,
+                      equal(Exam::getPeriod, Exam::getPeriod),
+                      equal(Exam::getRoom, Exam::getRoom),
+                      greaterThan(Exam::getId, Exam::getId))
+                .filter((leftExam, rightExam) -> rightExam.getTopicDuration() != leftExam.getTopicDuration())
+                .ifNotExists(Exam.class,
+                             equal((leftExam, rightExam) -> rightExam.getPeriod(), Exam::getPeriod),
+                             equal((leftExam, rightExam) -> rightExam.getRoom(), Exam::getRoom),
+                             greaterThan((leftExam, rightExam) -> rightExam.getId(), Exam::getId),
```
triceo: Personally, I like to see all `equals()` joiners first, but that's mostly aesthetic.

optaplanner-examples/src/main/resources/org/optaplanner/examples/examination/solver/examinationSolverConfig.xml
```diff
@@ -9,6 +9,7 @@
   <entityClass>org.optaplanner.examples.examination.domain.FollowingExam</entityClass>
 
   <scoreDirectorFactory>
+<!--    <constraintProviderClass>org.optaplanner.examples.examination.optional.score.examinationConstraintProvider</constraintProviderClass>-->
```
triceo: Formatting, please. See other examples.

optaplanner-examples/src/main/resources/org/optaplanner/examples/examination/solver/examinationSolverConfig.xml
```diff
@@ -9,6 +9,7 @@
...
     <scoreDrl>org/optaplanner/examples/examination/solver/examinationScoreRules.drl</scoreDrl>
```
triceo: Has this been through assertion score director and `FULL_ASSERT`? Since I don't see it commented out in the example, I assume not.

Do it. It will tell you whether or not your implementation is identical to DRL.

triceo:
> I've used a lot of filtering after noticing it is not supported yet, so I've put .filter() instead.

That problem should be gone on today's master.

> * What is the ideal way how to model the != null? Two options come to mind, either use .ifNotExists() or .filter(!=null). What is the ideal one performance-wise?

I made some suggestions inline. if/when applied, I don't think there will be many `null` checks left.

> * Regarding roomCapacityTooSmall() constraint, is the drools accumulate function properly converted? If not what is the ideal way to do this?

See my comment on `FULL_ASSERT`. That will give you a very exact answer.

> * Are there any obvious performance hindering problems that I am missing?

I made some suggestions. We'll see when they're solved.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//    $topicConflict : TopicConflict($leftTopic : leftTopic, $rightTopic : rightTopic)
+//    $leftExam : Exam(topic == $leftTopic, $period : period, period != null)
+//    $rightExam : Exam(topic == $rightTopic, period == $period)
+//    then
+//        scoreHolder.addHardConstraintMatch(kcontext, - $topicConflict.getStudentSize());
+//    end
+    private Constraint conflictingExamsInSamePeriod(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .filter(exam -> exam.getPeriod() != null)
+                .join(Exam.class, equal(Exam::getPeriod, Exam::getPeriod))
```
triceo: Also, these two joins can be replaced by `.fromUniquePair(Exam.class, equal(...))`

michaltomco:
Thanks a lot for the input, I will implement them first thing tomorrow!

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//    rule "periodDurationTooShort"
+//    when
+//    $exam : Exam(topicDuration > periodDuration, period != null)
+//    then
+//        scoreHolder.addHardConstraintMatch(kcontext, - $exam.getTopicStudentSize());
+//    end
+    private Constraint periodDurationTooShort(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .filter(exam -> exam.getTopicDuration() > exam.getPeriodDuration())
+                .ifExists(Period.class, equal(Exam::getPeriod, Function.identity()))
```
triceo: Most likelynot necessary, as the rule doesn't use it either.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//            Exam(period == $period, room == $room, $studentSize : topicStudentSize);
+//    $totalStudentSize : sum($studentSize);
+//    $totalStudentSize > $capacity
+//        )
+//    then
+//        scoreHolder.addHardConstraintMatch(kcontext, $capacity - $totalStudentSize);
+//    end
+    private Constraint roomCapacityTooSmall(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .join(Period.class, equal(Exam::getPeriod, Function.identity()))
```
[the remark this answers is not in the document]
triceo: Most likely it's a bug, though. The period is actually used in the DRL.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//    end
+    private Constraint twoExamsInADay(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TopicConflict.class)
+                .join(Exam.class,
+                      equal(TopicConflict::getLeftTopic, Exam::getTopic))
+                .filter((topicConflict, leftExam) -> leftExam.getPeriod() != null)
+                .join(Exam.class,
+                      equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                      equal((topicConflict, leftExam) -> leftExam.getDayIndex(), Exam::getDayIndex))
+                .filter((topicConflict, leftExam, rightExam) -> Math.abs(leftExam.getPeriodIndex() - rightExam.getPeriodIndex()) > 1)
```
triceo: Although this isn't functionally equivalent to the DRL, you could get away with replacing those joins with two `ifExists()`. Good for performance, but may not pass `FULL_ASSERT`.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
triceo: Also same comment regarding `ifExists()` as with `twoExamsInADay`.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//        scoreHolder.penalize(kcontext, $topicConflict.getStudentSize());
+//    end
+    private Constraint twoExamsInARow(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TopicConflict.class)
+                .join(Exam.class,
+                      equal(TopicConflict::getLeftTopic, Exam::getTopic))
+                .filter((topicConflict, leftExam) -> leftExam.getPeriod() != null)
+                .join(Exam.class,
+                      equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                      equal((topicConflict, leftExam) -> leftExam.getDayIndex(), Exam::getDayIndex))
```
triceo: Perhaps could be replaced by `ifExists()`, but then the constraint may no longer be functionally equivalent to the DRL. `FULL_ASSERT` would tell.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
triceo: Dtto. on `ifExists()`.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+//    // rightExam has the lowest id of the period+room+rightDuration combo
+//    not Exam(period == $period, room == $room, id < $rightId,
+//            topicDuration == $rightTopicDuration
+//    )
+//    then
+//        scoreHolder.penalize(kcontext);
+//    end
+    private Constraint mixedDurations(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .ifExists(Exam.class, equal(Exam::getPeriod, Exam::getPeriod))
```
triceo: Again, use the .join(Exam.class, equal(Exam::getPeriod)) shorthand.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+                .ifExists(Exam.class, equal(Exam::getRoom, Exam::getRoom))
```
triceo: Why is this here? I don't see it in the DRL.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+                .ifNotExistsOther(Exam.class,
+                                  equal(Exam::getPeriod, Exam::getPeriod),
+                                  equal(Exam::getRoom, Exam::getRoom),
+                                  lessThan(Exam::getId, Exam::getId))
+                .join(Exam.class,
+                      equal(Exam::getPeriod, Exam::getPeriod),
+                      equal(Exam::getRoom, Exam::getRoom),
```
triceo: Since both left and right are the same type and we're joining on the same field, `equal()` can lose its second argument.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+                      greaterThan(Exam::getId, Exam::getId))
```
triceo: Since both left and right are the same type and we're joining on the same field, `equal()` can lose its second argument.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,371 @@
...
+
+    //// Two exams in the same period which share students.
+//    rule "conflictingExamsInSamePeriod"
+//    when
+//    $topicConflict : TopicConflict($leftTopic : leftTopic, $rightTopic : rightTopic)
+//    $leftExam : Exam(topic == $leftTopic, $period : period, period != null)
+//    $rightExam : Exam(topic == $rightTopic, period == $period)
+//    then
+//        scoreHolder.addHardConstraintMatch(kcontext, - $topicConflict.getStudentSize());
+//    end
```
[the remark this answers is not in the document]
michaltomco: Yes, that was meant as a help for myself and potentially the reviewer. I will remove it after I fix all the issues.

michaltomco:
I still need to debug the CS, since it returns different score values than drools.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/optional/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,369 @@
...
+    private Constraint roomPenalty(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Room.class)
+                .filter(room -> room.getPenalty() != 0)
+                .ifExists(Exam.class, equal(Function.identity(), Exam::getRoom))
+                .penalize("Room penalty", HardSoftScore.ONE_SOFT, Room::getPenalty);
+    }
+
+    private int getAbsoluteValue(Exam leftExam, Exam rightExam) {
+        return Math.abs(leftExam.getPeriodIndex() - rightExam.getPeriodIndex());
+    }
```
triceo: Choose a better name for the method. Don't describe the code, describe the purpose.

michaltomco:
The performance is still very slow. I will let you know when I discover the culprit.

triceo:
> The performance is still very slow. I will let you know when I discover the culprit.

I suggest an experiment, @michaltomco : remove the one constraint where you use `groupBy()`. If the performance is magically fixed, don't look any further - you've hit the CS-D performance bug and the results are as bad as expected.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                .ifExists(Exam.class,
+                        equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                        equal((topicConflict, leftExam) -> leftExam.getDayIndex(), Exam::getDayIndex),
+                        // Find exams in a day, but not being held right after each other. That case is handled in the twoExamsInARow constraint.
+                        filtering((topicConflict, leftExam,
+                                rightExam) -> getPeriodIndexDifferenceBetweenExams(leftExam, rightExam) > 1))
+                .join(ExaminationConstraintConfiguration.class)
+                .penalize("Two exams in a day",
+                        HardSoftScore.ONE_SOFT,
+                        (topicConflict, leftExam, config) -> topicConflict.getStudentSize() * config.getTwoInADayPenalty());
```
triceo, line 173: I see that there is a `ConstraintConfiguration` available. Yet, you still use `penalize()` instead of `penalizeConfigurable()`. I suggest you change that.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                        (config, topicConflict, leftExam) -> topicConflict.getStudentSize());
+    }
+
+    protected Constraint mixedDurations(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .ifExists(Exam.class, equal(Exam::getPeriod, Exam::getPeriod))
+                .ifNotExistsOther(Exam.class,
+                        equal(Exam::getPeriod, Exam::getPeriod),
+                        equal(Exam::getRoom, Exam::getRoom),
+                        lessThan(Exam::getId, Exam::getId))
```
triceo, line 199: return constraintFactory.from(Exam.class)
                .ifNotExists(Exam.class,
                        equal(Exam::getPeriod),
                        equal(Exam::getRoom),
                        lessThan(Exam::getId))

But that won't change much, if anything.

triceo: Also, I'd put the `lessThan` filter first. Will eliminate many thing quickly.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                .join(Exam.class,
+                        equal(Exam::getPeriod),
+                        equal(Exam::getRoom),
+                        greaterThan(Exam::getId),
+                        filtering((leftExam, rightExam) -> rightExam.getTopicDuration() != leftExam.getTopicDuration()))
```
triceo, line 204: Also, I wonder why this entire construct isn't simply `.fromUniquePair(Exam.class, ...)`

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                .join(Exam.class,
+                        filtering((config, exam) -> exam.isTopicFrontLoadLarge() && exam.isPeriodFrontLoadLast()))
+                .penalize("Front load", HardSoftScore.ONE_SOFT, (config, exam) -> config.getFrontLoadPenalty());
+    }
+
+    protected Constraint periodPenalty(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .ifExists(Period.class,
+                        equal(Exam::getPeriod, Function.identity()),
+                        filtering((exam, period) -> period.getPenalty() != 0))
```
triceo, line 226: Why not `.filter(exam -> exam.getPeriod().getPenalty() != 0)`?

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                        equal((leftExam, rightExam) -> rightExam.getRoom(), Exam::getRoom),
+                        equal((leftExam, rightExam) -> rightExam.getTopicDuration(), Exam::getTopicDuration),
+                        greaterThan((leftExam, rightExam) -> rightExam.getId(), Exam::getId))
+                .join(ExaminationConstraintConfiguration.class)
+                .penalize("Mixed durations", HardSoftScore.ONE_SOFT,
+                        (leftExam, rightExam, config) -> config.getMixedDurationPenalty());
+    }
+
+    protected Constraint frontLoad(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(ExaminationConstraintConfiguration.class)
```
triceo, line 216: All uses of from/join on `ExaminationConstraintConfiguration` should theoretically be removable when converted to `penalizeConfigurable()`.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                        equal(Exam::getPeriod, Function.identity()),
+                        filtering((exam, period) -> period.getPenalty() != 0))
+                .penalize("Period penalty", HardSoftScore.ONE_SOFT, exam -> exam.getPeriod().getPenalty());
+    }
+
+    protected Constraint roomPenalty(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Exam.class)
+                .ifExists(Room.class,
+                        equal(Exam::getRoom, Function.identity()),
+                        filtering((exam, room) -> room.getPenalty() != 0))
```
triceo, line 234: Ditto. It appears you have a Room reference on `Exam` directly.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -0,0 +1,241 @@
...
+                        equal((topicConflict, leftExam) -> topicConflict.getRightTopic(), Exam::getTopic),
+                        equal((topicConflict, leftExam) -> leftExam.getPeriod(), Exam::getPeriod))
+                .penalize("Conflicting exams in same period",
+                        HardSoftScore.ONE_HARD,
+                        (topicConflict, leftExam) -> topicConflict.getStudentSize());
+    }
+
+    protected Constraint periodDurationTooShort(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(Exam.class)
+                .filter(exam -> exam.getPeriod() != null)
```
triceo, line 65: Why `fromUnfiltered()`, if you then filter anyway?

triceo:
Let's not leave the constraint configuration unfinished - there should be no `penalize()` method left.
Also, when you've got all the constraints in the configuration, please make sure that the DRL is using them, too.

michaltomco:
The performance issue causing constraints are definitely the twoExamsInARow, twoExamsInADay, and periodSpread. I'll try to rewrite them to see if it helps.

optaplanner-examples/src/main/resources/org/optaplanner/examples/examination/solver/examinationSolverConfig.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <solver>
-  <!--<environmentMode>FULL_ASSERT</environmentMode>--><!-- To slowly prove there are no bugs in this code -->
-  <moveThreadCount>AUTO</moveThreadCount><!-- To solve faster by saturating multiple CPU cores -->
+<!--  <environmentMode>FULL_ASSERT</environmentMode>&lt;!&ndash; To slowly prove there are no bugs in this code &ndash;&gt;-->
```
rsynek, line 3: It looks like accidentally wrapping one comment into another.

optaplanner-benchmark/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.optaplanner</groupId>
     <artifactId>optaplanner</artifactId>
-    <version>7.39.0-SNAPSHOT</version>
+    <version>7.39.0.CR1</version>
```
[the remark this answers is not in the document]
michaltomco: Oops, this commit must've sneaked there while rebasing. Fixed now!

optaplanner-examples/src/main/resources/org/optaplanner/examples/examination/solver/examinationSolverConfig.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <solver>
-  <!--<environmentMode>FULL_ASSERT</environmentMode>--><!-- To slowly prove there are no bugs in this code -->
-  <moveThreadCount>AUTO</moveThreadCount><!-- To solve faster by saturating multiple CPU cores -->
+<!--  <environmentMode>FULL_ASSERT</environmentMode> To slowly prove there are no bugs in this code -->
+  <!--<moveThreadCount>AUTO</moveThreadCount>--><!-- To solve faster by saturating multiple CPU cores -->
```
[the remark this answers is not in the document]
michaltomco: Done.

triceo:
Thanks, glad to see this merged!

PR #705, opened 2020-02-25
Move setBestSolutionTimeMillis inside "if (bestScore.isSolutionInitialized())"

Christopher-Chianelli:
This change causes the solver to terminates gracefully instead of crashing if a construction heuristic fails to initalize a solution within the Unimproved Spent Limit (if one is set).

Reason for crash before:

A construction heuristic does not update the best solution until it is finished. As such, even if the score is improving, the Unimproved Spent Limit Termination cannot tell until the Phase is finished (either when the solution is initalized or unimproved spent limit is passed). As a consequence, the timer get reset at the end of the Construction Heuristic phase (and not mid-way through it). Since the timer is reset, the Solver does not terminate and proceeds to the Local Search Phase (if one is configured). However, the Local Search Phase required the solution to be initalized, so a crash occurs if the Construction Heuristic could not initalize a solution within the unimproved spent limit.

For projects that use solverConfig's, this issue does not happen often since it is not recommended to use unimprovedSpentLimit on the entire solver. However, for spring-boot-autoconfigure (without solverConfig.xml), you cannot set unimprovedSpentLimit only on specific phases in application.properties, meaning you need to set it on the solver, which can lead to this crash.

Fix:

Only update best solution time if the solution is initalized

Alternatives:

- Redesign how Construction Heuristic update best solution
- Remove UnimprovedSpentLimit property from Spring Boot Autoconfigure


ge0ffrey:
I 'll look into this along with Christopher's chat log about this issue, hopefully later this week.

ge0ffrey:
Related jira: https://issues.redhat.com/browse/PLANNER-1872

ge0ffrey:
I am going to look into having the DefaultSolverScope's bestScore and bestScoreTimeMillis being updated without the bestSolution being updated and need to be cloned.

ge0ffrey (changes requested):
To make sure the unimproved termination doesn't kill the CH before it finally does it's one and only bestSolution event, we need a way to make that exception, not by leaving it null the first time.
We could have 2 CH's in some use cases, this should work for the 2nd CH too.

PR #706, opened 2020-02-26
PLANNER-1845 Improve multi-threaded shutdown performance

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/decider/MultiThreadedConstructionHeuristicDecider.java
```diff
@@ -111,12 +111,14 @@ public void phaseStarted(ConstructionHeuristicPhaseScope<Solution_> phaseScope)
...
+        // Clear non-started operations to improve shutdown performance. We're shutting down anyway, so it's fine.
+        operationQueue.clear();
         // Tell the move thread runners to stop
         DestroyOperation<Solution_> destroyOperation = new DestroyOperation<>();
         for (int i = 0; i < moveThreadCount; i++) {
             operationQueue.add(destroyOperation);
         }
-        // TODO This should probably be in a finally that spawns at least the entire phase, maybe even the entire solve
+        // TODO This should probably be in a finally that spans at least the entire phase, maybe even the entire solve
         ThreadUtils.shutdownAwaitOrKill(executor, logIndentation, "Multithreaded Local Search");
```
triceo: FYI I went looking around the code and this may not be achievable.
Most likely the entire lifecycle class hierarchy would have to be made `AutoClosable` and the `close()` call propagated downward.

triceo:
As I was looking around, I will add one comment that has nothing to do with this PR. I do not like the way how these deciders (and probably other classes in the lifecycle hierarchy) are designed to be "forever" classes.

At the end of the phase, the internal fields are cleared so that the instance can be recycled for when the phase is started again. This goes against modern thinking around GC, where objects are supposed to be short-lived. Also, it creates extra code that has to clear all those fields, leading us one step further away from immutability.

In my opinion, we should avoid that pattern in future.

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/decider/MultiThreadedConstructionHeuristicDecider.java
```diff
@@ -111,12 +111,14 @@ public void phaseStarted(ConstructionHeuristicPhaseScope<Solution_> phaseScope)
     @Override
     public void phaseEnded(ConstructionHeuristicPhaseScope<Solution_> phaseScope) {
         super.phaseEnded(phaseScope);
+        // Clear non-started operations to improve shutdown performance. We're shutting down anyway, so it's fine.
+        operationQueue.clear();
```
ge0ffrey: Just thinking of this now:
Can this cause a race condition?
- The ApplyStep operations have a `CyclicBarrier#await()` in MoveThreadRunner line 125.
- Presume there are 4 child threads
- At some piont, we do a step, adding 4 ApplyStep operations on the operationsQueue.
- 3 ApplyStep operations gets pulled quickly, but the 4th child thread is still working on a very big move form the previous step (which will be discarded when it finishes).
- Those first 3 ApplyStep operations hit the `moveThreadBarrier.await();`, waiting for the 4th one to come along.
- Meanwhile, time's up. Termination sets in. This `operationQueue.clear();` activates and 4 DestroyOperations are added.
- When the 4th child thread finishes his big move, its ApplyStep operation is already cleared. It pulls a DestroyOperation and kills itself.
- The other 3 child threads hang on the `moveThreadBarrier.await();` forever. This is a bug. Later, the thread interruptions kill the child threads. There is no destroy operations, so no info is given back, including an accurate "score calculation speed" etc.

ge0ffrey: This wasn't resolved and is proven to be accurate.

ge0ffrey:
PR replaced by https://github.com/kiegroup/optaplanner/pull/762

PR #707, opened 2020-02-26
Address Sonar-reported vulnerabilities

triceo:
These have to do with public access to fields, which can be externally mutated by bad people.
I have run all affected examples and made sure they continue to work as before.


optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/generator/LocationDataGenerator.java
```diff
@@ -68,53 +73,10 @@
             new LocationData("STR", 48.69, 9.221944),
             new LocationData("OPO", 41.235556, -8.678056),
             new LocationData("KBP", 50.344722, 30.893333),
-            new LocationData("VCE", 45.505278, 12.351944),
-    };
-
-    public static final LocationData[] EUROPE_CAPITALS = {
```
triceo: This appears to have been unused.

ge0ffrey: Please don't throw it away, it was used for the europe40 dataset and it was a PITA to collect that data.

PR #708, opened 2020-02-26
Override KIE group java conventions for OptaPlanner

rsynek:
Our code style differs from KIE group conventions on aligning method parameters. To avoid frequent comments on PRs about this code style difference, we should maintain our own code style definition.

### DIFF
--- a/ide-configuration/intellij/OptaPlanner Java Conventions.xml
+++ b/ide-configuration/intellij/OptaPlanner Java Conventions.xml
@@ -1,0 +1,117 @@
+<code_scheme name="OptaPlanner Java Conventions" version="173">
+  <option name="OTHER_INDENT_OPTIONS">
+    <value>
+      <option name="INDENT_SIZE" value="2" />
+      <option name="TAB_SIZE" value="2" />
+      <option name="SMART_TABS" value="true" />
+    </value>
+  </option>
+  <option name="LINE_SEPARATOR" value="&#xA;" />
+  <HTMLCodeStyleSettings>
+    <option name="HTML_ATTRIBUTE_WRAP" value="0" />
+    <option name="HTML_TEXT_WRAP" value="0" />
+  </HTMLCodeStyleSettings>
+  <JavaCodeStyleSettings>
+    <option name="CLASS_NAMES_IN_JAVADOC" value="3" />
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+    <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+      <value />
+    </option>
+    <option name="IMPORT_LAYOUT_TABLE">
+      <value>
+        <package name="java" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="javax" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="" withSubpackages="true" static="false" />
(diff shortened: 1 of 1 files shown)


Christopher-Chianelli:
Why does this code style difference exists? Would this also apply to OptaWeb?

rsynek:
@Christopher-Chianelli good questions;
"Why does this code style difference exists?"
There was a discussion about the code style in KIE group in the past - either this was an option that won or maybe it was not possible to enforce it for some other IDEs. I tried to look up the conversation, but haven't found it.
"Would this also apply to OptaWeb?" Depends on what we agree on. So far, I have seen this to be a struggle in PRs to OptaPlanner.

ge0ffrey:
Agreed that this indentation difference has been a pita.

Good for me to merge this PR as a step forward,
but we need to standardize on the Quarkus code style (which is the defacto Java standard with Red Hat), enforce it in the build (like they do) and just be done with it:
- Here it is https://github.com/quarkusio/quarkus/tree/master/independent-projects/ide-config/src/main/resources (what, no IntelliJ code style files? Let's ask on zulip - I can't believe that)
- Here's how the build enforces it: https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L320

triceo:
@ge0ffrey Doesn't it make sense then to move to the Quarkus code style instead of using this PR?

ge0ffrey:
Sure, I am happy to go there directly. But that will take a bit more work I imagine :)

rsynek:
One more question about how far we want to go:
1) just publish the IntelliJ code style to make it easier for folks opening PRs to OptaPlanner (no matter if it's this code style or the one used in Quarkus)
2) enforce it via checkstyle and format the entire repository
3) spread it to OptaWeb projects as well

ge0ffrey:
I believe: Yes, to 1), 2) and 3).

Christopher-Chianelli:
Note: eclipse has a maven auto formatter plugin; Intellij does not

rsynek:
> I believe: Yes, to 1), 2) and 3). Wydt?

It's the only way of achieving consistency across the projects. If we all agree on using that code style, it makes sense.

triceo:
@rsynek Assuming I can get the Quarkus code style in IDEA, I say let's go for it.

ge0ffrey:
@triceo I would be surprised if the Quarkus team haven't already done that. Note that the IDEA code styles are much more fine grained than Eclipse (Eclipse is a superset), so the Eclipse one doesn't define the IDEA one accuratly (but vica versa it does).

triceo:
@ge0ffrey Their solution is to configure IDEA to use the Eclipse formatter:
https://github.com/quarkusio/quarkus/blob/master/CONTRIBUTING.md#idea-setup

ge0ffrey:
> It's the only way of achieving consistency across the projects. If we all agree on using that code style, it makes sense.

Sounds good, as long as any voting is either in favor of using that code style or against it. Yes or no. No bikeshedding, no design by committee, no small variations, etc. I am putting my foot down on that. Other suggestions of well-accepted code styles in the Java ecosystem are welcome too - but again, no "tweaking" it.
Normally I don't believe in black-white acceptance voting like this, but 3+ previous experience with code style discussions make me force this exception...

triceo:
Yeah, let's just take what Quarkus does without changes. I've looked at their code and there is nothing I would be fundamentally opposed to.

rsynek:
Closing this PR, as its primary purpose was starting this discussion. I will open a new one with Quarkus code style. Hopefully in the next sprint.

Thanks everyone for your opinion!

PR #709, opened 2020-02-26
PLANNER-1875 Adjust docs and Javadocs following feedback

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -368,6 +369,7 @@
      * Runs all tuples of the stream through a given @{@link BiConstraintCollector} and converts them into a new
      * {@link UniConstraintStream} which only has a single tuple, the result of applying {@link BiConstraintCollector}.
      * @param collector never null, the collector to perform the grouping operation with
+     * (See {@link ConstraintCollectors} for a selection of ready-made collectors.)
```
ge0ffrey: soft suggestions:
- I'd remove the braces: "(See" -> "See"
- "for a selection of ready-made collectors.": I'd avoid the use of the word "selection" when it's not "move selection" or "entity selection" or something like that. "ready-made" is a tautology. Also, I 'd mention "count()" and "sum()" as that's much more tangible as to what I can find there (and what a collector actually is).
So suggestion => "See ConstraintCollectors for common operations, such as `count()`, `sum()` and others".

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -338,6 +338,13 @@ the prescribed ordering.
 For a full list of all supported `Joiner` implementations and their various overloads, refer to the
 `org.optaplanner.core.api.score.stream.Joiners` class.
 
+[NOTE]
+====
+In case you only need to act when a particular fact exists, and you don't need to access the fact itself, use
```
ge0ffrey: "a particlar fact": is that the first or the second fact?
"you don't need to" is indirection
This sentance can be more concise I think.
Also we need to mention the "one or more" difference.
How about something like

If the otherStream potentially matches multiple times, but it must only be propagated once, use <<constraintStreamsConditionalPropagation,ifExists>> instead.

PR #710, opened 2020-03-02
Rough Tennis ConstraintCollectors work

cuijulian:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -69,6 +69,21 @@
  */
 public final class ConstraintCollectors {
 
+    public static <A> DefaultUniConstraintCollector<A, ?, Object> loadBalanceByCount() {
```
ge0ffrey: Great that you've implemented it with a ConstraintCollector for loadBalancing.
However, given the complexity of loadBalancing and fairness - this is just the tip of the iceberg, see [1] - I am not comfortable adding it a method name into the backwards compatible API.
Let's have it ripe and mature in the example first. Can you put this code in the tennis example for now? 

[1] https://www.optaplanner.org/blog/2017/02/03/FormulaForMeasuringUnfairness.html

ge0ffrey: Note that even [1] doen't take into account the weird affinity-effects of the task assignment example.

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+    // ############################################################################
+
+    private Constraint oneAssignmentPerDatePerTeam(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TeamAssignment.class)
+                .ifExistsOther(TeamAssignment.class,
+                        Joiners.equal(TeamAssignment::getTeam),
+                        Joiners.equal(TeamAssignment::getDay),
+                        Joiners.lessThan(TeamAssignment::getId),
+                        Joiners.filtering((teamAssignment, otherTeamAssignment) -> teamAssignment.getTeam() != null))
+                .penalize("oneAssignmentPerDatePerTeam", HardMediumSoftScore.ONE_HARD);
```
ge0ffrey: This causes a "score trap" (see docs).

If 2 assignments conflict, the penality should be -1hard (AB)
If 3 assignements conflict, the penality should be -3hard (AB, BC, AC)
If 4 conflict, -6hard (AB, AC, AD, BC, BD, CD)

ge0ffrey: Furtermore, the justification won't mention which 2 teamAssignments conflict.

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+    // ############################################################################
+    // Hard constraints
+    // ############################################################################
+
+    private Constraint oneAssignmentPerDatePerTeam(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TeamAssignment.class)
+                .ifExistsOther(TeamAssignment.class,
+                        Joiners.equal(TeamAssignment::getTeam),
+                        Joiners.equal(TeamAssignment::getDay),
+                        Joiners.lessThan(TeamAssignment::getId),
```
ge0ffrey: Doesn't the "Other" in ifExistsOther already do that? @triceo wdyt?
Well not really: other means unequal, no uniqueness of pair, right?
This is confusing. How often do we want to use "other" without "uniqueness of pair"? I 'd argue never?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: +1.

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    // ############################################################################
+    // Medium constraints
+    // ############################################################################
+
+    // Faster, but does not combine well with other constraints in the same level
+    private Constraint fairAssignmentCountPerTeam(ConstraintFactory constraintFactory) {
+
+        //return (long) (Math.sqrt((double) squaredSum) * 1000);
```
ge0ffrey: code style: space after //

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+        return constraintFactory.from(TeamAssignment.class)
+                .filter(teamAssignment -> teamAssignment.getTeam() != null)
+                .groupBy(TeamAssignment::getTeam, loadBalanceByCount())
+                .penalize("fairAssignmentCountPerTeam", HardMediumSoftScore.ONE_MEDIUM,
+                          (team, result) -> (int) (Math.sqrt(((Long[])result)[1]) * 1000.0));
```
ge0ffrey: What happens if you do
`(Team team, long[] result) -> ...`?

1) This avoids the cast
2) This avoids unboxing a long[] into a Long[]

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
triceo: I was actually considering that when I was designing the API, we even discussed that. We've decided to stick to what `join()` does. With `join()`, you also have to handle uniqueness manually (or use `fromUniquePair()`).

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,80 @@
...
+    // ############################################################################
+    // Medium constraints
+    // ############################################################################
+
+    // Faster, but does not combine well with other constraints in the same level
+    private Constraint fairAssignmentCountPerTeam(ConstraintFactory constraintFactory) {
+
+        //return (long) (Math.sqrt((double) squaredSum) * 1000);
+        return constraintFactory.from(TeamAssignment.class)
+                .filter(teamAssignment -> teamAssignment.getTeam() != null)
```
triceo: This shouldn't be necessary, I think. `from()` in CS gives you the entities only after they've been already initialized.

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
ge0ffrey: Yes, I remember.
Seeing this example, I started to doubt our decision (a bit late as 7.33 is out).
We can leave it as is for now, and keep it in the back of our head.

But this example should drop the "Other" because it already has a "lessThan(id)"

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
ge0ffrey: yep, fromUninitialized() is the alternative that doesn't do that (but don't use that here)

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
cuijulian: @ge0ffrey I'd argue this is outside the scope of the PR, which is to add CS for the existing DRL file (which has the "score trap" from what I can tell), but I agree with your point about how multiple assignment conflicts are penalized. I will add a TODO for a change of this rule (and its equivalent in the DRL file).

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
cuijulian: I get a `Incompatible parameter types in lambda expression: expected Object but found long[]` complaint from the IDE. Though I can fix 2 by casting to `long[]`, not sure why I put `Long[]`.

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
ge0ffrey: The reasoning itself makes sense, but the DRL does it differently, the way I expected.

The DRL (= CS with join()) :

```
        TeamAssignment(team != null, $t : team, $d : day, $leftId : id)
        TeamAssignment(team == $t, day == $d, id > $leftId)
```

A DRL that matches the current PR (= CS with ifExists()) :

```
        TeamAssignment(team != null, $t : team, $d : day, $leftId : id)
        exists TeamAssignment(team == $t, day == $d, id > $leftId)
```

ge0ffrey: assertionScoreDirectoryFactory might not always detect this difference.

cuijulian: Ah, I see your point. Strange that a FULL_ASSERT didn't catch this, shouldn't the assertionScoreDirectoryFactory assert that the scores are penalized by the same values? In any case, I'll fix the CS.

ge0ffrey: They're not penalized by the same value for a conflict of 3 entities or more. However, because it's hard constraint, on most datasets, it never going to evaluate a conflict of 3+ entities: as soon as it sees a conflict of 2 entities, it's very unlikely that that step will with the CH or LS, at least with the default set of move selectors.

triceo:
Marking as stale, we do not expect to merge this before the code style changes are merged.

ge0ffrey:
From an SO reviewer of this code:
" it seems like in above reference example, the penalize() is incorrect. Instead of calculating sqrt((a^2) + (b^2)), it is calculating sqrt(a^2)+sqrt(b^2)"

cuijulian:
> From an SO reviewer of this code:
> " it seems like in above reference example, the penalize() is incorrect. Instead of calculating sqrt((a^2) + (b^2)), it is calculating sqrt(a^2)+sqrt(b^2)"

@ge0ffrey Quick question, what's an "SO reviewer"? And as for reference example, is the reviewer referring to the original DRL rule, or this PR? Or better yet, is there a link to this review/ticket?

triceo:
Rebased in #939 .

PR #712, opened 2020-03-03
PLANNER-1865 Disable Drools rules when constraint weight is zero

triceo:
I've benchmarked this to cause ~ 5 % performance drop when creating `KieSession`. (Went from 3300 sessions per second to ~ 3100.)

Considering that this operation is not performed frequently, and that it will be followed by a much longer-running solver operation, I do not consider the drop to be significant. Even if the solver only runs for 5 seconds, we have lost a fraction of a millisecond here.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetScoringBiConstraintStream.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetScoringBiConstraintStream.java
@@ -113,9 +113,6 @@ public final class BavetScoringBiConstraintStream<Solution_, A, B>
             if (intMatchWeigher != null) {
                 scoreImpacter = (A a, B b, Consumer<Score<?>> matchScoreConsumer) -> {
                     int matchWeight = intMatchWeigher.applyAsInt(a, b);
-                    if (matchWeight == 0) { // No need to include and no need to undo later.
-                        return () -> { /* NOOP */ };
-                    }
                     constraint.assertCorrectImpact(matchWeight);
                     return castedWeightedScoreImpacter.impactScore(matchWeight, matchScoreConsumer);
                 };
@@ -132,9 +129,6 @@ public final class BavetScoringBiConstraintStream<Solution_, A, B>
             if (longMatchWeigher != null) {
                 scoreImpacter = (A a, B b, Consumer<Score<?>> matchScoreConsumer) -> {
                     long matchWeight = longMatchWeigher.applyAsLong(a, b);
-                    if (matchWeight == 0L) { // No need to include and no need to undo later.
-                        return () -> { /* NOOP */ };
-                    }
                     constraint.assertCorrectImpact(matchWeight);
                     return castedWeightedScoreImpacter.impactScore(matchWeight, matchScoreConsumer);
                 };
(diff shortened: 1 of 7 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
```diff
@@ -42,30 +49,65 @@
...
-                matchJustificationsToOutput((List<Object>) justificationList, constraints.get(rule).getExpectedJustificationTypes()));
-        constraints.forEach((rule, constraint) -> scoreHolder.configureConstraintWeight(rule,
-                constraint.extractConstraintWeight(workingSolution)));
-        KieSession kieSession = kieBase.newKieSession();
-        ((RuleEventManager) kieSession).addEventListener(new OptaPlannerRuleEventListener()); // Enables undo in rules
+                matchJustificationsToOutput((List<Object>) justificationList,
+                        compiledRuleToConstraintMap.get(rule).getExpectedJustificationTypes()));
+        // Determine which rules to enable based on the fact that their constraints carry weight.
+        Score<?> zero = scoreDefinition.getZeroScore();
+        Set<String> enabledConstraintIdSet = new LinkedHashSet<>(compiledRuleToConstraintMap.size());
```
triceo: I'll refactor this to `disabledConstraintIdSet` - that will cause less operations with the collection, and therefore perform better.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetScoringBiConstraintStream.java
```diff
@@ -113,6 +113,9 @@ private BavetScoringBiConstraintStream(BavetConstraintFactory<Solution_> constra
             if (intMatchWeigher != null) {
                 scoreImpacter = (A a, B b, Consumer<Score<?>> matchScoreConsumer) -> {
                     int matchWeight = intMatchWeigher.applyAsInt(a, b);
+                    if (matchWeight == 0) { // No need to include and no need to undo later.
+                        return () -> { /* NOOP */ };
```
ge0ffrey: Soft: might be worth extracting a NO_OP_UNDO into AbstractConstraintStream?

ge0ffrey: I still wonder if it shouldn't fail fast if the matchWeight is zero.

Actually, thinking about this more - we probably want to have the indictment anyway, even if the matchWeigh is zero...  even if it's just for debugging purposes - for easy of debugging. Let's discuss.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DroolsCondition.java
```diff
@@ -175,20 +175,29 @@ protected DroolsCondition(T ruleStructure) {
 
     protected <S extends Score<S>, H extends AbstractScoreHolder<S>> void impactScore(DroolsConstraint<?> constraint,
             Drools drools, H scoreHolder, int impact) {
+        if (impact == 0) { // No need to include and no need to undo later.
+            return;
+        }
```
ge0ffrey: Same here: if a matchWeigth is zero, it probably shouldn't disable the indictment. For clarity: If the constraintWeight is zero, it should (and the entire constraint execution too, as it does in the rest of this PR).

triceo:
Jenkins please retest this.

triceo:
The `matchWeight` changes were reverted, as per our discussion offline. Coverage will be increased when we have the constraint verifier, which is next on my agenda. Merging.

PR #713, opened 2020-03-04
PLANNER-1861 ConstraintVerifier API

triceo:


optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.optaplanner.test.impl.score.stream.ConstraintVerifier.forConstraintProvider;
+
+public class NQueensConstraintProviderTest {
+
+    // Dummy data, the tests don't actually do anything.
+    private Queen queen1, queen2;
+    private Row row1, row2;
+
+    @Test
+    public void testSingleConstraint() {
```
ge0ffrey: test method should not start with "test", that's verbose (and junit 3 era)

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.optaplanner.test.impl.score.stream.ConstraintProviderVerifier;
+
+import static org.optaplanner.test.impl.score.stream.ConstraintVerifier.forConstraint;
+import static org.optaplanner.test.impl.score.stream.ConstraintVerifier.forConstraintProvider;
+
+public class NQueensConstraintProviderTest {
+
+    // Dummy data, the tests don't actually do anything.
+    private Queen queen1, queen2;
+    private Row row1, row2;
```
ge0ffrey: Let's write the code to initialize those in the test strawman too.
It might affect how we design methods, such as adding varargs params.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void testSingleConstraint() {
+        NQueensConstraintProvider constraintProvider = new NQueensConstraintProvider();
+        forConstraint(constraintProvider::ascendingDiagonalConflict)
+                .givenFacts(queen1, queen2, row1, row2)
+                .expectImpact(SimpleScore.ONE)  // No type check here. Wrong score type fails at runtime.
```
ge0ffrey: All ScoreVerifier checks should be regardless of the constraingWeight. Only the matchWeight needs to be tested.
This is for both fixed constraintWeight and penalizeConfigurable constraintWeights etc.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Test
+    public void testEntireConstraintProvider() {
+        forConstraintProvider(new NQueensConstraintProvider())
+                .givenPlanningSolution(new NQueens()) // You would read some XStream solution here.
+                .expectImpact(SimpleScore.ONE); // Does not allow to check constraint matches, pointless with Solution.
+    }
+
+    @Test
+    public void testVerifierReuse() {
+        ConstraintProviderVerifier constraintVerifier = forConstraintProvider(new NQueensConstraintProvider());
```
ge0ffrey: where do we set which constraint we're verifying?
The goal is to verify a single constraint in isolation, so that when refactoring the other constraints, this unit test is unaffected.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,63 @@
...
+                .expectConstraintMatches(queen1, queen2);
+        //      ^^^^^^ We shouldn't allow this until we have a clear definition and correct implementation of
+        //      constraint matches.
+        //      I suggest that this is out of scope for now, and therefore so is this method.
+    }
+
+    @Test
+    public void testEntireConstraintProvider() {
+        forConstraintProvider(new NQueensConstraintProvider())
+                .givenPlanningSolution(new NQueens()) // You would read some XStream solution here.
```
ge0ffrey: Makes sense probably to add indeed. But `givenFacts()` is going to be used far more frequently.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: Actually, I disagree.

We should assert on the final impact, not on the `matchWeight`. You want to know what becomes your score, not some partial calculation that is `matchWeight`.

In fact, that is why I added the method `givenPlanningSolution`, where the configurable constraint weights are defined - so that you can get the final result, not just the match weight.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: The goal of this method is to test the constraint provider as a whole. I do think that this is an interesting, useful use case - considering solution X, my score is Y.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
yurloc: I tend to agree with Geoffrey. In the context of a constraint provider unit test, I'd like to focus on testing purely the constraint behavior, that is:
- whether it's a penalty of a reward,
- how many times it triggers (typically to verify number of matching unique pairs in the given set of objects),
- whether the [`matchWeigher`](https://github.com/kiegroup/optaplanner/blob/7.34.0.Final/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java#L768) calculates the expected match weight (in cases it is affected by matching object properties, e.g. distance).

I would argue that `@ConstraintWeight` affects total score but not constraint behavior. So, if constraint weight configuration changes, constraints' behavior doesn't change, therefore the constraint provider unit test shouldn't fail.

In other words, I'd like to freely change my constraint weight configuration without breaking CP tests.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,60 @@
...
+        row.setId(0L);
+        row.setIndex(0);
+        Queen queen1 = new Queen();
+        queen1.setId(0L);
+        queen1.setRow(row);
+        queen1.setColumn(column1);
+        Queen queen2 = new Queen();
+        queen2.setId(1L);
+        queen2.setRow(row);
+        queen2.setColumn(column2);
```
triceo: This mess is what I hoped to avoid by allowing the user to simply deserialize a ready-made solution.
Keep in mind that this is a trival example - testing rea-world constraints will become a nightmare, unless we allow working with ready-made solutions.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.Test;
+import org.optaplanner.examples.nqueens.domain.Column;
+import org.optaplanner.examples.nqueens.domain.NQueens;
+import org.optaplanner.examples.nqueens.domain.Queen;
+import org.optaplanner.examples.nqueens.domain.Row;
+import org.optaplanner.test.impl.score.stream.ConstraintVerifier;
+import org.optaplanner.test.impl.score.stream.SingleConstraintVerifier;
+
+public class NQueensConstraintProviderTest {
```
triceo: This is the important bit that needs discussion, the test.
Don't focus on the code just yet.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,60 @@
...
+        Queen queen2 = new Queen();
+        queen2.setId(1L);
+        queen2.setRow(row);
+        queen2.setColumn(column2);
+        SingleConstraintVerifier<NQueens> horizontalConflictConstraintVerifier =
+                constraintVerifier.forConstraint(constraintProvider::horizontalConflict);
+        horizontalConflictConstraintVerifier.givenFacts(queen1, row, column1, column2)
+                .expectImpact(0);
+        horizontalConflictConstraintVerifier.givenFacts(queen1, queen2, row, column1, column2)
+                .expectImpact(1);
```
triceo: Now that I see this, I realize this needs a `message` argument, much like JUnit assertions.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -162,6 +162,7 @@ public SupplyManager getSupplyManager() {
 
     @Override
     public void setWorkingSolution(Solution_ workingSolution) {
+        Objects.requireNonNull(workingSolution);
```
yurloc: I like to do this:

output

triceo: Ah, it has a return type. Good to know, I thought it was `void`.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,75 @@
...
+        row.setIndex(0);
+        Queen queen1 = new Queen();
+        queen1.setId(0L);
+        queen1.setRow(row);
+        queen1.setColumn(column1);
+        // One queen
+        SingleConstraintVerifier<NQueens> horizontalConflictConstraintVerifier =
+                constraintVerifier.forConstraint(constraintProvider::horizontalConflict);
+        horizontalConflictConstraintVerifier.givenFacts(queen1, row, column1)
+                .expectImpact("No horizontal conflicts with just one queen.", 0);
```
yurloc: What is this number? Is it the sum of `matchWeight`s from all matches of this constraint?

yurloc: I'm thinking about two use cases:
1. I only want to verify how many times the constraint matched and I don't care about the match weights.
2. I want to verify the match weight was calculated correctly.

I think that use case 2 is more important and it also indirectly verifies the match count. So the question is: is use case 1 valid? Should it be supported as well? My proposed answer is yes. If it is, then the API should discern between these two verification types.

yurloc: Feel free to disagree. I probably shouldn't have looked at the implementation where the impact number is calculated as a sum of matches. The weight is disregarded. That made me ask the original question in this thread.

Now I think verifying the total match weight is enough.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,75 @@
...
+        // Three queens
+        Column column3 = new Column();
+        column2.setId(2L);
+        column2.setIndex(2);
+        Queen queen3 = new Queen();
+        queen3.setId(2L);
+        queen3.setRow(row);
+        queen3.setColumn(column3);
+        horizontalConflictConstraintVerifier.givenFacts(queen1, queen2, queen3, row, column1, column2, column3)
+                .expectImpact("Three pairs of queens on the same row.", 3);
```
yurloc: If `3` is the total match weight, I think it should be `SimpleScore.of(3)`. I think it will make the tests easier to understand although it requires more keystrokes than just providing the number.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: That's a good question.

I believe that the constraint is more than just a number of matches. The match weight is calculated, and in some cases it's calculated with a relatively complicated code. Therefore it needs to be tested.

You know I was originally arguing that we should evaluate the resulting score, not just the weights. I have since changed my mind on that. But I think that further reducing the problem to just match counts would be over-simplifying.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: I don't know what you mean. You've all made it very clear to me during the last meeting that we should not match scores, as constraint weights are externally configurable.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
yurloc: I'm thinking the same.

I still think verifying the match count *could be* useful in some cases but we can think about that later.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
yurloc: I'm not talking about the total score. I'm only talking about the match weight.

I went through the code again and I think I've probably confused constraint weight and match weight (in the API, not conceptually). So if I have:

```java

    Constraint horizontalConflict(ConstraintFactory factory) {
        return factory
                .fromUniquePair(Queen.class, equal(Queen::getRowIndex))
                .penalize("Horizontal conflict", SimpleScore.ONE);
    }
```
`SimpleScore.ONE` is constraint weight, which only determines score level and the base that can be weighted. In this case match weight is not provided, so I guess it defaults to 1.

Whereas:
```java
    Constraint computerCost(ConstraintFactory constraintFactory) {
        return constraintFactory.from(CloudComputer.class)
                .ifExists(CloudProcess.class, equal(Function.identity(), CloudProcess::getComputer))
                .penalize("computerCost",
                        HardSoftScore.ONE_SOFT,
                        CloudComputer::getCost);
    }
```
in this case constraint weight is `HardSoftScore.ONE_SOFT` and it's given an additional (dynamic) weight of `CloudComputer::getCost`.

If this is true then the question in the beginning of this thread doesn't make sense. Result of weight calculation is a number (`int`, `long`, `Big*`) so it makes sense as it is (`expectImpact(String msg, Number totalWeight)`).

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/AbstractConstraintVerifierAssertion.java
```diff
@@ -0,0 +1,58 @@
...
+    protected final V getParentConstraintVerifier() {
+        return parentConstraintVerifier;
+    }
+
+    abstract protected Number getImpact();
+
+    private void assertImpact(Number weight) {
+        Number impact = getImpact();
+        if (!weight.equals(impact)) {
+            throw new IllegalStateException("Expected " + weight + " (" + weight.getClass() + ") is not actual " + impact + " (" + impact.getClass() + ")");
```
rsynek: Why not using `AssertionError`? E.g. jUnit throws this exception in case of a test failure.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/AbstractConstraintVerifierAssertion.java
```diff
@@ -0,0 +1,82 @@
...
+    abstract protected Number getImpact();
+
+    private void assertImpact(Number weight, String message) {
+        Number impact = getImpact();
+        String constraintId = getParentConstraintVerifier().getConstraint().getConstraintId();
+        if (weight.equals(impact)) {
+            return;
+        }
+        if (message == null) {
+            throw new IllegalStateException("Broken expectation." + System.lineSeparator() +
```
yurloc: throw AssertionError? Frameworks like JUnit or AssertJ throw that or a subclass of it.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: Yes, exactly. The score (_HardSoftScore.ONE_SOFT_) is called "constraint weight", and can be configured externally using `ConstraintConfiguration`. The other value, often calculated (`CloudComputer::getCost`), is called "match weight", it is a numeric type and is defined within the constraint itself.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
yurloc: I think this can be solved by using a helper method in the test or factory methods in the domain model. But, since methods have just a single return value, we'd have to wrap the set of objects in some kind of container (the solution, probably (obviously?)).

`<TLDR>`
**So, no matter if we get the solution (or any kind of container) with a factory/helper method or by unmarshalling it, `givenFacts()` needs to (also) accept that instead of varargs of objects.**
`</TLDR>`

On the other hand, Geoffrey's argument against that approach is also valid. Sometimes the objects can be created very easily and wrapping them into a solution is an unnecessary ceremony.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSession.java
```diff
@@ -50,6 +50,8 @@
      */
     Map<Object, Indictment> getIndictmentMap();
 
+    ConstraintSessionFactory<Solution_> getSessionFactory();
```
yurloc: Unused.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -41,6 +42,7 @@
 
 public final class BavetConstraintSession<Solution_> implements ConstraintSession<Solution_> {
 
+    private final ConstraintSessionFactory<Solution_> sessionFactory;
```
yurloc: Seems to be unused.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintStream.java
```diff
@@ -63,7 +63,7 @@
 
     protected Function<Solution_, Score<?>> buildConstraintWeightExtractor(String constraintPackage, String constraintName, Score<?> constraintWeight) {
         validateConstraintId(constraintPackage, constraintName);
-        // Duplicates validation when the session is build, but this fail-faster when weights are hard coded
+        // Duplicates validation when the session is build, but this fail-fasts when weights are hard coded
```
yurloc:
```suggestion
        // Duplicates validation when the session is built, but this fails fast when weights are hard coded
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintSessionFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Arrays;
+
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import org.optaplanner.core.impl.score.definition.ScoreDefinition;
+import org.optaplanner.core.impl.score.stream.ConstraintSessionFactory;
+
+public abstract class AbstractConstraintSessionFactory<Solution_> implements ConstraintSessionFactory<Solution_> {
+
+    private final ScoreDefinition scoreDefinition;
```
yurloc: Raw type.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintSessionFactory.java
```diff
@@ -0,0 +1,61 @@
...
+
+    protected ScoreDefinition<?> getScoreDefinition() {
+        return scoreDefinition;
+    }
+
+    protected Score<?> getScoreZero() {
+        return scoreDefinition.getZeroScore();
+    }
+
+    protected <S extends Score<S>> S getScoreOneSoftest() {
```
yurloc: Candidate for a unit test.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+        Queen queen3 = new Queen(2, row3, column3);
+        constraintVerifier.forConstraint(constraintProvider::descendingDiagonalConflict)
+                .given(queen1, queen2, queen3)
+                .expectReward(3);
+    }
+
+    private NQueens readSolution(String resource) throws IOException {
+        Path tempFile = Files.createTempFile("optaplanner-test", "xml");
+        try (InputStreamReader inputStreamReader =
+                new InputStreamReader(NQueensConstraintProviderTest.class.getResourceAsStream(resource))) {
```
yurloc: Throws NPE on my machine. Can you confirm that?

yurloc: Maybe you forgot to commit `256queens_-30.xml`?

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+                .expectNoImpact();
+    }
+
+    @Test
+    public void horizontalConflictWithTwoQueens() {
+        Queen queen1 = new Queen(0, row1, column1);
+        Queen queen2 = new Queen(1, row1, column2);
+        constraintVerifier.forConstraint(constraintProvider::horizontalConflict)
+                .given(queen1, queen2)
+                .expectReward(1);
```
yurloc: Should be a penalty, shouldn't it?

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: Doesn't, tests pass both on my machine and on Jenkins.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirectorFactory.java
```diff
@@ -62,13 +65,21 @@ public ConstraintStreamScoreDirectorFactory(SolutionDescriptor<Solution_> soluti
...
     public ConstraintSession<Solution_> newConstraintStreamingSession(boolean constraintMatchEnabled, Solution_ workingSolution) {
         return constraintSessionFactory.buildSession(constraintMatchEnabled, workingSolution);
     }
 
+    // ************************************************************************
+    // Getters/setters
+    // ************************************************************************
+
+    public Constraint[] getConstraints() {
+        return Arrays.copyOf(constraints, constraints.length);
```
ge0ffrey: ConstraintStreamScoreDirectorFactory.getConstraints is non-public API.
We avoid defensive copies in non-public API, unless it's a probably-one-day-public-API, to avoid performance surprises and to reduce the code size.
This method will never be part of a public API I'd argue.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSessionFactory.java
```diff
@@ -23,7 +23,7 @@
     /**
      * This method is thread-safe.
      * @param constraintMatchEnabled  true if {@link ScoreDirector#isConstraintMatchEnabled()} should be true
-     * @param workingSolution never null
+     * @param workingSolution if null, uniform synthetic constraint weights may be applied
```
ge0ffrey: "may" or "will"?

ge0ffrey: to take that philosophy a step further: if it's indeed "may", then it should specify when and when it doesn't happen, if that's practical.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintSessionFactory.java
```diff
@@ -0,0 +1,61 @@
...
+        Number[] result = Arrays.copyOf(getScoreZero().toLevelNumbers(), scoreDefinition.getLevelsSize());
+        Number lastNumber = result[result.length - 1];
+        if (lastNumber instanceof Integer) {
+            lastNumber = 1;
+        } else if (lastNumber instanceof Long) {
+            lastNumber = 1L;
+        } else if (lastNumber instanceof Double) {
+            lastNumber = 1.0d;
```
ge0ffrey: code style: `d` in capital, just longs `L` are in capital (which get confused with one otherwise)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintSessionFactory.java
```diff
@@ -0,0 +1,61 @@
...
+        return scoreDefinition;
+    }
+
+    protected Score<?> getScoreZero() {
+        return scoreDefinition.getZeroScore();
+    }
+
+    protected <S extends Score<S>> S getScoreOneSoftest() {
+        Number[] result = Arrays.copyOf(getScoreZero().toLevelNumbers(), scoreDefinition.getLevelsSize());
+        Number lastNumber = result[result.length - 1];
```
ge0ffrey: -1 on this approach. It is not this classes's responsibility to figure out the "one soft score".
(This will cause pain if ever decide to build a HardLongSoftIntScore. Or with pareto scoring.)

Add a method on ScoreDefinition, just like we have getZeroScore() there.

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/domain/Queen.java
```diff
@@ -32,6 +32,16 @@
     // Planning variables: changes during planning, between score calculations.
     private Row row;
 
+    public Queen() {
+        // For XStream
```
ge0ffrey: If we add that comment, we should do it in the other examples that have a default constructor for this reason too.
Personally, I wouldn't add the comment.

Btw, the comment isn't accurate. OptaPlanner itself also needs a no-arg constructor to planning clone Queen.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
yurloc: Fails on Jenkins.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
ge0ffrey: Lukas, presuming this isn't a typo, I am playing chess again you next F2F!
I played Matej Cimbora a few years ago and he handed me my ***... Three times... (he has an ELO rating.) I need a win :)

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Test
+    public void horizontalConflictWithThreeQueens() {
+        Queen queen1 = new Queen(0, row1, column1);
+        Queen queen2 = new Queen(1, row1, column2);
+        Queen queen3 = new Queen(2, row1, column3);
+        constraintVerifier.forConstraint(constraintProvider::horizontalConflict)
+                .given(queen1, queen2, queen3)
+                .expectReward(3);
```
ge0ffrey: Code syntax and method parameters look great.
Now for the difficult part: naming.
Are we sure we want to settle on this names?
- `forConstraint()`: +1. alternatives: `ofConstraint()` - What does the mockito API use for spy's?
- `given`: +1 (short and sweet) alternatives: `givenFacts` +0 (less ambiguous)
- `expectReward`: +0 alternatives: `assertReward`: +1 (because of JUnit's `assertEquals(expeted, actual)`)

yurloc: I like it how it is now. When you read it out loud it makes sense. "For constraint *constraint*, given *facts*, expect penalty *value*.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: I suck at chess, that would be no challenge for you. :-)

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+        Queen queen1 = new Queen(0, row1, column1);
+        Queen queen2 = new Queen(1, row2, column2);
+        Queen queen3 = new Queen(2, row3, column3);
+        constraintVerifier.forConstraint(constraintProvider::descendingDiagonalConflict)
+                .given(queen1, queen2, queen3)
+                .expectReward(3);
+    }
+
+    private NQueens readSolution(String resource) throws IOException {
+        Path tempFile = Files.createTempFile("optaplanner-test", "xml");
```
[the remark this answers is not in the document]
triceo: I don't need a dataset, I need a solution that's at least partially solved. And I need it with a fixed score, that I am asserting against.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
triceo: I could swear it didn't. Fixing, thanks for pointing out.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.optaplanner.examples.nqueens.domain.Queen;
+import org.optaplanner.examples.nqueens.domain.Row;
+import org.optaplanner.persistence.common.api.domain.solution.SolutionFileIO;
+import org.optaplanner.persistence.xstream.impl.domain.solution.XStreamSolutionFileIO;
+import org.optaplanner.test.impl.score.stream.ConstraintVerifier;
+
+public class NQueensConstraintProviderTest {
+
+    private final NQueensConstraintProvider constraintProvider = new NQueensConstraintProvider();
+    private final ConstraintVerifier<NQueens> constraintVerifier = ConstraintVerifier.using(NQueens.class, Queen.class);
```
triceo: Prop A) using()
Prop B) of()
Prop C) withSolution().withEntities()
Prop D) forSolutionAndEntities()
Prop E) build()
Prop F) buildConstraintVerifier()

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+    private final Row row2 = new Row(1);
+    private final Row row3 = new Row(2);
+    private final Column column1 = new Column(0);
+    private final Column column2 = new Column(1);
+    private final Column column3 = new Column(2);
+
+    @Test
+    public void noHorizontalConflictWithOneQueen() {
+        Queen queen1 = new Queen(0, row1, column1);
+        constraintVerifier.forConstraint(constraintProvider::horizontalConflict)
```
triceo: Prop A) forConstraint() (geoffrey +1, chris +1)
Prop B) ofConstraint()
Prop C) of()
Prop D) forSingleConstraint()

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+                .given(queen1)
```
triceo: Prop A) given() (geoffrey +1, chris +1)
Prop B) givenFacts()

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+                .expectNoImpact();
+    }
+
+    @Test
+    public void horizontalConflictWithTwoQueens() {
+        Queen queen1 = new Queen(0, row1, column1);
+        Queen queen2 = new Queen(1, row1, column2);
+        constraintVerifier.forConstraint(constraintProvider::horizontalConflict)
+                .given(queen1, queen2)
+                .expectPenalty(1);
```
ge0ffrey: constraintVerifier.assertThat(constraintProvider::horizontalConflict)
.given(queen1, queen2)
.hasPenalty(1);

ge0ffrey: assert -> 
hasPenalty -> A) producesPenalty
hasPenalty -> B) inflictsPenalty
hasPenalty -> C) penalizesWith()/rewardsWith()
hasPenalty -> D) penalizes()/rewards()
hasPenalty -> E) penalizesBy()/rewardsBy()
hasPenalty -> E) penalizesBy()/rewardsWith()

yurloc: Optional description method example:
```java
        constraintVerifier
                .verifyThat(constraintProvider::ascendingDiagonalConflict)
                .given(queen1, queen2)
                .as("queen 1 attacks queen 2")
                .penalizesBy(2);
```

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/ConstraintProviderAssertion.java
```diff
@@ -0,0 +1,65 @@
...
+    public final void expectScore(Score<?> score, String message) {
+        if (actualScore.equals(score)) {
+            return;
+        }
+        Class<?> constraintProviderClass = getParentConstraintVerifier().getConstraintProvider().getClass();
+        if (message == null) {
+            throw new AssertionError("Broken expectation." + System.lineSeparator() +
+                    "    Constraint provider: " + constraintProviderClass + System.lineSeparator() +
+                    "         Expected score: " + score + " (" + score.getClass() + ")" + System.lineSeparator() +
+                    "           Actual score: " + actualScore + " (" + actualScore.getClass() + ")");
```
ge0ffrey: talk about matchWeight, not score.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/SingleConstraintAssertion.java
```diff
@@ -0,0 +1,223 @@
...
+                        actualImpactLabel, actualImpact, actualImpact.getClass());
+            }
+            return baos.toString();
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed assembling asserting message.", e);
+        }
+    }
+
+    private static String getImpactTypeLabel(ScoreImpactType scoreImpactType) {
+        switch (scoreImpactType) {
```
yurloc: Throws NPE if you `expectNoImpact()` but the constraint does a penalty or a reward. To reproduce, change any `expectPenalty()` in the NQueens test to `expectNoImpact()`.

```
java.lang.IllegalStateException: Failed assembling asserting message.

	at org.optaplanner.test.impl.score.stream.SingleConstraintAssertion.getAssertionErrorMessage(SingleConstraintAssertion.java:93)
	at org.optaplanner.test.impl.score.stream.SingleConstraintAssertion.assertImpact(SingleConstraintAssertion.java:62)
	at org.optaplanner.test.impl.score.stream.SingleConstraintAssertion.expectNoImpact(SingleConstraintAssertion.java:213)
	at org.optaplanner.test.impl.score.stream.SingleConstraintAssertion.expectNoImpact(SingleConstraintAssertion.java:220)
	at org.optaplanner.examples.nqueens.solver.score.NQueensConstraintProviderTest.horizontalConflictWithTwoQueens(NQueensConstraintProviderTest.java:62)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:230)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMainV2.main(AppMainV2.java:131)
Caused by: java.lang.NullPointerException
	at org.optaplanner.test.impl.score.stream.SingleConstraintAssertion.getImpactTypeLabel(SingleConstraintAssertion.java:98)
	at org.optaplanner.test.impl.score.stream.SingleConstraintAssertion.getAssertionErrorMessage(SingleConstraintAssertion.java:76)
	... 31 more
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreDefinition.java
```diff
@@ -44,6 +44,11 @@ public BendableScore getZeroScore() {
         return BendableScore.zero(hardLevelsSize, softLevelsSize);
     }
 
+    @Override
+    protected Number getOne() {
+        return 1;
```
ge0ffrey: Why BendableScoreDefinition.java has getOne() that returns Number, while HardSoftScoreDefinition.java has getOneSoftestScore() that returns Score?
I'd prefer them all to just have getOneSoftestScore() that returns Score and avoid any levelNumbers magic.
(LevelNumbers are error prone.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraintSessionFactory.java
ge0ffrey: It's raw on SolutionDescriptor too :(
I am really looking for to 8.0 to get reduce Score<S extends Score> to just Score,
that will give a fighting change against the raw types.

ge0ffrey: Ps: I don't recommending fixing SolutionDescriptor for this right now. I tried in the past and failed, eventually - we need to degenerify Score first.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/buildin/bendablebigdecimal/BendableBigDecimalScoreDefinitionTest.java
```diff
@@ -21,10 +21,24 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
 
 public class BendableBigDecimalScoreDefinitionTest {
 
+    @Test
+    public void getZeroScore() {
+        BendableBigDecimalScore score = new BendableBigDecimalScoreDefinition(1, 2).getZeroScore();
+        assertThat(score.toLevelNumbers()).containsOnly(BigDecimal.ZERO);
```
ge0ffrey: I recommend avoid toLevelNumbers if possible.
So I'd assert that getZeroScore equals `BendableScore.of(new int[]{0}, new int[]{0, 0})`

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/buildin/hardsoftdouble/HardSoftDoubleScoreDefinitionTest.java
```diff
@@ -21,10 +21,24 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
 
 public class HardSoftDoubleScoreDefinitionTest {
 
+    @Test
+    public void getZeroScore() {
+        HardSoftDoubleScore score = new HardSoftDoubleScoreDefinition().getZeroScore();
+        assertThat(score.toLevelNumbers()).containsOnly(0d);
```
ge0ffrey: code style: "0.0" instead of "0d"

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+                .expectPenalty(3);
+    }
+
+    private NQueens readSolution(String resource) throws IOException {
+        Path tempFile = Files.createTempFile("optaplanner-test", "xml");
+        try (InputStreamReader inputStreamReader =
+                new InputStreamReader(NQueensConstraintProviderTest.class.getResourceAsStream(resource))) {
+            List<String> lines = IOUtils.readLines(inputStreamReader);
+            Files.write(tempFile,lines);
+            final SolutionFileIO<NQueens> solutionFileIO = new XStreamSolutionFileIO<>(NQueens.class);
```
ge0ffrey: add new method XStreamSolutionFileIO.read(InputStream) and use that instead of writing a temp file.

optaplanner-examples/src/test/resources/org/optaplanner/examples/nqueens/solver/score/256queens_-30.xml
```diff
@@ -0,0 +1,3339 @@
+<NQueens id="1">
```
ge0ffrey: file name needs to be self-explaining.

ge0ffrey: How about 256queens-score30.xml?

optaplanner-persistence/optaplanner-persistence-xstream/src/main/java/org/optaplanner/persistence/xstream/impl/domain/solution/XStreamSolutionFileIO.java
```diff
@@ -61,11 +62,19 @@ public String getInputFileExtension() {
 
     @Override
     public Solution_ read(File inputSolutionFile) {
+        try(InputStream inputSolutionStream = Files.newInputStream(inputSolutionFile.toPath())) {
```
ge0ffrey: code style: space before (

optaplanner-persistence/optaplanner-persistence-xstream/src/main/java/org/optaplanner/persistence/xstream/impl/domain/solution/XStreamSolutionFileIO.java
```diff
@@ -61,11 +62,19 @@ public String getInputFileExtension() {
...
+            return read(inputSolutionStream);
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Failed reading inputSolutionFile (" + inputSolutionFile + ").", e);
```
ge0ffrey: This one is serious: if there an XStreamException or an IOException during `read(InputStream)`, the error message won't mention the inputSolutionFile, which is important as a breadcrumb (to figure out what 's wrong for the user).
Why? that read(InputStream) throws an IllegalSargumentException and that isn't caught.

xstream exceptions are very common btw. Either invalid xml or <localSearch> that has a <unexistingElement> (there is no XSD).

Thinking further:
if there is such an xml parsing or element nesting error for an resource (not file), then we still want the resource name.

Suggestion to fix all of this:
1) Don't add `read(InputStream()`. Add `readResource(String resource)` instead. That also decreases the boilerplate in the NQueensCosntraintProviderTest

optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/AbstractAssertion.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.test.api.score.stream;
+
+public abstract class AbstractAssertion<Solution_, A extends AbstractAssertion<Solution_, A, V>, V
+        extends AbstractConstraintVerifier<Solution_, A, V>> {
```
ge0ffrey: Mind renaming `V` to `Verifier_` and `A` to `Assertion_`?

Do we really need this Foo<F extends Foo<F>? I did that for Score and I regret it.
Do we really need types A and V?

Ok to see this merges already as is and discuss on master.

optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/ConstraintVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+    SolutionDescriptor<Solution_> getSolutionDescriptor() {
+        return solutionDescriptor;
+    }
+
+    /**
+     * All subsequent calls to {@link #verifyThat(Function)} and {@link #verifyThat(ConstraintProvider)}
+     * will use {@link ConstraintStreamImplType#BAVET} instead of the default {@link ConstraintStreamImplType#DROOLS}.
+     * @return this
+     */
+    public ConstraintVerifier<Solution_> useBavet() {
```
ge0ffrey: Can do `withConstraintStreamImplType(ConstraintStreamImplType constraintStreamImplType)` like ScoreDirectorFactoryConfig#withConstraintStreamImplType instead?

ge0ffrey: Consistency + allows us to switch the default if we every want to.

PR #714, opened 2020-03-05
[DROOLS-4900] bump to poi 4

mariofusco:


mariofusco:
@triceo @ge0ffrey I'm migrating to poi 4 and at that point a test using that specific spreadsheet started to randomly failing. I didn't do anything other than opening the file and save it again and at that point the test became green. I think it was just saved in a somewhat inconsistent state.

ge0ffrey:
Thanks for the fix, @mariofusco

PR #715, opened 2020-03-06
PLANNER-1883 default SolutionCloner should not reflect inside enums

ge0ffrey:
Might be needed for Graal (Quarkus native mode). Performance enhancement in any case.


ge0ffrey:
GMTA

PR #716, opened 2020-03-09
PLANNER-1871 Merge consecutive filter()s into a single predicate

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/InnerConstraintFactory.java
```diff
@@ -41,6 +49,19 @@
...
+    default <A> BiConstraintStream<A, A> fromUniquePair(Class<A> fromClass, BiJoiner<A, A> joiner) {
+        MemberAccessor planningIdMemberAccessor = ConfigUtils.findPlanningIdMemberAccessor(fromClass);
+        if (planningIdMemberAccessor == null) {
+            throw new IllegalArgumentException("The fromClass (" + fromClass + ") has no member with a @"
+                    + PlanningId.class.getSimpleName() + " annotation,"
+                    + " so the pairs can not be made unique ([A,B] vs [B,A]).");
+        }
+        // TODO In Bavet breaks node sharing + involves unneeded indirection
+        Function<A, Comparable> planningIdGetter = (fact) -> (Comparable<?>) planningIdMemberAccessor.executeGetter(fact);
+        return from(fromClass).join(fromClass, lessThan(planningIdGetter), joiner);
```
triceo: There is a problem here.

Originally, the order of `joiners` in this code was `joiner, lessThan(planningIdGetter)`. However, that breaks if the user wants to do:

    fromUniquePair(..., Joiners.filtering(...))

Because we fail fast when `filtering()` joiner is anything but last. This could be considered a bug that I only just now found out about.

However, the other way around (putting `lessThan()` before `filtering()`) breaks Bavet, as Bavet only supports `lessThan()` when it's the last joiner.

The fail-fasts of `filtering()` and `lessThan()` in Bavet are mutually incompatible. Drools is fine.
What do we want to do about this?

ge0ffrey: I see 3 proposals:
A) Leave as is (bad idea, fromUniquePair(..., Joiners.filtering(...)) doesn't work for CS-D and CS-B)
B) Put lessThan() before filtering() - Is it a regression for Bavet? does this break any of the existing Bavet working examples? 
C) Like B), but add Bavet syntactic sugar to support filtering() to avoid any regressions of B.

triceo: B) Yes, it actually breaks many tests.
C) Could be doable.

ge0ffrey: I am -1 for B if and only if B has regressions. Otherwise I am ok with all.

triceo:
Jenkins please retest this.

PR #718, opened 2020-03-12
Make JaxbAdapter tests compatible on Windows machines

cuijulian:


cuijulian:
Jenkins please test this

rsynek:
The failure message on lines 74 and 123 won't format well on Windows for the very same reason. The test should be fixed, though.

cuijulian:
@rsynek Fixed! Thanks for spotting that.

ge0ffrey:
> The failure message on lines 74 and 123 won't format well on Windows for the very same reason. The test should be fixed, though.

Are you sure? I 've been using "\n" in error message all across optaplanner-core...

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
```diff
@@ -103,22 +106,25 @@
             throw new IllegalStateException("Marshalling or unmarshalling for input (" + input + ") failed.", e);
         }
         assertEquals(expectedScore, output.getScore());
+
+        String newLine = System.lineSeparator();
         String regex;
         if (expectedScore != null) {
-            regex = "\\{\n" // Opening bracket
```
ge0ffrey: Proposal B)
How about we use "\r?\n" instead of "\n"?

Proposal C) Why do we assert the formatting of the text? Is that needed?
Normally "\\s*" includes all whitespace, including \r and \n.
Does this regex work for those 2 lines?

```
regex = "\\{\\s*\"([\\w]+)\"\\s:\\s\\{"
 + ...
```

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
```diff
@@ -103,22 +106,25 @@
             throw new IllegalStateException("Marshalling or unmarshalling for input (" + input + ") failed.", e);
         }
         assertEquals(expectedScore, output.getScore());
+
+        String newLine = System.lineSeparator();
```
ge0ffrey: Using System.lineSeperator() always makes me cringe. Can't remember/motivate why though...

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
ge0ffrey: If C) works, can we go with that? I don't think we need to test for formatting.

"\s*" (which is escaped written as "\\s*"), means 0 or more whitespace characters.

Christopher-Chianelli: Or Proposal D): The official Java regex for newline since Java 8: "\R" (https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#lineending)

Christopher-Chianelli: Wouldn't using jackson ObjectMapper (or similar) make sense here? We don't really care about how the JSON looks like, just that the JSON has certain fields/properties (which we can check via ObjectMapper:
```
JsonNode parent= new ObjectMapper().readTree(json);
String score = parent.get("score").asText();
```

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
rsynek: I haven't found anything wrong about it. Even OpenJDK seems to be switching to it:
https://bugs.openjdk.java.net/browse/JDK-8198645

rsynek:
> > The failure message on lines 74 and 123 won't format well on Windows for the very same reason. The test should be fixed, though.
> 
> Are you sure? I 've been using "\n" in error messages all across optaplanner-core... If that's indeed an issue, we have some clean up to do in optaplanner. Anyone seen this happen on windows?

@ge0ffrey I did a little test on Windows 10 in VM. It turned out that it depends on who is consuming such a message:
- cmd line in Windows prints \n as new lines just fine
- if you redirect stdout to a file and open it in the Notepad, the new lines are lost

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
cuijulian: +1 to using `\R` for Java regex. What would we use for new line characters in the error messages?
-1 to using `ObjectMapper`, I don't think we should introduce Jackson dependencies since this code deals with the JAXB JSON standard. We do use `ObjectMapper` in the `optaplanner-persistence-jackson` module. If we do decide to use it, it's outside the scope of this PR.

Christopher-Chianelli: String.format(...) is a thing, and use "%n" (https://dzone.com/articles/java-string-format-examples) (note it varargs are not type safe, it basically `sprintf`).

How about using `JAXBContext` to create a new unmarshaller to read the json?

cuijulian: `%n` looks reasonable. Let's see if we can get consensus on which newline char to use. As for the rest, a bit of scope creep here, the purpose of this PR was to fix the broken Windows tests. We can address how we assert the Json output in a separate PR.

rsynek:
@cuijulian are we still missing an agreement on the newline formatting? Tests keep failing on Windows.

cuijulian:
@rsynek No official consensus yet. Since there were no objections to using `\R` for newline regex and `%n`, I will go with those. Both of those options get a +1 from me. I'm pushing those changes soon, so speak now or forever hold your peace (until merge that is) :)

cuijulian:
Jenkins please test this

rsynek:
@cuijulian you have my blessing; not that it would mean much :)

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
```diff
@@ -71,7 +71,8 @@
                     + "<([\\w\\-\\.]+)/>"; // Start and end of element
         }
         if (!xmlString.matches(regex)) {
-            fail("Regular expression match failed.\nExpected regular expression: " + regex + "\nActual string: " + xmlString);
+            fail(String.format("Regular expression match failed.%nExpected regular expression: " + regex +
+                                       "%nActual string: " + xmlString));
```
ge0ffrey: If you use `format`, don't use concatenate `+`: `format("... Expected regular expression (%s) ...", regex)`

optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/AbstractScoreJaxbXmlAdapterTest.java
ge0ffrey: The JSON format is whitespace (and therefore newline) agnostic. Whether one puts newlines and 4 space indents into a JSON file is a matter of taste (= code style). 
We need to test if it's formatted as JSON.
Do we need to test if the JSON "code style" is adhered too? I am not sure. If we don't, then we can just put `\\s*`, this will match all and any whitespacing (including linefeed and newline characters). If we do want to test it, then this is fine.
Either way, +1 to merge.

cuijulian: Hmm, good points. I don't see why we shouldn't test for JSON code style. If anything it's a more strict test, so I am in favour of keeping it. Are there any examples where testing for "style" rather than just "format" is not a good choice?

cuijulian:
Jenkins please test this

PR #722, opened 2020-03-17
PLANNER-1894 SolverManager.terminateEarly() fix

ge0ffrey:
https://issues.redhat.com/browse/PLANNER-1894

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
@@ -48,7 +48,7 @@ public final class DefaultSolverJob<Solution_, ProblemId_> implements SolverJob<
     private final Consumer<? super Solution_> finalBestSolutionConsumer;
     private final BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler;
 
-    private AtomicReference<SolverStatus> solverStatusReference;
+    private final AtomicReference<SolverStatus> solverStatusReference;
     private CountDownLatch terminatedLatch;
 
     private Future<Solution_> future;
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
@@ -21,6 +21,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -47,7 +48,7 @@ public final class DefaultSolverJob<Solution_, ProblemId_> implements SolverJob<
     private final Consumer<? super Solution_> finalBestSolutionConsumer;
     private final BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler;
 
-    private SolverStatus solverStatus; // Access synchronized
+    private AtomicReference<SolverStatus> solverStatusReference;
     private CountDownLatch terminatedLatch;
 
     private Future<Solution_> future;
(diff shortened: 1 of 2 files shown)


ge0ffrey:
> Manual synchronization can, in this case, be replaced by replacing `SolverStatus` by `AtomicReference<SolverStatus>` and using `AtomicReference`'s `getAndSet(...)` methods. I believe that this mechanism is preferable in modern Java, as it will perform better and it makes it impossible for an unsynchronized operation to happen with `solverStatus` by accident.

I doubt it can be replaced, because the "NOT_SOLVING" change lock must include the map removal and the latch opening.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -77,13 +78,19 @@ public ProblemId_ getProblemId() {
...
     @Override
     public Solution_ call() {
-        solverStatus = SolverStatus.SOLVING_ACTIVE;
+        synchronized (this) {
+            if (solverStatus != SolverStatus.SOLVING_SCHEDULED) {
+                // This job has been canceled before it started
+                return problemFinder.apply(problemId);
+            }
+            solverStatus = SolverStatus.SOLVING_ACTIVE;
+        }
```
ge0ffrey: @triceo I actually originally tried an AtomicRefence, for the reasons you mentioned.
Trouble was the code above versus the terminateEarly() code.
That latter did change drastically, so maybe there's a way to reintroduce the AtomicReference, let me try.

triceo:
```suggestion
            SolverStatus previousSolverStatus = solverStatus.getAndSet(SOLVING_ACTIVE);
            if (previousSolverStatus != SolverStatus.SOLVING_SCHEDULED) {
                // This job has been canceled before it started
                return problemFinder.apply(problemId);
            }
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -96,12 +103,16 @@ public Solution_ call() {
...
-            solverStatus = SolverStatus.NOT_SOLVING;
-            terminatedLatch.countDown();
+            solvingTerminated();
         }
     }
 
+    private synchronized void solvingTerminated() {
+        solverManager.getProblemIdToSolverJobMap().remove(problemId);
+        solverStatus = SolverStatus.NOT_SOLVING;
+        terminatedLatch.countDown();
```
triceo:
```suggestion
        SolverStatus previousSolverStatus = solverStatus.getAndSet(NOT_SOLVING);
        if (previousSolverStatus != NOT_SOLVING) {
            solverManager.getProblemIdToSolverJobMap().remove(problemId);
            terminatedLatch.countDown();
        }
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -116,20 +127,26 @@ public Solution_ call() {
...
+            if (solverStatus == SolverStatus.SOLVING_SCHEDULED) {
+                solvingTerminated();
+            } else if (solverStatus == SolverStatus.SOLVING_ACTIVE) {
+                // Indirectly triggers solvingTerminated()
+                solver.terminateEarly();
+            } else if (solverStatus == SolverStatus.NOT_SOLVING) {
+                // Do nothing, solvingTerminated() already called
+            } else {
+                throw new IllegalStateException("Unsupported solverStatus (" + solverStatus + ").");
             }
```
triceo: This should be replacable by `switch (solverStatus.get())`, as all the actual state mutation happens in `solverTerminated()`/`solverTerminateEarly()` as I've suggested above.

ge0ffrey:
@triceo Agreed, adjusting as you suggested. Good catch!

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -47,8 +48,9 @@
     private final Consumer<? super Solution_> finalBestSolutionConsumer;
     private final BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler;
 
-    private volatile SolverStatus solverStatus;
+    private AtomicReference<SolverStatus> solverStatusReference;
```
triceo: Looks like this one could be `final`.

PR #723, opened 2020-03-18
PLANNER-1907 Add daemon to SolverProperties

yurloc:
This allows to set daemon mode through application.properties or system
properties.

Example application.properties:
```
optaplanner.solver.daemon=true
```

### DIFF
--- a/optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
+++ b/optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
@@ -28,6 +28,12 @@ public class SolverProperties {
      */
     private EnvironmentMode environmentMode;
     /**
+     * Enable daemon mode. In daemon mode, non-early termination pauses the solver instead of stopping it,
+     * until the next problem fact change arrives. This is often useful for real-time planning.
+     * Defaults to {@code false}.
+     */
+    private Boolean daemon;
+    /**
      * Enable multithreaded solving for a single problem, which increases CPU consumption.
      * Defaults to {@value SolverConfig#MOVE_THREAD_COUNT_NONE}.
      * Other options include {@value SolverConfig#MOVE_THREAD_COUNT_AUTO}, a number
@@ -48,6 +54,14 @@ public class SolverProperties {
 
     public void setEnvironmentMode(EnvironmentMode environmentMode) {
         this.environmentMode = environmentMode;
+    }
+
+    public Boolean getDaemon() {
+        return daemon;
+    }
+
+    public void setDaemon(Boolean daemon) {
+        this.daemon = daemon;
     }
 
     public String getMoveThreadCount() {
--- a/optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
+++ b/optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
@@ -450,6 +450,11 @@ Defaults to `solverConfig.xml`.
 If this property isn't specified, that file is optional.
 optaplanner.solver.environment-mode::
 Enable runtime assertions to detect common bugs in your implementation during development.
+optaplanner.solver.daemon::
+Enable <<daemon,daemon mode>>.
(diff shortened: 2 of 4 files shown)


optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
```diff
@@ -27,6 +27,10 @@
      * Defaults to {@link EnvironmentMode#REPRODUCIBLE}.
      */
     private EnvironmentMode environmentMode;
+    /**
+     * Enable daemon mode (disabled by default).
+     */
```
ge0ffrey: Can we make that more descriptive of what daemon means? How about something like:

```
In daemon mode, non-early termination pauses the solver instead of stopping it, until the next problem fact change arrives. This is often useful for real-time planning.
Defaults to {@code false}.

```

ge0ffrey:
Ow wait, it also needs to be added in the docs in the inventory of all properties:
https://docs.optaplanner.org/latestFinal/optaplanner-docs/html_single/#integrationWithSpringBoot

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/SolverProperties.java
yurloc: I like the idea of mentioning real-time planning. That gives a hint to someone that comes across this accidentally and has no idea what daemon mode could mean.

On the other hand I'd suggest to keep the documentation short and simple. This is not Javadoc. This is a text that will be used as a property description (without any processing, so things like `{@link}` that we use in Javadoc will appear there unprocessed). See https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-custom-starter-configuration-keys

yurloc: I would shorten it to
```
Enable daemon mode. This is often useful for real-time planning.
Disabled by default.
```
Wdyt?

ge0ffrey: "Disabled by default", the other javadocs and docs use "Defaults to ...", let's be consistent.

I am ok with the shorter description, I agree that the hint to "real-time planning" is what matters.

If the usage of @link etc in the javadoc for such properties causes issues somewhere for spring boot, like in IntelliJ auto-completion in application.properties, please create a jira with a screenshot (and if you're into it a PR to fix that across all our properties ;-)

As for our docs: our docs are not javadoc, but we need a listing of all sprng-boot-starter properties somewhere (same goes for quarkus). So far just putting them in the docs is the best I figured out: refactors welcome. Meanwhile do add it so it doesn't go out of sync.

ge0ffrey: Merged, I saw you went back to the longer description. Note that the shorter was was fine too (just was asking for the "Default to " consistency with others).

Thanks Jiri!

yurloc: Yes. In IDEA there are two modes of displaying the configuration key description.

One is in the hint window when you're choosing from available properties. In this case it only shows the first sentence. So it is good to have the first sentence as simple as possible.

optaweb-vehicle-routing  ~-src-github com-kiegroup-optaweb-vehicle-routing  -  -optaweb-vehicle-routing-backend-src-main-resources-application properties  Back-end _002

The second mode is when you display the documentation window for a property that is used in `application.properties`. In this case it displays the whole description and there's a plenty of space so it's actually no problem to display more details. Therefore I've decided to go with your proposal.

optaweb-vehicle-routing  ~-src-github com-kiegroup-optaweb-vehicle-routing  -  -optaweb-vehicle-routing-backend-src-main-resources-application properties  Back-end _003

PR #726, opened 2020-03-20
fix: fix issue:https://issues.redhat.com/browse/PLANNER-561

rihkddd:
A simple way to avoid issue PLANNER-561.


optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/forager/finalist/HighestScoreFinalistPodium.java
```diff
@@ -47,7 +48,7 @@ public void addMove(LocalSearchMoveScope moveScope) {
         }
         Score moveScore = moveScope.getScore();
         int scoreComparison = doComparison(moveScore);
-        if (scoreComparison > 0) {
+        if (scoreComparison > 0 || finalistList.size() > finalistMaxSize) {
```
ge0ffrey, line 51: Fixing this slightly different:
We can't accept a move just because some sort of hard threshold has been reached. That hard number is both too low and to big.
But we can limit the growth of the finalistList. Doing that.

ge0ffrey:
Replaced by alternative fix here: https://github.com/kiegroup/optaplanner/pull/763

Thanks again for raising this issue and pointing out the cause!

PR #728, opened 2020-03-23
School timetable input output

ge0ffrey:


manstis:
From my perspective the whole diagram is wrong.

All lessons should be assigned to "dad" in "the front room"!

ge0ffrey:
> The left "leg" of the "Problem" arrow is not very symmetrical to the right "leg", in its thickness.
> Other than that, looks good.

I tried fixing that, but it only gets worse. Good enough for now.

PR #731, opened 2020-03-30
Reproducer for PLANNER-1934

yurloc:
Not to be merged.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
```diff
@@ -87,8 +87,9 @@ public DroolsConstraintSessionFactory(SolutionDescriptor<Solution_> solutionDesc
         compiledRuleToConstraintMap.forEach((compiledRule, constraint) -> {
             // In constraint verifier API, we disregard constraint weights.
             // Yet we need to have them, otherwise the constraint would be ignored, so we add a synthetic weight.
-            Score<?> constraintWeight = workingSolution == null ?
-                    oneSoftestScore :
+            // DELETE this star ----------v   for a quick fix
+            Score<?> constraintWeight = /**/ workingSolution == null ?
```
yurloc: I've spent some time debugging to find out the reason why the actual score is 0/95soft instead of -95/0hard.

This should help save some time looking for the connection.

triceo:
Please see https://github.com/kiegroup/optaplanner/pull/735, @yurloc .

PR #732, opened 2020-04-01
schoolTimetablingArchitecture

ge0ffrey:


ge0ffrey:
@yurloc The schoolTimetable svg's are both for the quarkus and spring-boot integration. Both those guides use it. Our slides also use it. It just made sense to put it here for now, to centralize those svg's at least. There's a good chance the spring-boot one comes back here btw, as it's not getting accepted downstream.

ge0ffrey:
Mind if I merge?

PR #734, opened 2020-04-07
PLANNER-1942 Descriptive NPE when constraint configuration is null

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintWeightDescriptor.java
```diff
@@ -75,10 +77,14 @@ public MemberAccessor getMemberAccessor() {
...
-                .getConstraintConfigurationMemberAccessor();
+        SolutionDescriptor<Solution_> solutionDescriptor = constraintConfigurationDescriptor.getSolutionDescriptor();
+        MemberAccessor constraintConfigurationMemberAccessor =
+                solutionDescriptor.getConstraintConfigurationMemberAccessor();
         return (Solution_ solution) -> {
-            Object constraintConfiguration = constraintConfigurationMemberAccessor.executeGetter(solution);
+            Object constraintConfiguration =
+                    Objects.requireNonNull(constraintConfigurationMemberAccessor.executeGetter(solution),
+                            "Constraint configuration provider (" + constraintConfigurationMemberAccessor +
+                                    ") is null.");
```
ge0ffrey, line 87:
```suggestion
                                    ") returns null.");
```

the memberAccessor isn't null, it returns null.

ge0ffrey:
@DuncanDoyle This fix by Lukas improves the error message to avoid burning new users :)

PR #735, opened 2020-04-07
PLANNER-1934 Set custom constraint weights only for configurable cons…

triceo:
…traints

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraint.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraint.java
@@ -49,7 +49,7 @@ public abstract class AbstractConstraint<Solution_, ConstraintFactory extends In
         if (isConstraintWeightConfigurable && workingSolution == null) {
             /*
              * In constraint verifier API, we allow for testing constraint providers without having a planning solution.
-             * However, constraint weights may be configurable and in that case the solution is be required to read the
+             * However, constraint weights may be configurable and in that case the solution is required to read the
              * weights from.
              * For these cases, we set the constraint weight to the softest possible value, just to make sure that the
              * constraint is not ignored.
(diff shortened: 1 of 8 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractConstraint.java
```diff
@@ -32,17 +32,31 @@
...
         this.scoreImpactType = scoreImpactType;
+        this.isConstraintWeightConfigurable = isConstraintWeightConfigurable;
     }
 
-    public Score<?> extractConstraintWeight(Solution_ workingSolution) {
+    public final Score<?> extractConstraintWeight(Solution_ workingSolution) {
+        if (isConstraintWeightConfigurable && workingSolution == null) {
+            /*
+             * In constraint verifier API, we allow for testing constraint providers without having a planning solution.
+             * However, constraint weights may be configurable and in that case the solution is be required to read the
```
yurloc, line 52: Extra "be" in the sentence.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/score/NQueensConstraintProviderTest.java
```diff
@@ -129,9 +129,20 @@ private static NQueens readSolution(String resource) throws IOException {
...
     @Test
-    public void constraintProviderIntegrationTest() throws IOException {
+    public void constraintProviderSolutionBasedIntegrationTest() throws IOException {
         constraintVerifier.verifyThat(constraintProvider)
                 .given(readSolution("256queensScore-30.xml"))
                 .scores(SimpleScore.of(-30));
     }
+
+    @Test
+    public void constraintProviderFactBasedIntegrationTest() {
```
ge0ffrey, line 139: (soft) 0) do we need the prefix "constraitnProvider"? It in the class ...ConstraintProviderTest
1) What method are we testing? Answer: "given". I 'd start with that.
2) I am not a fan of having the word "test" in methods that already have @Test annotation.
3) Today I understand why this is an "integration" test. Because it does more integration than the other unit tests that just test one constraint. In 6 months from now I'll have forgotten that. Can we naming it something like "multipleConstraints"?

Therefore, these method name suggestions for these 2 tests:
givenSolutionMultipleConstraints
givenFactsMultipleConstraints

PR #738, opened 2020-04-09
School timetabling slides

ge0ffrey:


rsynek:
- the .svg files are missing the text when opened in a browser (at least for me), but Inkscape shows it correctly.
- why removing the schoolTimetablingArchitecture.png, if for the class diagram we keep the .png file?

ge0ffrey:
Inkscape files use more advanced SVG features that don't work in browser or linux default svg viewer. As we use the SVG purely to not loose the source, but always use png's in publishing formats (adoc), that is workable. but indeed not ideal.

ge0ffrey:
Good catch: schoolTimetablingArchitecture.png shouldn't be removed

ge0ffrey:
Not sure why github diff was saying it was removed, because it wasn't on that branch and hasn't been since merging. Proof: https://github.com/kiegroup/optaplanner/blob/master/optaplanner-docs/src/main/asciidoc/QuickStart/SchoolTimetabling/schoolTimetablingArchitecture.png

PR #739, opened 2020-04-09
PLANNER-1929 Constraint Verifier properly supports match weights

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
```diff
@@ -102,6 +102,22 @@
...
+     * should have no impact on the returned scale/precision.
+     * <p>
+     * Allows division by zero if and only if both levels are zero, in which case the resulting level will also be zero.
+     * Otherwise {@link ArithmeticException} is thrown.
+     * @param divisor value by which this Score is to be divided
+     * @return this / divisor
+     */
+    default Score_ divide(Score_ divisor) {
+        throw new UnsupportedOperationException();
+    }
```
triceo, line 119: Public API change.
This is a default method, so that this is 100 % backwards compatible.

triceo:
OK, this needs more work.

optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/SingleConstraintAssertion.java
```diff
@@ -39,9 +43,27 @@
...
+        Score<?> totalMatchWeightedScore = constraintMatchTotalMap.values().stream()
+                .map(s -> s.getScore().divide(s.getConstraintWeight()))
+                .reduce(zero, Score::add);
+        // Each level of the resulting score now has to be the same number, the matchWeight.
+        List<Number> matchWeightsFound = Arrays.stream(totalMatchWeightedScore.toLevelNumbers())
+                .distinct()
+                .collect(Collectors.toList());
+        if (matchWeightsFound.size() != 1) {
+            throw new IllegalStateException("Programming error: only one match weight should have been found ("
+                    + matchWeightsFound + ").");
```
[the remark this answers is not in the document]
triceo: The score here is `matchWeight * constraintWeight * matchCount`.
At this point, I have already divided by it `constraintWeight`.
What remains is therefore `matchWeight * matchCount`. And therefore the value in every level of the score must be the same.

I believe there already is an explanatory comment in the code.
(That said, you correctly point out a bug in case where one level of the score is `0`. Fixing that.)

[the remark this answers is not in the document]
triceo: For example, consider `.penalize("something", HardSoftScore.of(1, 2), something -> 2)`.
Further consider that this penalty is only applied once. In this case:

- `matchCount = 1`
- `matchWeight = 2`
- `constraintWeight = 1hard/2soft`

And therefore, from the end:
- `constraintMatchScore = 2hard/4soft`
- `constraintMatchScore / constraintWeight = 2hard/2soft`
- `matchWeightsFound = [2]`
- `result = 2`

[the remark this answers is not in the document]
yurloc: Beautiful explanation, thanks! Now it makes perfect sense.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractScore.java
```diff
@@ -34,6 +35,74 @@
 
     protected static final String INIT_LABEL = "init";
 
+    protected static int divideScoreLevel(int score, int scoreDivisor) {
+        return divideScoreLevel(score, scoreDivisor, null);
+    }
+
+    protected static int divideScoreLevel(int score, int scoreDivisor, String maybeLabel) {
+        String label = maybeLabel == null ? "" : maybeLabel;
```
yurloc, line 43: Nitpicking: You have just moved one instance of `SimpleScore`'s label to another place. What I was pointing out is that `""` is the score label of `SimpleScore` and I noticed the the string literal (`""`) is used multiple times in the `SimpleScore` class and therefore should be replaced by a constant (provided that all `""` occurrences have the same meaning, which I presume they do).

In another words, what would happen if we decided that `SimpleScore` will get a new label of `simp`? We would replace two `""` occurrences in `SimpleScore` with `"simp"` and we would probably miss this instance (on line 43).

triceo: What would happen is that the exception message would continue to say "X" instead of "Xsimp". 
I really don't consider this a problem - it was a problem before, where I was propagating something that should have been a constant. I don't anymore.

yurloc: OK, that's another, valid point of view that I didn't think of.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/AbstractScoreDefinition.java
```diff
@@ -30,6 +32,38 @@
...
+    protected static long divide(long dividend, long divisor) {
+        return (long) Math.floor(divide(dividend, (double) divisor));
+    }
+
+    protected static double divide(double dividend, double divisor) {
+        return dividend / divisor;
+    }
+
+    protected static BigDecimal divide(BigDecimal dividend, BigDecimal divisor) {
+        return dividend.divide(divisor, dividend.scale() - divisor.scale(), RoundingMode.FLOOR);
```
triceo, line 64: My spider sense is tingling, expecting comments here.
See table "Preferred Scales for Results of Arithmetic Operations" in [BigDecimal Javadoc.](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html)

ge0ffrey: This is fine. Our long existing method `Score.divide(boolean)` has a javadoc that says "When rounding is needed, it should be floored (as defined by {@link Math#floor(double)})." This is consistent with that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/ScoreDefinition.java
```diff
@@ -152,4 +152,14 @@ default boolean isNegativeOrZero(S score) {
      */
     S buildPessimisticBound(InitializingScoreTrend initializingScoreTrend, S score);
 
+    /**
+     * Return a Score whose every level is the result of dividing the matching levels in this and the divisor.
```
ge0ffrey, line 156: javadoc link to Score

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/buildin/hardsoft/HardSoftScoreDefinitionTest.java
```diff
@@ -89,14 +89,18 @@ public void buildPessimisticBoundOnlyUp() {
     }
 
     @Test
-    public void buildPessimisticBoundOnlyDown() {
+    public void divideBySanitizedDivisor() {
```
ge0ffrey, line 92: Did the buildPessimisticBoundOnlyDown() test get removed?

PR #740, opened 2020-04-11
avoid ConstraintProvider creation for tests

ge0ffrey:


optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/ConstraintVerifier.java
```diff
@@ -29,64 +28,70 @@
...
         this.solutionDescriptor = solutionDescriptor;
     }
 
     /**
      * Entry point to the API.
+     * @param constraintProvider never null, {@link PlanningEntity} used by the {@link PlanningSolution}
      * @param planningSolutionClass never null, {@link PlanningSolution}-annotated class associated with the constraints
-     * @param firstPlanningEntityClass never null, {@link PlanningEntity} used by the {@link PlanningSolution}
-     * @param otherPlanningEntityClasses optional, extra entity classes if {@link PlanningSolution} uses more than one
+     * @param entityClasses never null, at least one, {@link PlanningEntity} types used by the {@link PlanningSolution}
```
ge0ffrey, line 46: @triceo I merged these because of API's that would want to call this method.

ge0ffrey: (and for consistency with the methods on SolutionDescriptor)

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/optional/score/VehicleRoutingConstraintProvider.java
```diff
@@ -41,7 +41,7 @@
     // Hard constraints
     // ************************************************************************
 
-    private Constraint vehicleCapacity(ConstraintFactory factory) {
+    public Constraint vehicleCapacity(ConstraintFactory factory) {
```
triceo, line 44: Any reason why this one is public, yet all the others are protected?

PR #744, opened 2020-04-13
PLANNER-1876 Create a code style for OptaPlanner project

rsynek:
This PR introduces the Quarkus code style into the OptaPlanner project. By default, the codebase is automatically formatted during a build. In a CI environment, a validation takes place instead.

### DIFF
--- a/ide-configuration/eclipse.importorder
+++ b/ide-configuration/eclipse.importorder
@@ -1,0 +1,7 @@
+#Organize Import Order
+#Wed Jan 23 12:03:29 AEDT 2019
+0=java
+1=javax
+2=org
+3=com
+
--- a/pom.xml
+++ b/pom.xml
@@ -336,6 +336,7 @@
       </plugins>
     </pluginManagement>
     <plugins>
+      <!-- TODO: enable code formatting plugins once the team makes the switch.
       <plugin>
         <groupId>net.revelc.code.formatter</groupId>
         <artifactId>formatter-maven-plugin</artifactId>
@@ -344,6 +345,7 @@
         <groupId>net.revelc.code</groupId>
         <artifactId>impsort-maven-plugin</artifactId>
       </plugin>
+      -->
     </plugins>
   </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,11 @@
   <inceptionYear>2006</inceptionYear>
 
   <properties>
+    <!-- override in each parent module in the hierarchy -->
+    <format.directory>${project.basedir}/../ide-configuration</format.directory>
+    <formatter.skip>false</formatter.skip>
+    <formatter.goal>format</formatter.goal>
+    <impsort.goal>sort</impsort.goal>
     <jacoco.agent.argLine/>
     <spotbugs.failOnViolation>true</spotbugs.failOnViolation>
   </properties>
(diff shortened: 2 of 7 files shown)


ge0ffrey:
@rsynek Doing a diagonal scan of theses changes, I am fine with them.

Proposal to move forward in the 3 phases:
1) create a PR to add in the new files and commented out maven plugin during the build. Let's merge that first.  That shouldn't take long.
2) Get everyone in our team to configure his IDE accordingly. This can go in parallel.
3) Get all PR's merged, **freeze** the repo, apply it (and nothing else!), **unfreeze**. We'll need to announce a freeze date.

triceo:
Also, @rsynek, see Sonar - apparently these changes uncover a problem that would drop our rating to C.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
```diff
@@ -84,6 +85,7 @@
     /**
      * Reads an XML benchmark configuration from the classpath
      * and uses that {@link PlannerBenchmarkConfig} to do an aggregation.
+     *
```
yurloc, line 88: Trailing space.

yurloc: OK, this seems to be done by the Eclipse formatter. There's nothing we can do about it, I guess.

rsynek: Nice catch. This is truly done by the formatter - both the IDEA and maven plugin. If there is already an empty line before the `@param` mark, no space is added. However, when the formatter needs to include the empty line, it does include the trailing space as well.

Unfortunately, I don't see any systematic way of avoiding this. I quickly checked the Quarkus codebase and it has the same issue too.

yurloc: I checked again and crossed out my previous comment. After enabling Eclipse formatter plugin and formatting the code in my IDEA, empty Javadoc lines were inserted but without the trailing space.

It could be because I have this enabled:
Settings_005

I recommend this setting to everyone.

rsynek: +1 for using this setting. On the other hand, it does not prevent such a trailing space from occurring. The maven formatter plugin adds them too and so does Eclipse IDE.

pom.xml
```diff
@@ -279,8 +298,55 @@
...
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
+        <plugin>
+          <groupId>net.revelc.code</groupId>
+          <artifactId>impsort-maven-plugin</artifactId>
+          <version>1.3.2</version>
+          <configuration>
+            <groups>java.,javax.,org.,com.</groups>
```
yurloc, line 322: @rsynek This duplicates the content of `eclipse.importorder` - why?

rsynek: There is no way of passing the file into this plugin and at the same time, the formatter plugin does not care about imports. So yes, there is a duplication of the imports between the maven plugin use case and the IDE use case.

ide-configuration/eclipse.importorder
```diff
@@ -0,0 +1,7 @@
+#Organize Import Order
+#Wed Jan 23 12:03:29 AEDT 2019
+0=java
+1=javax
+2=org
+3=com
```
yurloc, line 6: What about packages starting with something else, like ch, it, cz, io, etc.?

rsynek: Packages that are not mentioned here go after the mentioned ones and are sorted lexicographically.

yurloc: Thanks!

PR #750, opened 2020-04-17
Fix a broken link in readme

rsynek:


README.adoc
```diff
@@ -57,7 +57,7 @@ If you want your pull requests (or patches) to be merged into master, please res
 
 OptaPlanner has adopted the https://github.com/quarkusio/quarkus[Quarkus] code style, enforces it, and automatically formats code during the build.
 To setup your IDE, please see the
-<<ide-configuration/ide-configuration.adoc, IDE Setup Instructions>>.
+<<ide-configuration/ide-configuration.adoc#, IDE Setup Instructions>>.
```
ge0ffrey, line 60: if this is an adoc, shouldn't that be `link:ide-configuration/ide-configuration.adoc[IDE Setup Instructions]` instead?

rsynek: According to https://asciidoctor.org/docs/asciidoc-writers-guide/#links-to-relative-files, it should not. As far as I understand it, the `link:` syntax should be used only for other kinds of resources than .adoc files.

PR #756, opened 2020-04-25
Fix balanceCost constraint

ge0ffrey:
I rewrote the test because it was testing at overcapacity. Also wanted to use the "different digits" trick so it was easier to follow what's going on. The test still failed after I refactored it, now it succeeds.

@rsynek I am not 100% sure that the refactored test counts as much as the previous one, please validate and point out if it's not :)

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/machinereassignment/solver/score/MachineReassignmentConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/machinereassignment/solver/score/MachineReassignmentConstraintProvider.java
@@ -16,10 +16,6 @@
 
 package org.optaplanner.examples.machinereassignment.solver.score;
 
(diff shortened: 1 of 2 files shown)


ge0ffrey:
@triceo @rsynek 
BTW, we should encourage use of the "different digits" trick more in our test cases.
It goes like this: **the tests number's arithmetic should be easy**. For example: `1000 + 200 + 30 + 4` is easy. But `7 + 5 + 3` is damn hard. Easy arithmetic makes it much, much easier to read and validate the test.

triceo:
If Radek approves, I am OK with this.
That said, I would be surprised if Radek had nothing to say about this - his own version of the `balanceCost` constraint had several `groupBy()` calls and still didn't do what it was supposed to.

ge0ffrey:
I didn't confirm equality with scoreDrl, especially if hard constraints are broken there might be a difference. @rsynek Would you mind confirming that part?

PR #757, opened 2020-04-25
Rename "ScoreRules.drl" to "Constraint.drl"

ge0ffrey:


optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
```diff
@@ -346,7 +346,7 @@
                             <div class="tab-pane active" id="summary_scoreCalculationSpeed">
                                 <h3>Score calculation speed summary</h3>
                                 <p>
-                                    Useful for comparing different score calculators and/or score rule implementations
+                                    Useful for comparing different score calculators and/or constraints implementations
```
triceo, line 349:
```suggestion
                                    Useful for comparing different score calculators and/or constraint implementations
```

optaplanner-benchmark/src/main/resources/org/optaplanner/benchmark/impl/report/benchmarkReport.html.ftl
```diff
@@ -411,7 +411,7 @@
                             <div class="tab-pane" id="summary_worstScoreCalculationSpeedDifferencePercentage">
                                 <h3>Worst score calculation speed difference percentage</h3>
                                 <p>
-                                    Useful for comparing different score calculators and/or score rule implementations
+                                    Useful for comparing different score calculators and/or constraints implementations
```
triceo, line 414:
```suggestion
                                    Useful for comparing different score calculators and/or constraint implementations
```

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/drools/ProblemFactCollectionProperty.java
```diff
@@ -20,24 +20,25 @@
...
+ * A problem fact must not change during solving (except through a {@link ProblemFactChange} event).
  * <p>
- * The problem facts will be added as facts in the {@link KieSession} of the {@link DroolsScoreDirector},
- * so the score rules can use them.
+ * A {@link ConstraintProvider} builds constraints from such problem facts
+ * with {@link ConstraintFactory#from(Class)}.
  * <p>
  * Do not annotate {@link PlanningEntity planning entities} as problem facts:
- * they are automatically inserted into the {@link KieSession}.
+ * they are automatically available as facts with {@link ConstraintFactory#from(Class)}.
```
triceo, line 41: This is not ideal either way. The original Javadoc assumed that we always use DRL, this one assumes we always use Constraint Streams. Maybe it should be more generic?

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/drools/ProblemFactProperty.java
```diff
@@ -19,25 +19,26 @@
...
- * so the score rules can use it.
+ * A {@link ConstraintProvider} builds constraints from such a problem fact
+ * with {@link ConstraintFactory#from(Class)}.
  * <p>
- * Do not annotate {@link PlanningEntity planning entities} or a {@link ConstraintConfiguration planning paramerization}
- * as problem facts: they are automatically inserted into the {@link KieSession}.
+ * Do not annotate a {@link PlanningEntity planning entity} or a {@link ConstraintConfiguration planning paramerization}
+ * as a problem fact: they are automatically available as facts with {@link ConstraintFactory#from(Class)}.
  * @see ProblemFactCollectionProperty
  */
```
triceo, line 43: Dtto.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/EnvironmentMode.java
```diff
@@ -47,7 +47,7 @@
     FULL_ASSERT,
     /**
      * This mode turns on several assertions (but not all of them)
-     * to fail-fast on a bug in a {@link Move} implementation, a score rule, the rule engine itself or something else
+     * to fail-fast on a bug in a {@link Move} implementation, a constraint, the engine itself or something else
      * at a horrible performance cost.
```
triceo, line 51:
```suggestion
     * to fail-fast on a bug in a {@link Move} implementation, a constraint, the engine itself or something else
     * at an overwhelming performance cost.
```

Somehow appears more professional to me.

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/drools/ProblemFactCollectionProperty.java
ge0ffrey: Given the philosophy "ConstraintStreams First, Score DRL Second", we can only skip Score DRL if it's not crucial. This is indeed a crucial concept - you're right. Let me see what I can make of it.

ge0ffrey: I just hate importing KieSession here. I want to avoid that.

PR #758, opened 2020-04-27
Machine reassignment snapshot test

rsynek:
Also removes insertion of unnecessary facts from the balanceCost tests.


rsynek:
@ge0ffrey Unfortunately, the comparison with DRL is not possible due to the accumulate function in machineUsage rule. It ends up with 0 (in contrast to a terminated stream in CSP) for ununitialized process assignment.

PR #763, opened 2020-04-28
PLANNER-561 Fix OutOfMemoryException in finalistList

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/forager/finalist/AbstractFinalistPodium.java
```diff
@@ -32,6 +32,8 @@
  */
 public abstract class AbstractFinalistPodium extends LocalSearchPhaseLifecycleListenerAdapter implements FinalistPodium {
 
+    protected static final int FINALIST_LIST_MAX_SIZE = 1_024_000;
```
triceo, line 35: Any particular reason why so many finalists would be a good idea?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/forager/finalist/AbstractFinalistPodium.java
```diff
@@ -44,6 +46,19 @@ public void stepStarted(LocalSearchStepScope stepScope) {
...
 
+    protected void clearAndAddFinalist(LocalSearchMoveScope moveScope) {
+        finalistList.clear();
+        finalistList.add(moveScope);
+    }
+
+    protected void addFinalist(LocalSearchMoveScope moveScope) {
+        if (finalistList.size() >= FINALIST_LIST_MAX_SIZE) {
+            // Avoid unbounded growth and OutOfMemoryException
+            return;
```
triceo, line 57: Wouldn't it make sense to instead remove the least recent finalist? Or perhaps the worst finalist?

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/forager/finalist/AbstractFinalistPodium.java
ge0ffrey: No, but with a normal amount of memory we can handle this.
Figuring out the MAX_SIZE on which these so many finalist would be 100% of the time a bad idea, is impossible.
It would be a good guesstimate at best, so if we get it wrong: an potential risk for silent, surprising misbehaviour in how the algorithm reacts in certain conditions. 
This MAX_SIZE isn't meant to distinguish between good and bad. It's to distinguish between bad without OOME trouble and bad with OOME.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/decider/forager/finalist/AbstractFinalistPodium.java
ge0ffrey: All moves in the finalistList have the same score (which is the reason that in most cases, it's typically size 1 to at most 100). There is no way to distinguish out the worst one.

I doubted about ignoring the oldest or the newest.
Removing the oldest in an ArrayList (not ArrayDeque) has a big0(n) perf cost. Ignoring the newest doesn't.
Both the first move evaluated and the last move evaluated in the same step with the same score, have equal right to be selected - recency doesn't count (there is no changes in the steps).
So I went for ignoring the newest.

PR #764, opened 2020-04-28
PLANNER-1003 Migrate tests to JUnit 5 API

yurloc:
## Depends on PR
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1288

## Checklist before undrafting
- [x] review executed test counts (on Jenkins)
  OK. **1 successful** test added (added missing `@Test` on a method). **381 skipped** tests removed (by reducing N param. test executions to 1 skipped execution on **turtle tests**).
- [x] review test names of dynamic and parameterized tests
  Almost OK. Display names of dynamic and parameterized test replaced correctly but the display names are not reflected in surefire report and therefore are not visible on Jenkins. This seems to be an issue on Surefire's side.
- [x] improve dynamic tests hierarchy by using `dynamicContainer()`
  Applied to `AbstractPhaseTest` that uses two-dimensional matrix (`unsolvedFileName` x `solverFactoryParam`).


yurloc:
Jenkins retest this.

optaplanner-core/pom.xml
```diff
@@ -93,6 +93,27 @@
...
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-migrationsupport</artifactId>
+      <scope>test</scope>
+    </dependency>
```
triceo, line 111: Ideally, the `@Rule`s that require this to be included would be gone, too.
Some are easier to replace than others.

rsynek: I think it can be addressed in a separate PR to avoid holding this one open for too long.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/buildin/hardmediumsoftbigdecimal/HardMediumSoftBigDecimalScoreTest.java
```diff
@@ -73,10 +74,12 @@ public void testToString() {
...
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void parseScoreIllegalArgument() {
-        HardMediumSoftBigDecimalScore.parseScore("-147.2");
-        HardMediumSoftBigDecimalScore.parseScore("-147.2hard/-258.3soft");
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> HardMediumSoftBigDecimalScore.parseScore("-147.2"));
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> HardMediumSoftBigDecimalScore.parseScore("-147.2hard/-258.3soft"));
```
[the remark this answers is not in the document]
rsynek: jUnit4 does not care - if at least one of the expressions throws an exception, the test passes. Now the test can actually work.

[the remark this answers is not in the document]
yurloc: Yes that is the drawback of the JUnit 4 approach. The test will pass no matter the line at which the exception is thrown. In this case, both statements throw but the second one was never reached.

yurloc:
Jenkins retest this.

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/decider/acceptor/simulatedannealing/SimulatedAnnealingAcceptorTest.java
```diff
@@ -102,11 +103,12 @@ public void lateAcceptanceSize() {
...
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void negativeSimulatedAnnealingSize() {
         SimulatedAnnealingAcceptor acceptor = new SimulatedAnnealingAcceptor();
-        acceptor.setStartingTemperature(HardMediumSoftScore.parseScore("1, -1, 2"));
-        acceptor.phaseStarted(null);
+        // FIXME this tests score parsing
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> acceptor.setStartingTemperature(HardMediumSoftScore.parseScore("1, -1, 2")));
```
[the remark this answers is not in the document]
yurloc: It was easier to add the fixme note than to make a decision immediately.

I've just reviewed the test and I think I can fix it. Thanks for pointing it out!

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/tracking/NQueensConstructionHeuristicTrackingTest.java
```diff
@@ -37,27 +36,15 @@
 import org.optaplanner.examples.nqueens.domain.NQueens;
 import org.optaplanner.examples.nqueens.persistence.NQueensGenerator;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertNotNull;
```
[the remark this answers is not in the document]
triceo: I may be wrong here, but I ran into something similar elsewhere. JUnit 5, in combination with Surefire, wouldjust silently skip some tests.

See https://stackoverflow.com/questions/36970384/surefire-is-not-picking-up-junit-5-tests for more.

rsynek:
FYI the GitHub Actions fail because the CI is not aware of a related PR to the kie-parent.

yurloc:
I've run one of the turtle tests locally and it worked.

yurloc:
The difference in executed test counts is related to skipping turtle tests. Basically, each concrete turtle test class is now skipped once before the `@TestFactory` method in its abstract parent generates all the combinations. I'll make a comment in the code so that you can see which line I'm talking about.

optaplanner-examples/src/test/java/org/optaplanner/examples/common/app/SolveAllTurtleTest.java
```diff
@@ -32,24 +36,36 @@
...
+    protected abstract CommonApp<Solution_> createCommonApp();
 
-    protected abstract Solution_ readProblem();
+    protected abstract ProblemFactory<Solution_> createProblemFactory(CommonApp<Solution_> commonApp);
 
-    @Test
-    public void runFastAndFullAssert() {
+    @TestFactory
+    Stream<DynamicTest> runFastAndFullAssert() {
         checkRunTurtleTests();
```
yurloc, line 58: Since the check is made here in the factory method, we have 1 skipped test for each turtle test class instead of skipping individual dynamic tests.

I just realized we can improve this even further by simply returning an empty stream from the test factory if turtle tests are not enabled. I think this is a nice use of the [dynamic test](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests) feature.

pom.xml
```diff
@@ -32,6 +32,7 @@
     <impsort.goal>sort</impsort.goal>
     <jacoco.agent.argLine/>
     <spotbugs.failOnViolation>true</spotbugs.failOnViolation>
+    <version.org.mockito>3.1.0</version.org.mockito>
```
ge0ffrey, line 35: In a different PR, later on, we'll want to moved into kie-parent

optaplanner-examples/src/test/java/org/optaplanner/examples/common/app/SolveAllTurtleTest.java
ge0ffrey: Thank you for pointing out. Makes sense.
Make sure that the skipped tests count comparison take this into account and ignores that difference - that's fine.

yurloc:
Jenkins retest this.

optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/solver/tracking/NQueensConstructionHeuristicTrackingTest.java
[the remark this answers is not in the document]
yurloc: Assertions are out of scope of this PR although it makes sense to completely drop dependency on JUnit 4. Maybe a simple regex on the import statement would solve that but I really want to avoid scope creep.

JUnit 5 migration tip:
> Note that you may continue to use assertion methods from org.junit.Assert or any other assertion library ...

yurloc:
Execution time unchanged on my machine.

optaplanner-core/pom.xml
yurloc: Done.

pom.xml
yurloc: Yes but upgrading the whole KIE to Mockito 3.x is a completely different story :slightly_smiling_face:

PR #767, opened 2020-05-01
Fix unchecked access to Optional

rsynek:


rsynek:
jenkins retest this

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/persistence/FlightCrewSchedulingXlsxFileIO.java
```diff
@@ -577,7 +577,9 @@ private void writeEmployeesView() {
                                                 .collect(joining(", ")));
                                 int maxArrivalHour = flightAssignmentList.stream()
                                         .map(a -> a.getFlight().getArrivalUTCTime().getHour())
-                                        .max(Comparator.naturalOrder()).get();
+                                        .max(Comparator.naturalOrder())
+                                        .orElseThrow(() -> new IllegalArgumentException(
+                                                "No flight with maximum arrival hour found."));
```
ge0ffrey, line 582: Given that there is a `!flightAssignmentList.isEmpty()` check earlier, and that LocalTime.getHour() never returns null (notice that you'd get an NPE if getFlight() or getArrivalUTCTime() returns null), this is an impossible situation (an execution branch that has been theoretically been proven to be impossible). So when an impossible situation happens, it's a bug.
We have the policy to start such exceptions with "Impossible state:".
So I recommend replacing

```
.orElseThrow(() -> new IllegalArgumentException(
                                                "No flight with maximum arrival hour found."));
```

with 

```
.orElseThrow(() -> new IllegalStateException(
                                                "Impossible state: There is not at least one arrival hour in a non-empty flightAssignmentList (" + flightAssignmentList + ")."));
```

rsynek: Thanks, I haven't noticed this policy yet.

ge0ffrey:
Thanks @rsynek !

PR #768, opened 2020-05-04
Remove unnecessary explicit type argument

rsynek:


rsynek:
> Looks good. Make sure it works build with JDK 8 too (if jenkins doesn't try that).

There truly was one case that didn't work in Java8. Interestingly, Idea is suggesting to remove it, despite having JDK 8 configured there.

rsynek:
Comment about the SonarCloud check: this PR does not add any testable logic and its goal is not to increase the existing coverage.

yurloc:
@rsynek What about `<T> Mockito.reset(T... mocks)` in `PhaseLifecycleTest`? It seems we can remove the type argument there as well.

rsynek:
@yurloc the `PhaseLifecycleTest` is the one which fails to compile on Java 8 after the type argument is removed - despite the fact that Idea suggests removing it.

yurloc:
> @yurloc the `PhaseLifecycleTest` is the one which fails to compile on Java 8 after the type argument is removed - despite the fact that Idea suggests removing it.

My mileage varies. IDEA doesn't suggest removing it but I can remove it and the test compiles and runs successfully.

Feel free to either check again or merge. Your call. I'll merge this on Wednesday after lunch if not merged earlier by someone else.

rsynek:
@yurloc thanks, you are right - I confused that occurrence with `ExhaustiveSearchPhaseConfig#inherit()`.

PR #771, opened 2020-05-10
PLANNER-1976 Fix decreasing the cardinality of some constraint streams

triceo:
@rsynek This fixes the bug that you've reported from your monstrous constraint provider. In some cases, there is a couple percentage point drop in performance - that has to be accepted, since the previous behavior was not correct and caused `ClassCastException`s.

(This has lurked in the back of my mind the entire week, until suddenly this morning, I woke up not only with an understanding of the problem, but also with an idea for the fix.)

Keep in mind I also renamed a whole bunch of implementation classes as their names were chosen months ago and their current purpose no longer matched their original name. :-)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/quad/DefaultQuadConstraintCollector.java
```diff
@@ -37,6 +38,13 @@ public DefaultQuadConstraintCollector(Supplier<ResultContainer_> supplier,
         this.finisher = finisher;
     }
 
+    public static <A, B, C, D, Result_> QuadConstraintCollector<A, B, C, D, ?, Result_> noop() {
```
ge0ffrey, line 41: Naming: noop() or noOp() or noOperation()?
Why do we need this method?
Does it need to be public?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetAbstractUniConstraintStream.java
```diff
@@ -63,6 +63,11 @@ public void addChildStream(BavetAbstractUniConstraintStream<Solution_, A> childS
         childStreamList.add(childStream);
     }
 
+    @Override
+    public int getCardinality() {
+        return 1;
```
ge0ffrey, line 68: I am ok with doing it like this, but we could just as well do it as a default method on UniConstraintStream. Either way, it's all good.

PR #773, opened 2020-05-11
ConstraintProvider: better error message if it returns null or null elements

ge0ffrey:


triceo:
@ge0ffrey It needs a code style fix. Feel free to merge after that.

PR #774, opened 2020-05-11
PLANNER-1579 Deprecate double-based score types

triceo:
Also see https://github.com/kiegroup/optaplanner-website/pull/232

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/hardsoftdouble/HardSoftDoubleScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/hardsoftdouble/HardSoftDoubleScoreHolder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


optaplanner-examples/pom.xml
```diff
@@ -73,6 +73,36 @@
         </plugin>
       </plugins>
     </pluginManagement>
+    <plugins>
+      <plugin> <!-- Shade plugin configuration shared by all distribution modules. -->
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>3.2.2</version>
```
ge0ffrey, line 79: Why are we adding the shade plugin?
I'd like to avoid the shade plugin at all cost.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/hardsoftdouble/HardSoftDoubleScoreHolder.java
```diff
@@ -25,14 +25,18 @@
...
 
 /**
  * WARNING: NOT RECOMMENDED TO USE DUE TO ROUNDING ERRORS THAT CAUSE SCORE CORRUPTION.
  * Use {@link HardSoftBigDecimalScoreHolder} instead.
  *
  * @see HardSoftDoubleScore
+ * @deprecated Double-based scores are deprecated as floating point numbers can't represent a decimal number correctly.
+ *             Use {@link HardSoftBigDecimalScoreHolder} or a scaled {@link HardSoftLongScoreHolder} instead.
  */
+@Deprecated(/* forRemoval = true */)
```
ge0ffrey, line 39: I don't mind the "forRemoval" comment, but note that we intend to remove all existing Deprecated methods in 8.0 and those don't have that forRemoval comment.

triceo:
Sorry about that shade plugin. No idea how that got in there, will be removed.

triceo:
I marked the Sonar failures as false positives. They complain of missing depracation tags in Javadocs, while they're clearly there.

PR #775, opened 2020-05-12
Class.getPackage() might return null

ge0ffrey:
Relates to https://stackoverflow.com/questions/61572757/optaplanner-throws-nullpointerexception-while-building-solver


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintConfigurationDescriptor.java
```diff
@@ -118,7 +118,8 @@ private void processPackAnnotation(DescriptorPolicy descriptorPolicy) {
         // So the ConstraintWeightDescriptors parse packAnnotation.constraintPackage() themselves.
         constraintPackage = packAnnotation.constraintPackage();
         if (constraintPackage.isEmpty()) {
-            constraintPackage = constraintConfigurationClass.getPackage().getName();
+            Package pack = constraintConfigurationClass.getPackage();
```
ge0ffrey, line 121: I wanted to call this variable "package", but that's a reserved keyword in Java.

PR #776, opened 2020-05-12
PLANNER-1978 Score-related deprecations

triceo:
Some of these changes are just a little bit invasive. Please see code review comments.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
```diff
@@ -60,15 +63,22 @@
...
      */
-    Score_ toInitializedScore();
+    @Deprecated
+    default Score_ toInitializedScore() {
+        if (isSolutionInitialized()) {
+            return (Score_) this;
+        } else {
+            return withInitScore(0);
+        }
+    }
```
triceo, line 75: I was able to replace methods on the individual `Score` implementations with this one.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
```diff
@@ -155,6 +165,22 @@
...
+     *
+     * @return true if the hard score is 0 or higher and the {@link #getInitScore()} is 0.
+     */
+    default boolean isFeasible() {
+        /*
+         * This exception will only be thrown for custom scores that did not implement FeasibilityScore.
+         * TODO Safe to remove this default implementation once FeasibilityScore has been removed in 8.0.
+         */
+        throw new UnsupportedOperationException();
+    }
```
triceo, line 182: Noteworthy, I anticipate comments here.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
```diff
@@ -60,15 +63,22 @@
...
+            return withInitScore(0);
+        }
+    }
 
     /**
      * For example {@code 0hard/-8soft} with {@code -7} returns {@code -7init/0hard/-8soft}.
      *
      * @param newInitScore always negative (except in statistical calculations), 0 if all planning variables are initialized
      * @return equals score except that {@link #getInitScore()} is set to {@code newInitScore}
-     * @throws IllegalStateException if the original {@link #getInitScore()} is not 0
```
triceo, line 71: Technically this is a backwards incompatible change. Used to throw, doesn't anymore. I think it should be fine.

ge0ffrey: EasyScoreDirector and IncrementalScoreDirector should fail fast if the EasyScoreCalculator etc return a Score with a non zero initScore. Before they did, now they don't I believe.

ge0ffrey: I'd add a fail-fast check in both those directors, if the score returned by calculateScore() has a non-zero initScore.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
```diff
@@ -155,6 +165,22 @@
      */
     boolean isCompatibleArithmeticArgument(Score otherScore);
 
+    /**
+     * A {@link PlanningSolution} is feasible if it has no broken hard constraints
+     * and {@link #isSolutionInitialized()} is true.
+     *
+     * Simple scores ({@link SimpleScore}, {@link SimpleLongScore}, {@link SimpleBigDecimalScore}) are always feasible.
```
ge0ffrey, line 172: not sure if it's worth adding, but "(unless their {@link #getInitScore()} isn't 0)"

optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
ge0ffrey: I am fine with this approach. When I added toShortString() in FEB 2017 I should have probably done it this way too with a default method until 8.0 is.

Please add this in the list of PLANNER-750 TODO's, to remove make this method abstract again.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/comparator/NaturalScoreComparator.java
```diff
@@ -23,7 +23,10 @@
 
 /**
  * Compares by the natural order of {@link Score}s.
+ * 
+ * @deprecated If you need this, it is trivial to implement.
```
ge0ffrey, line 27: "it is trivial to implement" says the rocket engineer on how to build a V2 missle :)
Mention mention how to build a comparator on natural comparison? Especially in the upgrade recipe, to save users from have to look that up on SO.

ge0ffrey: Ah, its just Comparable::compareTo apparently?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardmediumsoft/HardMediumSoftScoreDefinition.java
```diff
@@ -21,11 +21,13 @@
...
 import org.optaplanner.core.config.score.trend.InitializingScoreTrendLevel;
-import org.optaplanner.core.impl.score.definition.AbstractFeasibilityScoreDefinition;
+import org.optaplanner.core.impl.score.definition.AbstractScoreDefinition;
+import org.optaplanner.core.impl.score.definition.ScoreDefinition;
 import org.optaplanner.core.impl.score.inliner.ScoreInliner;
 import org.optaplanner.core.impl.score.trend.InitializingScoreTrend;
 
-public class HardMediumSoftScoreDefinition extends AbstractFeasibilityScoreDefinition<HardMediumSoftScore> {
+public class HardMediumSoftScoreDefinition extends AbstractScoreDefinition<HardMediumSoftScore>
+        implements ScoreDefinition<HardMediumSoftScore> {
```
ge0ffrey, line 30: AbstractScoreDefinition implements ScoreDefinition, no need to mention it twice. Remove the "implements".
Same in other files.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/AbstractBendableScoreDefinition.java
```diff
@@ -18,8 +18,8 @@
 
 import org.optaplanner.core.api.score.FeasibilityScore;
 
-public abstract class AbstractBendableScoreDefinition<S extends FeasibilityScore<S>>
-        extends AbstractFeasibilityScoreDefinition<S> {
+public abstract class AbstractBendableScoreDefinition<S extends FeasibilityScore<S>> extends AbstractScoreDefinition<S>
+        implements ScoreDefinition<S> {
```
ge0ffrey, line 22: same here, AbstractScoreDefinition already implements it

PR #779, opened 2020-05-15
PLANNER-1821 Simplify capacity constraints by using shadow variables

triceo:
- [x] Passes FULL_ASSERT.
- [x] Scores before and after this change are identical.
- [x] Performs a tiny bit better now than it used to.
- [ ] Completely identical to the easy calculator


optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Bus.java
```diff
@@ -79,6 +80,14 @@ public void setNextStop(BusStop nextStop) {
         this.nextStop = nextStop;
     }
 
+    public Integer getPassengerQuantityTotal() {
+        return passengerQuantityTotal == null ? 0 : passengerQuantityTotal;
+    }
+
+    public void setPassengerQuantityTotal(final Integer passengerQuantityTotal) {
```
triceo, line 87: I'd be great if I could make this an `int`,  but Planner fail-fasts, saying that it can not be an `int`, as Planner needs null to be able to say whether it's initialized or not.

I think that that's non-sense in this case.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Coach.java
```diff
@@ -42,6 +47,13 @@ public void setDestination(BusHub destination) {
     // Complex methods
     // ************************************************************************
 
+    @CustomShadowVariable(variableListenerClass = CoachPassengerQuantityTotalUpdatingVariableListener.class,
+            sources = { @PlanningVariableReference(entityClass = BusStop.class, variableName = "bus") })
+    @Override
+    public Integer getPassengerQuantityTotal() {
+        return super.getPassengerQuantityTotal();
+    }
```
triceo, line 55: This is sort of unnatural, and definitely unexpected. I had to override the method, because the annotation would not be inherited.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/AbstractPassengerQuantityTotalUpdatingVariableListener.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private void increasePassengerQuantityTotal(ScoreDirector scoreDirector, BusStop busStop) {
+        if (!isBusApplicable(busStop.getBus()) || busStop.getPassengerQuantity() == 0) {
+            return;
+        }
+        Bus bus = busStop.getBus();
+        scoreDirector.beforeVariableChanged(bus, "passengerQuantityTotal");
+        bus.setPassengerQuantityTotal(bus.getPassengerQuantityTotal() + busStop.getPassengerQuantity());
+        scoreDirector.afterVariableChanged(bus, "passengerQuantityTotal");
+    }
```
triceo, line 36: This may be too simplistic.

Even though FULL_ASSERT passes, this doesn't mimic what the EasyScoreCalculator does and there are even differences between the two. I have to investigate this more.

PR #782, opened 2020-05-19
PLANNER-1983 Final easy deprecations

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -112,6 +111,7 @@
     @XStreamConverter(KeyAsElementMapConverter.class)
     protected Map<String, String> incrementalScoreCalculatorCustomProperties = null;
 
+    @Deprecated(/* forRemoval = true */)
     protected String ksessionName = null;
```
triceo, line 115: We should be careful here.
If for some reason we have to make 8 compatible with 7 tooling, we can not remove this.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/ScoreManager.java
```diff
@@ -61,4 +63,33 @@
...
+     * Instead, to provide this information in a UI or a service, use {@link SolverFactory#getScoreDirectorFactory()}
+     * to retrieve {@link ScoreDirector#getConstraintMatchTotalMap()} and {@link ScoreDirector#getIndictmentMap()}
+     * and convert those into a domain specific API.
+     * <p>
+     * This method is thread-safe.
+     *
+     * @return null if {@link #getScore(Object)} returns null with the same solution
+     * @see ScoreDirector#explainScore()
+     */
+    String explainScore(Solution_ solution);
```
triceo, line 93: New API. Technically, this is a break of bwd compatibility for the interface.
Partically, no way to avoid it.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
```diff
@@ -72,4 +73,14 @@
...
 
+    /**
+     * Returns the {@link Duration} spent solving since the last start.
+     * If it hasn't started it yet, it returns {@link Duration#ZERO}.
+     * If it hasn't ended yet, it returns the time between the last start and now.
+     * If it has ended already, it returns the time between the last start and the ending.
+     *
+     * @return the {@link Duration} spent solving since the last (re)start, at least 0
+     */
+    Duration getTimeSpent();
```
triceo, line 84: New API. Technically, this is a break of bwd compatibility for the interface.
Partically, no way to avoid it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/scope/DefaultSolverScope.java
```diff
@@ -47,15 +48,15 @@
...
+    private volatile Long startingSystemTimeNanos;
+    private volatile Long endingSystemTimeNanos;
     protected long childThreadsScoreCalculationCount = 0;
 
     protected Score startingInitializedScore;
 
     protected volatile Solution_ bestSolution;
     protected volatile Score bestScore;
-    protected Long bestSolutionTimeMillis;
+    private Long bestSolutionTimeNanos;
```
triceo, line 59: I converted the class to use nanos instead of millis. No need not to use the precision, if we have it.
Also, I've hidden the fields, since they have getters and setters and this is not performance-sensitive code.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/ScoreManager.java
```diff
@@ -61,4 +63,33 @@
...
+     * Returns the {@link Score} of the solution.
+     * <p>
+     * This is useful for generic code, which doesn't know the type of the {@link PlanningSolution} to retrieve the
+     * {@link Score} from the solution easily.
+     * <p>
+     * This method is thread-safe.
+     *
+     * @return null if the {@link PlanningSolution} is still uninitialized
+     */
+    Score getScore(Solution_ solution);
```
ge0ffrey, line 76: This doesn't calculate the score, but it only extracts it right?

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/Solver.java
```diff
@@ -105,7 +109,9 @@
      * This method is thread-safe.
      *
      * @return null if the {@link PlanningSolution} is still uninitialized
+     * @deprecated in favor of {@link ScoreManager#explainScore(Object)}
```
ge0ffrey, line 112: copy paste bug: should be in favor of ScoreManager#getScore(...)

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
ge0ffrey: Maybe we should rename this method to getDurationSpent() or better yet, `getSolvingDuration()` Wdyt? Let's discuss on zulip.

optaplanner-core/src/main/java/org/optaplanner/core/config/domain/ReflectionsWorkaroundClasspathHelper.java
```diff
@@ -38,7 +38,10 @@
 /**
  * Workaround for bug in {@link ClasspathHelper}.
  * This workaround expires once https://github.com/ronmamo/reflections/pull/118 is fixed.
+ * 
+ * @deprecated for removal.
```
ge0ffrey, line 42: in favor of Quarkus extension or Spring Boot starter.

optaplanner-core/src/main/java/org/optaplanner/core/config/domain/ScanAnnotatedClassesConfig.java
```diff
@@ -37,7 +37,11 @@
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import com.thoughtworks.xstream.annotations.XStreamImplicit;
 
+/**
+ * @deprecated This functionality will be removed in the next major version of OptaPlanner.
```
ge0ffrey, line 41: in favor of Quarkus extension or Spring Boot starter.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
ge0ffrey: Yea, I don't think we can deprecate this just yet. There's no good alternative today (maybe quarkus extension but still we can't deprecate kie-server in 7.x).
Please undeprecate, but add a "// TODO remove for 8.0"

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -235,7 +235,11 @@ public static SolverConfig createFromXmlReader(Reader reader, ClassLoader classL
     protected Integer moveThreadBufferSize = null;
     protected Class<? extends ThreadFactory> threadFactoryClass = null;
 
+    /**
+     * @deprecated This functionality will be removed in the next major version of OptaPlanner.
```
ge0ffrey, line 239: deprecated in favor of the Quarkus extension or Spring Boot starter which handle annotation scanning more efficiently.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -354,10 +358,18 @@ public void setThreadFactoryClass(Class<? extends ThreadFactory> threadFactoryCl
         this.threadFactoryClass = threadFactoryClass;
     }
 
+    /**
+     * @deprecated This functionality will be removed in the next major version of OptaPlanner.
```
ge0ffrey, line 362: deprecated in favor of the Quarkus extension or Spring Boot starter which handle annotation scanning more efficiently.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -354,10 +358,18 @@ public void setThreadFactoryClass(Class<? extends ThreadFactory> threadFactoryCl
...
+     */
+    @Deprecated(/* forRemoval = true */)
     public ScanAnnotatedClassesConfig getScanAnnotatedClassesConfig() {
         return scanAnnotatedClassesConfig;
     }
 
+    /**
+     * @deprecated This functionality will be removed in the next major version of OptaPlanner.
```
ge0ffrey, line 370: deprecated in favor of the Quarkus extension or Spring Boot starter which handle annotation scanning more efficiently.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/DefaultScoreManager.java
```diff
@@ -38,10 +39,22 @@ public DefaultScoreManager(ScoreDirectorFactory<Solution_> scoreDirectorFactory)
 
     @Override
     public void updateScore(Solution_ solution) {
+        getScore(solution);
```
ge0ffrey, line 42: ok, so getScore() doesn't just extract it, but calculates it?
We can't have 2 methods that do exactly the same, so I guess updateScore() needs to be deprecated?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -125,17 +126,17 @@ public String explainBestScore() {
...
             // The solver hasn't started yet
             return 0L;
         }
-        Long endingSystemTimeMillis = solverScope.getEndingSystemTimeMillis();
-        if (endingSystemTimeMillis == null) {
+        Long endingSystemTimeNanos = solverScope.getEndingSystemTimeNanos();
+        if (endingSystemTimeNanos == null) {
             // The solver hasn't ended yet
-            endingSystemTimeMillis = System.currentTimeMillis();
+            endingSystemTimeNanos = System.nanoTime();
```
ge0ffrey, line 137: This is dangerous, from https://www.geeksforgeeks.org/java-system-nanotime-vs-system-currenttimemillis/

Cons:

    The result reflected doesn’t have any fixed reference point. According to Java documentation,

    The value returned represents nanoseconds since some fixed
    but arbitrary time (perhaps in the future, so values may be negative).

    Less accurate. This method provides nanosecond precision, but not necessarily nanosecond accuracy. No guarantees are made about how frequently values change.
    Depending on the system, it can take more than 100 cpu cycles to execute.
    Not thread safe. May return erroneous results if used between more than one threads.

ge0ffrey: Also, a Long can hold about 4 000 000 000 000 000 000 nanos.
So 4 000 000 000 seconds. (think milli, micro, nano)
So 46296 days. So 126 years. This means OptaPlanner might have a Y2K bug after 2096 (the epoch starts in 1970).
Ok, this is less important, but the previous comment highlights some risks. I don't think this change should be part of this PR, it should be in a seperate PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -61,7 +63,11 @@ public DefaultSolverJob(
             BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler) {
         this.solverManager = solverManager;
         this.problemId = problemId;
-        this.solver = solver;
+        if (!(solver instanceof DefaultSolver)) {
+            throw new IllegalStateException(
+                    "Impossile state: solver is not instance of " + DefaultSolver.class.getSimpleName() + ".");
```
ge0ffrey, line 68: typo impossile

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/solution/drools/ProblemFactCollectionProperty.java
```diff
@@ -40,10 +40,11 @@
  * Do not annotate {@link PlanningEntity planning entities} as problem facts:
  * they are automatically available as facts for {@link ConstraintFactory#from(Class)} or DRL.
  *
- * @see ProblemFactProperty
+ * @deprecated In favor of {@link org.optaplanner.core.api.domain.solution.ProblemFactCollectionProperty}.
  */
 @Target({ METHOD, FIELD })
 @Retention(RUNTIME)
+@Deprecated(/* forRemoval = true */)
 public @interface ProblemFactCollectionProperty {
```
ge0ffrey, line 48: Do we have a testdata unit test that still uses these deprecated ProblemFactCollectionProperty? (Including a // TODO remove in 8.0)
Almost all of our users use this, many won't refactor quickly. We need to make sure this keeps working during the course of the 7.x time.

optaplanner-core/src/main/java/org/optaplanner/core/config/domain/ReflectionsWorkaroundClasspathHelper.java
triceo: I actually re-read the comment and this class may no longer be necessary as the PR in question has been merged.

optaplanner-core/src/main/java/org/optaplanner/core/api/solver/SolverJob.java
ge0ffrey: agreed on getSolvingDuration()

optaplanner-core/src/main/java/org/optaplanner/core/api/score/ScoreManager.java
ge0ffrey: Removed, make updateScore return Score

PR #784, opened 2020-05-21
Change 2 broken links to Rostering competition

dupliaka:
https://www.kuleuven-kulak.be/nrpcompetition - outdated

To define example of rostering problem there was
a reference to kuleuven-kulak.be resource that now returns 503 Service unavailable.

However they do have an academic paper with the
full description of the competition in 2010.


ge0ffrey:
Thanks @dupliaka - great to see that link issues fixed, it's been there a while.

PR #789, opened 2020-05-25
PLANNER-1987, PLANNER-1575 More deprecations

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/definition/ScoreDefinition.java
```diff
@@ -185,4 +185,11 @@ default boolean isNegativeOrZero(S score) {
      */
     S divideBySanitizedDivisor(S dividend, S divisor);
 
+    /**
+     * @param score never null
+     * @return true if the otherScore is accepted as a parameter of {@link Score#add(Score)},
+     *         {@link Score#subtract(Score)} and {@link Score#compareTo(Object)} for scores of this score definition.
+     */
+    boolean isCompatibleArithmeticArgument(Score score);
```
triceo, line 193: Revapi complains here.
Why is `ScoreDefinition` a public API?

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/solver/MovableShiftAssignmentSelectionFilter.java
```diff
@@ -17,16 +17,19 @@
...
 
 import org.optaplanner.core.api.domain.entity.PinningFilter;
+import org.optaplanner.core.impl.heuristic.selector.common.decorator.SelectionFilter;
+import org.optaplanner.core.impl.score.director.ScoreDirector;
 import org.optaplanner.examples.nurserostering.domain.NurseRoster;
 import org.optaplanner.examples.nurserostering.domain.ShiftAssignment;
-import org.optaplanner.examples.nurserostering.domain.ShiftDate;
 
-public class MovableShiftAssignmentSelectionFilter implements PinningFilter<NurseRoster, ShiftAssignment> {
+public class MovableShiftAssignmentSelectionFilter implements SelectionFilter<NurseRoster, ShiftAssignment> {
```
triceo, line 25: I have to say... I don't like the fact that, in order to filter moves in selectors, users need to implement this interface.
Ideally, I'd like `ScoreDirector` to be an implementation detail. Yet, it keeps leaking.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/ranking/ResilientScoreComparator.java
```diff
@@ -20,11 +20,18 @@
...
 
 /**
  * Able to compare {@link Score}s of different types or nulls.
  */
-public class ResilientScoreComparator implements Comparator<Score>, Serializable {
+final class ResilientScoreComparator implements Comparator<Score>, Serializable {
+
+    private final ScoreDefinition aScoreDefinition;
+
+    public ResilientScoreComparator(ScoreDefinition aScoreDefinition) {
```
ge0ffrey, line 32: Are we sure about this?
Most users don't know about ScoreDefinition.
This means most users can never use a ResilientScoreComparator.
But on the other hand, I presume they will never need it, so it's fine?

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/PinningFilter.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.optaplanner.core.api.domain.entity;
+
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.impl.heuristic.move.Move;
+
+/**
+ * Decides on accepting or discarding a {@link PlanningEntity}.
+ * An immovable {@link PlanningEntity} is rejected and therefore never used in a {@link Move}.
```
ge0ffrey, line 24: "an immovable" => "a pinned"

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/PinningFilter.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <T> the entity type
+ */
+public interface PinningFilter<Solution_, T> {
+
+    /**
+     * @param solution working solution to which the selection belongs or applies to
+     * @param entity never null, a {@link PlanningEntity}
+     * @return true if the entity it is movable, false if the entity is immovable
```
ge0ffrey, line 34: true if the entity's planning variables can change, false if the entity is pinned

@triceo Red flag. `accept()` returns `false` if the entity is `pinned`... this is a serious issue.

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/PlanningEntity.java
```diff
@@ -51,13 +51,34 @@
...
+    @Deprecated(/* forRemoval = true */)
     interface NullMovableEntitySelectionFilter extends SelectionFilter {
     }
 
+    /**
+     * An immovable planning entity is never changed during planning,
+     * this is useful in repeated planning use cases (such as continuous planning and real-time planning).
+     * <p>
+     * This applies to all the planning variables of this planning entity.
+     * To make individual variables immovable, see https://issues.redhat.com/browse/PLANNER-124
```
ge0ffrey, line 69: find in path for all "immovable"

optaplanner-core/src/main/java/org/optaplanner/core/config/constructionheuristic/placer/QueuedValuePlacerConfig.java
```diff
@@ -87,6 +87,7 @@ public void setMoveSelectorConfig(MoveSelectorConfig moveSelectorConfig) {
     // Builder methods
     // ************************************************************************
 
+    @SuppressWarnings("deprecation")
```
ge0ffrey, line 90: Can we add a comment behind it "for what"? Mainly as an expiration date: when we can remove this line.
For example:

@SuppressWarnings("deprecation") // Remove in 8.0 when it doesn't call  foo.bar() any more

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -332,11 +338,24 @@ public void setAssertionScoreDirectorFactory(ScoreDirectorFactoryConfig assertio
...
+     * @deprecated for removal.
+     */
+    @Deprecated(/* forRemoval = true */)
     public void setGenerateDroolsTestOnError(Boolean generateDroolsTestOnError) {
+        if (BooleanUtils.isTrue(generateDroolsTestOnError)) {
+            System.setProperty(GENERATE_DROOLS_TEST_ON_ERROR_PROPERTY_NAME, "true");
+        } else {
+            System.clearProperty(GENERATE_DROOLS_TEST_ON_ERROR_PROPERTY_NAME);
+        }
         this.generateDroolsTestOnError = generateDroolsTestOnError;
```
ge0ffrey, line 359: This should be a simple setter (that is now deprecated). It shouldn't mess with the system property, in favor of which it is deprecated, because  System.setProperty() is dangerous: presume that a single JVM process runs 2 solver configs, constructed by java (not by xml) and the first one calls setGenerateDroolsTestOnError(true), but the second one doesn't. The second one would feel it too.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -159,10 +160,27 @@ private void processEntityAnnotations(DescriptorPolicy descriptorPolicy) {
...
-        }
-        if (movableEntitySelectionFilterClass != null) {
+        boolean hasSelectionFilter = movableEntitySelectionFilterClass != PlanningEntity.NullMovableEntitySelectionFilter.class;
+        Class<? extends PinningFilter> pinningFilterClass = entityAnnotation.pinningFilter();
+        boolean hasPinningFilter = pinningFilterClass != PlanningEntity.NullPinningFilter.class;
+        if (hasPinningFilter && hasSelectionFilter) {
+            throw new IllegalStateException("The entityClass (" + entityClass
+                    + ") uses both movableEntitySelectionFilter (" + movableEntitySelectionFilterClass +
+                    ") and pinningFilter (" + pinningFilterClass + "). " +
+                    "Use pinningFilter exclusively.");
```
ge0ffrey, line 170: Our error messages typically have 2 lines:
- line 1: what is wrong. This is always truth. It doesn't say how to fix it though. Ends with "\n"
- line 2 (optional): How to fix it. This isn't always the way to fix it, it is a recommendation. Therefore, it starts with "Maybe ..."

I'd also add a breadcrumb to @PlanningEntity

=>
```
            throw new IllegalStateException("The entityClass (" + entityClass
                    + ") uses both movableEntitySelectionFilter (" + movableEntitySelectionFilterClass +
                    ") and pinningFilter (" + pinningFilterClass + ").\n" +
                    "Maybe use only pinningFilterClass on your @" + PlanningEntity.class.getSimpleName() + " annotation.");
```

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/solver/MovableShiftAssignmentSelectionFilter.java
ge0ffrey: Yea, sounds like maybe SelectionFilter, instead of creating a new type of Filter like PinningFilter that doesn't work on moves (and shouldn't due to naming).

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/ranking/ResilientScoreComparator.java
triceo: This is no longer a public class. And even before now, it was not in public API.
We are fine.

ge0ffrey:
let's remove all suppresswarnigns deprecated

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -122,6 +122,10 @@
     protected List<File> scoreDrlFileList = null;
     @XStreamConverter(KeyAsElementMapConverter.class)
     protected Map<String, String> kieBaseConfigurationProperties = null;
```
triceo, line 124: I wonder... shouldn't this be deprecated too, along with the `scoreDrlFileList`? What's it for?

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/PinningFilter.java
```diff
@@ -0,0 +1,38 @@
...
+import org.optaplanner.core.impl.heuristic.move.Move;
+
+/**
+ * Decides on accepting or discarding a {@link PlanningEntity}.
+ * A pinned {@link PlanningEntity} is never used in a {@link Move}.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <T> the entity type
+ */
+public interface PinningFilter<Solution_, T> {
```
ge0ffrey, line 29: (soft) Consistency with other code: Rename `T` to `Entity_`. It reads much clearer if that's the return type somewhere, or if someone needs to extend/implement this class.

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/entity/PinningFilter.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.impl.heuristic.move.Move;
+
+/**
+ * Decides on accepting or discarding a {@link PlanningEntity}.
+ * A pinned {@link PlanningEntity} is never used in a {@link Move}.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <T> the entity type
```
ge0ffrey, line 27: (very soft) change to:
`@param <Entity_> the entity type, the class with the {@link PlanningEntity} annotation`

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
ge0ffrey: I'd argue no, because they can also be used in combination with scoreDrl.
For example to turn on switch that hows-it-called propagation from "ALLOWED" to "ALWAYS"

optaplanner-docs/src/main/asciidoc/MoveAndNeighborhoodSelection/MoveAndNeighborhoodSelection-chapter.adoc
```diff
@@ -927,7 +927,7 @@ public interface SelectionFilter<Solution_, T> {
 }
 ----
 
-Implement the `accept` method to return `false` on a discarded `selection` (see below). Filtered selection can happen on any Selector in the selector tree, including any ``MoveSelector``, `EntitySelector` or ``ValueSelector``.
+Implement the `accept` method to return `false` on a discarded `selection` (see below).Filtered selection can happen on any Selector in the selector tree, including any ``MoveSelector``, `EntitySelector` or ``ValueSelector``.
```
ge0ffrey, line 930: I believe this change is a typo

optaplanner-docs/src/main/asciidoc/PlannerConfiguration/PlannerConfiguration-chapter.adoc
```diff
@@ -523,24 +523,35 @@ Currently <<chainedPlanningVariable, chained>> planning variables are not compat
...
+  ...
+  <constructionHeuristics>
+    <queuedEntityPlacer>
+      <entitySelector id="foo">
+        <filterClass>...</filterClass>
+      </entitySelector>
+    </queuedEntityPlacer>
+    ...
+    <changeMoveselector>
+      <entitySelector mimicRef="foo" />
```
ge0ffrey, line 542: let's replace "foo" with something more meaningful, how about "entitySelector1"?

optaplanner-docs/src/main/asciidoc/RepeatedPlanning/continuousPlanningPatientAdmissionSchedule.svg
```diff
@@ -2,32 +2,31 @@
...
+    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+    width="800"
+    height="600"
+    id="svg2"
+    version="1.1"
+    inkscape:version="0.92.4 (unknown)"
+    sodipodi:docname="continuousPlanningPatientAdmissionSchedule.svg"
+    inkscape:export-filename="continuousPlanningPatientAdmissionSchedule.png"
+    inkscape:export-xdpi="90"
+    inkscape:export-ydpi="90">
```
ge0ffrey, line 20: (not for this PR) dpi is 96 these days - I 'll explain in one of our meetings, but don't worry about it for this PR.

PR #792, opened 2020-05-28
PLANNER-1961 Only process trailing entities when there's supply

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/SwapMoveSelector.java
```diff
@@ -90,26 +94,19 @@ public boolean supportsPhaseAndSolverCaching() {
...
+                            .demand(new SingletonInverseVariableDemand(variableDescriptor))));
         }
     }
 
     @Override
     public void solvingEnded(DefaultSolverScope solverScope) {
         super.solvingEnded(solverScope);
         if (anyChained) {
-            inverseVariableSupplyList = null;
+            inverseVariableSupplyMap = null;
```
triceo, line 109: And this is why you don't put types of variables in variable names.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/ChainedSwapMove.java
```diff
@@ -74,8 +77,8 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
                     scoreDirector.changeVariableFacade(variableDescriptor, leftEntity, oldRightValue);
                     scoreDirector.changeVariableFacade(variableDescriptor, rightEntity, oldLeftValue);
                 } else {
-                    Object oldLeftTrailingEntity = oldLeftTrailingEntityList.get(i);
-                    Object oldRightTrailingEntity = oldRightTrailingEntityList.get(i);
+                    Object oldLeftTrailingEntity = oldLeftTrailingEntityMap.get(variableDescriptor);
+                    Object oldRightTrailingEntity = oldRightTrailingEntityMap.get(variableDescriptor);
```
ge0ffrey, line 81: So when I have a plannning entity `LessonVisit` with a chained planning variable `previous` and a non-chained planning variable `room`, this code will execute for the `room`  too? That `room` should behave like a normal swap. Does it?

What happens to `room` if it's trailing entities are null, both left and right? I guess it behaves like a normal swap move indeed?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/SwapMoveSelector.java
```diff
@@ -90,26 +94,19 @@ public boolean supportsPhaseAndSolverCaching() {
...
         if (anyChained) {
-            inverseVariableSupplyList = new ArrayList<>(variableDescriptorList.size());
             SupplyManager supplyManager = solverScope.getScoreDirector().getSupplyManager();
-            for (GenuineVariableDescriptor variableDescriptor : variableDescriptorList) {
-                SingletonInverseVariableSupply inverseVariableSupply;
-                if (variableDescriptor.isChained()) {
-                    inverseVariableSupply = supplyManager.demand(
-                            new SingletonInverseVariableDemand(variableDescriptor));
-                } else {
-                    inverseVariableSupply = null;
```
triceo, line 101: This is where you put the `null` that is then thrown later.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/ChainedSwapMove.java
ge0ffrey: I think the code behaves correctly, but we should add a test for it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/ChainedSwapMove.java
```diff
@@ -31,26 +32,28 @@
...
     public ChainedSwapMove(List<GenuineVariableDescriptor<Solution_>> variableDescriptorList,
-            List<SingletonInverseVariableSupply> inverseVariableSupplyList, Object leftEntity, Object rightEntity) {
+            Map<GenuineVariableDescriptor<Solution_>, SingletonInverseVariableSupply> inverseVariableSupplyMap,
+            Object leftEntity, Object rightEntity) {
         super(variableDescriptorList, leftEntity, rightEntity);
-        oldLeftTrailingEntityList = new ArrayList<>(inverseVariableSupplyList.size());
-        oldRightTrailingEntityList = new ArrayList<>(inverseVariableSupplyList.size());
-        for (SingletonInverseVariableSupply inverseVariableSupply : inverseVariableSupplyList) {
-            oldLeftTrailingEntityList.add(inverseVariableSupply.getInverseSingleton(leftEntity));
-            oldRightTrailingEntityList.add(inverseVariableSupply.getInverseSingleton(rightEntity));
```
triceo, line 44: This is where the null would be thrown.
I fix that by not iterating over the cases where the null would be thrown. (Non-chained vars.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/ChainedSwapMove.java
```diff
@@ -74,8 +77,8 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
                     scoreDirector.changeVariableFacade(variableDescriptor, leftEntity, oldRightValue);
                     scoreDirector.changeVariableFacade(variableDescriptor, rightEntity, oldLeftValue);
                 } else {
-                    Object oldLeftTrailingEntity = oldLeftTrailingEntityList.get(i);
-                    Object oldRightTrailingEntity = oldRightTrailingEntityList.get(i);
+                    Object oldLeftTrailingEntity = oldLeftTrailingEntityMap.get(variableDescriptor);
```
ge0ffrey, line 80: These map calls are a performance loss. **We're replacing an indexed lookup by a hash lookup.**
I'd keep the index lookup (so the trailingEntityList has the same size as variableDescriptorList and their indexes match), but given that "Object oldLeftTrailingEntity " can be null, I 'd put a null as an element where we cache that (move constructor)

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/ChainedSwapMove.java
ge0ffrey: I'd iterate over them, to avoid changing an indexed lookup by a hashed lookup later on, but put in null if they don't exist, instead of throwing an NPE. wdyt?

triceo:
I changed the implementation and added a test that both the chained and unchained vars are swapped.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/chained/TestdataChainedEntity.java
```diff
@@ -39,7 +39,15 @@
...
 
+    public static GenuineVariableDescriptor<TestdataChainedSolution> buildVariableDescriptorForUnchainedObject() {
+        SolutionDescriptor<TestdataChainedSolution> solutionDescriptor = TestdataChainedSolution.buildSolutionDescriptor();
+        EntityDescriptor<TestdataChainedSolution> entityDescriptor = solutionDescriptor
+                .findEntityDescriptorOrFail(TestdataChainedEntity.class);
+        return entityDescriptor.getGenuineVariableDescriptor("unchainedObject");
+    }
+
     private TestdataChainedObject chainedObject;
+    private TestdataObject unchainedObject;
```
ge0ffrey, line 50: (very soft) I 'd have used TestdataValue as a type for unchainedObject, calling it unchainedValue instead, for consistency with other testdata. But it doesn't matter much.

PR #794, opened 2020-06-01
PLANNER-1991 ScoreHolder-related deprecations

triceo:
It's a bit unfortunate that I had to deprecate the ScoreHolder impl classes. But considering that there is no real IDE support for DRL, users hopefully won't even notice.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
@@ -31,7 +31,6 @@ import org.optaplanner.core.api.score.holder.AbstractScoreHolder;
 /**
  * @see BendableScore
  */
-@Deprecated
 public class BendableScoreHolder extends AbstractScoreHolder<BendableScore> {
 
     protected final Map<Rule, BiConsumer<RuleContext, Integer>> matchExecutorByNumberMap = new LinkedHashMap<>();
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,7 +30,9 @@ import org.optaplanner.core.api.score.holder.AbstractScoreHolder;
 
 /**
  * @see BendableScore
+ * @deprecated to be replaced by an interface of the same name, removing many methods
  */
+@Deprecated
 public class BendableScoreHolder extends AbstractScoreHolder<BendableScore> {
 
     protected final Map<Rule, BiConsumer<RuleContext, Integer>> matchExecutorByNumberMap = new LinkedHashMap<>();
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
@@ -28,6 +28,10 @@ import org.optaplanner.core.api.domain.constraintweight.ConstraintWeight;
 import org.optaplanner.core.api.domain.solution.PlanningScore;
 import org.optaplanner.core.api.score.holder.AbstractScoreHolder;
 
+/**
+ * @see BendableScore
+ */
+@Deprecated
 public class BendableScoreHolder extends AbstractScoreHolder<BendableScore> {
 
     protected final Map<Rule, BiConsumer<RuleContext, Integer>> matchExecutorByNumberMap = new LinkedHashMap<>();
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
@@ -28,11 +28,6 @@ import org.optaplanner.core.api.domain.constraintweight.ConstraintWeight;
 import org.optaplanner.core.api.domain.solution.PlanningScore;
 import org.optaplanner.core.api.score.holder.AbstractScoreHolder;
 
-/**
- * @see BendableScore
- * @deprecated to be replaced by an interface of the same name, removing many methods
- */
-@Deprecated
 public class BendableScoreHolder extends AbstractScoreHolder<BendableScore> {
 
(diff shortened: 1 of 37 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
```diff
@@ -30,7 +30,9 @@
 
 /**
  * @see BendableScore
+ * @deprecated to be replaced by an interface of the same name, removing many methods
  */
+@Deprecated
 public class BendableScoreHolder extends AbstractScoreHolder<BendableScore> {
```
ge0ffrey, line 36: I don't think we can deprecate the actual ScoreHolders, like this one. As that means we're forcing people in scoreDRL to use deprecated code, with no alternative to migrate to.

What we can do is deprecate the methods we don't want them to use, and once we branch 8.0, turn BendableScoreHolder etc into an interface with the exact same full class name while only remaining the methods that we allow them to use.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/ScoreHolder.java
```diff
@@ -50,15 +50,19 @@
      * @param initScore {@code <= 0}, managed by OptaPlanner, needed as a parameter in the {@link Score}'s creation
      *        method, see {@link Score#getInitScore()}
      * @return never null, the {@link Score} of the working {@link PlanningSolution}
+     * @deprecated for removal from public API
```
ge0ffrey, line 53: Let's reference ScoreManager because if they're using this, they probably want to learn about ScoreManager.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/ScoreHolder.java
```diff
@@ -81,7 +87,9 @@
      * @return never null
      * @throws IllegalStateException if {@link #isConstraintMatchEnabled()} is false
      * @see ScoreDirector#getConstraintMatchTotals()
+     * @deprecated for removal from public API
```
ge0ffrey, line 90: Let's reference ScoreManager because if they're using this, they probably want to learn about ScoreManager.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/ScoreHolder.java
```diff
@@ -92,7 +100,9 @@
      * @return never null
      * @throws IllegalStateException if {@link #isConstraintMatchEnabled()} is false
      * @see ScoreDirector#getConstraintMatchTotalMap()
+     * @deprecated for removal from public API
```
ge0ffrey, line 103: Let's reference ScoreManager because if they're using this, they probably want to learn about ScoreManager.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/ScoreHolder.java
```diff
@@ -103,7 +113,9 @@
      * @return never null
      * @throws IllegalStateException if {@link #isConstraintMatchEnabled()} returns false
      * @see ScoreDirector#getIndictmentMap()
+     * @deprecated for removal from public API
```
ge0ffrey, line 116: Let's reference ScoreManager because if they're using this, they probably want to learn about ScoreManager.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/ScoreHolder.java
ge0ffrey: Even though those getConstraintMatchTotals() methods aren't there yet, they will end up on ScoreManager. Might as well already reference it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreDefinition.java
```diff
@@ -109,7 +110,7 @@ public BendableScoreInliner buildScoreInliner(boolean constraintMatchEnabled) {
     }
 
     @Override
-    public BendableScoreHolder buildScoreHolder(boolean constraintMatchEnabled) {
+    public ScoreHolder buildScoreHolder(boolean constraintMatchEnabled) {
```
ge0ffrey, line 113: I am not sure about these changes.
In 8.0, where we have `DefaultBendableScoreHolder implements BendableScoreHolder`, we'll want to return `BendableScoreHolder` here (not `DefaultBendableScoreHolder` nor just `ScoreHolder`). Or wait - optaplanner-core will need something that has extractScore(), so it will need to return AbstractScoreHolder or InnerScoreHolder or DefaultBendableScoreHolder anyway...

ge0ffrey: ok to change like this or to leave as is. your call - the real work with be in the 8.x branch

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardsoftdouble/HardSoftDoubleScoreDefinition.java
```diff
@@ -20,11 +20,16 @@
...
 import org.optaplanner.core.api.score.buildin.hardsoftdouble.HardSoftDoubleScore;
 import org.optaplanner.core.api.score.buildin.hardsoftdouble.HardSoftDoubleScoreHolder;
+import org.optaplanner.core.api.score.holder.ScoreHolder;
 import org.optaplanner.core.config.score.trend.InitializingScoreTrendLevel;
 import org.optaplanner.core.impl.score.definition.AbstractScoreDefinition;
 import org.optaplanner.core.impl.score.inliner.ScoreInliner;
 import org.optaplanner.core.impl.score.trend.InitializingScoreTrend;
 
+/**
+ * @deprecated Double-based scores are deprecated as floating point numbers can't represent a decimal number correctly.
```
ge0ffrey, line 30: The real reason they are deprecated is because "floating point numbers are not associative (A + B + C == C + B + A) and therefore they are not compatible with incremental score calculation".

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/DroolsScoreDirector.java
```diff
@@ -82,6 +82,9 @@ private void resetKieSession() {
         }
     }
 
+    // The scoreHolder method is not actually deprecated for removal.
+    // It will only be moved to a different type at a time when we can make that change in public API.
+    @SuppressWarnings("deprecation")
```
ge0ffrey, line 87: Would it be possible to prefix such comments with `// TODO remove in 8.0`?
The reason I like the "8.0" there is because we'll do a find in path for "8.0" after branching

optaplanner-core/src/main/java/org/optaplanner/core/api/score/buildin/bendable/BendableScoreHolder.java
```diff
@@ -31,6 +31,7 @@
 /**
  * @see BendableScore
  */
+@Deprecated
```
ge0ffrey, line 34: It's still deprecated, but the others are not?

PR #795, opened 2020-06-04
Align Asciidoctor plugin Netty dependencies to kie-parent

almope:
Signed-off-by: Alberto Morales Pérez <almorale@redhat.com>

The Asciidoctor plugin mojo that uses the Netty dependency is actually not used in the project so the change should be safe.

### DIFF
--- a/optaplanner-docs/pom.xml
+++ b/optaplanner-docs/pom.xml
@@ -44,6 +44,13 @@
         <plugin>
           <groupId>org.asciidoctor</groupId>
           <artifactId>asciidoctor-maven-plugin</artifactId>
+          <dependencies>
+            <dependency>
+              <groupId>io.netty</groupId>
+              <artifactId>netty-codec-http</artifactId>
+              <version>${version.io.netty}</version>
+            </dependency>
+          </dependencies>
           <configuration>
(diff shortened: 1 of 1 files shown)


optaplanner-docs/pom.xml
```diff
@@ -44,6 +44,13 @@
         <plugin>
           <groupId>org.asciidoctor</groupId>
           <artifactId>asciidoctor-maven-plugin</artifactId>
+          <dependencies>
+            <dependency>
+              <groupId>io.netty</groupId>
+              <artifactId>netty-codec-http</artifactId>
+              <version>${version.io.netty}</version>
+            </dependency>
+          </dependencies>
```
ge0ffrey, line 53: You can add this dependency in kie-parent instead:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L1324

ge0ffrey:
@almope Wydt?

almope:
Hi @ge0ffrey,
thanks for your review. I'm still working with similar changes in other repositories. Following your recommendation I'll move this changes to KIE parent after I complete my checks. I've moved this to draft to reflect that this is a work in progress. Thanks!

PR #798, opened 2020-06-08
PLANNER-750 Backwards incompatible changes to the API

triceo:


.github/workflows/os_java_ci.yml
```diff
@@ -5,10 +5,10 @@ name: OptaPlanner CI
 on:
   push:
     branches:
-      - master
+      - future-8.x
   pull_request:
     branches:
-      - master
+      - future-8.x
```
ge0ffrey, line 11: These should probably stay master I guess? cc @rsynek
The future-8.x branch doesn't exist in the optaplanner repo, only on your fork.

rsynek: Here, yes. On the fork, it must be the future-8.x branch. I wonder if we just can have both of them (master, future-8.x) in the yaml file until this PR is ready to be merged to master.

triceo: Yes, we can.

rsynek: Actually, we don't necessarily have to: the OS x JDK matrix is available on the small, incremental PRs, while Jenkins PR check and the SonarCloud are available here.

@ge0ffrey I would just leave it as it is and before we merge this PR, we put the master branch back in the yaml file.

ge0ffrey: Sounds good :)

PR #799, opened 2020-06-09
Fix issues with artifact transfer in CI

rsynek:
Should fix the connection reset issues with transferring dependencies.

### DIFF
--- a/.github/workflows/jpa_integration_ci.yml
+++ b/.github/workflows/jpa_integration_ci.yml
@@ -11,6 +11,7 @@ on:
       - master
 
 env:
+  # To prevent build failures due to "Connection reset" during artifact download.
   MVN_CONNECTION_PROPS: -Dmaven.wagon.httpconnectionManager.ttlSeconds=60 -Dmaven.wagon.http.retryHandler.count=5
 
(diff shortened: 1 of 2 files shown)


triceo:
Doesn't appear to be fixed - in fact, quite the contrary.

rsynek:
@triceo ironic, isn't it? It seems that `.` in a property name has a special meaning in PowerShell, which runs the command on Windows.

rsynek:
The CI passed, let me re-run it a couple of times to increase the probability that the issue is gone.

.github/workflows/jpa_integration_ci.yml
```diff
@@ -10,6 +10,13 @@ on:
     branches:
       - master
 
+env:
+  MVN_CONNECTION_PROPS: -Dmaven.wagon.httpconnectionManager.ttlSeconds=60 -Dmaven.wagon.http.retryHandler.count=5
```
triceo, line 14: I'd add a comment here, explaining why those are necessary.

rsynek: good idea

PR #801, opened 2020-06-13
Every time a new problem fact change is added, reset is requested. A …

mayurhole:
Every time a new problem fact change is added, reset is requested. An already running session shall be disposed of (also done in DroolsScoreDirector), else originalKieBase and currentKieBase in DroolsConstraintSessionFactory keep accumulating sessions in concurrent hash map leading to memory leaks.


triceo:
ok to test

triceo:
Thanks for the PR, @mayurhole. I'll review next week.

triceo:
This will either make 7.39.0.Final or 7.40.0.Final, as the cut-off is any minute now. :-)
Thank you for the fix!

mayurhole:
Thanks for reviewing the PR and integrating the fix. Looking forward to having the next versions out :)

PR #802, opened 2020-06-16
PLANNER-2021 Don't use JUnit 4's Assert class anymore

triceo:
Manual review of this PR is not practical. I suggest we merge this when:

- [x] CI shows no failures and
- [x] Sonar shows no drop in test coverage.
- [ ] Maven build passes even if `kie-parent` doesn't have `junit` dependency (needs some local building) (Jiri's suggestion).


rsynek:
@triceo I think I've found one more: https://github.com/kiegroup/optaplanner/blob/master/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMoveTest.java#L42

triceo:
Ah, good, I didn't search for that one. Will fix.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/constraint/IndictmentTest.java
```diff
@@ -32,28 +32,28 @@ public void getScoreTotal() {
         TestdataEntity e2 = new TestdataEntity("e2");
         TestdataEntity e3 = new TestdataEntity("e3");
         Indictment indictment = new Indictment(e1, SimpleScore.ZERO);
-        assertEquals(SimpleScore.ZERO, indictment.getScore());
+        assertThat(indictment.getScore()).isEqualTo((org.optaplanner.core.api.score.Score) SimpleScore.ZERO);
```
yurloc, line 35: Wrong cast again. See also children of `AbstractConstraintStreamTest`.

optaplanner-core/src/test/java/org/optaplanner/core/config/heuristic/selector/entity/EntitySelectorConfigTest.java
```diff
@@ -39,9 +37,11 @@ public void phaseOriginal() {
         EntitySelector entitySelector = entitySelectorConfig.buildEntitySelector(
                 buildHeuristicConfigPolicy(),
                 SelectionCacheType.JUST_IN_TIME, SelectionOrder.RANDOM);
-        assertInstanceOf(FromSolutionEntitySelector.class, entitySelector);
-        assertNotInstanceOf(ShufflingEntitySelector.class, entitySelector);
-        assertEquals(SelectionCacheType.PHASE, entitySelector.getCacheType());
+        assertThat((Object) entitySelector)
```
yurloc, line 40: Redundant cast. It seems this is done automatically by your IDE. Please check if you can avoid that.

optaplanner-core/src/test/java/org/optaplanner/core/config/heuristic/selector/move/MoveSelectorConfigTest.java
```diff
@@ -53,8 +50,12 @@ public AssertingMoveSelectorConfig(MoveSelector baseMoveSelector,
...
                 HeuristicConfigPolicy configPolicy,
                 SelectionCacheType minimumCacheType, boolean randomSelection) {
-            assertEquals(expectedMinimumCacheType, minimumCacheType);
-            assertEquals(expectedRandomSelection, randomSelection);
+            assertThat(minimumCacheType).isEqualTo(expectedMinimumCacheType);
+            if (expectedRandomSelection) {
+                assertThat(randomSelection).isTrue();
+            } else {
+                assertThat(randomSelection).isFalse();
+            }
```
yurloc, line 58: Why this `if`?

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/constraintweight/descriptor/ConstraintWeightDescriptorTest.java
```diff
@@ -55,10 +53,12 @@ public void extractionFunction() {
...
         solution.setConstraintConfiguration(constraintConfiguration);
 
-        assertSame(constraintConfiguration,
-                solutionDescriptor.getConstraintConfigurationMemberAccessor().executeGetter(solution));
-        assertEquals(SimpleScore.ZERO, firstWeightDescriptor.createExtractor().apply(solution));
-        assertEquals(SimpleScore.of(7), secondWeightDescriptor.createExtractor().apply(solution));
+        assertThat(solutionDescriptor.getConstraintConfigurationMemberAccessor().executeGetter(solution))
+                .isSameAs(constraintConfiguration);
+        assertThat(firstWeightDescriptor.createExtractor().apply(solution)).isEqualTo(
+                (org.optaplanner.core.api.score.Score<? extends org.optaplanner.core.api.score.Score>) SimpleScore.ZERO);
```
yurloc, line 59: Redundant cast.

optaplanner-core/src/test/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfigTest.java
```diff
@@ -184,13 +181,13 @@ public SimpleScore calculateScore() {
     @Test
     public void testGenerateDroolsTestOption() {
         ScoreDirectorFactoryConfig config = new ScoreDirectorFactoryConfig();
-        assertNull(config.isGenerateDroolsTestOnError());
+        assertThat(config.isGenerateDroolsTestOnError()).isNull();
         config.setGenerateDroolsTestOnError(true);
-        assertTrue(config.isGenerateDroolsTestOnError());
+        assertThat((boolean) config.isGenerateDroolsTestOnError()).isTrue();
```
yurloc, line 186: Redundant cast.

optaplanner-core/src/test/java/org/optaplanner/core/impl/solver/thread/OrderByMoveIndexBlockingQueueTest.java
```diff
@@ -170,19 +169,24 @@ public void addExceptionIsNotEatenIfNextStepStartsBeforeTaken() throws Interrupt
...
     private void assertResult(String moveCode, boolean doable,
             OrderByMoveIndexBlockingQueue.MoveResult<TestdataSolution> result) {
         assertCode(moveCode, result.getMove());
-        assertEquals(doable, result.isMoveDoable());
+        boolean y = result.isMoveDoable();
+        if (doable) {
+            assertThat(y).isTrue();
+        } else {
+            assertThat(y).isFalse();
+        }
```
yurloc, line 189: Is this `if` intentional?

optaplanner-examples/src/test/java/org/optaplanner/examples/common/app/SolverPerformanceTest.java
```diff
@@ -121,27 +119,28 @@ private void runSpeedTest(
...
         InnerScoreDirectorFactory<Solution_> scoreDirectorFactory = (InnerScoreDirectorFactory<Solution_>) solverFactory
                 .getScoreDirectorFactory();
         Score bestScore = scoreDirectorFactory.getSolutionDescriptor().getScore(bestSolution);
         ScoreDefinition scoreDefinition = scoreDirectorFactory.getScoreDefinition();
         Score bestScoreLimit = scoreDefinition.parseScore(bestScoreLimitString);
-        assertTrue("The bestScore (" + bestScore + ") must be at least the bestScoreLimit (" + bestScoreLimit + ").",
-                bestScore.compareTo(bestScoreLimit) >= 0);
+        assertThat(bestScore)
+                .isGreaterThanOrEqualTo(bestScoreLimit)
+                .as("The bestScore (" + bestScore + ") must be at least the bestScoreLimit (" + bestScoreLimit + ").");
```
yurloc, line 130: `as()` needs to go before assertions, otherwise it's unreachable.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfigurationTest.java
```diff
@@ -131,46 +130,49 @@ public void solverProperties() {
...
                     SolverConfig solverConfig = context.getBean(SolverConfig.class);
-                    assertEquals("2", solverConfig.getMoveThreadCount());
-                    assertNotNull(context.getBean(SolverFactory.class));
+                    assertThat(solverConfig.getMoveThreadCount()).isEqualTo("2");
+                    assertThat(context.getBean(SolverFactory.class)).isNotNull();
                 });
     }
 
+    private void assertEquals(EnvironmentMode fullAssert, EnvironmentMode environmentMode) {
+    }
```
yurloc, line 152: Committed by mistake?

yurloc:
@triceo `optaplanner-core` still depends on JUnit 4 (`Assume`). There may be more things like that. I haven't checked other modules. I suggest that you test this locally by removing `junit` dependency from `kie-parent` and build the whole optaplanner repo to prove all dependencies have been removed.

My assumption is that the goal of this PR is to get rid of JUnit 4 dependency completely. Correct me if I'm wrong and this is only about `Assert`.

I've added an item for this to the check list, if you agree. Otherwise, looks good!

triceo:
I did not actually try to remove the dependency. Good point about `Assume`, I did forget that exists.

triceo:
Removing JUnit 4 will not work, as we need to be able to continue using it with `optaplanner-test`. That will only be changed in 8.

Besides that, I've done all I could.

PR #804, opened 2020-06-17
Avoid testing exceptions with try/catch

yurloc:


triceo:
(When the tests pass.)

PR #806, opened 2020-06-18
PLANNER-2021 Remove the remaining uses of JUnit 4

triceo:
JUnit 4 is still on the classpath. That will be only fixed when we stop depending on kie-parent.


triceo:
Mistakenly opened for the wrong repo.

PR #808, opened 2020-06-22
WIP: Convert MeetingScheduling example to ConstraintStreams

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-1820

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
@@ -1,0 +1,222 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.meetingscheduling.optional.score;
+
+import static org.optaplanner.core.api.score.stream.Joiners.equal;
+import static org.optaplanner.core.api.score.stream.Joiners.filtering;
+import static org.optaplanner.core.api.score.stream.Joiners.greaterThan;
+import static org.optaplanner.core.api.score.stream.Joiners.lessThan;
+
+import org.optaplanner.core.api.score.stream.Constraint;
+import org.optaplanner.core.api.score.stream.ConstraintFactory;
+import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.examples.meetingscheduling.domain.Attendance;
+import org.optaplanner.examples.meetingscheduling.domain.MeetingAssignment;
+import org.optaplanner.examples.meetingscheduling.domain.PreferredAttendance;
+import org.optaplanner.examples.meetingscheduling.domain.RequiredAttendance;
+import org.optaplanner.examples.meetingscheduling.domain.Room;
+import org.optaplanner.examples.meetingscheduling.domain.TimeGrain;
+
+public class MeetingSchedulingConstraintProvider implements ConstraintProvider {
+
+    @Override
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
+        return new Constraint[] {
+                roomConflict(constraintFactory),
+                avoidOvertime(constraintFactory),
+                requiredAttendanceConflict(constraintFactory),
+                requiredRoomCapacity(constraintFactory),
+                startAndEndOnSameDay(constraintFactory),
+                requiredAndPreferredAttendanceConflict(constraintFactory),
+                preferredAttendanceConflict(constraintFactory),
+                doMeetingsAsSoonAsPossible(constraintFactory),
+                oneBreakBetweenConsecutiveMeetings(constraintFactory),
+                overlappingMeetings(constraintFactory),
+                assignLargerRoomsFirst(constraintFactory),
+                roomStability(constraintFactory)
+        };
+    }
+
+    // ************************************************************************
+    // Hard constraints
+    // ************************************************************************
+
+    protected Constraint roomConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MeetingAssignment.class)
+                .filter(leftAssignment -> leftAssignment.getRoom() != null)
+                .join(MeetingAssignment.class,
+                        equal(MeetingAssignment::getRoom),
+                        lessThan(MeetingAssignment::getId),
+                        filtering((leftAssignment, rightAssignment) -> rightAssignment.calculateOverlap(leftAssignment) > 0))
+                .penalizeConfigurable("Room conflict",
+                        (leftAssignment, rightAssignment) -> rightAssignment.calculateOverlap(leftAssignment));
+    }
+
+    protected Constraint avoidOvertime(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MeetingAssignment.class)
+                .filter(meetingAssignment -> meetingAssignment.getStartingTimeGrain() != null)
+                .ifNotExists(TimeGrain.class,
+                        equal(MeetingAssignment::getLastTimeGrainIndex, TimeGrain::getGrainIndex))
+                .penalizeConfigurable("Don't go in overtime", MeetingAssignment::getLastTimeGrainIndex);
+    }
+
+    protected Constraint requiredAttendanceConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUniquePair(RequiredAttendance.class,
+                equal(RequiredAttendance::getPerson))
+                .join(MeetingAssignment.class,
+                        equal((leftRequiredAttendance, rightRequiredAttendance) -> leftRequiredAttendance.getMeeting(),
+                                MeetingAssignment::getMeeting),
+                        filtering((leftRequiredAttendance, rightRequiredAttendance,
+                                leftAssignment) -> leftAssignment.getStartingTimeGrain() != null))
+                .join(MeetingAssignment.class,
+                        equal((leftRequiredAttendance, rightRequiredAttendance, leftAssignment) -> rightRequiredAttendance
+                                .getMeeting(),
+                                MeetingAssignment::getMeeting),
+                        filtering((leftRequiredAttendance, rightRequiredAttendance, leftAssignment,
+                                rightAssignment) -> rightAssignment.calculateOverlap(leftAssignment) > 0))
+                .penalizeConfigurable("Required attendance conflict",
+                        (leftRequiredAttendance, rightRequiredAttendance, leftAssignment, rightAssignment) -> rightAssignment
+                                .calculateOverlap(leftAssignment));
+    }
+
+    protected Constraint requiredRoomCapacity(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MeetingAssignment.class)
+                .filter(meetingAssignment -> meetingAssignment.getRequiredCapacity() > meetingAssignment.getRoomCapacity())
+                .penalizeConfigurable("Required room capacity",
+                        meetingAssignment -> meetingAssignment.getRequiredCapacity() - meetingAssignment.getRoomCapacity());
+    }
+
+    protected Constraint startAndEndOnSameDay(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MeetingAssignment.class)
+                .filter(meetingAssignment -> meetingAssignment.getStartingTimeGrain() != null)
+                .join(TimeGrain.class,
+                        equal(MeetingAssignment::getLastTimeGrainIndex, TimeGrain::getGrainIndex),
+                        filtering((meetingAssignment,
+                                timeGrain) -> meetingAssignment.getStartingTimeGrain().getDay() != timeGrain.getDay()))
+                .penalizeConfigurable("Start and end on same day");
+    }
+
+    // ************************************************************************
+    // Medium constraints
+    // ************************************************************************
+
+    protected Constraint requiredAndPreferredAttendanceConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(RequiredAttendance.class)
+                .join(PreferredAttendance.class,
+                        equal(RequiredAttendance::getPerson, PreferredAttendance::getPerson))
+                .join(MeetingAssignment.class,
+                        equal((requiredAttendance, preferredAttendance) -> requiredAttendance.getMeeting(),
+                                MeetingAssignment::getMeeting),
+                        filtering((requiredAttendance, preferredAttendance,
+                                leftAssignment) -> leftAssignment.getStartingTimeGrain() != null))
+                .join(MeetingAssignment.class,
+                        equal((requiredAttendance, preferredAttendance, leftAssignment) -> preferredAttendance.getMeeting(),
+                                MeetingAssignment::getMeeting),
+                        filtering((requiredAttendance, preferredAttendance, leftAssignment,
+                                rightAssignment) -> rightAssignment.calculateOverlap(leftAssignment) > 0))
+                .penalizeConfigurable("Required and preferred attendance conflict");
+    }
+
+    protected Constraint preferredAttendanceConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUniquePair(PreferredAttendance.class,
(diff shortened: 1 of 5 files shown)


optaplanner-examples/src/test/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProviderTest.java
```diff
@@ -0,0 +1,616 @@
...
+        meeting.setDurationInGrains(4);
+
+        MeetingAssignment meetingAssignment = new MeetingAssignment();
+        meetingAssignment.setStartingTimeGrain(assignmentTimeGrain);
+        meetingAssignment.setMeeting(meeting);
+
+        constraintVerifier.verifyThat(MeetingSchedulingConstraintProvider::avoidOvertime)
+                .given(meetingAssignment)
+                .penalizesBy(3);
+    }
```
triceo, line 133: The problem lies in `from(MeetingAssignment.class)`. There are no such instances, because you never set any value to `MeetingAssignment.room` planning variable. `from()` therefore treats those instances as uninitialized, and skips them. Two solutions:

- use `fromUnfiltered()` in the constraint provider or
- initialize your entities properly in the test.

Also, I suggest you make sure your `@PlanningId`s are never null, because the following PR will start enforcing that pretty soon:
https://github.com/triceo/optaplanner/pull/14

I have not reviewed any other constraints as I assumed the same problem would be plagueing them as well.

optaplanner-examples/src/test/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProviderTest.java
```diff
@@ -0,0 +1,616 @@
...
+
+        Meeting meeting = new Meeting();
+        meeting.setDurationInGrains(4);
+
+        MeetingAssignment meetingAssignment = new MeetingAssignment();
+        meetingAssignment.setStartingTimeGrain(assignmentTimeGrain);
+        meetingAssignment.setMeeting(meeting);
+
+        constraintVerifier.verifyThat(MeetingSchedulingConstraintProvider::avoidOvertime)
+                .given(meetingAssignment)
```
triceo, line 131: Please remember to include in the `given(...)` call all the facts relevant to the problem. Not including them here means you won't be able to use them in `join()`, `exists()` etc.

optaplanner-examples/src/test/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProviderTest.java
triceo: I suggest you create constructors for the facts/entities so that you are forced to always provide all the information that must not be null.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    protected Constraint preferredAttendanceConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PreferredAttendance.class)
+                .join(PreferredAttendance.class,
+                        equal(PreferredAttendance::getPerson, PreferredAttendance::getPerson),
+                        lessThan(PreferredAttendance::getId, PreferredAttendance::getId))
+                .join(MeetingAssignment.class,
+                        filtering((leftAttendance, rightAttendance,
+                                leftAssignment) -> leftAttendance.getMeeting() == leftAssignment.getMeeting()),
```
triceo, line 141: `Joiners.equal(...)`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+                        filtering((leftAttendance, rightAttendance,
+                                leftAssignment) -> leftAssignment.getStartingTimeGrain() != null))
+                .join(MeetingAssignment.class,
+                        filtering((leftAttendance, rightAttendance, leftAssignment,
+                                rightAssignment) -> rightAttendance.getMeeting() == rightAssignment.getMeeting()),
```
triceo, line 146: `Joiners.equal(...)`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+                        filtering((requiredAttendance, preferredAttendance, leftAssignment,
+                                rightAssignment) -> rightAssignment.calculateOverlap(leftAssignment) > 0))
+                .penalizeConfigurable("Required and preferred attendance conflict");
+    }
+
+    protected Constraint preferredAttendanceConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PreferredAttendance.class)
+                .join(PreferredAttendance.class,
+                        equal(PreferredAttendance::getPerson, PreferredAttendance::getPerson),
+                        lessThan(PreferredAttendance::getId, PreferredAttendance::getId))
```
triceo, line 138: `fromUniquePair(...)`?

Also, when operating on objects of the same type, `equal(PreferredAttendance::getPerson)` is syntactic sugar for `equal(PreferredAttendance::getPerson, PreferredAttendance::getPerson)`.

I believe the same is true for `lessThan()`.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    protected Constraint oneBreakBetweenConsecutiveMeetings(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MeetingAssignment.class)
+                .filter(meetingAssignment -> meetingAssignment.getStartingTimeGrain() != null)
+                .join(MeetingAssignment.class,
+                        filtering((leftAssignment, rightAssignment) -> rightAssignment.getStartingTimeGrain() != null),
+                        filtering((leftAssignment,
+                                rightAssignment) -> leftAssignment
+                                        .getLastTimeGrainIndex() == rightAssignment.getStartingTimeGrain().getGrainIndex() - 1))
```
triceo, line 170: I think this should be doable via `equal()` as well.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+                .penalizeConfigurable("One TimeGrain break between two consecutive meetings");
+    }
+
+    protected Constraint overlappingMeetings(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MeetingAssignment.class)
+                .filter(meetingAssignment -> meetingAssignment.getStartingTimeGrain() != null)
+                .join(MeetingAssignment.class,
+                        filtering((leftAssignment, rightAssignment) -> rightAssignment.getStartingTimeGrain() != null),
+                        filtering((leftAssignment,
+                                rightAssignment) -> rightAssignment.getMeeting().getId() < leftAssignment.getMeeting().getId()),
```
triceo, line 180: `lessThan()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+
+    protected Constraint roomStability(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Attendance.class)
+                .join(Attendance.class,
+                        equal(Attendance::getPerson, Attendance::getPerson),
+                        filtering((leftAttendance,
+                                rightAttendance) -> leftAttendance.getMeeting() != rightAttendance.getMeeting()))
+                .join(MeetingAssignment.class,
+                        filtering((leftAttendance, rightAttendance,
+                                leftAssignment) -> leftAttendance.getMeeting() == leftAssignment.getMeeting()),
```
triceo, line 205: `equals()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+                        filtering((leftAttendance, rightAttendance,
+                                leftAssignment) -> leftAssignment.getStartingTimeGrain() != null))
+                .join(MeetingAssignment.class,
+                        filtering((leftAttendance, rightAttendance, leftAssignment,
+                                rightAssignment) -> rightAttendance.getMeeting() == rightAssignment.getMeeting()),
```
triceo, line 210: `equals()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+                        filtering((leftAttendance, rightAttendance, leftAssignment,
+                                rightAssignment) -> rightAssignment.getStartingTimeGrain() != null),
+                        filtering((leftAttendance, rightAttendance, leftAssignment,
+                                rightAssignment) -> leftAssignment.getStartingTimeGrain().getGrainIndex() < rightAssignment
+                                        .getStartingTimeGrain().getGrainIndex()),
```
triceo, line 215: From the looks of it, this should be doable with `lessThan(...)`.

optaplanner-examples/src/main/resources/org/optaplanner/examples/meetingscheduling/solver/meetingSchedulingSolverConfig.xml
```diff
@@ -7,7 +7,11 @@
   <entityClass>org.optaplanner.examples.meetingscheduling.domain.MeetingAssignment</entityClass>
 
   <scoreDirectorFactory>
+    <!--<constraintProviderClass>org.optaplanner.examples.meetingscheduling.optional.score.MeetingSchedulingConstraintProvider</constraintProviderClass>-->
     <scoreDrl>org/optaplanner/examples/meetingscheduling/solver/meetingSchedulingConstraints.drl</scoreDrl>
+    <!--<assertionScoreDirectorFactory>-->
+      <!--<scoreDrl>org/optaplanner/examples/meetingscheduling/solver/meetingSchedulingConstraints.drl</scoreDrl>-->
+    <!--</assertionScoreDirectorFactory> -->
```
triceo, line 14: I'll take this as you've made sure the DRL and the Provider give matching results. Good.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    protected Constraint requiredAttendanceConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(RequiredAttendance.class)
+                .join(RequiredAttendance.class,
+                        equal(RequiredAttendance::getPerson, RequiredAttendance::getPerson),
+                        lessThan(RequiredAttendance::getId, RequiredAttendance::getId))
+                .join(MeetingAssignment.class,
+                        filtering((leftRequiredAttendance, rightRequiredAttendance,
+                                leftAssignment) -> leftRequiredAttendance.getMeeting() == leftAssignment.getMeeting()),
```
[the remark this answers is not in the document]
cuijulian: How does this work on `TriConstraintStreams`? I'm guessing something like: `equal(RequiredAttendance::getMeeting, MeetingAssignment::getMeeting)`. But in this case wouldn't the `RequiredAttendance` being compared be ambiguous, since there are two of them joined in this stream?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -0,0 +1,224 @@
...
+                        filtering((leftRequiredAttendance, rightRequiredAttendance,
+                                leftAssignment) -> leftAssignment.getStartingTimeGrain() != null))
+                .join(MeetingAssignment.class,
+                        filtering((leftRequiredAttendance, rightRequiredAttendance, leftAssignment,
+                                rightAssignment) -> rightRequiredAttendance.getMeeting() == rightAssignment.getMeeting()),
```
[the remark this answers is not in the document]
cuijulian: Same question here, if the Joiner used here is `equal(RequiredAttendance::getMeeting, MeetingAssignment::getMeeting)`, wouldn't the instances being compared be ambiguous?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
cuijulian: Ah, good to know, wasn't aware of the `equal(PreferredAttendance::getPerson)` syntactic sugar. As for `fromUniquePair`, I actually decided against using it since doing the `join` manually reads better for me. I don't have an issue to changing it, but in general is syntactic sugar preferred?

optaplanner-examples/src/main/resources/org/optaplanner/examples/meetingscheduling/solver/meetingSchedulingSolverConfig.xml
cuijulian: Yup, ran `FULL_ASSERT` for a few minutes, no issues.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
cuijulian: I don't see it, is there a way to use `Predicate`s with `equal()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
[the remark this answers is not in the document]
triceo: The joiner will always take all the arguments from the left, and the single argument from the right.
So this would be something like `equal((lra, rra) -> lra.getMeeting(), MeetingAssignment::getMeeting)`.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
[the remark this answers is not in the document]
triceo: Same answer. :-)

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
cuijulian: Similar to `equal()`, how to put lambda expression in `lessThan()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
cuijulian: Same as above, how to put lambda expression in `lessThan()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
triceo: You need to move past method references. Simply use full lambda syntax - `(a, b) -> c`. That will allow you to write all joiners without restrictions.

cuijulian:
> Ran the tests locally, they pass.
> 
> Some of the filters possibly could be replaced by indexed joiners. You may have good reasons for using filters instead, but they are not immediately obvious to me.
> 
> OK to go when comments are resolved, either by explanation or a fix.
> Good job!

Regarding the comments replacing the `filtering()` methods with `equal()`, I don't see a way to do it for constraint streams with cardinality above two, since the Class instances would be ambiguous. The rest of the comments I addressed separately.

PR #814, opened 2020-06-24
PLANNER-2011 Somewhat improve Examination Constraint Provider performance

triceo:
Minor improvements in index performance. Also re-implements parts of the constraint provider to do the same thing as DRL does.
This speeds up Examination from 6 to ~140. So still bad, but not apocalyptic anymore.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
@@ -55,7 +55,6 @@ public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
 
     @Override
     public Function<A, Object> getLeftMapping(int index) {
-        assertMappingIndex(index);
         return (Function<A, Object>) leftMappings[index];
     }
 
@@ -73,7 +72,6 @@ public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
 
     @Override
     public Function<B, Object> getRightMapping(int index) {
-        assertMappingIndex(index);
         return (Function<B, Object>) rightMappings[index];
     }
 
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
@@ -24,7 +24,7 @@ import org.optaplanner.core.impl.score.stream.common.JoinerType;
 
 public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
 
-    private final List<SingleBiJoiner<A, B>> joiners;
+    private final List<SingleBiJoiner<A, B>> joinerList;
     private final JoinerType[] joinerTypes;
     private final Function<A, ?>[] leftMappings;
     private final Function<B, ?>[] rightMappings;
@@ -33,7 +33,7 @@ public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
         if (joinerList.isEmpty()) {
             throw new IllegalArgumentException("The joinerList (" + joinerList + ") must not be empty.");
         }
-        this.joiners = joinerList;
+        this.joinerList = joinerList;
         this.joinerTypes = joinerList.stream()
                 .map(SingleBiJoiner::getJoinerType)
                 .toArray(JoinerType[]::new);
@@ -46,7 +46,7 @@ public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
     }
 
     public List<SingleBiJoiner<A, B>> getJoinerList() {
-        return joiners;
+        return joinerList;
     }
 
     // ************************************************************************
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
@@ -19,13 +19,13 @@ package org.optaplanner.core.impl.score.stream.bi;
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Function;
-import java.util.stream.IntStream;
 
 import org.optaplanner.core.impl.score.stream.common.JoinerType;
 
 public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
 
-    private final List<SingleBiJoiner<A, B>> joinerList;
+    private final List<SingleBiJoiner<A, B>> joiners;
+    private final JoinerType[] joinerTypes;
     private final Function<A, ?>[] leftMappings;
     private final Function<B, ?>[] rightMappings;
 
@@ -33,7 +33,10 @@ public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
         if (joinerList.isEmpty()) {
             throw new IllegalArgumentException("The joinerList (" + joinerList + ") must not be empty.");
         }
-        this.joinerList = joinerList;
+        this.joiners = joinerList;
+        this.joinerTypes = joinerList.stream()
+                .map(SingleBiJoiner::getJoinerType)
+                .toArray(JoinerType[]::new);
         this.leftMappings = joinerList.stream()
                 .map(SingleBiJoiner::getLeftMapping)
                 .toArray(Function[]::new);
@@ -43,7 +46,7 @@ public final class CompositeBiJoiner<A, B> extends AbstractBiJoiner<A, B> {
     }
 
     public List<SingleBiJoiner<A, B>> getJoinerList() {
-        return joinerList;
+        return joiners;
     }
 
     // ************************************************************************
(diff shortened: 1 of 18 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractJoiner.java
```diff
@@ -18,14 +18,6 @@
...
 public abstract class AbstractJoiner {
 
-    protected void assertMappingIndex(int index) {
-        int mappingCount = getJoinerTypes().length;
-        if (index >= mappingCount) {
-            throw new IllegalArgumentException(
-                    "Joiner only has (" + mappingCount + ") mappings, requested index (" + index + ")");
-        }
-    }
-
```
triceo, line 28: I am removing this as it just replaces an exception with another one.
This brings no value and costs precious time on the CS-D hot path.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
```diff
@@ -52,39 +55,29 @@
 
     @Override
     public Function<A, Object> getLeftMapping(int index) {
-        assertMappingIndex(index);
```
ge0ffrey, line 55: Why can we get rid of these fail fasts? Because it throws an ArrayOutOfBoundsException?
If only OptaPlanner code can trigger that, then the ArrayOutOfBoundsException is certainly fine. If a user can do something wrong that easily triggers it, it might not be ideal. But I can't immediately think of something, so +1 to remove the assert.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/AbstractJoiner.java
ge0ffrey: Agreed.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bi/CompositeBiJoiner.java
ge0ffrey: Retracting this comment. See below.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -65,12 +83,10 @@ protected Constraint periodDurationTooShort(ConstraintFactory constraintFactory)
...
     protected Constraint roomCapacityTooSmall(ConstraintFactory constraintFactory) {
-        return constraintFactory.from(Room.class)
-                .join(Exam.class, equal(Function.identity(), Exam::getRoom))
-                .join(Period.class, equal((room, exam) -> exam.getPeriod(), Function.identity()))
-                .groupBy((room, exam, period) -> room,
-                        (room, exam, period) -> period,
-                        ConstraintCollectors.sum((period, exam, room) -> exam.getTopicStudentSize()))
+        return constraintFactory.from(Exam.class)
+                .filter(exam -> exam.getRoom() != null)
+                .filter(exam -> exam.getPeriod() != null)
```
ge0ffrey, line 88: a from() guarantees that all genuine planning variables aren't null.
So the room filter is definitely obsolete, because room is a planning variable on Exam.

For period its more complex. It's a genuine planning variable on LeadingExam, and a shadow variable on FollowingExam - and nothing on Exam. So I think it doesn't get filtered by default, because we're doing "from" Exam.
I would recommend adding a comment above that filter, to signal why were're filtering it.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -99,8 +115,8 @@ protected Constraint periodPenaltyExclusion(ConstraintFactory constraintFactory)
                         filtering((periodPenalty, leftExam) -> leftExam.getPeriod() != null))
                 .join(Exam.class,
                         equal((periodPenalty, leftExam) -> periodPenalty.getRightTopic(), Exam::getTopic),
-                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null),
-                        filtering((periodPenalty, leftExam, rightExam) -> leftExam.getPeriod().equals(rightExam.getPeriod())))
+                        equal((periodPenalty, leftExam) -> leftExam.getPeriod(), Exam::getPeriod),
+                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null))
```
ge0ffrey, line 119: 2 lines above says that the leftExam's period isn't null.
The line above says that the rightExam period is the same one.
Therefore, checking if the rightExam isn't null is obsolete.
If A != null && A == B then B != null

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -114,9 +130,8 @@ protected Constraint periodPenaltyAfter(ConstraintFactory constraintFactory) {
                         filtering((periodPenalty, leftExam) -> leftExam.getPeriod() != null))
                 .join(Exam.class,
                         equal((periodPenalty, leftExam) -> periodPenalty.getRightTopic(), Exam::getTopic),
-                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null),
-                        filtering((periodPenalty, leftExam,
-                                rightExam) -> leftExam.getPeriod().getPeriodIndex() <= rightExam.getPeriod().getPeriodIndex()))
+                        lessThanOrEqual((periodPenalty, leftExam) -> leftExam.getPeriodIndex(), Exam::getPeriodIndex),
+                        filtering((periodPenalty, leftExam, rightExam) -> rightExam.getPeriod() != null))
```
ge0ffrey, line 134: Doesn't this crash? rightExam.getPeriodIndex() can throw an NPE, right? Or return null that isn't Comparable?

Ideally, the rightExam period != null filtering must happen before the join:

```
.join(f.from(Exam).filter(exam.getPeriod != null)),
       lessThanOrEqual(....)
```

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -170,30 +185,31 @@ protected Constraint periodSpread(ConstraintFactory constraintFactory) {
...
                         filtering((config, topicConflict, leftExam,
                                 rightExam) -> getPeriodIndexDifferenceBetweenExams(leftExam,
                                         rightExam) < (config.getPeriodSpreadLength() + 1)))
                 .penalizeConfigurable("periodSpread",
-                        (config, topicConflict, leftExam) -> topicConflict.getStudentSize());
+                        (config, topicConflict, leftExam, rightExam) -> topicConflict.getStudentSize());
     }
 
     protected Constraint mixedDurations(ConstraintFactory constraintFactory) {
-        // 4 mixed durations of 100, 150, 200 and 200 should only result in 2 penalties (for 100&150 and 100&200).
```
ge0ffrey, line 184: I'd keep this comment.
Otherwise, it's the only reason why this rule is so complex, because of the ITC 2007 spec.
Normally, this would result in 5 penalties, probably mutliplied by the number of students.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -207,15 +223,19 @@ protected Constraint frontLoad(ConstraintFactory constraintFactory) {
     }
 
     protected Constraint periodPenalty(ConstraintFactory constraintFactory) {
-        return constraintFactory.from(Exam.class)
-                .filter(exam -> exam.getPeriod().getPenalty() != 0)
-                .penalizeConfigurable("periodPenalty", exam -> exam.getPeriod().getPenalty());
```
ge0ffrey, line 212: I prefer the original rule.
It reads better and it's probably faster too (no joins).

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
```diff
@@ -207,15 +223,19 @@ protected Constraint frontLoad(ConstraintFactory constraintFactory) {
...
 
     protected Constraint roomPenalty(ConstraintFactory constraintFactory) {
-        return constraintFactory.from(Exam.class)
-                .filter(exam -> exam.getRoom().getPenalty() != 0)
-                .penalizeConfigurable("roomPenalty", exam -> exam.getRoom().getPenalty());
+        return constraintFactory.from(Room.class)
+                .filter(room -> room.getPenalty() != 0)
+                .join(Exam.class,
+                        equal(Function.identity(), Exam::getRoom))
+                .penalizeConfigurable("roomPenalty", (room, exam) -> room.getPenalty());
```
ge0ffrey, line 238: Same here, see zulip "perf guidlines writing CS"

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
triceo: `getPeriodIndex()` has a built-in `null` check. That's how I can afford to do this.
But the `null` check here is redundant anyway, since `lessThanOrEqual()` would filter those situations out anyway.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/solver/score/ExaminationConstraintProvider.java
triceo: This is actually a performance improvement - there is only a small number of `Period`s, and many of them will be ignored due to the penalty being zero. Therefore, the subsequent `join` will result in *much less* tuples that have to be traversed, than if I start will all the `Exam`s.

I'm keeping this as is.

PR #815, opened 2020-06-25
JUnit 5 OptaPlannerExtension

yurloc:


optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/optional/score/VehicleRoutingConstraintProviderTest.java
```diff
@@ -19,25 +19,29 @@
...
 import org.optaplanner.examples.vehiclerouting.domain.Vehicle;
 import org.optaplanner.examples.vehiclerouting.domain.VehicleRoutingSolution;
 import org.optaplanner.examples.vehiclerouting.domain.location.AirLocation;
 import org.optaplanner.examples.vehiclerouting.domain.location.Location;
 import org.optaplanner.examples.vehiclerouting.domain.timewindowed.TimeWindowedCustomer;
 import org.optaplanner.examples.vehiclerouting.domain.timewindowed.TimeWindowedDepot;
 import org.optaplanner.test.api.score.stream.ConstraintVerifier;
+import org.optaplanner.test.junit.ConstraintProviderTest;
 
+@ConstraintProviderTest(solverConfig = "org/optaplanner/examples/vehiclerouting/solver/vehicleRoutingSolverConfig.xml")
```
yurloc, line 31: `solverConfig` is optional if you have a project with the default solver config at `/solverConfig.xml`.

optaplanner-examples/src/test/java/org/optaplanner/examples/vehiclerouting/optional/score/VehicleRoutingConstraintProviderTest.java
```diff
@@ -19,25 +19,29 @@
...
     private final Location location1 = new AirLocation(1L, 0.0, 0.0);
     private final Location location2 = new AirLocation(2L, 0.0, 4.0);
     private final Location location3 = new AirLocation(3L, 3.0, 0.0);
 
+    private final ConstraintVerifier<VehicleRoutingConstraintProvider, VehicleRoutingSolution> constraintVerifier;
+
+    public VehicleRoutingConstraintProviderTest(
+            ConstraintVerifier<VehicleRoutingConstraintProvider, VehicleRoutingSolution> constraintVerifier) {
+        this.constraintVerifier = constraintVerifier;
+    }
```
yurloc, line 43: The constructor is a boilerplate code. That's a downside. But still it's an improvement because you don't have to provide:
- constraint provider instance,
- solution and entity classes.

You still need to provide constraint provider and solution type argument (another boilerplate) but that's given by the `ConstraintVerifier` API. Maybe we should have made it non-generic.

triceo: So, it replaces one boilerplate with another, plus extra complexity in the form of an extension. 
Personally, I don't see a reason to do it.

ge0ffrey: Interesting experiment, good idea to try it out. Looking at the result of this experiment (based on what's possible), I agree that the before and after of this `VehicleRoutingConstraintProviderTest` class isn't convincing. About as much code, but far more magic.
Given than all enterprise deverlopers will use Quarkus or Spring, this technology would:
- be for newbie java developers (straight from school). Especially for those we want to avoid too much magic.
- create more choice for users. Do I use a the spring injection or the junit injection? More choice is not always a good thing.

yurloc: More complexity on our side but more user friendliness on the other side - that's the reason. Having to _repeat_ the entity classes, that are already defined in `solverConfig.xml` is jumping through hoops. Solver config should be the single source of truth, ideally.

On the other hand, the implementation in this PR is not ideal either. The assumption of default solver config file being `/solverConfig.xml` is duplicated by the Spring Boot and Quarkus extensions. If you're using non-default config location you have to specify it twice - once in `application.properties`, once in `@ConstraintProviderTest(solverConfig = ...)`.

ge0ffrey:
Closing this PR without merging because this POC shows that the benefit does not outweight the complexity, as discussed above. Thank you for building this POC to prove that.

PR #819, opened 2020-06-30
Change master version to 8.0.0-SNAPSHOT (8.x - step 1)

ge0ffrey:
- [X] change versions

- [x]  tests

- [x] optapplanner-bom transitive influence

- [x]  quickstarts

Not in scope of this PR: remove the usage of kie-parent 7 (that will be step 3), but don't loose it's valuable contents (enforcer checks etc)


optaplanner-bom/pom.xml
```diff
@@ -0,0 +1,487 @@
...
+    <!-- But for that to work, optaplanner-parent must have an empty dependencyManagement (so no kie-parent parent) -->
+    <groupId>org.kie</groupId>
+    <artifactId>kie-user-bom-parent</artifactId>
+    <version>7.40.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.optaplanner</groupId>
+  <artifactId>optaplanner-bom</artifactId>
+  <packaging>pom</packaging>
+  <version>8.0.0-SNAPSHOT</version>
```
ge0ffrey, line 17: This file was duplicated from droolsjbpm-build-bootstrap, which has the 7.x version (and cannot host the 8.x version because there is no 8 there).

Advantages of moving it here:
- Once it extends optaplanner-parent, half of the file can be removed (url, scm, issueManagement, etc)
- Less cross-repo PR's.
- Easier maintenance

optaplanner-build-parent/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.optaplanner</groupId>
+    <artifactId>optaplanner-parent</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>optaplanner-build-parent</artifactId>
+  <packaging>pom</packaging>
```
ge0ffrey, line 13: So, why this pom?
- The `optaplanner-parent` will (after step 3) hold all the metadata (url, scm, repo's, etc), maven plugin configuration (think enforcer, rules etc) etc. **But it cannot hold the dependencies because optaplanner-bom extends it.**
- The `optaplanner-build-parent` will (after step 3) extend optaplanner-parent and hold all dependencies, and also import optaplanner-bom.
- optaplanner-core, etc extend optaplanner-build-parent (not optaplanner-parent)

optaplanner-bom/pom.xml
```diff
@@ -0,0 +1,487 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <!-- TODO In step 3, this must extend optaplanner-parent instead and stop duplicating this metadata -->
+    <!-- But for that to work, optaplanner-parent must have an empty dependencyManagement (so no kie-parent parent) -->
+    <groupId>org.kie</groupId>
+    <artifactId>kie-user-bom-parent</artifactId>
+    <version>7.40.0-SNAPSHOT</version>
+  </parent>
```
yurloc, line 12: Should have empty <relativePath/> otherwise:

[WARNING] 'parent.relativePath' of POM org.optaplanner:optaplanner-bom:8.0.0-SNAPSHOT (/home/jlocker/src/github.com/kiegroup/optaplanner/optaplanner-bom/pom.xml) points at org.optaplanner:optaplanner-parent instead of org.kie:kie-user-bom-parent, please verify your project structure @ line 6, column 11

ge0ffrey: Good point, fixed :)

optaplanner-build-parent/pom.xml
rsynek: If I understand it correctly, the `optaplanner-parent` should server as a parent pom for both the `optaplanner-bom` and the `optaplanner-build-parent`.

What is the reason for the intention to define plugin configuration in the `optaplanner-parent`? Plugin configuration is specific for the build of the OptaPlanner project, so I would rather expect them to be defined in the `optaplanner-build-parent`.

ge0ffrey: Good point. Agreed. When we do step 3, plugin configuration should go into optaplanner-build-parent, not in optaplanner-parent.

ge0ffrey:
jenkins please retest this

ge0ffrey:
One of the github actions is failing on

```
 [ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for org.optaplanner:optaplanner-bom:8.0.0-SNAPSHOT: Could not find artifact org.kie:kie-user-bom-parent:pom:7.40.0-SNAPSHOT and 'parent.relativePath' points at no local POM @ line 6, column 11
```

ge0ffrey:
My last commit was a rather big change, after everyone already approved the PR:
 https://github.com/kiegroup/optaplanner/pull/819/commits/a0e6b68e50d4dfe98088e5699d97a12c7f7596b1
It should fix the CI issues.

ge0ffrey:
@rsynek Would you mind taking a look at my last commit here? Just so I don't sneak in the replacement of the parent of optapanner-bom :)

ge0ffrey:
That JPA CI job is failing due to a false negative caused by https://issues.redhat.com/browse/PLANNER-2049
Rest is green.

PR #826, opened 2020-07-03
PLANNER-1921 Increase test coverage for solver configs

rsynek:
Please note that JAXB changes the order of sibling elements - based on the position of corresponding fields in the class. It matters only for serialization. During deserialization, JAXB does not care about the order of sibling elements (unless told otherwise).


rsynek:
@mbiarnes Hi Michael, it seems the new PR check job was not triggered. Could you please advise what might be the issue?

rsynek:
Jenkins retest this

PR #832, opened 2020-07-09
Build optaplanner-core in optaplanner-persistence-jpa CI

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-2049

Accounts for `optaplanner-core` api and version changes by building it before running `optaplanner-persistence-jpa` CI container tests.

### DIFF
--- a/.github/workflows/jpa_integration_ci.yml
+++ b/.github/workflows/jpa_integration_ci.yml
@@ -49,4 +49,4 @@ jobs:
 
       # Builds the JPA module and runs tests in a PostgreSQL container
       - name: Build with Maven
-        run: cd optaplanner-persistence/optaplanner-persistence-jpa && mvn -B clean install -Ppostgresql ${{ env.MVN_CONNECTION_PROPS }}
+        run: mvn -B clean install -pl :optaplanner-persistence-jpa -am -Ppostgresql ${{ env.MVN_CONNECTION_PROPS }}


.github/workflows/jpa_integration_ci.yml
```diff
@@ -47,6 +47,10 @@ jobs:
         with:
           java-version: 11
 
+      # Builds optaplanner-core which optaplanner-persistence-jpa depends on, in case of changes
+      - name: Build optaplanner-core
```
ge0ffrey, line 51: 1) this should include all modules before optaplanner-persistence-jpa, including optaplanner-parent, optaplanner-bom, optaplanner-build-parent, optaplanner-persistence-common, ...
2) This is brittle: if we rename one of those modules or add one or remove one, it is broken.

Luckily, there is a fix, I think: when building with maven from the root of the repository, there is a command line argument to build a specific module (optaplanner-persistence-jpa) and anything that _needs_ to be build before it (so anything it depends on). That would ideal.

ge0ffrey: Here's that proposal: https://stackoverflow.com/a/3899772/472109

rsynek: Nice, I was not aware of this `-am` option. Thanks!

cuijulian: This is exactly what I'm looking for. Thanks @ge0ffrey for the tip! According to this [article on Maven advanced options](https://blog.sonatype.com/2009/10/maven-tips-and-tricks-advanced-reactor-options/) you can also specify projects that depend on the project list with `-amd, --also-make-dependents`.

cuijulian:
Jenkins retest this

PR #833, opened 2020-07-10
PLANNER-2061 Remove solution and a cloner from benchmark results

rsynek:
Tested with the benchmark aggregator on VRP benchmark config.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/loader/InstanceProblemProvider.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/loader/InstanceProblemProvider.java
@@ -16,17 +16,22 @@
 
 package org.optaplanner.benchmark.impl.loader;
 
+import java.util.Objects;
+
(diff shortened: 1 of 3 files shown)


rsynek:
@ge0ffrey pushing one unrelated commit because GitHub actions do not work on this branch yet.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/loader/InstanceProblemProvider.java
```diff
@@ -57,22 +62,24 @@ public void writeSolution(Solution_ solution, SubSingleBenchmarkResult subSingle
...
     @Override
     public int hashCode() {
-        return problem.hashCode();
+        return Objects.hash(problemName, problem, solutionCloner);
     }
 
     @Override
     public String toString() {
-        return problem.toString();
+        return problem != null ? problem.toString() : problemName;
```
ge0ffrey, line 82: This should probably just return problemName instead of problem.toString(), already before this change too. wdyt?

ge0ffrey:
Good to merge.

PR #824, opened 2020-07-02
PLANNER-1996: Enable dependency check

yurloc:
Use maven-dependency-plugin to fail the build if any used undeclared
or unused declared dependencies are found.

The plugin is configured to ignore non-compile dependencies. This avoids
having to ignore runtime and test-runtime dependencies individually.

Explanation:
- The plugin cannot detect if a declared runtime dependency is unused.
- The plugin cannot detect if an undeclared runtime dependency is used
  transitively. It is contributor's responsibility to declare runtime
  dependencies even though they are provided transitively.
- Missing runtime dependencies should be detected by tests. It's
  contributor's responsibility to write tests using runtime dependencies
  or to document why a runtine dependency is declared in case it
  cannot be covered by tests.
- The plugin will detect unused declared compile-scope dependencies.
- The plugin will detect used undeclared (transitive) compile-scope
  dependencies.

Exception: Spring Boot Starters are designed to provided transitive
dependencies. So the OptaPlanner Spring Boot Starter is allowed to have
unused declared dependencies and projects depending on Starters are not
checked at all. The same applies for Quarkus integration modules.

PR #835, opened 2020-07-14
Depend on Kogito Drools instead of Drools 7

ge0ffrey:
- [X] Depend optaplanner-core on kogito-drools

- [X] Remove junit 4 relics

- [x] adjust quarkus integration accordingly

### DIFF
--- a/optaplanner-benchmark/pom.xml
+++ b/optaplanner-benchmark/pom.xml
@@ -29,10 +29,6 @@
   <dependencies>
     <!-- Internal dependencies -->
     <dependency>
-      <groupId>org.kie</groupId>
-      <artifactId>kie-api</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-core</artifactId>
     </dependency>
@@ -59,17 +55,6 @@
     <dependency>
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-persistence-jaxb</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.drools</groupId>
-      <artifactId>drools-core</artifactId>
-      <scope>runtime</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.drools</groupId>
-      <artifactId>drools-compiler</artifactId>
-      <type>test-jar</type>
-      <scope>test</scope>
     </dependency>
 
     <!-- External dependencies -->
--- a/optaplanner-build-parent/pom.xml
+++ b/optaplanner-build-parent/pom.xml
@@ -112,6 +112,11 @@
       </dependency>
       <dependency>
         <groupId>org.kie.kogito</groupId>
+        <artifactId>drools-core-static</artifactId>
+        <version>${version.org.kie.kogito}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
         <artifactId>drools-decisiontables</artifactId>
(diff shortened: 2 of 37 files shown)


optaplanner-build-parent/pom.xml
```diff
@@ -19,8 +19,8 @@
     <!-- Dependencies -->
     <!-- ************************************************************************ -->
 
-    <!-- SNAPSHOT dependency versions -->
-    <version.org.kie>7.40.0-SNAPSHOT</version.org.kie>
+    <!-- SNAPSHOT and KIE dependency versions -->
+    <version.org.kie.kogito>0.11.1</version.org.kie.kogito>
```
triceo, line 23: 0.12 is out already.

optaplanner-persistence/optaplanner-persistence-jpa/src/test/java/org/optaplanner/persistence/jpa/impl/AbstractScoreJpaTest.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.optaplanner.persistence.jpa;
+package org.optaplanner.persistence.jpa.impl;
```
triceo, line 17: Although I don't have anything against this change, I don't see how it's related to the stated goal of this PR.

optaplanner-build-parent/pom.xml
triceo: Also, we should discuss how we become part of the Kogito train.
(Not for this PR though.)

optaplanner-benchmark/pom.xml
```diff
@@ -28,10 +28,6 @@
 
   <dependencies>
     <!-- Internal dependencies -->
-    <dependency>
```
rsynek, line 31: There will be a conflict with https://github.com/kiegroup/optaplanner/pull/824

optaplanner-persistence/optaplanner-persistence-jpa/src/test/java/org/optaplanner/persistence/jpa/impl/AbstractScoreJpaTest.java
rsynek: It comes from the same commit as the rest of the changes, so it will stay like that in the git history. Could we at least make it a separate commit?

optaplanner-build-parent/pom.xml
ge0ffrey: Upgraded to 0.12. Yes, we must become part of the kogito release train before September. Ideally, the fix their drools 7 dependency issue first.

optaplanner-persistence/optaplanner-persistence-jpa/src/test/java/org/optaplanner/persistence/jpa/impl/AbstractScoreJpaTest.java
ge0ffrey: Sorry about that. Next time I'll make it a separate commit, but sorting that out now for a git history that few people look at, isn't worth the hassle, I 'd argue.

optaplanner-benchmark/pom.xml
ge0ffrey: Yes, unfortunate. Feel free to add to tomorrow's meeting if there's way to avoid this.

triceo:
Jenkins did not pass, but that's expected these days.

PR #836, opened 2020-07-15
PLANNER-2034 Use jakarta EE API for JAXB

rsynek:
- switches to jakarta.xml.bind:jakarta.xml.bind-api
- ecxludes any dependency that brings duplicate classes (ban-duplicated-classes enforce check must pass)

### DIFF
--- a/optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
+++ b/optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
@@ -53,10 +53,15 @@
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-core</artifactId>
       <exclusions>
-        <!-- Clash with jakarta.activation:jakarta.activation-api -->
+        <!-- Duplicates classes of jakarta.activation:jakarta.activation-api -->
         <exclusion>
           <groupId>javax.activation</groupId>
           <artifactId>javax.activation-api</artifactId>
+        </exclusion>
+        <!-- Duplicates classes of jakarta.xml.bind:jakarta.xml.bind-api -->
+        <exclusion>
+          <groupId>javax.xml.bind</groupId>
+          <artifactId>jaxb-api</artifactId>
         </exclusion>
       </exclusions>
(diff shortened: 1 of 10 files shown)


rsynek:
> Although I would have thought that Quarkus have their dependencies in order, this exclusion business is a mess.

I agree. @ge0ffrey told me the Quarkus team has switching to Jakarta API artifacts on their radar, so hopefully, we can get rid of these exclusions soon.

rsynek:
Regarding Quarkus getting rid of the jboss spec api: https://github.com/quarkusio/quarkus/issues/1394.

rsynek:
Available PR checks passed, locally the enforcer plugin's checks passed too. Merging.

optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
```diff
@@ -53,11 +53,16 @@
...
+        <!-- Duplicates classes of jakarta.activation:jakarta.activation-api -->
         <exclusion>
           <groupId>javax.activation</groupId>
           <artifactId>javax.activation-api</artifactId>
         </exclusion>
+        <!-- Duplicates classes of jakarta.xml.bind:jakarta.xml.bind-api -->
+        <exclusion>
+          <groupId>javax.xml.bind</groupId>
+          <artifactId>jaxb-api</artifactId>
+        </exclusion>
```
ge0ffrey, line 65: Can we avoid duplicating the exclusions by doing them in the optaplanner-build-parent?

rsynek: Got the same question from Lukas. The answer is both yes and no:

The vast majority of the exclusions come from `io.quarkus` artifacts. If we define the exclusions in dependency management, we have to define a version of every such `io.quarkus` artifact. Effectively, we duplicate (and potentially) override the version coming from quarkus BOM.

For this reason, I would rather keep the exclusions at individual dependencies. Chances are there will be much less of them after #824 is merged.

PR #837, opened 2020-07-16
Move sonar project binding properties to root

rsynek:


rsynek:
SonarCloud is back -> merging

PR #839, opened 2020-07-16
PLANNER-1997 Spring Boot starter doesn't detect @Planning annotations when they are in abstract classes or interfaces

ge0ffrey:
Also adds test coverage for this bug, both for our Spring Boot Starter and our Quarkus Extension.

### DIFF
--- a/optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/chained/domain/TestdataChainedSpringSolution.java
+++ b/optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/chained/domain/TestdataChainedSpringSolution.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.spring.boot.autoconfigure.chained.domain;
+
+import java.util.List;
+
+import org.optaplanner.core.api.domain.solution.PlanningEntityCollectionProperty;
+import org.optaplanner.core.api.domain.solution.PlanningScore;
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.solution.ProblemFactCollectionProperty;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+
+@PlanningSolution
+public class TestdataChainedSpringSolution {
+
+    @ProblemFactCollectionProperty
+    @ValueRangeProvider(id = "chainedAnchorRange")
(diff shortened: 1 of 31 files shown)


rsynek:
Will take a look tomorrow morning, if that's ok.

triceo:
Personally, I don't like this.
This is a bug in Spring, not in our code. (As evidenced by the fact that we need to implant Spring code to fix it.)
The fix increases our maintenance burden, as opposed to making Spring accountable for their own.
If it were my decision, I'd detect this situation and fail fast with a clear explanation of where the bug is, and what the workaround is.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/chained/domain/TestdataChainedSpringSolution.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.optaplanner.core.api.domain.solution.PlanningEntityCollectionProperty;
+import org.optaplanner.core.api.domain.solution.PlanningScore;
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.solution.ProblemFactCollectionProperty;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+
+@PlanningSolution
+public class TestdataChainedSpringSolution {
```
rsynek, line 29: Naming: is it really a Spring-specific test domain or is it a domain with annotation on an interface?

ge0ffrey: It's a testdata (= test domain) specifically for Spring. We can't reuse the normal TestdataSolution etc because it needs to be located in the package that spring scans.

ge0ffrey:
> Personally, I don't like this.

Me neither, but it's the lesser evil.

> This is a bug in Spring, not in our code. (As evidenced by the fact that we need to implant Spring code to fix it.)

See Stéphane Nicole, of the Spring Boot team's answer here:
  https://stackoverflow.com/a/62936555/472109
We're doing exactly what he recommends. Wether it's a bug or an API gap is debatable. I think it's an API gap.

> The fix increases our maintenance burden, as opposed to making Spring accountable for their own.

Yes. If we truly believe that we should submit a PR on spring-boot to make the EntityScanner more flexible, so we don't duplicate the code once that's released. I'll take a look at that.

> If it were my decision, I'd detect this situation and fail fast with a clear explanation of where the bug is, and what the workaround is.

1) That would kill most VRP etc use cases out of the box without a solver config on Spring. This is a bug that was reported at least twice in a little time the spring boot starter is out. In both cases, it failed silently because the Standstill interface wasn't registered, so in both cases the real cause was hard to determine: shadow variables simply didn't behave correctly. This _burns_ new users. (a burned user poisons other users against our project)
2) Detecting it and failing fast would pretty much require the same code.

We need to merge this, but I 'll submit a PR upstream so at least we have a good chance of getting rid of this dirtyness over time.

optaplanner-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/chained/domain/TestdataChainedSpringSolution.java
ge0ffrey: What name do you think would be better?

rsynek: We test a specific situation - annotations on an interface. How about `TestdataChainedAnnotatedInterface*`?

ge0ffrey:
jenkins retest this

rsynek:
@ge0ffrey would you mind rebasing on the current master branch? It seems GH actions still fail on the dependency:analyze due to undeclared dependency on XStream, which has been already resolved on the master branch.

PR #840, opened 2020-07-17
PLANNER-2066 Remove any presence of XStream from SolverConfig and BenchmarkConfig

rsynek:
Let's pay XStream our last respects.

Note: XStream is still used by the optaplanner-examples and optaplanner-persistence-xstream. I will check if we still can get warnings about illegal reflection access from there.


rsynek:
SonarCloud:
- 1 code smell resolved, others waived
- the uncovered lines are mostly getters/setters

triceo:
@rsynek FYI I just realized that `investment` and `scrabble` examples can not be benchmarked anymore. They have no `SolutionIO` impl and the benchmarker doesn't like it to be null. Can you please fix that?

PR #841, opened 2020-07-17
Fix import statement

cuijulian:


triceo:
Already fixed by #824.
But thanks anyway, Julian!

PR #842, opened 2020-07-20
Fix issues found by the dependency analysis

rsynek:
Removes XStream from remaining places with one exception - PlannerTestUtils. That will be handled by a separate PR.


optaplanner-core/pom.xml
```diff
@@ -80,6 +80,10 @@
       <scope>test</scope>
     </dependency>
     <!-- XML -->
+    <dependency>
+      <groupId>com.thoughtworks.xstream</groupId>
+      <artifactId>xstream</artifactId>
+    </dependency>
```
yurloc, line 86: This should be either `<scope>test</scope>` or, even better, removed completely.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -112,14 +110,8 @@ public static SolverConfig createFromXmlResource(String solverConfigResource, Cl
...
-        } catch (ConversionException e) {
-            String lineNumber = e.get("line number");
-            throw new IllegalArgumentException("Unmarshalling of solverConfigResource (" + solverConfigResource
-                    + ") fails on line number (" + lineNumber + ")."
-                    + (Objects.equals(e.get("required-type"), "java.lang.Class")
-                            ? "\n  Maybe the classname on line number (" + lineNumber
-                                    + ") is surrounded by whitespace, which is invalid."
-                            : ""),
+        } catch (XmlUnmarshallingException e) {
+            throw new IllegalArgumentException("Unmarshalling of solverConfigResource (" + solverConfigResource + ") failed.",
```
ge0ffrey, line 114: I would replace paste tense with present tense: "failed" -> "fails", for consistency.

optaplanner-core/pom.xml
ge0ffrey, line 86: We still need it, despite the `ConversionException` import removal?

yurloc: Only for PlanningTestUtils.

ge0ffrey:
@rsynek Merged, regardless of comments because it fixes the build.

rsynek:
The remaining concerns are addressed by #846.

PR #845, opened 2020-07-21
PLANNER-2058 Init ScoreDirector to fix NPE in examples

triceo:


yurloc:
No unit test/reproducer?

triceo:
@yurloc Good point, added.

PR #848, opened 2020-07-21
PLANNER-2060 Fix Javadoc warnings

triceo:


optaplanner-build-parent/pom.xml
```diff
@@ -1014,6 +1017,18 @@
         <groupId>net.revelc.code</groupId>
         <artifactId>impsort-maven-plugin</artifactId>
       </plugin>
+      <plugin> <!-- Make sure the build fails-fast on Javadoc issues. -->
```
triceo, line 1020: Javadoc now runs in every module. Fails fast, as opposed to once every blue moon when `optaplanner-distribution` is being built.

optaplanner-build-parent/pom.xml
```diff
@@ -1230,9 +1245,6 @@
       <activation>
         <jdk>[11,)</jdk>
       </activation>
-      <properties>
-        <javadoc.additional.params>-Xdoclint:none</javadoc.additional.params>
-      </properties>
```
triceo, line 1235: Let's lint, I'd argue.

optaplanner-build-parent/pom.xml
```diff
@@ -759,8 +759,11 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
+          <version>3.2.0</version>
```
triceo, line 762: To get `<failOnWarnings>`.

yurloc:
On my machine, the build time of the whole repository increased from 7:36 to 8:13, that's roughly +35s. When you skip tests, the 35 seconds become +66% (0:59 -> 1:38).

Let's take a look at `optaplanner-core` in which we spend most of the time:

Selection_013

So `mvn clean install` takes almost 60s.

You can build `optaplanner-core` in 30s, including test execution if you skip revapi, javadoc, checkstyle and formatter.

I don't know how often people in our team build the project locally and whether the increasing build time hurts productivity.

We should discuss this and decide whether we want a slow but reliable build that executes everything by default or we want a quick build by default and leave some extra checks for the CI.

triceo:
Personally, I don't see this as a problem. Checkstyle will go, which means Javadoc only adds 6 seconds.
Compared to the time I will waste pushing (and waiting for) something that's going to fail later in CI, that's nothing.

optaplanner-build-parent/pom.xml
ge0ffrey: We inherit the maven-javadoc-plugin from jboss-parent, so we would set the version with a property `<version.javadoc.plugin>3.2.0</>` too. Not saying we should, but note that we do do that for surefire/failsafe sometimes... (=> consistency dilemma)

optaplanner-build-parent/pom.xml
ge0ffrey: (CI does do -Dfull always.)

Anyway, why not, let's see how this impacts speed build. I am happy to see spotbugs and checkstyle go for build speedup :)

PR #858, opened 2020-07-27
BAQE-1488 Update Revapi for 7.39.0.Final

triceo:
@MarianMacik I'm not sure how this happened, but your earlier Revapi PR appears to have broken the build.
Was the PR green when merging? If so, this is mysterious.

EDIT: Something is fishy here. GHA is using 7.33 as the baseline, even though it absolutely shouldn't.


triceo:
Problem has been identified. kie-parent SNAPSHOT has not been updated in Maven for a few days. Therefore this issue.

PR #859, opened 2020-07-27
PLANNER-1950 ConstraintProvider tests for School Timetabling quickstarts

triceo:
Identical domains for both quickstarts = identical constraint provider tests.
Essentially, this is one code twice. Perfect code duplication.

### DIFF
--- a/optaplanner-quickstarts/quarkus-school-timetabling/src/main/java/org/acme/schooltimetabling/domain/Timeslot.java
+++ b/optaplanner-quickstarts/quarkus-school-timetabling/src/main/java/org/acme/schooltimetabling/domain/Timeslot.java
@@ -54,6 +54,11 @@ public class Timeslot extends PanacheEntityBase {
         this.endTime = endTime;
     }
 
+    public Timeslot(long id, DayOfWeek dayOfWeek, LocalTime startTime) {
+        this(dayOfWeek, startTime, startTime.plusMinutes(50));
+        this.id = id;
+    }
+
     public Long getId() {
         return id;
     }
--- a/optaplanner-quickstarts/quarkus-school-timetabling/src/test/java/org/acme/schooltimetabling/solver/TimeTableConstraintProviderTest.java
+++ b/optaplanner-quickstarts/quarkus-school-timetabling/src/test/java/org/acme/schooltimetabling/solver/TimeTableConstraintProviderTest.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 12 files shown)


optaplanner-quickstarts/quarkus-school-timetabling/src/test/java/org/acme/schooltimetabling/solver/TimeTableConstraintProviderTest.java
```diff
@@ -0,0 +1,111 @@
...
+import java.time.LocalTime;
+
+import org.acme.schooltimetabling.domain.Lesson;
+import org.acme.schooltimetabling.domain.Room;
+import org.acme.schooltimetabling.domain.TimeTable;
+import org.acme.schooltimetabling.domain.Timeslot;
+import org.junit.jupiter.api.Test;
+import org.optaplanner.test.api.score.stream.ConstraintVerifier;
+
+class TimeTableConstraintProviderTest {
```
Christopher-Chianelli, line 29: Not public?

triceo: Since JUnit 5, it needn't be. And I would argue it shouldn't be, as it doesn't export nothing to any other class, nor does it need to be visible.

Most of Optaplanner code makes the test class public. However, the constraint provider in facility maintenance quickstart doesn't. So, faced with a choice, I picked the rule that was closer and made more sense.

optaplanner-quickstarts/quarkus-school-timetabling/src/main/java/org/acme/schooltimetabling/domain/Timeslot.java
```diff
@@ -54,6 +54,11 @@ public Timeslot(DayOfWeek dayOfWeek, LocalTime startTime, LocalTime endTime) {
         this.endTime = endTime;
     }
 
+    public Timeslot(long id, DayOfWeek dayOfWeek, LocalTime startTime) {
```
yurloc, line 57: In a real project, this constructor would deserve a documentation explaining how long a timeslot it creates if you do not provide an end time. Not sure if it's worth it for the quickstart. Your call.

Christopher-Chianelli: The end time doesn't really matter anyways; all that matter if time slots are exactly the same or not (the constraints do not check for overlap of time slots, but equality of time slots)

Christopher-Chianelli: (actually, it does matter for one constraint - teacherTimeEfficiency and studentGroupSubjectVariety, where we check for > 30 minutes gaps). There is an implict assumption that timeslots do not overlap

triceo: Considering that this would be the only Javadoc in the entire project, I decided against it. Merging.

PR #863, opened 2020-07-30
PLANNER-1935 Only allow to populate constraint stream from known types

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/InnerConstraintFactory.java
```diff
@@ -73,6 +78,34 @@
         }
     }
 
+    public <A> void assertValidFromClass(Class<A> fromClass) {
```
yurloc, line 81: It seems this could be `protected` to indicate it's only meant to be called by subclasses.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/quad/QuadConstraintStreamTest.java
```diff
@@ -153,6 +154,21 @@ public void filterConsecutive() {
     // If (not) exists
     // ************************************************************************
 
+    @TestTemplate
+    public void ifExists_unknownClass() {
+        assumeDrools();
+        Assertions.assertThatThrownBy(() -> buildScoreDirector((factory) -> {
```
yurloc, line 160: Static imports are preferred for assertions, mock methods, etc. in tests. I know this is not enforced and what's not enforced that's allowed but I think it's a good idea to follow some extra rules even they're not enforced and more importantly be consistent about them.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/InnerConstraintFactory.java
triceo: Unfortunately not, I also need to call this from the streams.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/InnerConstraintFactory.java
```diff
@@ -96,15 +98,19 @@
                             ProblemFactCollectionProperty.class, entry.getKey());
                 });
         Set<Class> allAcceptedClassSet = concat(concat(entityClassStream, factClassStream), factCollectionClassStream)
-                .collect(Collectors.toSet());
+                .collect(toSet());
```
yurloc, line 101: I like this. It improves readability. I used to be afraid of static imports in the past because they make it difficult to guess where the method comes from but in some cases, like this one, they're not confusing at all. In the context of a stream it's always clear what `toSet()` or `toList()` means. I'm going to apply this style in my code as well.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStreamTest.java
```diff
@@ -404,6 +404,16 @@ public void join_mixedEqualsAndFiltering() {
...
     // ************************************************************************
 
+    @TestTemplate
+    public void ifExists_unknownClass() {
+        assumeDrools();
+        assertThatThrownBy(() -> buildScoreDirector((factory) -> {
+            return factory.fromUniquePair(TestdataLavishEntity.class)
+                    .ifExists(Integer.class)
+                    .penalize(TEST_CONSTRAINT_NAME, SimpleScore.ONE);
+        })).isInstanceOf(IllegalArgumentException.class);
```
rsynek, line 414: If the code throws the `IllegalArgumentException` in a different situation, the test won't notice. The `IAE` is one of the most used exceptions in the optaplanner codebase.
Suggestion: add a check for a specific substring of the exception message.

PR #865, opened 2020-07-30
PLANNER-2078: Add Overlaps Joiners

Christopher-Chianelli:
Also refactored the Joiners test to remove static field imports to make it clear what type the constants refer to.

TODO: Other temporal/interval joiners (contains, consecutive)
TODO: TriJoiner, QuadJoiner, PentaJoiner versions


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -151,6 +152,23 @@
     // join(..., planningVariableContainsCached(Talk::getPeriod, (Period a, Period b) -> a.overlaps(b)))
     // get the period value range, does a cartesian product on it, so it maps every period to an overlapping periodList
     // then keep an index from every period to all talks in an overlapping period (possible the same period)
+    public static <A, B, Property_ extends Comparable<Property_>> BiJoiner<A, B> overlapsOrMeets(
+            Function<A, Property_> leftStartMapping,
+            Function<A, Property_> leftEndMapping,
+            Function<B, Property_> rightStartMapping,
+            Function<B, Property_> rightEndMapping) {
+        return AbstractBiJoiner.merge(lessThanOrEqual(leftStartMapping, rightEndMapping),
```
triceo, line 160: Considering that these are not simple Joiners anymore, I would:

- Extract their implementations into their own class.
- So that the tests for each also get their own class.
- And in that class, I'd make sure I cover all the boundary conditions in different test methods.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
```diff
@@ -300,9 +308,163 @@ public void greaterThanOrEqualPenta() {
...
+                return new Interval<A, B>(constructorMap.apply(start), constructorMap.apply(end), map);
+            }
+        }
+    }
+
+    private final Interval.IntervalFactory<BigInteger, Long, Long> bigIntInterval =
+            Interval.createFactory(BigInteger::valueOf, BigInteger::longValue);
+
+    private final Interval.IntervalFactory<BigDecimal, Long, Long> bigDecimalInterval =
+            Interval.createFactory(BigDecimal::valueOf, BigDecimal::longValue);
```
triceo, line 367: Personally, I think this a bit over-engineered. You don't need the factory, just use the `Interval` constructor directly.

These are tests - we're not looking for memory efficiency or performance, we're looking for maintainable code that is easy to understand.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
```diff
@@ -300,9 +308,163 @@ public void greaterThanOrEqualPenta() {
...
+        Function<Interval<BigDecimal, Long>, Long> rightStartMapping = Interval::getStart;
+        Function<Interval<BigDecimal, Long>, Long> rightEndMapping = Interval::getEnd;
+
+        AbstractBiJoiner<Interval<BigInteger, Long>, Interval<BigDecimal, Long>> joiner =
+                (AbstractBiJoiner<Interval<BigInteger, Long>, Interval<BigDecimal, Long>>) Joiners.overlaps(leftStartMapping,
+                        leftEndMapping, rightStartMapping, rightEndMapping);
+
+        assertSoftly(softly -> {
+            // Case A = B
+            softly.assertThat(joiner.matches(bigIntInterval.of(1L, 5L),
```
triceo, line 382: Now that the Joiners are actually relatively complex, I think that the comment before these assertions (`Case A = B`) should instead be turned into `.as(...)` on the assertion itself. The assertion messages will become so much better for it.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -151,6 +152,23 @@
     // join(..., planningVariableContainsCached(Talk::getPeriod, (Period a, Period b) -> a.overlaps(b)))
     // get the period value range, does a cartesian product on it, so it maps every period to an overlapping periodList
     // then keep an index from every period to all talks in an overlapping period (possible the same period)
```
triceo, line 154: Let's make sure we have proper Javadocs for these factory methods. This is public API after all.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -151,6 +152,23 @@
...
+            Function<B, Property_> rightEndMapping) {
+        return AbstractBiJoiner.merge(lessThanOrEqual(leftStartMapping, rightEndMapping),
+                greaterThanOrEqual(leftEndMapping, rightStartMapping));
+    }
+
+    public static <A, B, Property_ extends Comparable<Property_>> BiJoiner<A, B> overlaps(
+            Function<A, Property_> leftStartMapping,
+            Function<A, Property_> leftEndMapping,
+            Function<B, Property_> rightStartMapping,
+            Function<B, Property_> rightEndMapping) {
```
triceo, line 168: Did you consider any other way of specifying the intervals?
(I'm not saying this one is bad. Just making sure we went over all the options before we picked a public API to be with us forever.)

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
triceo: (To clarify: I prefer `BigDecimalInterval` and `BigIntegerInterval` to the factory that effectively does the same thing, only with a level of indirection.)

triceo:
Even thought I don't have anything against the naming which is pretty standard (`overlaps`, `meets` etc.), I think we should follow the standard operating principles for designing API. Put a draft on paper, call a meeting to discuss.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
triceo: (Which wouldn't be necessary if each of the cases had their own properly named test methods, as I suggest above. Either is fine with me.)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
Christopher-Chianelli: An other way would be `startA, startB, endA, endB`
When generalizing this to higher cardinality, it looks like this:
```
overlaps((a,b) -> a.start, c -> c.start, (a,b) -> a.end, c -> c.end)
```

Meanwhile, with the current definition, it be generalized like this:
```
overlaps((a,b) -> a.start, (a,b) -> a.end, c -> c.start, c -> c.end)
```

I see these as the two main contenders if we use a Function -> Comparable API for the following reasons:
1. The other joiners have the "main" stream as the first argument (this eliminates (start2,  ...) (end2, ...))
2. The main convention I see in other code is either (start1,end1,start2,end2)  (first interval followed by second interval) or  (start1,start2,end1,end2) (start points followed by end points)

The alternative way of defining it is to have a "Duration" (ex: 2 hour shift that start at 3am), but that requires an additional method to "add" a point and the point must still be comparable, so I don't see any advantage to that approach.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
Christopher-Chianelli: I personally think calling assert with `new Interval(...)` doesn't look pretty, which was the main reason why I created the factories.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
Christopher-Chianelli: But then how will people know these Joiners exists? People might already be used to all the Joiners being in the class `Joiners`. If we move it to a new class, we should at least mention it in the Javadocs of Joiners so they know there are other classes that contain Joiners (but not everyone read Javadocs; I suspect some people just use their IDE auto-complete which will hide the existence of these Joiners)

Christopher-Chianelli:
@triceo I agree; that why this is a draft, to generate discussion + get feedback before creating the higher level joiners and the other interval joiner types.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
triceo: Don'ŧ get me wrong. I think that the factory methods should stay in this class. But the implementations should be outside.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/JoinersTest.java
triceo: Therefore my second comment - if you create factory methods directly on `Interval`, the code will still be easier to understand. The current `IntervalFactory` approach creates indirection that is harder to see through.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
triceo: The idea of `(startA, durationA, startB, durationB)` is intriguing. Not as the only option, but as an alternative API alongside the `(start1,end1,start2,end2)`. The former can just delegate to the latter. Why?

In order to support a wider range of data models:
- People who have a data model with start date and job length, they use the former, without having to have complex lambdas like `a -> // add duration to the start date`.
- People who have a data model with a start/end dates, they use the former.

A win-win, I'd say.

Christopher-Chianelli: That requires the duration to be of type `Duration`. Some people might use int, doubles etc. for durations. So we need an 'add' method for this to be generic, or we need to create a Temporal, Duration special case.

triceo: I think it's a safe bet that, if they have `startDate`  as `OffsetDateTime`, they'd have a job length as `Duration`. Otherwise, if it's just an `int` for example, they're making life harder for themselves - and we'll support that anyway. So yes, I do think we should have a specialization for `Duration`. (I think we already do that in `ConstraintCollectors`.)

Christopher-Chianelli:
@ge0ffrey @triceo Added Javadoc and during joiner from spec. Move implementation to own file and addressed suggestions. I am debating about how to generalize the Bi test to the four other tests, since the checks are the same, but cannot be reused since Bi, Tri, etc. take a different number of arguments.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/IntervalJoiners.java
```diff
@@ -0,0 +1,314 @@
...
+     * @return An indexed joiner that filters the constraint stream to only include
+     *         elements (A,B,C,D,E) where (A,B,C,D)'s interval contains E's interval (as defined by the function mapping)
+     */
+    public static <A, B, C, D, E, Property_ extends Comparable<Property_>> PentaJoiner<A, B, C, D, E> during(
+            QuadFunction<A, B, C, D, Property_> leftStartMapping,
+            QuadFunction<A, B, C, D, Property_> leftEndMapping,
+            Function<E, Property_> rightStartMapping,
+            Function<E, Property_> rightEndMapping) {
+        return AbstractPentaJoiner.merge(Joiners.lessThanOrEqual(leftStartMapping, rightStartMapping),
+                Joiners.greaterThanOrEqual(leftEndMapping, rightEndMapping));
```
triceo, line 309: In fact, now that I see how little code there actually is in this class, makes me think that maybe `IntervalJoiners` class is unnecessary and the methods could just be inlined into `Joiners`?

When I was making this suggestion, I was somehow expecting that there would be more code here.

triceo (commented):
In general, the API looks good. I have left some comments.

When it comes to testing of joiners, I understand it is a lot of repetition - that said, I think we should take the hit. The existing tests for the `Joiners` class already do that, and I think we should stay consistent.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +148,103 @@
...
+     * @param <Property_> The type used to define the interval, comparable
+     * 
+     * @param leftStartMapping Maps the first argument to the first interval start point
+     * @param leftEndMapping Maps the first argument to the first interval end point
+     * @param rightStartMapping Maps the second argument to the second interval start point
+     * @param rightEndMapping Maps the second argument to the second interval end point
+     * 
+     * @return An indexed joiner that filters the constraint stream to only include
+     *         elements (A,B) where A's and B's intervals (as defined by the function mapping)
+     *         overlap.
```
triceo, line 171: We usually start with `never null, an indexed joiner...` And since it doesn't start with a capital letter, it doesn't end with a full stop.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/common/IntervalJoiners.java
Christopher-Chianelli: Probably from the 13 cases required to ensure full checking of joiner correctness (`4! = 24`, but pairs (a,b) where a > b are invalid, which cut it in half, and one additional case for equals). Simple code may not have so simple tests.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+     * words, there exists an x such that x in [a,b) and x in [c,d)).
+     * These are exactly the pairs where {@code a < d} and {@code b > c} (or in other words, when the first interval
+     * starts before the second one ends and the first interval ends after the second interval begins).
+     * 
+     * @param <A> the type of the first argument
+     * @param <B> the type of the second argument
+     * @param <Property_> the type used to define the interval, comparable
+     * 
+     * @param leftStartMapping maps the first argument to the first interval start point
+     * @param leftEndMapping maps the first argument to the first interval end point
```
triceo, line 165:
```suggestion
     * @param leftStartMapping maps the first argument to the first interval start point (inclusive)
     * @param leftEndMapping maps the first argument to the first interval end point (exclusive)
```

The method description marks the intervals `[a, b)`, which is the correct mathematical notation for an interval open on the right, but let's make it crystal clear.

triceo (commented):
I like it! Let's update the constraint providers and I'd say we're done here.

triceo (approved):
LGTM.

Side note: these examples show that we could use a Planner-specific type for "Event" that would have a start and an end. If we had that implemented in these examples, it would get so much simpler...

    overlaps(t -> t.getTimeslot().getStartDateTime(), t -> t.getTimeslot().getEndDateTime()))

Becomes:

    overlaps(t -> t.getTimeslot())

And similarly for the other examples.

triceo:
@Christopher-Chianelli FYI Geoffrey asked to be included in the review process.

Also, we had a discussion and please replace `overlaps` with `overlapping` - it works better, since the joiner names will generally be adjectives/adverbs.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
-    // then keep an index from every period to all talks in an overlapping period (possible the same period)
+    /**
+     * For pairs of intervals [a,b) and [c,d), returns only those pairs where [a,b) and [c,d) overlap (or, in other
+     * words, there exists an x such that x in [a,b) and x in [c,d)).
+     * These are exactly the pairs where {@code a < d} and {@code b > c} (or in other words, when the first interval
+     * starts before the second one ends and the first interval ends after the second interval begins).
+     * 
+     * @param <A> the type of the first argument
+     * @param <B> the type of the second argument
+     * @param <Property_> the type used to define the interval, comparable
```
ge0ffrey, line 162: style: type params are typically the last in the javadocs @ list

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+     * 
+     * @param leftStartMapping maps the first argument to the first interval start point (inclusive)
+     * @param leftEndMapping maps the first argument to the first interval end point (exclusive)
+     * @param rightStartMapping maps the second argument to the second interval start point (inclusive)
+     * @param rightEndMapping maps the second argument to the second interval end point (exclusive)
+     * 
+     * @return never null, an indexed joiner that filters the constraint stream to only include elements (A,B) where
+     *         A's and B's intervals (as defined by the function mapping) overlap
+     */
+    public static <A, B, Property_ extends Comparable<Property_>> BiJoiner<A, B> overlaps(
```
ge0ffrey, line 172: method name => overlapping (because it needs to be an adjective like other method names filtering, intersecting, equal, ...)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+     * For pairs of intervals [a,b) and [c,d), returns only those pairs where [a,b) and [c,d) overlap (or, in other
+     * words, there exists an x such that x in [a,b) and x in [c,d)).
+     * These are exactly the pairs where {@code a < d} and {@code b > c} (or in other words, when the first interval
+     * starts before the second one ends and the first interval ends after the second interval begins).
+     * 
+     * @param <A> the type of the first argument
+     * @param <B> the type of the second argument
+     * @param <Property_> the type used to define the interval, comparable
+     * 
+     * @param leftStartMapping maps the first argument to the first interval start point (inclusive)
```
ge0ffrey, line 164: "maps the first argument to the first interval start point " the double "first" could be confusing, suggestion:
=> maps the first argument to its interval start point"

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
 
-    // TODO
-    // join(..., planningVariableContainsCached(Talk::getPeriod, (Period a, Period b) -> a.overlaps(b)))
-    // get the period value range, does a cartesian product on it, so it maps every period to an overlapping periodList
-    // then keep an index from every period to all talks in an overlapping period (possible the same period)
+    /**
+     * For pairs of intervals [a,b) and [c,d), returns only those pairs where [a,b) and [c,d) overlap (or, in other
+     * words, there exists an x such that x in [a,b) and x in [c,d)).
+     * These are exactly the pairs where {@code a < d} and {@code b > c} (or in other words, when the first interval
+     * starts before the second one ends and the first interval ends after the second interval begins).
```
ge0ffrey, line 158: This second sentance seems to repeat the first sentence. I don't think it brings additional clarity, so I'd remove it.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
      * }
      */
 
-    // TODO
-    // join(..., planningVariableContainsCached(Talk::getPeriod, (Period a, Period b) -> a.overlaps(b)))
-    // get the period value range, does a cartesian product on it, so it maps every period to an overlapping periodList
-    // then keep an index from every period to all talks in an overlapping period (possible the same period)
+    /**
```
ge0ffrey, line 154: All joiners should probably have a first general line concept line, how about something like:
"Joins every A and B that overlap for an interval which is specified by a start and end property on both A and B."

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+            Function<B, Property_> rightEndMapping) {
+        return AbstractBiJoiner.merge(Joiners.lessThan(leftStartMapping, rightEndMapping),
+                Joiners.greaterThan(leftEndMapping, rightStartMapping));
+    }
+
+    /**
+     * Syntactic sugar for {@link #overlaps(Function, Function, Function, Function)} where
+     * both elements of the tuple (A,B) are the same type and have the same interval
+     * defining function. Identical to
+     * {@code overlaps(startMapping, endMapping, startMapping, endMapping)}.
```
ge0ffrey, line 185: We should probably use the same trick here as we do for all overloaded methods in the UniConstraintStream API javadocs etc: simply put `As defined by {@link #overlapping(...)}.` and then still specify the @param and @return etc.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+     * 
+     * @param <A> the type of both the first and second argument
+     * @param <Property_> the type used to define the interval, comparable
+     * 
+     * @param startMapping maps the argument to the start point of its interval (inclusive)
+     * @param endMapping maps the argument to the end point of its interval (exclusive)
+     * 
+     * @return see {@link #overlaps(Function, Function, Function, Function)}
+     */
+    public static <A, Property_ extends Comparable<Property_>> BiJoiner<A, A> overlaps(
```
ge0ffrey, line 195: Method ordering: the other methods in the class such as equal(A) are ordered before equal(A, B). Let's keep that consistent for easier maintenance?

ge0ffrey: Note that I expect that like equal, this method will be use far more than the (A, B) one, and therefor this deserves the javadoc and the other the "as defined by".

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+            Function<A, Property_> startMapping,
+            Function<A, Property_> endMapping) {
+        return overlaps(startMapping, endMapping, startMapping, endMapping);
+    }
+
+    /**
+     * For pairs of intervals [a,b) and [c,d), returns only those pairs where [a,b) contains [c,d) (or, in other
+     * words, for any x in [c, d), x is also in [a,b)).
```
ge0ffrey, line 203: Adding x complicates things. I am sure mathematicians love the "for any x" definition, but our audience, programmers, are better off with a simple statement a <= c and b <= d (or is it the other way around? see comment below).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+     * starts before the second one starts and the first interval ends after the second interval ends).
+     * 
+     * @param <A> the type of the first argument
+     * @param <B> the type of the second argument
+     * @param <Property_> the type used to define the interval, comparable
+     * 
+     * @param leftStartMapping maps the first argument to the first interval start point (inclusive)
+     * @param leftEndMapping maps the first argument to the first interval end point (exclusive)
+     * @param rightStartMapping maps the second argument to the second interval start point (inclusive)
+     * @param rightEndMapping maps the second argument to the second interval end point (exclusive)
```
ge0ffrey, line 214: Unlike in overlapping (and addition), for duration (like for substraction) it really important as to which is which.
Which of these is true?
* during(summer 2020, year 2020)
* during(year 2020, summer 2020)
The method name and the javadocs must make that crystal clear with no verbosity

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -147,10 +151,99 @@
...
+     * 
+     * @return never null, an indexed joiner that filters the constraint stream to only include elements (A,B) where
+     *         A's interval contains B's interval (as defined by the function mapping)
+     */
+    public static <A, B, Property_ extends Comparable<Property_>> BiJoiner<A, B> during(
```
ge0ffrey, line 219: -1 on "during" as method name because it presumes the notion of time.
This method applies not just to LocalDate etc, but also Integer, Long, BigDecimal and anything else that is Comparable.
Not just intervals, but any type of range.

ge0ffrey: I've opened a zulip chat on this, this is something we need to resolve before merging I think, as it gets written in stone in the API.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
```diff
@@ -0,0 +1,951 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.optaplanner.core.api.function.QuadFunction;
+import org.optaplanner.core.api.function.TriFunction;
+import org.optaplanner.core.impl.score.stream.bi.AbstractBiJoiner;
+import org.optaplanner.core.impl.score.stream.penta.AbstractPentaJoiner;
+import org.optaplanner.core.impl.score.stream.quad.AbstractQuadJoiner;
+import org.optaplanner.core.impl.score.stream.tri.AbstractTriJoiner;
+
+public class IntervalJoinersTest {
```
ge0ffrey, line 33: All unit tests for the class Joiners should be in JoinersTests normally, so it's easier to find them. It's harder to know that this class exists, because it's not classname +"Test"

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
```diff
@@ -0,0 +1,951 @@
...
+    private static final class Interval<A, B> {
```
ge0ffrey, line 34: Please create a Testdata class for this instead, in the testdata package.
@triceo Do we have any testdata classes already that are only for ConstraintStream testing and don't have any @Planning* annotations?

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
```diff
@@ -0,0 +1,951 @@
...
+                    .as("A meets B").isFalse();
+
+            softly.assertThat(joiner.matches(Interval.ofBigInt(3L, 7L),
+                    Interval.ofBigDecimal(1L, 3L)))
+                    .as("B meets A").isFalse();
+        });
+    }
+
+    @Test
+    public void overlapsBiShort() {
```
ge0ffrey, line 140: How does overlapsBiShort() improve the test coverage over overlapsBi?

optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/optional/score/FlightCrewSchedulingConstraintProvider.java
```diff
@@ -35,8 +35,9 @@ private Constraint requiredSkill(ConstraintFactory constraintFactory) {
     }
 
     private Constraint flightConflict(ConstraintFactory constraintFactory) {
-        return constraintFactory.fromUniquePair(FlightAssignment.class, Joiners.equal(FlightAssignment::getEmployee))
-                .filter((first, second) -> second.getFlight().overlaps(first.getFlight()))
+        return constraintFactory.fromUniquePair(FlightAssignment.class, Joiners.equal(FlightAssignment::getEmployee),
+                Joiners.overlaps(fa -> fa.getFlight().getDepartureUTCDateTime(),
+                        fa -> fa.getFlight().getArrivalUTCDateTime()))
```
ge0ffrey, line 40: Great to see this!

Small suggestion: revmove `Joiners.` for consistency because the other ConstraintProviders seem to do a static import of all Joiners methods.

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
```diff
@@ -86,8 +89,12 @@ protected Constraint requiredAttendanceConflict(ConstraintFactory constraintFact
...
-                                rightAssignment) -> rightAssignment.calculateOverlap(leftAssignment) > 0))
+                        overlaps((attendee1, attendee2, assignment) -> assignment.getStartingTimeGrain().getGrainIndex(),
+                                (attendee1, attendee2, assignment) -> assignment.getStartingTimeGrain().getGrainIndex() +
+                                        assignment.getMeeting().getDurationInGrains(),
+                                assignment -> assignment.getStartingTimeGrain().getGrainIndex(),
+                                assignment -> assignment.getStartingTimeGrain().getGrainIndex() +
+                                        assignment.getMeeting().getDurationInGrains()))
                 .penalizeConfigurable("Required attendance conflict",
                         (leftRequiredAttendance, rightRequiredAttendance, leftAssignment, rightAssignment) -> rightAssignment
                                 .calculateOverlap(leftAssignment));
```
ge0ffrey, line 100: (not a change request)
This brings up an interesting question: often the overlap is penalized. For example in conference scheduling, the overlap of 2 somehow conflicting talks is penalized.
We must have the overlap/duration API to allow for indexing. Should we also provide a helper method to calculate the overlap during penalization? Probably we can't: LocalDate and Integer don't have a common interface with a "subtract" method.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
triceo: No, and personally I don't see the need to do that. I'd rather keep single use classes limited in scope to where they're declared.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
Christopher-Chianelli: It for the short form of the `overlapping` method for the same type (i.e. `overlapping(Function, Function)` NOT `overlapping(Function, Function, Function, Function)`)

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
Christopher-Chianelli: Issue is the Interval Joiner tests are a lot more complicated than the normal joiner tests as there are 13 test cases we need to go through instead of the typical 2. So adding it to JoinersTests will make it massive: ~800 lines. As we add more and more Joiners, eventually JoinersTests will be a couple thousand lines and get harder and harder to maintain. Since 800 isn't too bad, I'll merge it, but we should separate the tests if we add more joiners (in particular, those with non-trivial testing)

optaplanner-examples/src/main/java/org/optaplanner/examples/meetingscheduling/optional/score/MeetingSchedulingConstraintProvider.java
Christopher-Chianelli: If there was such an interface, it would have to allow the following operations: `LocalDate - LocalDate -> Period`, `Integer - Integer -> Integer`, `LocalDateTime - LocalDateTime -> Duration`, etc. `overlapping` can be done for any `Comparable` class, `overlappingAmount` cannot; it requires an additional interface `difference(a,b)`. No such interface exists that all implement, although each class has their subtract method.

What we can do is help something like this:
`Pair<T,T> overlappingRange(interval1, interval2)`
which returns (max(A.start, B.start), min(A.end, B.end)) (which is the "area" of overlap). Then they can do the subtract operation on those two points (or can provide a function that that turns it into a number) See https://stackoverflow.com/a/36035369

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/IntervalJoinersTest.java
Christopher-Chianelli: Changed to `overlapsDifferentTypes` and `overlapsSameTypes`

PR #869, opened 2020-08-05
BXMSPROD-912 openjdk11 build execution added to pull_request.yml

Ginxo:
https://issues.redhat.com/browse/BXMSPROD-912

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -14,7 +14,7 @@ jobs:
       with:
         parent-dependencies: 'drools@7.x:master'
         child-dependencies: 'jbpm@7.x:master'
-        build-command: 'mvn -e clean install -nsu -Prun-code-coverage,wildfly -Dfull  -Dintegration-tests=true -Dmaven.test.failure.ignore=true -DjvmArgs="-Xms1g -Xmx4g" | mvn -nsu generate-resources -Psonarcloud-analysis -DjvmArgs="-Xms1g -Xmx4g"'
+        build-command: 'mvn -e clean install -nsu -Prun-code-coverage,wildfly -Dfull  -Dintegration-tests=true -Dmaven.test.failure.ignore=true -DjvmArgs="-Xms1g -Xmx4g"'
         build-command-upstream: "mvn -e clean install -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true"
         workflow-file-name: "pull_request.yml"
       env:
@@ -30,7 +30,7 @@ jobs:
       with:
         parent-dependencies: 'drools@7.x:master'
         child-dependencies: 'jbpm@7.x:master'
-        build-command: 'mvn -e clean install -nsu -Prun-code-coverage,wildfly -Dfull  -Dintegration-tests=true -Dmaven.test.failure.ignore=true -DjvmArgs="-Xms1g -Xmx4g" | mvn -nsu generate-resources -Psonarcloud-analysis -DjvmArgs="-Xms1g -Xmx4g"'
+        build-command: 'mvn -e clean install -nsu -Prun-code-coverage,wildfly -Dfull  -Dintegration-tests=true -Dmaven.test.failure.ignore=true -DjvmArgs="-Xms1g -Xmx4g"'
         build-command-upstream: "mvn -e clean install -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true"
         workflow-file-name: "pull_request.yml"
       env:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -16,6 +16,7 @@ jobs:
         child-dependencies: 'jbpm@7.x:master'
         build-command: 'mvn -e clean install -nsu -Prun-code-coverage,wildfly -Dfull  -Dintegration-tests=true -Dmaven.test.failure.ignore=true -DjvmArgs="-Xms1g -Xmx4g" | mvn -nsu generate-resources -Psonarcloud-analysis -DjvmArgs="-Xms1g -Xmx4g"'
         build-command-upstream: "mvn -e clean install -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true"
+        workflow-file-name: "pull_request.yml"
       env:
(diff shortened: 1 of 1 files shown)


mareknovotny:
this needs to be probably discussed with @ge0ffrey @rsynek @triceo to replace their already existing Github Actions for https://github.com/kiegroup/optaplanner/blob/7.x/.github/workflows/os_java_ci.yml which certainly doesn't have the upstream/downstream chain implemented ;)

.github/workflows/pull_request.yml
```diff
@@ -1,18 +1,39 @@
...
+  build-chain-openjdk11:
     runs-on: ubuntu-latest
-    name: Build Pull Request
+    name: Pull Request openjdk11
     steps:
     - uses: actions/checkout@v2
     - name: Build Chain
       id: build-chain
-      uses: kiegroup/github-action-build-chain@master
+      uses: kiegroup/github-action-build-chain@openjdk11
```
triceo, line 29: It is unfortunate that you need to specify JDK like this. If you could use the standard way (see https://github.com/kiegroup/optaplanner/blob/master/.github/workflows/os_java_ci.yml#L29), you'd be able to use a standard matrix build, avoiding this large config duplication.

.github/workflows/pull_request.yml
```diff
@@ -1,18 +1,39 @@
...
+    runs-on: ubuntu-latest
+    name: Pull Request openjdk8
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Chain
+      id: build-chain
+      uses: kiegroup/github-action-build-chain@openjdk8
+      with:
+        parent-dependencies: 'drools@7.x:master'
+        child-dependencies: 'jbpm@7.x:master'
```
triceo, line 16: Is there any doc I could read on how this particular config works? I'm confused by both "7.x" and "master" being there.

mareknovotny: look into the GA extension https://github.com/kiegroup/github-action-build-chain I think @Ginxo will add documentation as it goes. Right now fast speed implementation :)

.github/workflows/pull_request.yml
mareknovotny: I think @Ginxo wrote about some limitation on GA side there is some bug which blocks this easier way of config

from @Ginxo`s msg:
> just to let you know I tried to use matrix in order not to duplicate commands for openjdk8 and 11, but it's not possible to use it in usues like uses: kiegroup/github-action-build-chain@{{ matrix.build-chain-version}}
the limitation is already reported and there's no way of work around it (apart from duplicate it again)
The other thing you should pay attention to is workflow-file-name since the name of the flow is defined there's no way about how to get file-name information (I already reported it to github) (edited)

.github/workflows/pull_request.yml
rsynek: If I understand it correctly, according to [1], it says our `7.x` depends on the other repo's `master` branch.

[1] https://github.com/kiegroup/github-action-build-chain/blob/master/src/lib/common.js#L65

.github/workflows/pull_request.yml
```diff
@@ -1,18 +1,39 @@
...
+        build-command: 'mvn -e clean install -nsu -Prun-code-coverage,wildfly -Dfull  -Dintegration-tests=true -Dmaven.test.failure.ignore=true -DjvmArgs="-Xms1g -Xmx4g" | mvn -nsu generate-resources -Psonarcloud-analysis -DjvmArgs="-Xms1g -Xmx4g"'
```
rsynek, line 17: The profiles `run-code-coverage` and `sonarcloud-analysis` should be never in the same maven execution:
`mvn -e clean install -nsu -Prun-code-coverage` runs tests with measuring the coverage
`mvn validate -Psonarcloud-analysis` merges the coverage information into a single XML coverage file and triggers the sonarcloud

rsynek: Also, as I see the same command in the `build-chain-openjdk11`, SonarCloud should be triggered only once per any change on the PR. There is no point in running it inside multiple checks.

.github/workflows/pull_request.yml
Ginxo: @triceo that's it, the thing is build-chain tool runs over a docker image (java8+maven3.6.6, java11+maven3.6.3, gradle, whatever) but there's not way to dynamically select the build-chain version you want to use. So for now the only way is to duplicate it.

.github/workflows/pull_request.yml
Ginxo: yes, the documentation still missing https://issues.redhat.com/browse/BXMSPROD-913 sorry for that
but basically whatever comes after `@` means `branch mapping`, so it will map the current `7.x` branch to jbpm/drools `master`
The same here but opposite way https://github.com/kiegroup/jbpm/blob/5e3cb75e2111668d4e093605f5c82df2b9c53183/.github/workflows/pull_request.yml#L15

.github/workflows/pull_request.yml
rsynek: Question: do the GitHub Actions inject SonarCloud auth token into the build? Last time I experimented with GA and SonarCloud, it didn't work because GA deliberately avoid injecting secrets into builds of PRs opened from a forked repo. The motivation was to protect the secret from being read by a malicious PR.

.github/workflows/pull_request.yml
Ginxo: https://github.community/t/using-strategy-matrix-values-in-the-specification-of-actions-to-be-used-in-a-workflow/16706

.github/workflows/pull_request.yml
Ginxo: thanks @rsynek `build-command` specifies multiple commands (maybe `build-commands` is a better name :thinking: ) in this case there are two commands split by `|` so two different maven executions. I already removed sonar execution from the rest of the projects but due to sonar execution is in `build-command` (only executed when the PR is for the current project) and the optaplanner event is `workflow_dispatch` this will never be triggered by a PR.
About sonarcloud credentials I already open a JIRA for it https://issues.redhat.com/browse/BXMSPROD-895
I'm going to remove sonarcloud execution anyway until BXMSPROD-895 is fixed.

Ginxo: just one thing `with` variables don't allow arrays, that's why it's defined as a string split by `|`

.github/workflows/pull_request.yml
mareknovotny: > If I understand it correctly, according to [1], it says our `7.x` depends on the other repo's `master` branch.
> 
> [1] https://github.com/kiegroup/github-action-build-chain/blob/master/src/lib/common.js#L65

this is what Optaplanner with having 8.x version on its master made up ;) @rsynek

.github/workflows/pull_request.yml
rsynek: Thanks, I missed the `|` in the command string.

Ginxo: sonar analysis command removed

PR #872, opened 2020-08-07
Planner-1823 Add ConstraintProvider for Patient Admission Scheduling example

dupliaka:
Created constraints based on drl rules
http://github.com/dupliaka/optaplanner/blob/8a77adbe40cd58339450dc3ec51b8258fbb4dd2e/optaplanner-examples/src/main/resources/org/optaplanner/examples/pas/solver/patientAdmissionScheduleConstraints.drl#L1-L1

Previously all of the constraints were defined via a set of rules declared in drl file, but now we have  [constraint streams](https://docs.optaplanner.org/latest/optaplanner-docs/html_single/index.html#constraintStreams).

I repeated the drl constraints logic and created some tests + fixes abug in drl rule

The [example doc](https://people.cs.kuleuven.be/~wim.vancroonenburg/pas/)

[PLANNER-1823](https://issues.redhat.com/browse/PLANNER-1823)

Note:
Will change patientAdmissionScheduleSolverConfig.xml  after approving to turn off constraints stream by default


optaplanner-examples/src/test/java/org/optaplanner/examples/pas/solver/BedDesignationDO.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.optaplanner.examples.pas.domain.AdmissionPart;
+import org.optaplanner.examples.pas.domain.Bed;
+import org.optaplanner.examples.pas.domain.BedDesignation;
+import org.optaplanner.examples.pas.domain.Night;
+import org.optaplanner.examples.pas.domain.Patient;
+import org.optaplanner.examples.pas.domain.Room;
+import org.optaplanner.examples.pas.domain.Specialism;
+
+public class BedDesignationDO extends BedDesignation {
```
triceo, line 11: Personally, I see this class as unnecessary complexity - I'd simply change `BedDesignation` to add the constructors etc.
The other examples do it the same way, so I'd appreciate if we stayed consistent.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,180 @@
...
+                roomSpecialismNotExistsConstraint(constraintFactory),
+                roomSpecialismNotFirstPriorityConstraint(constraintFactory),
+                preferredPatientEquipmentConstraint(constraintFactory)
+        };
+    }
+
+    public Constraint sameBedInSameNightConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUniquePair(BedDesignation.class)
+                .filter((left, right) -> left.getBed() != null
+                        && right.getBed() != null
```
triceo, line 53: `null` checks for planning variables are not necessary in Constraint Streams. `fromUniquePair` will make sure `bed` is never `null`.
Please check for this in the other constraints as well.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,180 @@
...
+                        && left.getBed() == right.getBed()
```
triceo, line 54: Please use `Joiners.equal(BedDesignation::getBed)` for this.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,180 @@
...
+    public Constraint differentGenderInSameGenderRoomInSameNightConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(BedDesignation.class)
+                .filter(bd -> bd.getRoomGenderLimitation() == GenderLimitation.SAME_GENDER
+                        && bd.getBed() != null)
+                .join(BedDesignation.class,
+                        equal(BedDesignation::getRoom),
+                        lessThan(BedDesignation::getId),
+                        filtering((left, right) -> right.getRoomGenderLimitation() == GenderLimitation.SAME_GENDER
+                                && left.getAdmissionPart().calculateSameNightCount(right.getAdmissionPart()) > 0
+                                && left.getPatient().getGender() != right.getPatient().getGender()))
```
triceo, line 83: Personally, I'd switch the order of operations to first do the `==`, then `!=` and only then do the `calculateSameNightCount(...)`. This way, you make sure that the (possibly expensive) calculation is only done when the quick checks are already true, never wasting time.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,180 @@
...
+                .penalize("differentGenderInSameGenderRoomInSameNight", HardMediumSoftScore.ofHard(1000),
+                        (left, right) -> left.getAdmissionPart().calculateSameNightCount(right.getAdmissionPart()));
```
triceo, line 85: Considering how often `calculateSameNightCount()` is used, I'd consider caching the calculation somehow. But maybe it's not a problem, doing a quick profiler run would tell you.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public Constraint departmentMinimumAgeConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Department.class)
+                .filter(d -> d.getMinimumAge() != null)
+                .join(BedDesignation.class,
+                        equal(d -> d, BedDesignation::getDepartment),
+                        filtering((d, bd) -> bd.getPatientAge() < d.getMinimumAge()))
```
triceo, line 93: Note: we have `Joiners.lessThan(...)`.

triceo:
Also, please make sure the files are properly formatted - CI is currently failing on account of that.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,180 @@
...
+                                && left.getPatient().getGender() != right.getPatient().getGender()))
+                .penalize("differentGenderInSameGenderRoomInSameNight", HardMediumSoftScore.ofHard(1000),
+                        (left, right) -> left.getAdmissionPart().calculateSameNightCount(right.getAdmissionPart()));
+    }
+
+    public Constraint departmentMinimumAgeConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Department.class)
+                .filter(d -> d.getMinimumAge() != null)
+                .join(BedDesignation.class,
+                        equal(d -> d, BedDesignation::getDepartment),
```
triceo, line 92: `d -> d` should be the same as `Function.identity()`.

optaplanner-examples/src/main/resources/org/optaplanner/examples/pas/solver/patientAdmissionScheduleSolverConfig.xml
```diff
@@ -1,13 +1,19 @@
...
   <!--<moveThreadCount>AUTO</moveThreadCount>--><!-- To solve faster by saturating multiple CPU cores -->
 
   <solutionClass>org.optaplanner.examples.pas.domain.PatientAdmissionSchedule</solutionClass>
   <entityClass>org.optaplanner.examples.pas.domain.BedDesignation</entityClass>
 
   <scoreDirectorFactory>
-    <scoreDrl>org/optaplanner/examples/pas/solver/patientAdmissionScheduleConstraints.drl</scoreDrl>
+<!--    <scoreDrl>org/optaplanner/examples/pas/solver/patientAdmissionScheduleConstraints.drl</scoreDrl>-->
+<!--    Uncomment if you want to use constraints stream-->
+<!--    <initializingScoreTrend>ONLY_DOWN</initializingScoreTrend>-->
```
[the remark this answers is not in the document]
dupliaka: Sorry, maybe  I should comment it in the code next time, not in the PR description. I left this like this so you can pull and test it if you want to check something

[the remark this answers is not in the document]
triceo: I mean specificall the `ONLY_DOWN` part - that should not be commented out, as it's a property of the construction heuristic. It has nothing to do with the scoring function.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
dupliaka: Fix 94c60080

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
dupliaka: BTW is that worth to mention that `fromUniquePair` is a syntax sugar as here: http://github.com/dupliaka/optaplanner/blob/PLANNER-1823/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintFactory.java#L95-L95
Since I was beliving that sugared syntax should behave absolutely the same under the hood.

dupliaka: I mean since fromUniquePair implemented in the way that it is much slower

triceo: I believe that the Javadoc is correct - since `join`, indeed, also filters out uninitialized entities.

dupliaka: +I have a concern about that: if I will use regular join, then since bed is nullable I need to check both (left and right) streams for bed!=null, but I will have a problem there bcs I need to do that before joining and I can not do that

dupliaka:
```    public Constraint sameBedInSameNightConstraint(ConstraintFactory constraintFactory) {
        return constraintFactory.from(BedDesignation.class)
                .filter((left)->left.getBed()!=null) // cannot filter right
                .join(BedDesignation.class,
                        equal(BedDesignation::getBed),
                        lessThan(BedDesignation::getId))
                .penalize("sameBedInSameNight", HardMediumSoftScore.ofHard(1000)));
    }```

Comment (triceo) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java:54]: That's the point though, you don't need to do that. The `bed != null` check is already done for you. This will be enough:

     return constraintFactory.from(BedDesignation.class)
                .join(BedDesignation.class,
                        equal(BedDesignation::getBed),
                        lessThan(BedDesignation::getId))
                .penalize("sameBedInSameNight", HardMediumSoftScore.ofHard(1000)));

In fact, now that this constraint is so simple, you probably just need:

     return constraintFactory.fromUniquePair(BedDesignation.class,
                equal(BedDesignation::getBed))
                .penalize("sameBedInSameNight", HardMediumSoftScore.ofHard(1000)));

Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java:54]: Brilliant idea

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java (hunk 2 above)
Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java:53]: Ignoring since bed is nullable

File: optaplanner-examples/src/test/java/org/optaplanner/examples/pas/solver/BedDesignationDO.java (hunk 1 above)
Comment (dupliaka) [reply optaplanner-examples/src/test/java/org/optaplanner/examples/pas/solver/BedDesignationDO.java:11]: Ok, just a proposal. Will add with methods

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java (hunk 7 above)
Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java]: Fix https://github.com/kiegroup/optaplanner/pull/872/commits/94c60080ed05dd144db456a18e268b029fdd3328

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java (hunk 6 above)
Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java]: https://github.com/kiegroup/optaplanner/pull/872/commits/94c60080ed05dd144db456a18e268b029fdd3328

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java (hunk 4 above)
Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java]: https://github.com/kiegroup/optaplanner/pull/872/commits/7dd155181c6f1f7f867667d82eb53de3aa1d5a38

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java (hunk 3 above)
Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java]: https://github.com/kiegroup/optaplanner/pull/872/commits/7dd155181c6f1f7f867667d82eb53de3aa1d5a38

File: optaplanner-examples/src/test/java/org/optaplanner/examples/pas/solver/BedDesignationDO.java (hunk 1 above)
Comment (dupliaka) [reply optaplanner-examples/src/test/java/org/optaplanner/examples/pas/solver/BedDesignationDO.java]: https://github.com/kiegroup/optaplanner/pull/872/commits/7dd155181c6f1f7f867667d82eb53de3aa1d5a38

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java (hunk 5 above)
Comment (dupliaka) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java:87]: Is that still relevant?

Comment (triceo) [reply optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java:87]: No.

File: optaplanner-examples/src/main/java/org/optaplanner/examples/pas/domain/BedDesignation.java [hunk 9]
```diff
@@ -118,4 +118,49 @@ public String toString() {
...
+
+    public BedDesignation withRoom(Room room) {
+        this.bed.setRoom(room);
+        return this;
+    }
+
+    public BedDesignation withSpecialism(Specialism spec1) {
+        this.admissionPart.setSpecialism(spec1);
+        return this;
+    }
```
triceo, line 164: In my original comment, I meant that these methods would simply become setters. That is `public void set...(...)`.
This way, it becomes consistent with the other examples. (Although unfortunately the test code gets a bit uglier.)

Sorry for not being clear earlier.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,182 @@
...
+                roomSpecialismNotFirstPriorityConstraint(constraintFactory),
+                preferredPatientEquipmentConstraint(constraintFactory)
+        };
+    }
+
+    public Constraint sameBedInSameNightConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUniquePair(BedDesignation.class,
+                        equal(BedDesignation::getBed))
+                .filter((left,right) -> left.getAdmissionPart().calculateSameNightCount(right.getAdmissionPart()) > 0
+                && left.getBed()!=null && right.getBed()!=null)
```
triceo, line 57: Please put the `null` checks first in the filter.
Also, since `BedDesignation::getBed` is `equal`, you only need to check if `left.getBed() != null`.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
```diff
@@ -0,0 +1,182 @@
...
+                .ifNotExists(RoomEquipment.class,
+                        equal((rpe, bd) -> bd.getRoom(), RoomEquipment::getRoom),
+                        equal((rpe, bd) -> rpe.getEquipment(), RoomEquipment::getEquipment))
+                .penalize("requiredPatientEquipment", HardMediumSoftScore.ofHard(50),
+                        (rpe, bd) -> bd.getAdmissionPartNightCount());
+    }
+
+    //Medium
+    public Constraint assignEveryPatientToABedConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(BedDesignation.class)
```
triceo, line 124: Since `bed` is `nullable`, `fromUnfiltered()` is no different than `from()`. You might as well use `from()`.

triceo:
Also, it appears there are still formatting issues.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
dupliaka: Yes, leftover.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
dupliaka: Yeh, the last one I got a call from somewhere inside, thanks for correcting.

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/solver/score/PatientAdmissionMoveConstraintProvider.java
dupliaka: https://github.com/kiegroup/optaplanner/pull/872/commits/1f9f2b888293c0e1f544dfc79931c17386a57b41

optaplanner-examples/src/main/java/org/optaplanner/examples/pas/domain/BedDesignation.java
dupliaka: https://github.com/kiegroup/optaplanner/pull/872/commits/1f9f2b888293c0e1f544dfc79931c17386a57b41

PR #873, opened 2020-08-10
Upgrade Quarkus to 1.7.0.Final

rsynek:
Do not merge before io.quarkus:quarkus-bom-deployment:1.7.0.Final is available.

Locally optaplanner-quarkus-integration tests passed with 1.7.0.Final core artifacts (and quarkus-bom-deployment:1.6.0.Final).

### DIFF
--- a/optaplanner-quarkus-integration/pom.xml
+++ b/optaplanner-quarkus-integration/pom.xml
@@ -24,8 +24,6 @@
   <url>https://www.optaplanner.org</url>
 
   <properties>
-    <version.io.quarkus>1.7.0.Final</version.io.quarkus>
-
     <!-- TODO Quarkus is riddled with duplicate classes, see https://github.com/quarkusio/quarkus/issues/9834 -->
     <enforcer.failOnDuplicatedClasses>false</enforcer.failOnDuplicatedClasses>
   </properties>
--- a/optaplanner-quarkus-integration/pom.xml
+++ b/optaplanner-quarkus-integration/pom.xml
@@ -24,7 +24,7 @@
   <url>https://www.optaplanner.org</url>
 
   <properties>
-    <version.io.quarkus>1.6.0.Final</version.io.quarkus>
+    <version.io.quarkus>1.7.0.Final</version.io.quarkus>
 
     <!-- TODO remove if https://issues.redhat.com/browse/PLANNER-2005 is fixed for 8.x -->
     <!-- TODO Extending kie-parent 7 is extremely dangerous and messy because it's not aligned to quarkus -->
--- a/optaplanner-quarkus-integration/pom.xml
+++ b/optaplanner-quarkus-integration/pom.xml
@@ -43,13 +43,6 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-      <dependency>
-        <groupId>io.quarkus</groupId>
-        <artifactId>quarkus-bom-deployment</artifactId>
-        <version>${version.io.quarkus}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
     </dependencies>
   </dependencyManagement>
   <build>
--- a/optaplanner-quarkus-integration/pom.xml
+++ b/optaplanner-quarkus-integration/pom.xml
@@ -26,41 +26,8 @@
   <properties>
     <version.io.quarkus>1.7.0.Final</version.io.quarkus>
 
-    <!-- TODO remove if https://issues.redhat.com/browse/PLANNER-2005 is fixed for 8.x -->
-    <!-- TODO Extending kie-parent 7 is extremely dangerous and messy because it's not aligned to quarkus -->
-    <version.asm>7.3.1</version.asm>
-    <version.commons-cli>1.4</version.commons-cli>
-    <version.commons-codec>1.13</version.commons-codec>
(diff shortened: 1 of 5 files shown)


optaplanner-quarkus-integration/pom.xml
```diff
@@ -24,43 +24,10 @@
...
-    <version.org.sonatype.plexus.plexus-sec-dispatcher>1.4</version.org.sonatype.plexus.plexus-sec-dispatcher>
-    <version.org.wildfly.common>1.5.4.Final-format-001</version.org.wildfly.common>
-    <version.org.ow2.asm>8.0.1</version.org.ow2.asm>
-    <version.org.jboss.jandex>2.1.3.Final</version.org.jboss.jandex>
-
-
-    <version.org.jboss.resteasy>4.5.3.Final</version.org.jboss.resteasy>
-    <version.com.fasterxml.jackson>2.10.3</version.com.fasterxml.jackson>
-    <version.com.fasterxml.jackson.databind>2.10.3</version.com.fasterxml.jackson.databind>
+    <version.io.quarkus>1.7.0.Final</version.io.quarkus>
```
ge0ffrey, line 27: The quarkus version should be defined in optaplanner-build-parent I think, but we can't import the quarkus-bom there, not even in depMgmt, because it's a platform bom (not a user bom), so it contains groupId's from other artifacts, messing up our dep tree.

optaplanner-quarkus-integration/pom.xml
```diff
@@ -24,43 +24,10 @@
...
-    <version.org.junit>5.6.2</version.org.junit>
-    <version.org.sonatype.plexus.plexus-sec-dispatcher>1.4</version.org.sonatype.plexus.plexus-sec-dispatcher>
-    <version.org.wildfly.common>1.5.4.Final-format-001</version.org.wildfly.common>
-    <version.org.ow2.asm>8.0.1</version.org.ow2.asm>
-    <version.org.jboss.jandex>2.1.3.Final</version.org.jboss.jandex>
-
-
-    <version.org.jboss.resteasy>4.5.3.Final</version.org.jboss.resteasy>
-    <version.com.fasterxml.jackson>2.10.3</version.com.fasterxml.jackson>
-    <version.com.fasterxml.jackson.databind>2.10.3</version.com.fasterxml.jackson.databind>
```
ge0ffrey, line 56: Great to see all these removed. Are we sure for each of these that our dependency tree if it contains it somewhere (for example optaplanner-jackson also depends on jackson) that it uses a version that is the same or higher (but not lower) than the versions we're removing here?

rsynek: Checked with the dependency:tree; also tests passed (locally).

rsynek:
@ge0ffrey good point, let me create a ticket for that:
https://issues.redhat.com/browse/PLANNER-2112

optaplanner-quickstarts/quarkus-school-timetabling/build.gradle
```diff
@@ -14,9 +14,10 @@ repositories {
 }
 
 dependencies {
-    implementation enforcedPlatform("io.quarkus:quarkus-universe-bom:1.4.2.Final")
-    implementation 'io.quarkus:quarkus-optaplanner'
-    implementation 'io.quarkus:quarkus-optaplanner-jackson'
+    implementation platform("io.quarkus:quarkus-universe-bom:1.7.0.Final")
```
ge0ffrey, line 17: What's the difference between "enforcedPlatform" and "platform"?

optaplanner-quickstarts/quarkus-school-timetabling/build.gradle
```diff
@@ -14,9 +14,10 @@ repositories {
...
+    implementation "org.optaplanner:optaplanner-quarkus:${optaplannerVersion}"
+    implementation "org.optaplanner:optaplanner-quarkus-jackson:${optaplannerVersion}"
+    implementation "org.optaplanner:optaplanner-test:${optaplannerVersion}"
```
ge0ffrey, line 20: The gradle build was broken due to this missing test dependency.

Yes, we need CI to somehow run the gradle build for the quickstarts too, maybe from a maven profile in the directory above? Separate issue/PR in any case.

ge0ffrey: @rsynek This separate issue/PR, is this something you could take a look at some time? (no rush)

optaplanner-quickstarts/quarkus-school-timetabling/build.gradle
rsynek: "platform" works pretty much the same way as importing a BOM in maven. "enforcedPlatform" overrides whatever is declared before or after that => the BOM always wins.

optaplanner-quickstarts/quarkus-school-timetabling/build.gradle
rsynek: Quarkus does it this way for their gradle support (the quarkus gradle plugin):
https://github.com/quarkusio/quarkus/blob/master/devtools/gradle/pom.xml#L75

For quickstarts, they use only maven.

rsynek:
CI passed, locally tests passed with native image compilation (both the quickstarts and the quarkus-integration). Merging.

PR #874, opened 2020-08-10
PLANNER-1966 Improve Constraint Verifier assertion messages

triceo:


optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultSingleConstraintAssertion.java
```diff
@@ -16,16 +16,21 @@
 
 package org.optaplanner.test.impl.score.stream;
 
+import static java.util.Objects.*;
```
rsynek, line 19:
```suggestion
import static java.util.Objects.requireNonNull;
```

PR #875, opened 2020-08-11
PLANNER-1433 Handle logical OR in DRL correctly

triceo:
This may have some small impact on constraint matching performance, but that should not matter as the previous behavior was incorrect.
The performance impact could be eliminated entirely, if we can live with the fact that the duplicate constraint matches will be removed on the first instance, and not on the last. This will allow me to turn the counter `Map` back into the original `Set` and remove the counting logic altogether.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/constraint/DefaultConstraintMatchTotal.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/constraint/DefaultConstraintMatchTotal.java
@@ -16,6 +16,10 @@
 
 package org.optaplanner.core.impl.score.constraint;
 
+import static java.util.Comparator.comparing;
+import static java.util.Objects.hash;
+import static java.util.Objects.requireNonNull;
+
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/constraint/DefaultConstraintMatchTotal.java
```diff
@@ -85,25 +86,36 @@ public Score getScore() {
...
 
     public ConstraintMatch addConstraintMatch(List<Object> justificationList, Score score) {
-        this.score = this.score.add(score);
-        ConstraintMatch constraintMatch = new ConstraintMatch(constraintPackage, constraintName,
-                justificationList, score);
-        boolean added = constraintMatchSet.add(constraintMatch);
-        if (!added) {
-            throw new IllegalStateException("The constraintMatchTotal (" + this
-                    + ") could not add constraintMatch (" + constraintMatch
-                    + ") to its constraintMatchSet (" + constraintMatchSet + ").");
```
ge0ffrey, line 95: I've seen many score corruptions that were detected due to this fail fast exception. If this is removed, we're going to have more "ghosts in the machine" (= burn users).

triceo:
Decided with Geoffrey to not merge this.

PR #876, opened 2020-08-11
PLANNER-496 Ensure that nearby selection code is correct

triceo:
I'm adding a fail-fast and some comments to explain why the code - although it looks weird - actually works as intended.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/nearby/NearEntityNearbyEntitySelector.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/nearby/NearEntityNearbyEntitySelector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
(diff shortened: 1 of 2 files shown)


ge0ffrey (approved):

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/nearby/NearEntityNearbyEntitySelector.java
```diff
@@ -42,7 +43,12 @@
     public NearEntityNearbyEntitySelector(EntitySelector childEntitySelector, EntitySelector originEntitySelector,
             NearbyDistanceMeter nearbyDistanceMeter, NearbyRandom nearbyRandom, boolean randomSelection) {
         this.childEntitySelector = childEntitySelector;
-        this.originEntitySelector = originEntitySelector;
+        if (!(originEntitySelector instanceof MimicReplayingEntitySelector)) {
+            // In order to select a nearby entity, we must first have something to be near by.
+            throw new IllegalStateException("Programming error: Nearby entity selector (" + this +
```
ge0ffrey, line 48: "Programming error" isn't a term we use often. It's not an "impossible state", I guess?

PR #879, opened 2020-08-12
Reduce raw types usage

rsynek:
Also:
- renames `resolvedActiveThreadCount` to `resolveActiveThreadCount`
- fixes error message by passing correct instance to `ConfigUtils.newInstance`.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseFactory.java
@@ -53,10 +53,11 @@ public class DefaultPartitionedSearchPhaseFactory<Solution_>
             BestSolutionRecaller<Solution_> bestSolutionRecaller, Termination solverTermination) {
         HeuristicConfigPolicy phaseConfigPolicy = solverConfigPolicy.createPhaseConfigPolicy();
         ThreadFactory threadFactory = solverConfigPolicy.buildThreadFactory(ChildThreadType.PART_THREAD);
-        DefaultPartitionedSearchPhase phase = new DefaultPartitionedSearchPhase(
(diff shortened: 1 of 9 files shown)


triceo:
Although it may be worth getting rid of some of the Sonar warnings.

rsynek:
> Although it may be worth getting rid of some of the Sonar warnings.

Right, these were false positives:
- a licence header detected as a code duplication
- fragments of error messages, again, marked as code duplication

after removing them, the maintainability is back to A (although the check would need to rerun). Merging.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseFactory.java
```diff
@@ -76,13 +77,13 @@ public DefaultPartitionedSearchPhaseFactory(
         return phase;
     }
 
-    private SolutionPartitioner buildSolutionPartitioner() {
+    private SolutionPartitioner<Solution_> buildSolutionPartitioner() {
         if (phaseConfig.getSolutionPartitionerClass() != null) {
-            SolutionPartitioner<?> solutionPartitioner = ConfigUtils.newInstance(this,
-                    "solutionPartitionerClass", phaseConfig.getSolutionPartitionerClass());
+            SolutionPartitioner<?> solutionPartitioner =
+                    ConfigUtils.newInstance(phaseConfig, "solutionPartitionerClass", phaseConfig.getSolutionPartitionerClass());
```
yurloc, line 83: Why have you replaced `this` with `phaseConfig`?

rsynek: If you take a look at the `ConfigUtils.newInstance()`, the reference is used to properly communicate an error during instantiation of the `SolutionPartitionerClass`, which is a property of the `phaseConfig` object.

PR #882, opened 2020-08-13
PLANNER-2114 Extract build methods from Config

rsynek:
The scope of this PR is move selectors.

**JIRA**: [PLANNER-2114](https://issues.redhat.com/browse/PLANNER-2114)

**referenced Pull Requests**: None

**How to retest or run**:


rsynek:
I reviewed the test coverage and don't see how to improve it further without testing fail-fast checks (already did for some more complex checks).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -727,6 +727,32 @@ public ScoreDefinition getScoreDefinition() {
     // Model methods
     // ************************************************************************
 
+    public EntityDescriptor deduceEntityDescriptor(Class<?> entityClass) {
```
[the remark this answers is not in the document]
triceo: Please don't access collections directly. Leaking references to collections leads to a virtual impossibility of figuring out who has the opportunity to read (and more importantly write) it.

The config code is full of places where collections are being exported to other places, where they are being modified indiscriminately. Refactoring such code becomes a nightmare, as you are never sure if you've covered all the possible escape routes. 

Therefore I would like to suggest that exporting a modifiable collection should be prohibited.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/MoveSelectorFactory.java
```diff
@@ -0,0 +1,93 @@
...
+     * Builds {@link MoveSelector} from the {@link MoveSelectorConfig} and provided parameters.
+     * 
+     * @param configPolicy never null
+     * @param minimumCacheType never null, If caching is used (different from {@link SelectionCacheType#JUST_IN_TIME}),
+     *        then it should be at least this {@link SelectionCacheType} because an ancestor already uses such caching
+     *        and less would be pointless.
+     * @param inheritedSelectionOrder never null
+     * @return never null
+     */
+    MoveSelector buildMoveSelector(HeuristicConfigPolicy configPolicy, SelectionCacheType minimumCacheType,
```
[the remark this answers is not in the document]
rsynek: This is an interface, maybe you meant the same method on the `AbstractMoveSelectorFactory`?. Added, thanks.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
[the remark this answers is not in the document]
rsynek: As for the first suggestion, I will look into it in the next PR, which will be focused on the entity and value selection config. As for the second suggestion, I would end up duplicating a part of the `getGenuineEntityDescriptors` method, with the drawbacks @triceo has mentioned (although the reference does not leak out the class).

@yurloc could you please explain what is your concern about not accessing the map directly?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -516,6 +516,58 @@ public boolean hasAnyChainedGenuineVariables() {
...
+                        + ") has no configured variableName (" + variableName
+                        + ") for entityClass (" + getEntityClass()
+                        + ") and because there are multiple variableNames ("
+                        + getGenuineVariableNameSet()
+                        + "), it cannot be deduced automatically.");
+            }
+            variableDescriptor = variableDescriptors.iterator().next();
+        }
+        return variableDescriptor;
+    }
```
triceo, line 543: This method, it's actually two methods. One accepts the argument, the other doesn't need it. Let's do it like that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -516,6 +516,58 @@ public boolean hasAnyChainedGenuineVariables() {
...
+
+    public List<GenuineVariableDescriptor<Solution_>> deduceVariableDescriptorList(List<String> variableNameIncludeList) {
```
triceo, line 545: I'd prefer a no-arg overload to just putting `null` here.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -727,6 +727,32 @@ public ScoreDefinition getScoreDefinition() {
...
+                        + ") has no entityClass (" + entityClass
+                        + ") configured and because there are multiple in the entityClassSet ("
+                        + getEntityClassSet()
+                        + "), it cannot be deduced automatically.");
+            }
+            entityDescriptor = entityDescriptors.iterator().next();
+        }
+        return entityDescriptor;
+    }
+
```
triceo, line 755: Same comment on the `entityClass` nullity. This is two methods masquerading as one.

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/move/MoveSelectorFactoryTest.java
```diff
@@ -0,0 +1,302 @@
...
+    @Test
+    void stepRandom() {
+        final MoveSelector baseMoveSelector = SelectorTestUtils.mockMoveSelector(DummyMove.class);
+        DummyMoveSelectorConfig moveSelectorConfig = new DummyMoveSelectorConfig();
+        MoveSelectorFactory moveSelectorFactory =
+                new AssertingMoveSelectorFactory(moveSelectorConfig, baseMoveSelector, SelectionCacheType.STEP, false);
+        moveSelectorConfig.setCacheType(SelectionCacheType.STEP);
+        moveSelectorConfig.setSelectionOrder(SelectionOrder.RANDOM);
+        MoveSelector moveSelector = moveSelectorFactory.buildMoveSelector(
+                buildHeuristicConfigPolicy(),
```
triceo, line 121: Is this what the code formatter does? Pretty sure it fits on one 120-char line.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
rsynek: Regarding the last three comments - I agree, my plan was to look into it in the next PR, which focuses on the remaining part of the config code - including config of `ValueSelector` and `EntitySelector`. For this PR, I just moved these existing methods to - in my opinion - a more suitable place without changing their internals.

I should have clearly stated my intentions before opening the PR for review.

optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/move/MoveSelectorFactoryTest.java
rsynek: No, the formatter is not guilty; there must have been a new line already and the formatter didn't touch it, precisely as we wanted. Some of the test methods were moved from a different test class.

Good catch, thanks!

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
rsynek: will be addressed in the next PR, see the comment below.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
rsynek: will be addressed in the next PR, see the comment below.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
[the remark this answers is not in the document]
yurloc: The original post in this thread was a note for myself. I understand the second suggestion might be a bit confusing. Let's continue on line 733.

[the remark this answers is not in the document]
yurloc: I fully agree with what Lukas have said.

PR #884, opened 2020-08-18
Always use the same JDK for Javadoc as for Maven

triceo:
For systems where Maven may be running with different `JAVA_HOME` than is the default (such as Fedora), this fixes a whole lot of issues.


triceo:
And of course it doesn't work, because Java 8 has a different dir structure than Java 11.
Not worth the time.

PR #887, opened 2020-08-18
[7.x] RHDM-1430 Make the quickstart part of the build

triceo:


pom.xml
```diff
@@ -128,22 +129,9 @@
       <modules>
         <!-- Until Kogito is productized, neither can optaplanner-quarkus be productized. -->
         <module>optaplanner-quarkus-integration</module>
+        <module>optaplanner-quickstarts/quarkus-school-timetabling</module>
```
ge0ffrey, line 132: This approach is a bit unconvential (although it's less code so that's a good thing). Not sure how IntelliJ "import maven project" will react, or people would look for it here instead of the quarkus-quickstarts pom.xml (principle of least surprise), but I don't mind because it's not in master and as long as the quarkus quickstart is build on 7.x in community, but not in product, which it seems to do.

ge0ffrey (approved):
LGTM to merge.

pom.xml
triceo: Yeah, I tried both `productized` and not, and it works as expected.

PR #888, opened 2020-08-19
PLANNER-1771 Make Quarkus extension exit if no OptaPlanner annotated classes are found

ge0ffrey:


ge0ffrey:
@geoand @gsmet Would either of you have a moment to review these changes to the Quarkus extension of OptaPlanner?

geoand:
Hey @ge0ffrey ,

I'll be back from PTO next week, so if no one has beat me to it, I'll check it then

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -81,18 +86,37 @@ HotDeploymentWatchedFileBuildItem watchScoreDrl() {
...
+        if (hasNoOptaPlannerContent(indexView)) {
+            return;
+        }
         // The bean encapsulating the SolverFactory
         additionalBeans.produce(new AdditionalBeanBuildItem(OptaPlannerBeanProvider.class));
     }
 
+    private boolean hasNoOptaPlannerContent(IndexView indexView) {
+        // Only skip this extension if everything is missing. Otherwise, if some parts are missing, fail fast later.
+        return indexView.getAnnotations(DotNames.PLANNING_SOLUTION).isEmpty()
```
rsynek, line 101: Do we care about DRL (less than about easy score calculation)?

rsynek (approved):
Looks good, one question to think about.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
ge0ffrey: We could do add a `classLoader.getResource(SolverBuildTimeConfig.DEFAULT_SCORE_DRL_URL) == null` check?

rsynek: That partially helps; now people might override the default location by solverConfig.xml. Where is the borderline of what else to check? What about the opposite approach: check just the annotations and provide a debug-level message explaining what's going on.
Because if we check the project just partially, for a couple of corner cases it will be very difficult for a user to investigate why the extension does just nothing.

gsmet (approved):

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
ge0ffrey: Very good point about the debug message!

ge0ffrey:
@rsynek I've adjusted as you suggested, and also did a few other refactors based upon threads in the quarkus zulip. Would you mind reviewing again?

rsynek (approved):
Looks good, thanks!

PR #893, opened 2020-08-20
PLANNER-2116 Externalize build methods of entity and value selectors

rsynek:
**JIRA**: [PLANNER-2116](https://issues.redhat.com/browse/PLANNER-2116)

**How to retest or run**:


optaplanner-core/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
```diff
@@ -0,0 +1,115 @@
...
+
+    protected final Config_ config;
+
+    public AbstractFromConfigFactory(Config_ config) {
+        this.config = config;
+    }
+
+    protected EntityDescriptor deduceEntityDescriptor(SolutionDescriptor solutionDescriptor, Class<?> entityClass) {
+        EntityDescriptor entityDescriptor;
+        entityDescriptor = solutionDescriptor.getEntityDescriptorStrict(Objects.requireNonNull(entityClass));
```
triceo, line 40: Please merge the assignment with the declaration.

optaplanner-core/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
```diff
@@ -0,0 +1,115 @@
...
+                    + ") has no entityClass configured and because there are multiple in the entityClassSet ("
+                    + solutionDescriptor.getEntityClassSet()
+                    + "), it cannot be deduced automatically.");
+        }
+        return entityDescriptors.iterator().next();
+    }
+
+    protected GenuineVariableDescriptor deduceVariableDescriptor(EntityDescriptor entityDescriptor, String variableName) {
+        GenuineVariableDescriptor variableDescriptor;
+        variableDescriptor = entityDescriptor.getGenuineVariableDescriptor(Objects.requireNonNull(variableName));
```
triceo, line 65: Dtto.

optaplanner-core/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+            if (!found) {
+                throw new IllegalArgumentException("The config (" + this
+                        + ") has a variableNameInclude (" + variableNameInclude
+                        + ") which does not exist in the entity (" + entityDescriptor.getEntityClass()
+                        + ")'s variableDescriptorList (" + variableDescriptorList + ").");
+            }
+        }
+        return resolvedVariableDescriptorList;
```
triceo, line 113: This could be a nice stream.

    return variableNameIncludeList.stream()
        .map(n -> variableDescriptorList.stream()
            .filter(d -> d.gerVariableName().equals(n))
            .findFirst()
            .orElseThrow(...))
        .collect(Collectors.toList());

In my opinion, this reads better as you don't need to mentally track the state of `found`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/NearbyRandomFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        boolean linearDistributionEnabled = nearbySelectionConfig
+                .getNearbySelectionDistributionType() == NearbySelectionDistributionType.LINEAR_DISTRIBUTION
+                || nearbySelectionConfig.getLinearDistributionSizeMaximum() != null;
+        boolean parabolicDistributionEnabled = nearbySelectionConfig
+                .getNearbySelectionDistributionType() == NearbySelectionDistributionType.PARABOLIC_DISTRIBUTION
+                || nearbySelectionConfig.getParabolicDistributionSizeMaximum() != null;
+        boolean betaDistributionEnabled =
+                nearbySelectionConfig.getNearbySelectionDistributionType() == NearbySelectionDistributionType.BETA_DISTRIBUTION
+                        || nearbySelectionConfig.getBetaDistributionAlpha() != null
+                        || nearbySelectionConfig.getBetaDistributionBeta() != null;
```
triceo, line 52: How about:

    boolean distributionEnabled = false;
    switch (nearbySelectionConfig.getNearbySelectionDistributionType()) {
         case LINEAR_DISTRIBUTION:
              distributionEnabled = nearbySelectionConfig.getLinearDistributionSizeMaximum() != null;
              break;
         ...
    }

Less repetition, and in my opinion more clarity.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/NearbyRandomFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        if (!randomSelection) {
+            if (blockDistributionEnabled || linearDistributionEnabled || parabolicDistributionEnabled
+                    || betaDistributionEnabled) {
+                throw new IllegalArgumentException("The nearbySelectorConfig (" + nearbySelectionConfig
+                        + ") with randomSelection (" + randomSelection
+                        + ") has distribution parameters.");
+            }
+            return null;
+        }
+        if (blockDistributionEnabled && linearDistributionEnabled) {
```
triceo, line 62: Implementing my previous comment would complicate this. But I suppose you could set these variables in the `switch` directly.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/NearbyRandomFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+        if (linearDistributionEnabled && betaDistributionEnabled) {
+            throw new IllegalArgumentException("The nearbySelectorConfig (" + nearbySelectionConfig
+                    + ") has both linearDistribution and betaDistribution parameters.");
+        }
+        if (parabolicDistributionEnabled && betaDistributionEnabled) {
+            throw new IllegalArgumentException("The nearbySelectorConfig (" + nearbySelectionConfig
+                    + ") has both parabolicDistribution and betaDistribution parameters.");
+        }
+        if (blockDistributionEnabled) {
```
triceo, line 86: Personally, I'd make this part of the `switch`, too. There should only be one place where decisions are made from a single piece of data.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
```diff
@@ -0,0 +1,396 @@
...
+                || config.getSelectionOrder() != null
+                || config.getNearbySelectionConfig() != null
+                || config.getFilterClass() != null
+                || config.getSorterManner() != null
+                || config.getSorterComparatorClass() != null
+                || config.getSorterWeightFactoryClass() != null
+                || config.getSorterOrder() != null
+                || config.getSorterClass() != null
+                || config.getProbabilityWeightFactoryClass() != null
+                || config.getSelectedCountLimit() != null) {
```
triceo, line 143: Good that you're not doing mutation testing. You'd grow old testing all the permutations of this particular condition.

triceo: But to be constructive here - I'd add all of the calls into a stream, and then filter anything non null. That way, the code would  more readily express what you're trying to do.

triceo (commented):
I have added some suggestions, but nothing critical.

optaplanner-core/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
rsynek: I didn't touch the existing code if I wasn't changing it. That said, if there is an opportunity to improve it, I will take a look at this one and other suggestions you made.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/NearbyRandomFactory.java
rsynek: Let me just answer this and the next two comments together, as they are connected with each other:
The method consists of 3 steps:

1. discovers which distribution types are enabled
2. fails fast if multiple distribution types have been configured
3. returns the computed distribution

From your comments, I understand you suggest implementing all these steps in a single switch statement.

To implement step 2, the subsequent branches of the switch statement would have to depend on preceding ones. Each case would contain:
- the logic for discovering its distribution type
- fail fast condition(s)
- instantiation of the NearbyRandom

Since the implementation of fail fast conditions requires avoiding `break` after any `case`, we cannot use the default branch, which is supposed to return `LinearDistributionNearbyRandom(Integer.MAX_VALUE)`.

To sum this up, I don't think doing all 3 steps in the switch statement (while there will be additional if conditions after it anyway) is as an improvement. But I like the idea of using switch for the step 1.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
rsynek: Good point, I used the same "trick" for resolving terminationConfigs; here it suits pretty well too.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/NearbyRandomFactory.java
rsynek: Actually even using switch just for step 1 wouldn't help, as the individual conditions are connected with logical OR.

triceo (approved):
Good to go when green.

PR #902, opened 2020-08-26
PLANNER-2093 Schema Definition for SolverConfig and BenchmarkConfig

rsynek:
**JIRA**: [PLANNER-2093](https://issues.redhat.com/browse/PLANNER-2093)

**How to retest or run**:


optaplanner-build-parent/pom.xml
```diff
@@ -840,6 +840,31 @@
...
+            </execution>
+          </executions>
+          <configuration>
+            <clearOutputDir>true</clearOutputDir>
+            <createJavaDocAnnotations>false</createJavaDocAnnotations>
+            <generateEpisode>false</generateEpisode>
+            <outputDirectory>${project.build.outputDirectory}</outputDirectory>
+            <transformSchemas>
+              <transformSchema>
+                <uri>https://www.optaplanner.org/xsd/optaplanner</uri>
```
rsynek, line 862: This is the single namespace we have for both the `<solver/>` and `<benchmark/>`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+        Objects.requireNonNull(reader);
+        try {
+            return (T) createUnmarshaller().unmarshal(reader);
+        } catch (JAXBException jaxbException) {
+            String errMessage = String.format(ERR_MSG_READ, rootClass.getName());
+            throw new OptaPlannerXmlSerializationException(errMessage, jaxbException);
+        }
+    }
+
+    public T readAndValidate(Reader reader, String schemaResource) {
```
rsynek, line 104: As we agreed, we don't validate the input solver config XML. We use the validation just in tests to make sure the solver config XML is valid by the generated XSD.

triceo (approved):
LGTM. Just fix the PR title, please.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+import org.xml.sax.SAXException;
+import org.xml.sax.XMLFilter;
+import org.xml.sax.XMLReader;
+import org.xml.sax.helpers.XMLFilterImpl;
+
+public final class GenericJaxbIO<T> implements JaxbIO<T> {
+    private static final int DEFAULT_INDENTATION = 2;
+
+    private static final String ERR_MSG_WRITE = "Unable to write the %s to XML.";
+    private static final String ERR_MSG_READ = "Unable to read the (%s) from XML.";
```
ge0ffrey, line 66: soft: I think putting error messages in a constant - if it's only used once - is a form of unneeded indirection, making the code harder to read (code is read 10 times more than it it written) because it reduces "vertical density" (that which belongs together is together).

In this case, the ERR_MSG_READ is used more than once, but that's often a sign that multiple different error causes are being reported as the same error, providing less information to the user. See examples below.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+    public void writeWithoutNamespaces(T root, Writer writer) {
+        Objects.requireNonNull(root);
+        Objects.requireNonNull(writer);
+        DOMResult domResult = new DOMResult();
+        final String errMessage = String.format(ERR_MSG_WRITE, rootClass.getName());
+        try {
+            marshaller.marshal(root, domResult);
+        } catch (JAXBException jaxbException) {
+            throw new OptaPlannerXmlSerializationException(errMessage, jaxbException);
+        }
```
ge0ffrey, line 231: DRY with the first few lines of `write(T root, Writer writer)`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+
+        formatXml(new DOMSource(domResult.getNode()), null, writer);
+    }
+
+    public void writeWithoutNamespaces(T root, Writer writer) {
+        Objects.requireNonNull(root);
+        Objects.requireNonNull(writer);
+        DOMResult domResult = new DOMResult();
+        final String errMessage = String.format(ERR_MSG_WRITE, rootClass.getName());
```
ge0ffrey, line 226: The errorMessage String should only be created if it actually fails, so inside the catch

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
ge0ffrey: These error message formatting are not consistent with our normal formatting "... parameterName (parameterValue) ..."
So
  "Unable to write the %s to XML."
=> "Unable to write rootClass (%s) to XML."

ge0ffrey: - A rootClass isn't written to XML, it's marshalled to XML.
- We typically favor "Failed to" instead of "Unable to".
=> "Failed to marshall rootClass (%s) to XML."

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+        TransformerFactory transformerFactory = TransformerFactory.newInstance();
+        try {
+            Transformer transformer = transformationTemplate == null ? transformerFactory.newTransformer()
+                    : transformerFactory.newTransformer(transformationTemplate);
+            // See https://stackoverflow.com/questions/46708498/jaxb-marshaller-indentation.
+            transformer.setOutputProperty(OutputKeys.INDENT, "yes");
+            transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", String.valueOf(indentation));
+            transformer.transform(source, new StreamResult(writer));
+        } catch (TransformerException transformerException) {
+            String errMessage = String.format(ERR_MSG_WRITE, rootClass.getName());
```
ge0ffrey, line 256: Here, it didn't fail to marshall the rootClass to XML, so the error message should be different.
=> "Failed to format XML for rootClass(...)." maybe?

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        try (InputStream xsltInputStream = getClass().getResourceAsStream("removeNamespaces.xslt")) {
+            formatXml(new DOMSource(domResult.getNode()), new StreamSource(xsltInputStream), writer);
+        } catch (IOException e) {
+            throw new OptaPlannerXmlSerializationException(errMessage, e);
+        }
+    }
+
+    private void formatXml(Source source, Source transformationTemplate, Writer writer) {
```
ge0ffrey, line 240: Do we need to bring in an XSLT formatter to format the JAXB output?
Can we not use this instead?

`marshaller.setProperty(Marshaller.JAXB_FORMATTED_OUTPUT, Boolean.TRUE);`

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+     * @param reader input XML {@link Reader}; never null
+     * @param elementNamespaceOverrides never null
+     * @return deserialized object representation of the XML.
+     */
+    public T readOverridingNamespace(Reader reader, ElementNamespaceOverride... elementNamespaceOverrides) {
+        Objects.requireNonNull(reader);
+        Objects.requireNonNull(elementNamespaceOverrides);
+
+        final String errMessage = String.format("Unable to read the (%s) from XML with overriding elements' namespaces: %s.",
+                rootClass.getName(), Arrays.toString(elementNamespaceOverrides));
```
ge0ffrey, line 164: Performance nitpick: Construct error message only when it fails.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+     * 
+     * @param reader input XML {@link Reader}; never null
+     * @param elementNamespaceOverrides never null
+     * @return deserialized object representation of the XML.
+     */
+    public T readOverridingNamespace(Reader reader, ElementNamespaceOverride... elementNamespaceOverrides) {
+        Objects.requireNonNull(reader);
+        Objects.requireNonNull(elementNamespaceOverrides);
+
+        final String errMessage = String.format("Unable to read the (%s) from XML with overriding elements' namespaces: %s.",
```
ge0ffrey, line 163: error style, always use braces to mention values at runtime.
=> "... the rootClass (%s) ... namespaces (%s)."

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+                activeNamespace.push(namespaceOverride);
+            }
+
+            String resultingUri = activeNamespace.isEmpty() ? uri : activeNamespace.peek();
+            super.startElement(resultingUri, localName, qName, atts);
+        }
+
+    }
+
+    private static final class ValidationEventStringCollector extends ValidationEventCollector {
```
ge0ffrey, line 300: Do we need a custom class for this?
We can probably just do 

`String s = validationEventCollector.getEvents().stream().map(Event::getMessage).collect(Collectors.joining("\n"))`

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/OptaPlannerXmlSerializationException.java
```diff
@@ -16,21 +16,21 @@
 
 package org.optaplanner.core.impl.io;
 
-public class XmlUnmarshallingException extends RuntimeException {
+public class OptaPlannerXmlSerializationException extends RuntimeException {
```
ge0ffrey, line 19: open question (maybe for another PR): Is this public API? Should SolverFactory.createFromXmlResource()'s javadoc have a "@throws OptaPlannerXmlSerializationException" ? In that case we might want to move it into the api package.

ge0ffrey: Naming:
- We avoid class names that contain the word "OptaPlanner" because everything is OptaPlanner.
- Is it Serialization or Marshalling? Much of the other code naming and error messages seem to talk about marshalling, not serialization. Let's use one term and stick with that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+
+import org.optaplanner.core.impl.io.OptaPlannerXmlSerializationException;
+import org.xml.sax.Attributes;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+import org.xml.sax.XMLFilter;
+import org.xml.sax.XMLReader;
+import org.xml.sax.helpers.XMLFilterImpl;
+
+public final class GenericJaxbIO<T> implements JaxbIO<T> {
```
ge0ffrey, line 62: Naming consistency: the only implementation of an interface Foo is normally called DefaultFoo in the rest of the optaplanner repo. This makes it easier to guess when doing ctrl-n.

=> rename GenericJaxbIO to DefaultJaxbIO

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -0,0 +1,318 @@
...
+            throw new OptaPlannerXmlSerializationException(errMessage, jaxbException);
+        }
+
+        try {
+            return (T) unmarshaller.unmarshal(reader);
+        } catch (JAXBException jaxbException) {
+            String errMessage = String.format(ERR_MSG_READ, rootClass.getName());
+            if (validationEventHandler.hasEvents()) {
+                String errMessageWithValidationEvents = errMessage + "\n" + validationEventHandler.reportAll();
+                throw new OptaPlannerXmlSerializationException(errMessageWithValidationEvents, jaxbException);
```
ge0ffrey, line 133: Read didn't fail here. Validation failed. The error message should reflect that.

ge0ffrey:
In optaplanner 7, we had the guarantee that if `SolverConfig.createFromXMLResource("fooConfig.xml")` was called, the error message mentioned `fooConfig.xml` to easily spot typo's. Do we still have that behavior?

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
rsynek: Unfortunately, the `Marshaller.JAXB_FORMATTED_OUTPUT` does always 4 spaces of indentation. We use the XSLT to make it 2 (or any other number, if needed).

For the "backward compatibility mode", when writing the elements, we remove their namespaces via an XSLT, so we need a transformation anyway.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
rsynek: It's not the only implementation; it's just the only one that takes a generic type parameter. Other implementations deal e.g. with the `SolverConfig` and use this generic one under the hood.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/OptaPlannerXmlSerializationException.java
rsynek: The same question applies to all the fail-fast checks: should we mention there the `IllegalArgumentException`? Maybe it should. In that case, I would definitely do it in a separate PR.

rsynek:
> In optaplanner 7, we had the guarantee that if `SolverConfig.createFromXMLResource("fooConfig.xml")` was called, the error message mentioned `fooConfig.xml` to easily spot typo's. Do we still have that behavior?
> 
> This prevents a lot of StackOverflow questions due to typo's :)

That's a responsibility of the `SolverFactory` - nothing has changed there in this regard.

optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
rsynek: Only messages that repeat are defined as constants. However, I changed the message in a couple of places to make it more specific.

Thanks.

PR #904, opened 2020-08-27
Refactor ExaminationConstraintProviderTest.mixedDurationsTest

triceo:
Prevents NPE when the test fails. Also, breaks down a single mega test into multiple smaller tests.


rsynek (approved):
LGTM

PR #906, opened 2020-08-27
PLANNER-2137 Guarantee unique name for CS-D variables

triceo:
All variable creation now happens in one place. That one place appends auto-incrementing numbers to the requested variable names. Therefore variable uniqueness is guaranteed.

The old CS-D did the same thing, but the feature got somehow lost in the refactor.


triceo:
@rsynek or @ge0ffrey , either of you, review please.
This is the bug I've been chasing for the past couple of days.

rsynek (approved):
Looks good.

PR #907, opened 2020-08-27
Updating the github PR template

rsynek:
- simplify the Jenkins triggering phrases and make them rather a subtle hint than a major part of the template
- add a checklist of typical things we should think about that are not automated as a part of the PR checks

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,23 +1,23 @@
-**Thank you for submitting this pull request**
+**JIRA**: [link](https://www.example.com) 
 
-**JIRA**: _(please edit the JIRA link if it exists)_ 
+**Referenced Pull Requests** (if any):
 
-[link](https://www.example.com)
+**Checklist**:
 
-**referenced Pull Requests**: _(please edit the URLs of referenced pullrequests if they exist)_
+- [ ] Documentation updated if applicable
+- [ ] Upgrade recipe provided if applicable
 
-paste the link(s) from GitHub here
+**Thank you for submitting this pull request!**
 
-**How to retest or run**:
+---
+<pre>
+<b>To run tests, please add a comment:</b>
 
-* a pull request please add comment: regex **[.\*[j|J]enkins,?.\*(retest|test) this.\*]**
-
-* a full downstream build please add comment: regex **[.*\[j|J]enkins,?.\*(execute|run|trigger|start|do) fdb.\*]**
-
-* a compile downstream build please  add comment: regex **[.\*[j|J]enkins,?.\*(execute|run|trigger|start|do) cdb.\*]**
-
-* a full production downstream please add comment: regex **[.\*[j|J]enkins,?.\*(execute|run|trigger|start|do) product fdb.\*]**
-
-* an upstream build please add comment: regex **[.\*[j|J]enkins,?.\*(execute|run|trigger|start|do) upstream.\*]**
-
-i.e for running a full downstream build =  **Jenkins do fdb**
+<ul>
+    <li><b>jenkins test this</b> (for a PR check)</li>
+    <li><b>jenkins run fdb</b> (for a full downstream build)</li>
+    <li><b>jenkins run cdb</b> (for a compile downstream build)</li>
+    <li><b>jenkins run product fdb</b> (a full production downstream)</li>
+    <li><b>jenkins run upstream</b> (for an upstream build)</li>
+</ul>
+</pre>


.github/pull_request_template.md
```diff
@@ -1,23 +1,23 @@
...
-**referenced Pull Requests**: _(please edit the URLs of referenced pullrequests if they exist)_
+- [ ] Documentation updated if applicable
+- [ ] Upgrade recipe provided if applicable
 
-paste the link(s) from GitHub here
+**Thank you for submitting this pull request!**
 
-**How to retest or run**:
+---
+<pre>
```
triceo, line 13: Is this intentional? I don't really see a reason for the entire following text to be preformatted.
The commands themselves, yes. But everything?

.github/pull_request_template.md
```diff
@@ -1,23 +1,23 @@
...
 
-[link](https://www.example.com)
+**Checklist**:
 
-**referenced Pull Requests**: _(please edit the URLs of referenced pullrequests if they exist)_
+- [ ] Documentation updated if applicable
+- [ ] Upgrade recipe provided if applicable
 
-paste the link(s) from GitHub here
+**Thank you for submitting this pull request!**
```
triceo, line 10: Personally, I think that this is a good thing to start with.

.github/pull_request_template.md
```diff
@@ -1,23 +1,23 @@
-**Thank you for submitting this pull request**
+**JIRA**: [link](https://www.example.com) 
 
-**JIRA**: _(please edit the JIRA link if it exists)_
```
triceo, line 3: The previous version made it clear that this was optional. Do we actually want people to file JIRAs for every PR they provide?

I don't really see the template as something for us. We know the rules. The template is for the outsiders, and shouldn't scare them away.

.github/pull_request_template.md
```diff
@@ -1,23 +1,23 @@
-**Thank you for submitting this pull request**
+**JIRA**: [link](https://www.example.com) 
 
-**JIRA**: _(please edit the JIRA link if it exists)_ 
+**Referenced Pull Requests** (if any):
 
-[link](https://www.example.com)
```
triceo, line 5: Referenced why? Again, you and I know that, but does the target audience?
I understand the intention to keep it short. But if that's the intention, we might as well just not include this section. The outsiders won't be doing cross-repo PRs.

.github/pull_request_template.md
```diff
@@ -1,23 +1,23 @@
...
-* an upstream build please add comment: regex **[.\*[j|J]enkins,?.\*(execute|run|trigger|start|do) upstream.\*]**
-
-i.e for running a full downstream build =  **Jenkins do fdb**
+<ul>
+    <li><b>jenkins test this</b> (for a PR check)</li>
+    <li><b>jenkins run fdb</b> (for a full downstream build)</li>
+    <li><b>jenkins run cdb</b> (for a compile downstream build)</li>
+    <li><b>jenkins run product fdb</b> (a full production downstream)</li>
+    <li><b>jenkins run upstream</b> (for an upstream build)</li>
+</ul>
```
triceo, line 22: If we're shortening, surely we don't need all of those commands in there.
I myself only ever used two of these. :-)

triceo (commented):
Some suggestions. Don't feel particularly strong about either.
But I do think we should figure out who these templates are for, and tailor the content to them.

.github/pull_request_template.md
yurloc: We should use `<!-- Comments -->` for instructions, explanation etc.

.github/pull_request_template.md
yurloc: Suggestion: use `<details>`:
<details>
<summary>How to retest</summary>

- PR build
- FDB
- etc
</details>

See [Using `<details>` in GitHub](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d).

.github/pull_request_template.md
yurloc: In my opinion we can keep it but comment it out. The person opening a PR will see it but it will remain invisible after the PR is submitted, reducing noise.

yurloc:
My proposal: https://github.com/yurloc/optaweb-vehicle-routing/compare/master...yurloc:test?expand=1

rsynek:
> My proposal: https://github.com/yurloc/optaweb-vehicle-routing/compare/master...yurloc:test?expand=1

The comments and the <details/> section works much better than my proposal; it has all the tips and hints while showing only the necessary information once the PR is created.

@yurloc Please open a PR, I am going to close mine. Thanks!

PR #908, opened 2020-08-28
Remove XSD from JAR files

rsynek:
Removes XSD and JAXB metadata from the JAR files.


optaplanner-core/pom.xml
```diff
@@ -142,6 +142,10 @@
         <artifactId>jaxb2-maven-plugin</artifactId>
         <configuration>
           <sources>
+            <!--
+              The XSD generated in this module contains only its types and thus is usable only for the solver but not for
+              the benchmark configuration XML.
+            -->
             <source>${project.build.sourceDirectory}/org/optaplanner/core/config/solver</source>
```
rsynek, line 149: This is going to be an issue if we start validating in the future: the XSDs inside optaplanner-core.jar and optaplanner-benchmark.jar are different although they declare the same namespace.

XSD in the optaplanner-benchmark is a superset of the optaplanner-core XSD.

optaplanner-build-parent/pom.xml
```diff
@@ -730,6 +730,23 @@
                 <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
               </manifest>
             </archive>
+            <excludes>
+              <!-- Remove unnecessary metadata generated by the jaxb2-maven-plugin. -->
+              <exclude>META-INF/JAXB/</exclude>
+              <!-- The XSD should be available only on the https://www.optaplanner.org. -->
+              <exclude>optaplanner.xsd</exclude>
```
ge0ffrey, line 737: Doesn't the jaxb maven plugin have a property to not put those files in the jar file?

optaplanner-core/pom.xml
triceo: Another argument for my proposal in #909. The fact that both Solver and Benchmark are the same physical XSD is nothing but trouble.

optaplanner-build-parent/pom.xml
rsynek: No; it can be moved to a different location, though, but it needs to be on the classpath so that we can test it.

optaplanner-core/pom.xml
rsynek: I fully agree this is a potential problem. The only solution to this problem is designing the XSDs right: one for the solver config and another one for the benchmark config, each in a different namespace. 
However, if we generate the solver config in a different namespace than the benchmark config, there is no way of having the solver config elements in the benchmark namespace. There will be `<xs:import/>` and users creating a benchmark config XML will have to qualify solver config elements to the right namespace by using prefix:
```
<s:solver>
...
</s:solver>
```

rsynek: Let me summarize a conversation we had with @triceo:
- we would like to both avoid prefixes and single namespace
- as there is no way of doing this via JAXB, let's generate each XSD in a separate namespace and replace `<xs:import/>` in the `benchmark.xsd` by the content of `solver.xsd`

rsynek:
Closing in favor of https://github.com/kiegroup/optaplanner/pull/911.

PR #909, opened 2020-08-28
PLANNER-2136 Use XSD in optaplanner-examples

rsynek:
**JIRA**: [PLANNER-2136](https://issues.redhat.com/browse/PLANNER-2136)

### DIFF
--- a/optaplanner-examples/src/main/resources/org/optaplanner/examples/app/benchmark/generalOptaPlannerBenchmarkConfig.xml
+++ b/optaplanner-examples/src/main/resources/org/optaplanner/examples/app/benchmark/generalOptaPlannerBenchmarkConfig.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<plannerBenchmark>
+<plannerBenchmark xmlns="https://www.optaplanner.org/xsd/optaplanner" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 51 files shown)


optaplanner-examples/src/main/resources/org/optaplanner/examples/app/benchmark/generalOptaPlannerBenchmarkConfig.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<plannerBenchmark>
+<plannerBenchmark xmlns="https://www.optaplanner.org/xsd/optaplanner" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="https://www.optaplanner.org/xsd/optaplanner https://www.optaplanner.org/xsd/optaplanner/optaplanner.xsd">
```
triceo, line 3: I don't like the fact that both Benchmark and Solver share a namespace and are both in the same XSD.
In my opinion, they should be two different XSDs, with two different namespaces.
Where the Benchmarker references the Solver, that defintion would still be included in the Benchmarker XSD, no linking or anything like that.

Actually, I thought that that's what we were already agreed to do.

rsynek: Let me answer at https://github.com/kiegroup/optaplanner/pull/908 so that we continue the discussion in a single place.

rsynek:
Closing in favor of https://github.com/kiegroup/optaplanner/pull/911

PR #910, opened 2020-08-31
[7.x] [KOGITO-2852] descoping of mvel/asm from the core engine

mariofusco:


ge0ffrey (approved):
Good to merge if CI is green.

ge0ffrey:
@mariofusco This is the error on CI:
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.drools:drools-mvel:jar is missing. @ line 47, column 17
[ERROR] 'dependencies.dependency.version' for org.drools:drools-mvel:test-jar is missing. @ line 51, column 17
[ERROR] 'dependencies.dependency.version' for org.drools:drools-mvel:jar is missing. @ line 68, column 17
[ERROR] 'dependencies.dependency.version' for org.drools:drools-mvel:test-jar is missing. @ line 73, column 17
```

OptaPlanner 7 extends kie-parent, so it should inherit the drools-mvel version from the drools-bom on the build-bootstrap repo.
Presuming that that is a new module, please either submit a PR there with the same branch name, or - what I find easier - simply merge adding that to the drools-bom already (it's pretty harmless) and retrigger the build here.

mariofusco:
> OptaPlanner 7 extends kie-parent, so it should inherit the drools-mvel version from the drools-bom on the build-bootstrap repo.
> Presuming that that is a new module, please either submit a PR there with the same branch name.

@ge0ffrey I think I already did this ( see https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1448/files ) and indeed optaplanner compiles correctly for me locally. Is this probably a CI issue or am I missing something else?

triceo:
As I already explained on Zulip, this PR will never be green. Github Actions can not use PRs from the parent repos, and therefore will always fail as the BOMs will not have the changes we need.

Since the Jenkins build works, and that one actually uses the parent PRs, we should be good to merge this.

PR #911, opened 2020-09-01
Merge solver.xsd and benchmark.xsd

rsynek:
An alternative to having a single namespace: merging solver.xsd into benchmark.xsd, so that:
- optaplanner-core.jar can contain solver.xsd under its namespace
- optaplanner-benchmark.jar can contain benchmark.xsd under its own namespace without importing the solver.xsd (a thus without the need of using prefixes in the benchmarkConfig.xml)

Related PR:
https://github.com/kiegroup/optaplanner-website/pull/271


triceo (approved):
The best thing possible, given the circumstances.
Question: don't all the Optaplanner Examples configs need updating too? I remember seeing a PR earlier, changing them to use the XSD which now changes location.

rsynek:
> The best thing possible, given the circumstances.
> Question: don't all the Optaplanner Examples configs need updating too? I remember seeing a PR earlier, changing them to use the XSD which now changes location.

Sure they do; I wanted to have some feedback first before I change all of them. Also, similar cleanup of the JAR files as in https://github.com/kiegroup/optaplanner/pull/908 will be needed too (we want to keep the benchmark.xsd, but remove the benchmark-with-import.xsd).

optaplanner-examples/src/main/resources/org/optaplanner/examples/cheaptime/solver/cheapTimeSolverConfig.xml
```diff
@@ -14,7 +15,7 @@
     <!--<scoreDrl>org/optaplanner/examples/cheaptime/solver/cheapTimeConstraints.drl</scoreDrl>-->
     <initializingScoreTrend>ONLY_DOWN</initializingScoreTrend>
     <!--<assertionScoreDirectorFactory>-->
-      <!--<easyScoreCalculatorClass>org.optaplanner.examples.cheaptime.solver.score.CheapTimeEasyScoreCalculator</easyScoreCalculatorClass>-->
```
triceo, line 17: Unnecessary change. This PR actually has plenty of them.
(Although does it matter? I leave it up to you, @rsynek .)

rsynek: Good catch, thanks.

ge0ffrey:
Interesting, let me look at this tomorrow, after OptaPlanner Week event.

PR #912, opened 2020-09-04
Improve pull request template

yurloc:
Supersedes #907.

The goal is to
1. Remove noise from the PR description as it appears after submitting it.
2. Keep instructions and comments that make it easier for an external contributor to provide all relevant information.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,23 +1,38 @@
-**Thank you for submitting this pull request**
+<!--
+Thank you for submitting this pull request.
 
-**JIRA**: _(please edit the JIRA link if it exists)_ 
+Please provide all relevant information as outlined below. Feel free to delete
+a section if that type of information is not available.
+-->
 
-[link](https://www.example.com)
+### JIRA
 
-**referenced Pull Requests**: _(please edit the URLs of referenced pullrequests if they exist)_
+<!-- Add a JIRA ticket link if it exists. -->
+<!-- Example: https://issues.redhat.com/browse/PLANNER-1234 -->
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,23 +1,38 @@
...
+<!-- Example:
+* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1234
+* https://github.com/kiegroup/drools/pull/3000
+* https://github.com/kiegroup/optaplanner/pull/899
+* etc.
+-->
+
+### Checklist
+- [ ] Documentation updated if applicable.
+- [ ] Upgrade recipe provided if applicable.
```
yurloc, line 26: Not sure about the checklist. I'm afraid it will be confusing most of the times because it will stay in the description unchecked. In reality both checklist items have 3 states: -1) not applicable, 0) required but not included, 1) included. So I'm afraid that for reviewer it will be confusing whether an empty checkbox means 0 and the PR is blocked on that or it means -1.

rsynek: That's why it's formulated as "*Something* updated if applicable." - if there is no update necessary, it should be checked. Of course, it doesn't guarantee that what you describe cannot happen.

rsynek: We can give it a try and if we see it's a source of confusion, we can always remove the checklist.

rsynek (approved):

triceo (approved):

ge0ffrey (approved):
This changes are good, but I am not sure if a pull request template it is worth it.
The more hoops we set up, the less likely external contributors will take the time to contribute.

PR #914, opened 2020-09-07
Disable Maven deployment of quickstarts

triceo:


mareknovotny (approved):
lgtm

optaplanner-quickstarts/quarkus-school-timetabling/pom.xml
```diff
@@ -202,5 +203,26 @@
...
+          <plugin>
+            <artifactId>maven-deploy-plugin</artifactId>
+            <version>${deploy-plugin.version}</version>
+            <configuration>
+              <skip>true</skip>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
```
ge0ffrey, line 224: We don't want to litter the quickstarts with this kind of stuff. Users do look at all the code in the quickstart and want to grok it.
However, if this only affects 7.x, and in 8.x we're solving this with the optaplanner-quickstart seperate repo - which seems to be the place - then I have no problem with this temp workaround.

PR #915, opened 2020-09-07
PLANNER-2135 Remove support for JavaScript expressions from configs

triceo:
Verified to work correctly on OpenJDK 15.
This will not be backported to 7.x - we'll have to use a different approach there.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/PlannerBenchmarkConfig.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/PlannerBenchmarkConfig.java
@@ -584,8 +584,7 @@ public class PlannerBenchmarkConfig {
      * <p>
      * If there aren't enough processors available, it will be decreased.
      *
-     * @return null, a number, {@value #PARALLEL_BENCHMARK_COUNT_AUTO} or a JavaScript calculation using
-     *         {@value org.optaplanner.core.config.util.ConfigUtils#AVAILABLE_PROCESSOR_COUNT}.
+     * @return null, a number or {@value #PARALLEL_BENCHMARK_COUNT_AUTO}.
      */
     public String getParallelBenchmarkCount() {
         return parallelBenchmarkCount;
@@ -806,8 +805,8 @@ public class PlannerBenchmarkConfig {
         } else if (parallelBenchmarkCount.equals(PARALLEL_BENCHMARK_COUNT_AUTO)) {
             resolvedParallelBenchmarkCount = resolveParallelBenchmarkCountAutomatically(availableProcessorCount);
         } else {
-            resolvedParallelBenchmarkCount = ConfigUtils.resolveThreadPoolSizeScript(
-                    "parallelBenchmarkCount", parallelBenchmarkCount, PARALLEL_BENCHMARK_COUNT_AUTO);
+            resolvedParallelBenchmarkCount = ConfigUtils.resolvePoolSize("parallelBenchmarkCount",
+                    parallelBenchmarkCount, PARALLEL_BENCHMARK_COUNT_AUTO);
         }
         if (resolvedParallelBenchmarkCount < 1) {
             throw new IllegalArgumentException("The parallelBenchmarkCount (" + parallelBenchmarkCount
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
@@ -107,9 +107,7 @@ public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchP
      * Use {@value #ACTIVE_THREAD_COUNT_UNLIMITED} to give it all CPU cores.
      * This is useful if you're handling the CPU consumption on an OS level.
      *
-     * @return null, a number, {@value #ACTIVE_THREAD_COUNT_AUTO}, {@value #ACTIVE_THREAD_COUNT_UNLIMITED}
-     *         or a JavaScript calculation using
-     *         {@value org.optaplanner.core.config.util.ConfigUtils#AVAILABLE_PROCESSOR_COUNT}.
+     * @return null, a number, {@value #ACTIVE_THREAD_COUNT_AUTO} or {@value #ACTIVE_THREAD_COUNT_UNLIMITED}.
      */
     public String getRunnablePartThreadLimit() {
         return runnablePartThreadLimit;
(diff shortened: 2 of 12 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/config/PlannerBenchmarkConfig.java
```diff
@@ -800,27 +799,9 @@ protected void generateSolverBenchmarkConfigNames() {
...
-        if (resolvedParallelBenchmarkCount > availableProcessorCount) {
-            logger.warn("Because the resolvedParallelBenchmarkCount ({}) is higher "
-                    + "than the availableProcessorCount ({}), it is reduced to "
-                    + "availableProcessorCount.", resolvedParallelBenchmarkCount, availableProcessorCount);
-            resolvedParallelBenchmarkCount = availableProcessorCount;
-        }
-        return resolvedParallelBenchmarkCount;
+        return ConfigUtils.resolveThreadPoolSize("parallelBenchmarkCount", parallelBenchmarkCount,
+                () -> resolveParallelBenchmarkCountAutomatically(availableProcessorCount),
+                () -> availableProcessorCount, logger);
```
ge0ffrey, line 805: If PARALLEL_BENCHMARK_COUNT_AUTO isn't used any more, it should be removed too.

But that's still used in javadocs, and replacing it with ConfigUtils#THREAD_COUNT_AUTO makes no sense, because not a THREAD_COUNT. One parallel benchmark or 1 solverManager.parallel may actually use multiple threads. So calling it threadCount is confusing. Also,  resolveThreadPoolSize() is a bad name I guess - **we can rename that to resolvePoolSize()?**

I see 2 options:
1) rename - ConfigUtils.THREAD_COUNT_AUTO to just AUTO. But then every resolveThreadPoolSize() call must have an AUTO option (which is currently the case, so I guess that's fine).
2) Don't generalize "auto", like it is before this PR (doesn't impact the other changes in this PR).

optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
```diff
@@ -45,7 +45,6 @@
 public class PartitionedSearchPhaseConfig extends PhaseConfig<PartitionedSearchPhaseConfig> {
 
     public static final String XML_ELEMENT_NAME = "partitionedSearch";
-    public static final String ACTIVE_THREAD_COUNT_AUTO = "AUTO";
     public static final String ACTIVE_THREAD_COUNT_UNLIMITED = "UNLIMITED";
```
ge0ffrey, line 48: This is why I tend to prefer option 2, as sometimes there are extra values, like UNLIMITED.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -29,29 +27,33 @@
...
+
+import static org.optaplanner.core.impl.domain.common.accessor.MemberAccessorFactory.MemberAccessorType.FIELD_OR_READ_METHOD;
 
 public class ConfigUtils {
 
+    /**
+     * Name of the variable that represents {@link Runtime#availableProcessors()}.
+     */
+    public static final String THREAD_COUNT_AUTO = "AUTO";
+    public static final String AVAILABLE_PROCESSOR_COUNT = "availableProcessorCount";
```
ge0ffrey, line 56: Let's also remove availableProcessorCount:
While "availableProcessorCount / 2" can be usefull (despite no one using it), "availableProcessorCount" is always a bad idea (freezes the computer). It's a loaded gun tied to your leg aimed at your foot. There's no way to miss.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -261,35 +263,31 @@ public static int ceilDivide(int dividend, int divisor) {
...
-                    + "Maybe try running in a normal JVM.");
+    public static int resolveThreadPoolSize(String propertyName, String script, IntSupplier defaultValue,
+            IntSupplier availableProcessors, Logger logger) {
+        int resolved;
+        if (script == null || script.equals(THREAD_COUNT_AUTO)) {
+            resolved = defaultValue.getAsInt();
+        } else {
+            try {
+                resolved = Integer.parseInt(script);
+            } catch (Exception e) {
```
ge0ffrey, line 274: We don't catch Exception or Throwable. Catching Throwable could inadvertadly catch an OutOfMemoryException 98% GC usage. Catching exception is less dangerous, but still undesired. Let's just catch NumberFormatException.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -261,35 +263,31 @@ public static int ceilDivide(int dividend, int divisor) {
...
-        try {
-            scriptResult = scriptEngine.eval(script);
-        } catch (ScriptException e) {
-            throw new IllegalArgumentException("The " + propertyName + " (" + script
-                    + ") is not in magicValues (" + Arrays.toString(magicValues)
-                    + ") and cannot be parsed in " + scriptLanguage
-                    + " with the variables ([" + AVAILABLE_PROCESSOR_COUNT + "]).", e);
+        if (resolved < 1) {
+            throw new IllegalArgumentException("The " + propertyName + " (" + script + ") resolved to (" + resolved +
+                    ") that is lower than 1.");
```
ge0ffrey, line 281: Suggestion: change "that ..." to "that must be at least 1."

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -261,35 +263,31 @@ public static int ceilDivide(int dividend, int divisor) {
...
+            IntSupplier availableProcessors, Logger logger) {
+        int resolved;
+        if (script == null || script.equals(THREAD_COUNT_AUTO)) {
+            resolved = defaultValue.getAsInt();
+        } else {
+            try {
+                resolved = Integer.parseInt(script);
+            } catch (Exception e) {
+                throw new IllegalStateException("The " + propertyName + " (" + script +
+                        ") resolved to neither '" + THREAD_COUNT_AUTO + "' nor a number.");
```
ge0ffrey, line 276: this is lacking UNLIMITED for one of the use cases. This is related to the removal of magicValues parameter.

optaplanner-core/src/main/java/org/optaplanner/core/config/partitionedsearch/PartitionedSearchPhaseConfig.java
ge0ffrey: (But if we find good consistent naming, I don't veto option 1 though.)

triceo:
I have addressed all comments by switching to a much dumber version which reuses nearly no code.
Waiting for build to turn green.

PR #919, opened 2020-09-09
PLANNER-2139 Allow filtering joiner as a second joiner in fromUniquePair()

triceo:


rsynek (approved):
Looks good, thanks for a quick fix!

PR #921, opened 2020-09-10
PLANNER-829 Enforce working solution same class as annotated

triceo:


optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
```diff
@@ -179,25 +175,6 @@ public void exceptionInConsumer() {
         assertThat(solverJob1.getSolverStatus()).isEqualTo(NOT_SOLVING);
     }
 
-    @Test
-    @Timeout(60)
-    public void solveGenerics() throws ExecutionException, InterruptedException {
```
triceo, line 184: Had to remove this, as it was testing precisely the situation that we've now made impossible.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -172,9 +172,20 @@ public SupplyManager getSupplyManager() {
     // Complex methods
     // ************************************************************************
 
+    private void assertWorkingSolutionClassSameAsAnnotatedClass(Solution_ workingSolution) {
+        Class<Solution_> actualSolutionClass = (Class<Solution_>) Objects.requireNonNull(workingSolution).getClass();
+        Class<Solution_> expectedSolutionClass = getSolutionDescriptor().getSolutionClass();
+        if (!actualSolutionClass.equals(expectedSolutionClass)) {
```
ge0ffrey, line 178: The workingSolution is also allowed to be a subclass of the annotated class (but not the other way around).
Use `Class.isAssignableFrom`.

**Why doesn't TestdataUnannotatedExtendedSolution trigger this fail fast?**

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -172,9 +172,20 @@ public SupplyManager getSupplyManager() {
...
+            throw new IllegalStateException("Working solution class (" + actualSolutionClass.getCanonicalName() +
+                    ") different from " + PlanningSolution.class + "-annotated class (" +
```
ge0ffrey, line 180: style: other error messages output " from the @PlanningSolution annotated class",
instead of " from PlanningSolution-annotated class".

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -172,9 +172,20 @@ public SupplyManager getSupplyManager() {
     // Complex methods
     // ************************************************************************
 
+    private void assertWorkingSolutionClassSameAsAnnotatedClass(Solution_ workingSolution) {
+        Class<Solution_> actualSolutionClass = (Class<Solution_>) Objects.requireNonNull(workingSolution).getClass();
```
ge0ffrey, line 176: nitpick: I'd give the requireNonNull() it's own line, without capturing the return value - that reads better. I'd keep it in the setWorkingSolution() for the stacktrace if it fails and because that assert belongs there more I'd argue.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
ge0ffrey: We need to keep this, this is valid usages. Even though our polymorphism epic has some lacking feature gaps and testings for those - and it's pushed on the backlog - we don't want to reduce our support for it.

Some examples to think about, presume CatSolution extends AnimalSolution. BeerSolution does not.
- If AnimalSolution has the @PlanningSolution annotation, then it should fail fast on BeerSolution, but not AnimalSolution nor CatSolution.
- If CatSolution has the @PlanningSolution annotation, then it should fail fast on BeerSolution and AnimalSolution, but not CatSolution.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
triceo: It does, which is why the test was removed.

triceo:
829 speaks of A extends B, and that it should not be allowed.
Because that would mean annotated stuff from A would go unnoticed.

optaplanner-core/src/test/java/org/optaplanner/core/api/solver/SolverManagerTest.java
ge0ffrey: PLANNER-829 is for when `@PlanningSolution Cat extends @PlanningSolution Animal` and `Cat` is solved with `Solver<Animal>`.
For TestdataUnannotatedExtendedSolution, it's different though, which shouldn't fail fast:
it's when `Cat extends @PlanningSolution Animal` and `Cat` is solved with `Solver<Animal>`, but Cat doesn't have a `@PlanningSolution` annotation.

ge0ffrey: That's a valid use case. Especially if hibernate decided to CGLib proxy the Solution class.

ge0ffrey: Let's discuss on bluejeans - this stuff is complicated and 2 heads are better than one.

triceo:
With this new implementation, I think that it's not worth the performance hit it'd cause.
I suggest we close the issue.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -172,9 +175,51 @@ public SupplyManager getSupplyManager() {
...
+            Class<?> actualFirstPlanningSolutionAnnotatedSuperClass = firstPlanningSolutionAnnotatedSuperClass.get();
+            if (!actualFirstPlanningSolutionAnnotatedSuperClass.equals(expectedSolutionClass)) {
+                throw new IllegalStateException("Working solution class (" + actualSolutionClass.getCanonicalName() +
+                        ") has a super class (" + actualFirstPlanningSolutionAnnotatedSuperClass.getCanonicalName() +
+                        ") that is @" + PlanningSolution.class + " annotated and different from the planning solution" +
+                        " (" + expectedSolutionClass.getCanonicalName() + ") for which the solver was created.");
+            }
+        } else {
+            // Superclasses are OK, but we need to check implemented interfaces as well.
+            Class<?> firstPlanningSolutionAnnotatedInterface = ClassUtils.getAllInterfaces(actualSolutionClass)
```
ge0ffrey, line 204: I agree, the solution is worse than the problem.

ge0ffrey:
@triceo I agree with your assessment. "If it looks bad now, just wait until we've solved it" :)
The perf cost is too expensive and I can't think of any good way to avoid it easily.
Closing as suggested.

PR #923, opened 2020-09-11
PLANNER-1391 Degenerify Score<?> to Score

ge0ffrey:
* This change affects 4 lines in optaplanner-examples.
* **It fixes 179 raw usages of Score, reducing our java warnings count**.
```
Score s = HardSoftScore.ofSoft(-1);
s.add(HardMediumSoftScore.ofSoft(-20); // Runtime failure (ClassCastExceptions) before+after
```
* It _fixes_ 324 usages of `Score<?>`. Note that `?` is like a parameter or field that is typed `Object`, it smells.
```
Score<?> s = HardSoftScore.ofSoft(-1);
s.add(HardMediumSoftScore.ofSoft(-20); // Runtime failure (ClassCastExceptions) before
```
* As for compile time errors that have now been potentially turned into runtime ClassCastExceptions:
there are _only_ 10 occurrences of "<Score_ extends ", both as class as method types,
for which exactly 1 case the compile time error could become a runtime error - in theory at least.
It's in the benchmarker:
```
public class ScoreDifferencePercentage {

    public static <Score_ extends Score> ScoreDifferencePercentage calculateScoreDifferencePercentage(
            Score_ baseScore, Score_ valueScore) {
```

And most importantly, it solves this blocker for generifying ConstraintMatch, ConstraintMatchTotal and Indictment:
* `List<ConstraintMatch<Score>>` mixes non-generics and `List<ConstraintMatch<Score<?>>`
works only partially: you can't assign it something that returns `List<ConstraintMatch<Score>`.


triceo (changes requested):
This change further complicates backporting to 7.x and would only be acceptable if it brought substantial benefits.

Unfortunately, the only tangible benefit this brings, the `List<...>` thing, might as well be fixed by casting before return. Since this "benefit" is pretty much exclusive to our internal code and users would barely ever see it, I say it doesn't outweigh the side effects of this change.

ge0ffrey:
Closing this PR because we decided to reject this change. New Pr coming soon.

PR #926, opened 2020-09-14
use rhino version defined in kie-parent

mariofusco:


triceo (approved):
CI is green, feel free to merge.

PR #927, opened 2020-09-14
PLANNER-2151 Add generic score type to various APIs

triceo:
This is a very quick PoC. I only did the MVP to make sure the code still compiles.
I can **make it nicer and even more type-safe** if I go deeper and add `<Score_>` on other types as well.

There is **no change to public interfaces**, except where requested by the JIRA.
According to my IDEA, in the files I touched, I have **removed more raw type warnings than I have caused**.

One test in **Quarkus integration is failing**, but that would've failed regardless - there's something wrong with injection when `ScoreManager` has two generic types, I don't understand the problem (yet).

**600 of the lines changed is actually Revapi.** I have added ~20 lines of code.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractScore.java
```diff
@@ -31,7 +31,7 @@
  * @see Score
  * @see HardSoftScore
  */
-public abstract class AbstractScore<S extends Score> implements Score<S>,
+public abstract class AbstractScore<S extends Score<S>> implements Score<S>,
```
triceo, line 34: Technically, this is not necessary, but I did it anyway. It's the right thing to do.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/DefaultScoreExplanation.java
```diff
@@ -25,16 +25,18 @@
 import org.optaplanner.core.api.score.constraint.ConstraintMatchTotal;
 import org.optaplanner.core.api.score.constraint.Indictment;
 
-public final class DefaultScoreExplanation<Solution_> implements ScoreExplanation<Solution_> {
+public final class DefaultScoreExplanation<Solution_, Score_ extends Score<Score_>>
```
triceo, line 28: This is a nice example of where the generification actually helps.

We are now bound to `<Score_>` and this makes it crystal clear that the entire code is supposed to work with instances of the same type.

Generalization is good. More generalization is better.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -77,7 +77,7 @@
      * @throws IllegalStateException if {@link #isConstraintMatchEnabled()} returns false
      * @see #getIndictmentMap()
      */
-    Map<String, ConstraintMatchTotal> getConstraintMatchTotalMap();
+    <Score_ extends Score<Score_>> Map<String, ConstraintMatchTotal<Score_>> getConstraintMatchTotalMap();
```
triceo, line 80: IMO we really *should* add `<Score_>` on `InnerScoreDirector`. I could then start cleaning up *a lot more* of the internals.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultSingleConstraintAssertion.java
```diff
@@ -112,16 +113,16 @@ private void assertImpact(ScoreImpactType scoreImpactType, Number matchWeightTot
     }
 
     private Number deduceImpact() {
```
triceo, line 115: Another example of a nicer code, strongly coupled by the same generic type.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/InnerScoreDirector.java
rsynek: +1 as it would reduce the number of duplicate `Score_` type definitions on methods.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirector.java
```diff
@@ -82,7 +82,7 @@ public boolean isConstraintMatchEnabled() {
     }
 
     @Override
-    public Map<String, ConstraintMatchTotal> getConstraintMatchTotalMap() {
+    public <Score_ extends Score<Score_>> Map<String, ConstraintMatchTotal<Score_>> getConstraintMatchTotalMap() {
```
rsynek, line 85: This is probably the only thing that makes me a little bit nervous: the type definition is duplicated over multiple methods, effectively being a different type each time.

triceo: True. I can go deeper and add `<Score_>` on more and more classes, but this is the MVP that we agreed I'd do.

rsynek (commented):
Nothing really shocking for me, this PR rather helps than hurts. Let's just avoid (or at least reduce) duplicate definitions of the generic type definition, where possible.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/DefaultScoreExplanation.java
ge0ffrey: I agree that here it's a good thing, no doubt about that :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/constraint/DefaultConstraintMatchTotal.java
```diff
@@ -29,26 +29,26 @@
...
 
-public final class DefaultConstraintMatchTotal implements ConstraintMatchTotal,
-        Comparable<DefaultConstraintMatchTotal> {
+public final class DefaultConstraintMatchTotal<Score_ extends Score<Score_>> implements ConstraintMatchTotal<Score_>,
+        Comparable<DefaultConstraintMatchTotal<Score_>> {
 
-    private static final Comparator<DefaultConstraintMatchTotal> COMPARATOR =
-            comparing(DefaultConstraintMatchTotal::getConstraintPackage)
+    private static final Comparator<DefaultConstraintMatchTotal<?>> COMPARATOR =
+            comparing((DefaultConstraintMatchTotal<?> constraintMatchTotal) -> constraintMatchTotal.constraintPackage)
```
ge0ffrey, line 36: This one drove me mad in my version of this. This is a good approach!

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -379,7 +379,9 @@ public static String explainScore(Score workingScore,
 
     @Override
     public String explainScore() {
-        return explainScore(calculateScore(), getConstraintMatchTotalMap().values(), getIndictmentMap().values());
+        Collection constraintMatchTotalCollection = getConstraintMatchTotalMap().values();
+        Collection indictmentCollection = getIndictmentMap().values();
+        return explainScore(calculateScore(), constraintMatchTotalCollection, indictmentCollection);
```
ge0ffrey, line 384: This is the raw typing trick as discussed. Here we have raw values of `Collection`

triceo: Correct. And i had to use it on all 2 places in entire `core`. :-)
If `InnerScoreDirector` had `<Score_>`, it would go away.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/InnerScoreDirector.java
ge0ffrey: I think it would be a good experiment, but it might cause more harm than good. In any case, only the experiment will (dis)prove it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
triceo: But now that I think about it, there is another way. Stay tuned.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirector.java
```diff
@@ -112,25 +112,25 @@ public boolean isConstraintMatchEnabled() {
...
+    public <Score_ extends Score<Score_>> Map<Object, Indictment<Score_>> getIndictmentMap() {
         if (!isConstraintMatchEnabled()) {
             throw new IllegalStateException("When constraintMatchEnabled (" + isConstraintMatchEnabled()
                     + ") is disabled in the constructor, this method should not be called.");
         }
         Map<Object, Indictment> incrementalIndictmentMap =
                 ((ConstraintMatchAwareIncrementalScoreCalculator<Solution_>) incrementalScoreCalculator).getIndictmentMap();
         if (incrementalIndictmentMap != null) {
-            return incrementalIndictmentMap;
+            return (Map) incrementalIndictmentMap;
```
ge0ffrey, line 123: Raw typing trick

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirector.java
ge0ffrey: Indeed, a Score_ on InnerScoreDirector can fix that, but the side-effects of that can be severe.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/persistence/AbstractXlsxSolutionFileIO.java
```diff
@@ -378,8 +378,8 @@ public AbstractXlsxWriter(Solution_ solution, String solverConfigResource) {
             try (InnerScoreDirector<Solution_> scoreDirector = scoreDirectorFactory.buildScoreDirector()) {
                 scoreDirector.setWorkingSolution(solution);
                 score = scoreDirector.calculateScore();
-                constraintMatchTotalsMap = scoreDirector.getConstraintMatchTotalMap();
-                indictmentMap = scoreDirector.getIndictmentMap();
+                constraintMatchTotalsMap = (Map) scoreDirector.getConstraintMatchTotalMap();
+                indictmentMap = (Map) scoreDirector.getIndictmentMap();
```
ge0ffrey, line 382: Not sure why the raw type trick is here

optaplanner-examples/src/main/java/org/optaplanner/examples/common/swingui/SolutionPanel.java
```diff
@@ -196,7 +197,8 @@ public String determinePlanningEntityTooltip(Object planningEntity) {
             return "<html>No indictment</html>";
         }
         StringBuilder s = new StringBuilder("<html>Indictment: ").append(indictment.getScore().toShortString());
-        for (ConstraintMatch constraintMatch : indictment.getConstraintMatchSet()) {
+        Set<ConstraintMatch> constraintMatchSet = indictment.getConstraintMatchSet();
+        for (ConstraintMatch constraintMatch : constraintMatchSet) {
```
ge0ffrey, line 201: We can probably use <?> here to avoid the raw types

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/persistence/util/VehicleRoutingDistanceTypeComparison.java
```diff
@@ -36,7 +36,7 @@
 
 public class VehicleRoutingDistanceTypeComparison extends LoggingMain {
 
-    private final ScoreManager<VehicleRoutingSolution> scoreManager;
+    private final ScoreManager<VehicleRoutingSolution, ?> scoreManager;
```
ge0ffrey, line 39: We want to use the actual score types in these cases, which can simply be looked up by clicking on VehicleRoutingSolution

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultMultiConstraintVerification.java
```diff
@@ -51,7 +51,7 @@ public final DefaultMultiConstraintAssertion given(Object... facts) {
     public final DefaultMultiConstraintAssertion givenSolution(Solution_ solution) {
         try (InnerScoreDirector<Solution_> scoreDirector = scoreDirectorFactory.buildScoreDirector(true, true)) {
             scoreDirector.setWorkingSolution(Objects.requireNonNull(solution));
-            return new DefaultMultiConstraintAssertion<>(constraintProvider, scoreDirector.calculateScore(),
+            return new DefaultMultiConstraintAssertion(constraintProvider, scoreDirector.calculateScore(),
```
ge0ffrey, line 54: Any reason we're using less generics and more raw types here?

triceo: I didn't want to touch the public API, which I'd have to do to add the `<Score_>` type there.
I could do that, but again - MVP.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
ge0ffrey: Meanwhile, I 've found a way to avoid the raw usages:

```
    @Override
    public <Score_ extends Score<Score_>> String explainScore() {
        Map<String, ConstraintMatchTotal<Score_>> constraintMatchTotalMap = getConstraintMatchTotalMap();
        Map<Object, Indictment<Score_>> indictmentMap = getIndictmentMap();
        return explainScore(calculateScore(), constraintMatchTotalMap.values(), indictmentMap.values());
    }
```

The trick is assigning the map to a local variable, typing it correctly, before calling values()

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirector.java
```diff
@@ -112,25 +112,25 @@ public boolean isConstraintMatchEnabled() {
...
 
     @Override
-    public Map<Object, Indictment> getIndictmentMap() {
+    public <Score_ extends Score<Score_>> Map<Object, Indictment<Score_>> getIndictmentMap() {
         if (!isConstraintMatchEnabled()) {
             throw new IllegalStateException("When constraintMatchEnabled (" + isConstraintMatchEnabled()
                     + ") is disabled in the constructor, this method should not be called.");
         }
         Map<Object, Indictment> incrementalIndictmentMap =
                 ((ConstraintMatchAwareIncrementalScoreCalculator<Solution_>) incrementalScoreCalculator).getIndictmentMap();
```
ge0ffrey, line 121: If we're BC breaking ScoreManager for justification API gain, we should also do it to ConstraintMatchAwareIncrementalScoreCalculator, which one one uses and isn't part of the api package. So I am doing that now :)

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultMultiConstraintVerification.java
ge0ffrey: Good point - I managed to limit adding the Score_ param to the impl classes only, but we probably want to make MultiConstraintAssertion Score_ sensitive? Different issue.

ge0ffrey:
Closing in favor of https://github.com/kiegroup/optaplanner/pull/928 which includes the commit(s) of this PR.

PR #928, opened 2020-09-16
Add generic type Score to SolverManager, SE, CMT, CM and Indictment

ge0ffrey:
Merge these together:

* https://github.com/kiegroup/optaplanner/pull/928
* https://github.com/kiegroup/optaplanner-quickstarts/pull/3
* https://github.com/kiegroup/optaplanner-website/pull/279


ge0ffrey:
Once this is merged, we can do a separate experiment to make InnerScoreDirectory have a Score_ type.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractBendableScore.java
```diff
@@ -28,13 +28,13 @@
...
  */
-public abstract class AbstractBendableScore<S extends Score<S>> extends AbstractScore<S> {
+public abstract class AbstractBendableScore<Score_ extends AbstractBendableScore<Score_>> extends AbstractScore<Score_> {
 
     protected static final String HARD_LABEL = "hard";
     protected static final String SOFT_LABEL = "soft";
     protected static final String[] LEVEL_SUFFIXES = new String[] { HARD_LABEL, SOFT_LABEL };
 
-    protected static String[][] parseBendableScoreTokens(Class<? extends Score> scoreClass, String scoreString) {
+    protected static String[][] parseBendableScoreTokens(Class<? extends Score<?>> scoreClass, String scoreString) {
```
triceo, line 37: I'm not exactly sure of what this method does, but shouldn't the first argument rather be `Class<? extends AbstractBendableScore<?>>`?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/AbstractScore.java
```diff
@@ -27,16 +27,16 @@
...
  * @see HardSoftScore
  */
-public abstract class AbstractScore<S extends Score> implements Score<S>,
+public abstract class AbstractScore<Score_ extends AbstractScore<Score_>> implements Score<Score_>,
         Serializable {
 
     protected static final String INIT_LABEL = "init";
 
-    protected static String[] parseScoreTokens(Class<? extends Score> scoreClass,
+    protected static String[] parseScoreTokens(Class<? extends Score<?>> scoreClass,
```
triceo, line 39: `Class<? extends AbstractScore<?>>`?
I'm thinking if we should just make an instance method, and use `Score_` instead.
(Likewise for other similar methods on this class.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -679,8 +683,8 @@ private void determineGlobalShadowOrder() {
     /**
      * @return the {@link Class} of {@link PlanningScore}
      */
-    public Class<? extends Score> extractScoreClass() {
-        return (Class<? extends Score>) scoreMemberAccessor.getType();
+    public Class<? extends Score<?>> extractScoreClass() {
```
triceo, line 686: Now wouldn't it be nice if `SolutionDescriptor` also had a `Score_`? But maybe that'd just go too far.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/supply/SupplyManager.java
```diff
@@ -22,13 +22,15 @@
...
+     * If the {@link org.optaplanner.core.impl.domain.variable.supply.Supply} doesn't exist yet (as part of the domain model or
+     * externalized), it creates and attaches it.
      *
      * @param demand never null
-     * @param <S> Subclass of {@link Supply}
+     * @param <Supply> Subclass of {@link org.optaplanner.core.impl.domain.variable.supply.Supply}
      * @return never null
      */
-    <S extends Supply> S demand(Demand<S> demand);
+    <Supply extends org.optaplanner.core.impl.domain.variable.supply.Supply> Supply demand(Demand<Supply> demand);
```
triceo, line 34: First of all, `<Supply_ extends Supply> Supply_` should work, right?
Second, wouldn't it make sense to do `SupplyManager<Supply_>` directly?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/ScoreUtils.java
```diff
@@ -56,6 +57,8 @@ public static Score parseScore(Class<? extends Score> scoreClass, String scoreSt
             return HardMediumSoftScore.parseScore(scoreString);
         } else if (HardMediumSoftLongScore.class.equals(scoreClass)) {
             return HardMediumSoftLongScore.parseScore(scoreString);
+        } else if (HardMediumSoftBigDecimalScore.class.equals(scoreClass)) {
+            return HardMediumSoftBigDecimalScore.parseScore(scoreString);
```
triceo, line 61: This was missing all the time?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -378,8 +378,12 @@ public static String explainScore(Score workingScore,
     }
 
     @Override
-    public String explainScore() {
-        return explainScore(calculateScore(), getConstraintMatchTotalMap().values(), getIndictmentMap().values());
+    public <Score_ extends Score<Score_>> String explainScore() {
```
triceo, line 381: Now wouldn't it be lovely if `AbstractScoreDirector` had a generic <Score_> type? :-) Inner and Abstract score directors, and their subclasses, are officially private code in 8, so adding more generics there does not affect users.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -76,8 +76,9 @@
      *         (to create one, use {@link ConstraintMatchTotal#composeConstraintId(String, String)}).
      * @throws IllegalStateException if {@link #isConstraintMatchEnabled()} returns false
      * @see #getIndictmentMap()
+     * @param <Score_> the actual score type
      */
-    Map<String, ConstraintMatchTotal> getConstraintMatchTotalMap();
+    <Score_ extends Score<Score_>> Map<String, ConstraintMatchTotal<Score_>> getConstraintMatchTotalMap();
```
triceo, line 81: IMO `InnerScoreDirector` is really screaming for a `<Score_>` generic type.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirector.java
```diff
@@ -99,38 +99,41 @@ public boolean isConstraintMatchEnabled() {
...
+                        .getIndictmentMap();
         if (incrementalIndictmentMap != null) {
             return incrementalIndictmentMap;
         }
-        Map<Object, Indictment> indictmentMap = new LinkedHashMap<>(); // TODO use entitySize
-        Score zeroScore = getScoreDefinition().getZeroScore();
-        for (ConstraintMatchTotal constraintMatchTotal : getConstraintMatchTotalMap().values()) {
-            for (ConstraintMatch constraintMatch : constraintMatchTotal.getConstraintMatchSet()) {
+        Map<Object, Indictment<Score_>> indictmentMap = new LinkedHashMap<>(); // TODO use entitySize
+        Score_ zeroScore = (Score_) getScoreDefinition().getZeroScore();
```
triceo, line 127: The cast should not be necessary, as `ScoreDefinition` also has `<Score_>`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSession.java
```diff
@@ -40,15 +40,17 @@
      * As defined by {@link InnerScoreDirector#getConstraintMatchTotalMap()}.
      *
      * @return never null
+     * @param <Score_> the {@link Score} type
      */
-    Map<String, ConstraintMatchTotal> getConstraintMatchTotalMap();
+    <Score_ extends Score<Score_>> Map<String, ConstraintMatchTotal<Score_>> getConstraintMatchTotalMap();
```
triceo, line 45: Considering how often these two methods (constraint match total map, indictment map) repeat in various unrelated interfaces, perhaps these should be extracted into their own interface, which the other interfaces would then extend?

triceo (approved):
Love it!
Added some suggestions and observations.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
ge0ffrey: I fear this will go to far. That's a rabbit hole that will lead very deep.
Separate PR/experiment in any case. And let's first figure out if InnerScoreDirector should be score generified...

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/supply/SupplyManager.java
ge0ffrey: 1) Yes, was a mistake
2) No: SupplyManager handles multiple Supple types.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/ScoreUtils.java
ge0ffrey: Yes. Scary :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
ge0ffrey: Yes, I am starting to think so too. Make InnerScoreDirector (not ScoreDirector!) Score generic. Separate PR in any case. Want to create a jira for it?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/incremental/IncrementalScoreDirector.java
ge0ffrey: But as long as AbstractScoreDirector doesn't have a generic type, it ScoreDefinition isn't typed either. So can't fix in this PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSession.java
ge0ffrey: I think this is premature optimizaton, that if we clean up some architecture, will become much less.

They repeat 3 times:
- ScoreExplenation - which is public API, so the extra interface inheritence will complicate it for users
- ScoreDirector
- ConstraintSession

I believe ConstraintSession is a needless indirection (that maybe even has a perf impact).
Same for ConstraintSessionFactory and ConstraintStreamScoreDirectorFactory.
We could maybe replace it by [Drools/Bavet]ConstraintStreamScoreDirector(Factory) extends AbstractConstraintStreamScoreDirector(Factory).
That experiment needs to be looked into first.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
triceo: https://issues.redhat.com/browse/PLANNER-2164

ge0ffrey:
Feedback applied, waiting for CI now.

PR #932, opened 2020-09-17
Improve CI performance by caching local Maven repo

triceo:
Implemented Github Actions' Maven caching mechanism. The goals:

- Maven repos are only downloaded incrementally, not the entire world every single time.
- This reduces the likelihood of Maven Central error failing the build.
- And should also speed up CI significantly.

Results once cached:

- Main CI went from 20 min to 15 min, **25 % run time improvement**.
- Persistence CI went from 7.5 min to 3.25 min, **50 % run time improvement**.

There were **no Maven Central-related failures**.

### DIFF
--- a/.github/workflows/jpa_integration_ci.yml
+++ b/.github/workflows/jpa_integration_ci.yml
@@ -47,6 +47,14 @@ jobs:
         with:
           java-version: 11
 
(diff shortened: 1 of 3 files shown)


rsynek (approved):
Very nice improvement, thank you.

PR #793, opened 2020-05-28
PLANNER-1873 Refactor capacity constraints with shadow variables

triceo:
Got further than before. Now it actually runs for a couple seconds, before it throws.

PR #691, opened 2020-02-16
[PLANNER-1821] Convert CoachShuttleGathering example to constraint streams

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-1821

PR #938, opened 2020-09-18
PLANNER-1821 & PLANNER-1873: Add Constraint Provider for Coach Shuttle Gathering

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-1873
https://issues.redhat.com/browse/PLANNER-1821

### Checklist

<details>
<summary>
</summary>

</details>

Replaces https://github.com/kiegroup/optaplanner/pull/691 and https://github.com/kiegroup/optaplanner/pull/793

Coach Capacity was a fun one; it would be easy if we had an accumlate that works on empty streams, but since we don't, I need to implement it as three constraints: One when no streams are empty, one that handles when coach is overfilled without shuttle, and one to remove double counting.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Shuttle.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Shuttle.java
@@ -56,7 +56,8 @@ public class Shuttle extends Bus {
 
     @Override
     @CustomShadowVariable(variableListenerClass = ShuttlePassengerCountTotalUpdatingVariableListener.class,
-            sources = { @PlanningVariableReference(entityClass = BusStop.class, variableName = "bus") })
+            sources = { @PlanningVariableReference(entityClass = BusStop.class, variableName = "bus"),
+                    @PlanningVariableReference(entityClass = Shuttle.class, variableName = "destination") })
     public Integer getPassengerQuantityTotal() {
         return super.getPassengerQuantityTotal();
     }
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Shuttle.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Shuttle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,9 +17,12 @@
 package org.optaplanner.examples.coachshuttlegathering.domain;
 
 import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.CustomShadowVariable;
 import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.api.domain.variable.PlanningVariableReference;
 import org.optaplanner.examples.coachshuttlegathering.domain.location.RoadLocation;
 import org.optaplanner.examples.coachshuttlegathering.domain.solver.DepotAngleBusStopDifficultyWeightFactory;
+import org.optaplanner.examples.coachshuttlegathering.domain.solver.ShuttlePassengerCountTotalUpdatingVariableListener;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 
@@ -49,6 +52,13 @@ public class Shuttle extends Bus {
 
     public void setDestination(StopOrHub destination) {
         this.destination = destination;
+    }
+
+    @Override
+    @CustomShadowVariable(variableListenerClass = ShuttlePassengerCountTotalUpdatingVariableListener.class,
+            sources = { @PlanningVariableReference(entityClass = BusStop.class, variableName = "bus") })
+    public Integer getPassengerQuantityTotal() {
+        return super.getPassengerQuantityTotal();
     }
 
     // ************************************************************************
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/BusPassengerCountTotalUpdatingVariableListener.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/BusPassengerCountTotalUpdatingVariableListener.java
@@ -69,9 +69,6 @@ public abstract class BusPassengerCountTotalUpdatingVariableListener implements 
             return;
         }
         int difference = increase ? busStop.getPassengerQuantity() : -busStop.getPassengerQuantity();
-        if (difference == 0) {
-            return;
-        }
         scoreDirector.beforeVariableChanged(bus, "passengerQuantityTotal");
         bus.setPassengerQuantityTotal(bus.getPassengerQuantityTotal() + difference);
         scoreDirector.afterVariableChanged(bus, "passengerQuantityTotal");
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/BusPassengerCountTotalUpdatingVariableListener.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/BusPassengerCountTotalUpdatingVariableListener.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.coachshuttlegathering.domain.solver;
+
+import org.optaplanner.core.api.score.director.ScoreDirector;
+import org.optaplanner.core.impl.domain.variable.listener.VariableListener;
+import org.optaplanner.examples.coachshuttlegathering.domain.Bus;
+import org.optaplanner.examples.coachshuttlegathering.domain.BusStop;
+import org.optaplanner.examples.coachshuttlegathering.domain.CoachShuttleGatheringSolution;
+
+public abstract class BusPassengerCountTotalUpdatingVariableListener implements VariableListener<Object> {
+
+    @Override
+    public void beforeEntityAdded(ScoreDirector scoreDirector, Object busStop) {
+        // Do nothing
+    }
+
+    @Override
+    public void afterEntityAdded(ScoreDirector scoreDirector, Object entity) {
+        if (entity instanceof BusStop) {
+            updateBusPassengerCount(scoreDirector, (BusStop) entity, true);
+        }
+    }
+
+    @Override
+    public void beforeVariableChanged(ScoreDirector scoreDirector, Object entity) {
+        if (entity instanceof BusStop) {
+            updateBusPassengerCount(scoreDirector, (BusStop) entity, false);
+        }
+    }
+
+    @Override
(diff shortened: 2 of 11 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/CoachPassengerCountTotalUpdatingVariableListener.java
```diff
@@ -0,0 +1,84 @@
...
+        scoreDirector.beforeVariableChanged(bus, "passengerQuantityTotal");
+        bus.setPassengerQuantityTotal(bus.getPassengerQuantityTotal() + difference);
+        scoreDirector.afterVariableChanged(bus, "passengerQuantityTotal");
+        if (bus.getPassengerQuantityTotal() < 0) {
+            throw new IllegalStateException("Passenger quantity in " + bus + " got under zero here.");
+        }
+    }
+
+    private static void adjust(ScoreDirector<CoachShuttleGatheringSolution> scoreDirector, BusStop busStop,
+            boolean increase) {
```
ge0ffrey, line 41: Please follow the naming of the other VariableListeners for consistency across examples, such as ArrivalTimeUpdatingVariableListener. These 2 methods would be "update..." and be below the before/after calls.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/CoachPassengerCountTotalUpdatingVariableListener.java
```diff
@@ -0,0 +1,84 @@
...
+        adjustBus(scoreDirector, bus, increase ? busStop.getPassengerQuantity() : -busStop.getPassengerQuantity());
+    }
+
+    private static void increase(ScoreDirector<CoachShuttleGatheringSolution> scoreDirector, BusStop busStop) {
+        adjust(scoreDirector, busStop, true);
+    }
+
+    private static void decrease(ScoreDirector<CoachShuttleGatheringSolution> scoreDirector, BusStop busStop) {
+        adjust(scoreDirector, busStop, false);
+    }
```
ge0ffrey, line 55: method naming: increase/decrease what?
Also, I wouldn't extract them into separate methods, just call true/false directly from before/after.

ge0ffrey (approved):
Looks good. See some naming/method order suggestions, see comments.

Before merging, please validate that this is a pure refactor: check if the current master (DRL without this new shadow var) and this branch (CS with this new shadow var) deliver the same line in DEBUG logging on Local Search step 100 (except for the time spent of course). Chris, please try that out and report your findings here. If they don't match, let's figure out why.

triceo (changes requested):
After Geoffrey's comments are addressed and the solution is proven in the way that he suggests, changes need to be made to the solver config to restore it to its original form.

Christopher-Chianelli:
@ge0ffrey the 100 LS line was identical. Performance.... kinda bad
```
Master using old Drools
12:52:26.696 [l-3-thread-1] DEBUG     LS step (100), time spent (498), score (0hard/-400050soft),     best score (0hard/-384040soft), accepted/selected move count (1/1), picked move (S5 {B2 -> B1}).

New Branch using new Drools w/ Shadow Variables
13:14:19.973 [l-3-thread-1] DEBUG     LS step (100), time spent (2384), score (0hard/-400050soft),     best score (0hard/-384040soft), accepted/selected move count (1/1), picked move (S5 {B2 -> B1}).

New Branch using CP + Shadow Variables
13:08:35.521 [l-3-thread-1] DEBUG     LS step (100), time spent (3169), score (0hard/-400050soft),     best score (0hard/-384040soft), accepted/selected move count (1/1), picked move (S5 {B2 -> B1}).
```

I tested the new Drools and ConstraintProvider against the EasyScoreCalculator on full assert; no issues came up.
2384 / 498 = 4.787..., so almost 5 time slower using new Drools
3169 / 498 = 6.363..., so over 6 times slower using ConstraintProvider.

My guess: the variable update listener eats a lot of time and is called a lot (although it only increments/decrements a value)

triceo:
Actually, 30 % down for CS-D ain't that bad, considering that this is before the new groupBy implementation.

Did you see this results with `FULL_ASSERT` enabled? I'd argue that this would not really be relevant, as constraint matches add a lot of that wouldn't otherwise be used. (Although I could also argue that this would make relative results better, not worse.)

Christopher-Chianelli:
@triceo result is with FULL_ASSERT enable.... probably should rerun

Christopher-Chianelli:
New drools w/o full assert
```
14:10:33.371 [l-3-thread-1] DEBUG     LS step (100), time spent (495), score (0hard/-400050soft),     best score (0hard/-384040soft), accepted/selected move count (1/1), picked move (S5 {B2 -> B1}).
```
About same speed (compared to old drools)

Constraint Provider w/o full assert
```
14:12:23.113 [l-3-thread-1] DEBUG     LS step (100), time spent (711), score (0hard/-400050soft),     best score (0hard/-384040soft), accepted/selected move count (1/1), picked move (S5 {B2 -> B1}).
```
So about 43% performance loss (compared to old drools)

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Shuttle.java
```diff
@@ -51,6 +54,14 @@ public void setDestination(StopOrHub destination) {
         this.destination = destination;
     }
 
+    @Override
+    @CustomShadowVariable(variableListenerClass = ShuttlePassengerCountTotalUpdatingVariableListener.class,
+            sources = { @PlanningVariableReference(entityClass = BusStop.class, variableName = "bus"),
+                    @PlanningVariableReference(entityClass = Shuttle.class, variableName = "destination") })
```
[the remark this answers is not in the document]
Christopher-Chianelli: Can't do - leads to double counting it seems (despite code being exactly the same); was able to put it in an abstract class and have two variable listeners extend it (the abstract method is literally an instanceof check)

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/BusPassengerCountTotalUpdatingVariableListener.java
```diff
@@ -0,0 +1,81 @@
...
+    private void updateBusPassengerCount(ScoreDirector<CoachShuttleGatheringSolution> scoreDirector, BusStop busStop,
+            boolean increase) {
+        Bus bus = busStop.getBus();
+        if (!isCorrectBusInstance(bus)) {
+            return;
+        }
+        int difference = increase ? busStop.getPassengerQuantity() : -busStop.getPassengerQuantity();
+        if (difference == 0) {
+            return;
+        }
```
triceo, line 74: We can probably improve this:

    int quantity = busStop.getPassengerQuantity();
    if (quantity == 0) {
        return;
    }
    int difference = increase ? quantity : -quantity;

If this code is called often, this will bring a small perf improvement.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/solver/CoachShuttleGatheringConstraintProvider.java
```diff
@@ -0,0 +1,142 @@
...
+                distanceFromPrevious(constraintFactory),
+                distanceBusStopToBusDestination(constraintFactory),
+                distanceCoachDirectlyToDestination(constraintFactory)
+        };
+    }
+
+    Constraint coachStopLimit(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(Coach.class)
+                .join(BusStop.class, Joiners.equal(coach -> coach, BusStop::getBus))
+                .groupBy((coach, busStop) -> coach, ConstraintCollectors.countBi())
```
triceo, line 40: Let's `import static ...ConstraintCollectors...`
AFAIK the other constraint providers do it too.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/solver/CoachShuttleGatheringConstraintProvider.java
```diff
@@ -0,0 +1,142 @@
...
+                .filter((coach, stopCount) -> stopCount > coach.getStopLimit())
+                .penalizeLong(CONSTRAINT_PACKAGE, "coachStopLimit", HardSoftLongScore.ONE_HARD,
```
triceo, line 42: Not sure what the convention here is, but maybe `import static ... HardSoftLongScore...` as well?

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/solver/CoachShuttleGatheringConstraintProvider.java
```diff
@@ -0,0 +1,142 @@
...
+                        (coach, shuttlePassengerQuantityTotal) -> (coach.getPassengerQuantityTotal()
+                                + shuttlePassengerQuantityTotal - coach.getCapacity()) * 1000L);
+    }
+
+    // Correct the double counting
+    // Explanation: groupBy is like accumlate, but it doesn't trigger on empty streams.
+    // We need something like
+    // .accumlate(Function<ConstraintStream, UniConstraintStream<T>>, T defaultValue): ConstraintStream+1
+    // To change it from 3 separate constraints (one for the normal case, one in the case of empty stream,
+    // one to remove double counting)
```
triceo, line 74:
```suggestion
    /* 
       Correct the double counting
       Explanation: groupBy is like accumulate, but it doesn't trigger on empty streams.
       We need something like
        .accumulate(Function<ConstraintStream, UniConstraintStream<T>>, T defaultValue): ConstraintStream+1
       To change it from 3 separate constraints (one for the normal case, one in the case of empty stream,
       one to remove double counting).
     */
```

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/solver/CoachShuttleGatheringConstraintProvider.java
```diff
@@ -0,0 +1,142 @@
...
+        return constraintFactory.from(BusStop.class)
+                .filter(bus -> bus.getPreviousBusOrStop() != null)
+                .penalizeLong(CONSTRAINT_PACKAGE, "distanceFromPrevious", HardSoftLongScore.ONE_SOFT,
+                        BusStop::getDistanceFromPreviousCost);
+    }
+
+    Constraint distanceBusStopToBusDestination(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(BusStop.class)
+                .filter(busStop -> busStop.getNextStop() == null)
+                .join(Bus.class, Joiners.equal(BusStop::getBus, Function.identity()))
```
triceo, line 129: Another static import on `Joiners`?

triceo (commented):
One last going through the code.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/solver/BusPassengerCountTotalUpdatingVariableListener.java
Christopher-Chianelli: As a side note, can a bus stop have 0 passengers? I don't think it can (and if it does, it an error; passengerQuantity is NOT a shadow variable or a planning variable)

triceo: Good point. Maybe we should remove that check altogether.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/solver/CoachShuttleGatheringConstraintProvider.java
Christopher-Chianelli: No, all constraint providers in `optaplanner-examples` use `ScoreType.field`, not a static import of field from ScoreType.

triceo (approved):
LGTM. I'll merge tomorrow afternoon unless there's any more comments by then.
Good job, Chris!

triceo:
Sonar makes a good point - this PR is missing a test for the constraint provider.
Would you mind adding one, @Christopher-Chianelli?

triceo:
Also, the 6 code smells are trivial to address.

Christopher-Chianelli:
@triceo I'll add a test for the constraint provider tomorrow

PR #939, opened 2020-09-20
PLANNER-1830 Add Constraint Provider for Tennis example

triceo:
`FULL_ASSERT` against score DRL passed.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -69,6 +69,21 @@ import org.optaplanner.core.impl.score.stream.uni.DefaultUniConstraintCollector;
  */
 public final class ConstraintCollectors {
 
+    public static <A> DefaultUniConstraintCollector<A, ?, Object> loadBalanceByCount() {
+        return new DefaultUniConstraintCollector<>(
+                () -> new long[2],
+                (resultContainer, a) -> {
+                    resultContainer[0]++;
+                    long count = resultContainer[0];
+                    resultContainer[1] += (2 * count - 1);
+                    return () -> {
+                        resultContainer[1] -= (2 * resultContainer[0] - 1);
+                        resultContainer[0]--;
+                    };
+                },
+                resultContainer -> resultContainer);
+    }
+
     // ************************************************************************
     // count
     // ************************************************************************
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
@@ -69,21 +69,6 @@ import org.optaplanner.core.impl.score.stream.uni.DefaultUniConstraintCollector;
  */
 public final class ConstraintCollectors {
 
-    public static <A> DefaultUniConstraintCollector<A, ?, Object> loadBalanceByCount() {
-        return new DefaultUniConstraintCollector<>(
-                () -> new long[2],
-                (resultContainer, a) -> {
-                    resultContainer[0]++;
-                    long count = resultContainer[0];
-                    resultContainer[1] += (2 * count - 1);
-                    return () -> {
-                        resultContainer[1] -= (2 * resultContainer[0] - 1);
-                        resultContainer[0]--;
-                    };
-                },
-                resultContainer -> resultContainer);
-    }
-
     // ************************************************************************
     // count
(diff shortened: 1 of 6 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -69,6 +69,21 @@
...
+                    resultContainer[0]++;
+                    long count = resultContainer[0];
+                    resultContainer[1] += (2 * count - 1);
+                    return () -> {
+                        resultContainer[1] -= (2 * resultContainer[0] - 1);
+                        resultContainer[0]--;
+                    };
+                },
+                resultContainer -> resultContainer);
+    }
```
ge0ffrey, line 85: I am not sure if I am comfortable putting this into the API directly. I wouldn't mind having it in a TennisCollector first to let it ripe a bit :)

ge0ffrey: Especially because it doesn't combine well with other constraints on the same level.

triceo: Agree. This is just a rebase of an old PR, on which you made the same comment. :-)
I will work on it next week.

ge0ffrey: Ah lol - happy to hear that I am consistent (this time at least) with my old self :)

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,121 @@
...
+import org.optaplanner.core.impl.score.stream.uni.DefaultUniConstraintCollector;
+import org.optaplanner.examples.common.solver.drools.functions.LoadBalanceByCountAccumulateFunction;
+import org.optaplanner.examples.common.solver.drools.functions.LoadBalanceByCountAccumulateFunction.LoadBalanceByCountData;
+import org.optaplanner.examples.common.solver.drools.functions.LoadBalanceByCountAccumulateFunction.LoadBalanceByCountResult;
+import org.optaplanner.examples.tennis.domain.TeamAssignment;
+import org.optaplanner.examples.tennis.domain.UnavailabilityPenalty;
+
+public final class TennisConstraintProvider implements ConstraintProvider {
+
+    private static final String CONSTRAINT_PACKAGE = "org.optaplanner.examples.tennis.solver";
```
ge0ffrey, line 40: Why do we need that for the tennis example, but not for CloudBalancingConstraintProvider which is also in the "optional.score" package?

ge0ffrey: Any way to get rid of having to pass the constraintPackage to the penalize methods?

triceo: Yes, two options:

- Adapt both this and the DRL to `ConstraintConfiguration`.
- Relax the requirement that `FULL_ASSERT` must pass for both, which it won't, when the packages differ.

optaplanner-examples/src/main/java/org/optaplanner/examples/tennis/optional/score/TennisConstraintProvider.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static <A> DefaultUniConstraintCollector<A, ?, LoadBalanceByCountResult>
+            loadBalanceByCount(Function<A, Object> groupKey) {
+        LoadBalanceByCountAccumulateFunction accumulateFunction = new LoadBalanceByCountAccumulateFunction();
+        return new DefaultUniConstraintCollector<>(
+                () -> {
+                    LoadBalanceByCountData resultContainer = accumulateFunction.createContext();
```
ge0ffrey, line 47: This is ugly. Can we not simply duplicate `LoadBalanceByCountData` in here? Or use `Pair<...>` or something?
(I can't wait for java records).

ge0ffrey: Motivation: the ConstraintProvider should be code-wise indepenent of the scoreDRL (including that custom accumulate function which I consider to be owned by the drl)

ge0ffrey (approved):
Good work. See comment about accumulate function reuse

PR #942, opened 2020-09-22
PLANNER-1817 Finish the Conference Scheduling constraint provider

triceo:
`FULL_ASSERT` against score DRL passes.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/domain/Room.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/domain/Room.java
@@ -98,11 +98,6 @@ public class Room extends AbstractPersistable {
         return this;
     }
 
-    public Room withTagSet(Set<String> tagSet) {
-        this.tagSet = tagSet;
-        return this;
-    }
-
     public Room withTalkTypeSet(Set<TalkType> talkTypeSet) {
         this.talkTypeSet = talkTypeSet;
         return this;
@@ -112,4 +107,10 @@ public class Room extends AbstractPersistable {
         this.unavailableTimeslotSet = unavailableTimeslotTest;
         return this;
     }
+
+    public Room withTagSet(Set<String> tagSet) {
+        this.tagSet = tagSet;
+        return this;
+    }
+
 }
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/domain/Room.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/domain/Room.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -92,6 +92,11 @@ public class Room extends AbstractPersistable {
     // ************************************************************************
     // With methods
     // ************************************************************************
+
+    public Room withTagSet(Set<String> tagSet) {
+        this.tagSet = tagSet;
+        return this;
+    }
 
     public Room withTalkTypeSet(Set<TalkType> talkTypeSet) {
         this.talkTypeSet = talkTypeSet;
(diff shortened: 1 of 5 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/conferencescheduling/domain/Room.java
```diff
@@ -93,6 +93,16 @@ public void setTagSet(Set<String> tagSet) {
...
 
+    public Room withCapacity(int capacity) {
+        this.capacity = capacity;
+        return this;
+    }
+
+    public Room withTagSet(Set<String> tagSet) {
+        this.tagSet = tagSet;
+        return this;
+    }
```
ge0ffrey, line 104: Nitpick: ideally the with methods are in the same order as the fields (and the getter/setters), so tagSet after withUnavailableTimeslotSet. It's just a bit nicer to read, and some OCD example users love such detail :)

ge0ffrey: Btw, a bit below in this class, there is a parameter named unavailableTimeslotTest incorrectly.

optaplanner-examples/src/test/java/org/optaplanner/examples/conferencescheduling/optional/score/ConferenceSchedulingConstraintProviderTest.java
```diff
@@ -0,0 +1,919 @@
...
+    private final ConstraintVerifier<ConferenceSchedulingConstraintProvider, ConferenceSolution> constraintVerifier =
+            ConstraintVerifier.build(new ConferenceSchedulingConstraintProvider(), ConferenceSolution.class,
+                    Talk.class);
+
+    private static final LocalDateTime START = LocalDateTime.ofInstant(Instant.EPOCH, ZoneId.systemDefault());
+
+    private static final Timeslot TIMESLOT1 = new Timeslot(1)
+            .withStartDateTime(START)
+            .withEndDateTime(START.plusHours(1))
+            .withTagSet(singleton("a"));
```
ge0ffrey, line 49: A variable name like TIMESLOT1 etc makes the tests harder to understand.
Now I first have to build a mental mapping of timeslot 1, 2, 3 and 4 on how they overlap and they gaps, and then read the constraint test to validate if the assertions are asserting the right thing.

Also, a conference starting at midnight 1 jan 1970 isn't a realistic test I'd argue.

Proposal: instead of TIMESLOT_1 can we make it MONDAY_9_TO_10 and use "1-Feb 2000 9:00" as a value?
If a constraint test then checks if a speaker assigned to talks on MONDAY_9_TO_10 and MONDAY_9_TO_11 overlap, that's much easier to read.

optaplanner-examples/src/test/java/org/optaplanner/examples/conferencescheduling/optional/score/ConferenceSchedulingConstraintProviderTest.java
```diff
@@ -0,0 +1,919 @@
...
+            .withEndDateTime(TIMESLOT1.getEndDateTime().plusHours(1))
+            .withTagSet(singleton("b"));
+    private static final Timeslot TIMESLOT3 = new Timeslot(3)
+            .withStartDateTime(TIMESLOT2.getEndDateTime())
+            .withEndDateTime(TIMESLOT2.getEndDateTime().plusHours(1))
+            .withTagSet(singleton("c"));
+    private static final Timeslot TIMESLOT_ANOTHER_DAY = new Timeslot(3)
+            .withStartDateTime(TIMESLOT1.getStartDateTime().plusDays(1))
+            .withEndDateTime(TIMESLOT1.getStartDateTime().plusDays(1).plusHours(1))
+            .withTagSet(singleton("c"));
```
ge0ffrey, line 61: Potential missing test coverage:
The conf scheduling example allows for this:
- Timeslot A is from 9:00 to 10:00 with talk type "breakout".
- Timeslot B is from 9:20 to 11:20 with talk type "lab".
Notice how the first is 60 minutes, the second 120 minutes and they overlap for 40 minutes.

I don't see that kind of complexity in these timeslots as far as I call tell.

Some of the constraints will penalize based on the overlapping time, so by a factor of 40.
In that case it should be asserted that the matchPenality is 40, not 60 or 120 or 160 or 180.
Other constraints will penalize on the union time, so 160, or the summed time, so 180.

ge0ffrey (approved):
LGTM, see timeslot naming and test coverage comments.

PR #945, opened 2020-09-24
PLANNER-2174 Wildcard types in solution property collections wrongly trigger fail fast

ge0ffrey:


triceo (approved):

triceo:
Good when green.

ge0ffrey:
@triceo Test coverage was missing, so I've added that, which in turn caused me to improve some error messages.

PR #950, opened 2020-09-30
PLANNER-2187 Use java 11 minimum for OptaPlanner 8

rsynek:
### JIRA

https://issues.redhat.com/browse/PLANNER-2187

### Referenced pull requests
https://github.com/kiegroup/optaplanner-quickstarts/pull/6
https://github.com/kiegroup/optaplanner-website/pull/285

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/.github/workflows/jpa_integration_ci.yml
+++ b/.github/workflows/jpa_integration_ci.yml
@@ -6,11 +6,9 @@ on:
   push:
     branches:
       - master
-      - 7.x
   pull_request:
     branches:
       - master
-      - 7.x
 
 defaults:
   run:
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/DefaultPlannerBenchmark.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/DefaultPlannerBenchmark.java
@@ -140,14 +140,7 @@ public class DefaultPlannerBenchmark implements PlannerBenchmark {
         long timeLeftTotal = plannerBenchmarkResult.getWarmUpTimeMillisSpentLimit();
         int parallelBenchmarkCount = plannerBenchmarkResult.getParallelBenchmarkCount();
         int solverBenchmarkResultCount = plannerBenchmarkResult.getSolverBenchmarkResultList().size();
-        /*
-         * cyclesCount needs to be long, as opposed to its natural int,
-         * otherwise JDK 11 compiler uses Math.floorDiv(long, int), which is only available on JDK 9+.
-         * When such compiled code is then used on Java 8, there is a NoSuchMethodError.
-         * All is fine When the code is compiled on JDK 8, the compiler will use the good old Math.floorDiv(long, long).
-         * TODO Remove when JDK 8 is no longer supported.
-         */
-        long cyclesCount = ConfigUtils.ceilDivide(solverBenchmarkResultCount, parallelBenchmarkCount);
+        int cyclesCount = ConfigUtils.ceilDivide(solverBenchmarkResultCount, parallelBenchmarkCount);
         long timeLeftPerCycle = Math.floorDiv(timeLeftTotal, cyclesCount);
         Map<ProblemBenchmarkResult, List<ProblemStatistic>> originalProblemStatisticMap = new HashMap<>(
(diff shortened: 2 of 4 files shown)


triceo (changes requested):
As far as I know Kogito and Quarks still support Java 8. In my opinion, we should not do this before them, and if we do, I believe we will eventually be asked to revert it.

@ge0ffrey This is your call.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/DefaultPlannerBenchmark.java
```diff
@@ -140,14 +140,7 @@ private void warmUp() {
...
         int solverBenchmarkResultCount = plannerBenchmarkResult.getSolverBenchmarkResultList().size();
-        /*
-         * cyclesCount needs to be long, as opposed to its natural int,
-         * otherwise JDK 11 compiler uses Math.floorDiv(long, int), which is only available on JDK 9+.
-         * When such compiled code is then used on Java 8, there is a NoSuchMethodError.
-         * All is fine When the code is compiled on JDK 8, the compiler will use the good old Math.floorDiv(long, long).
-         * TODO Remove when JDK 8 is no longer supported.
-         */
-        long cyclesCount = ConfigUtils.ceilDivide(solverBenchmarkResultCount, parallelBenchmarkCount);
+        int cyclesCount = ConfigUtils.ceilDivide(solverBenchmarkResultCount, parallelBenchmarkCount);
```
ge0ffrey, line 143: Nice one that this is cleaned up!!

optaplanner-build-parent/pom.xml
```diff
@@ -356,6 +357,12 @@
   <build>
     <pluginManagement>
       <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-dependency-plugin</artifactId>
+          <!-- The first version that supports analysis of classes compiled by Java 11. -->
+          <version>3.1.2</version>
```
ge0ffrey, line 364: Overwrite property `<version.dependency.plugin>` instead, just like we have this property:


    <version.surefire.plugin>2.22.2</version.surefire.plugin>

ge0ffrey:
> As far as I know Kogito and Quarks still support Java 8. In my opinion, we should not do this before them, and if we do, I believe we will eventually be asked to revert it.
> 
> @ge0ffrey This is your call.

Kogito is also dropping Java 8 support for Kogito 1.0 AFAIK.
code.quarkus.io doesn't support java 8 no more. If you use JDK 8 with quarkus, it shows **warnings** in the console.

ge0ffrey (approved):

.github/workflows/jpa_integration_ci.yml
```diff
@@ -6,11 +6,9 @@ on:
   push:
     branches:
       - master
-      - 7.x
```
ge0ffrey, line 9: How do this work?
Should the master branch's file mention all branches?
Or should each branch's file mention its own branch?

triceo (approved):
I was not aware of the Kogito decision. If they are indeed switching to 11 for their 1.0, I agree we should follow suit.
I was merely objecting to us being the only ones to do this.

That said, won't this cause Jenkins trouble? AFAIK Jenkins still builds with Java 8. (Hence the loss of Sonar.)

.github/workflows/jpa_integration_ci.yml
rsynek: Each branch should have its own .yaml file(s) with a filter for that particular branch. I think the multiple branches mentioned in the .yaml file on the master branch is a shortcut if you want to build multiple branches using the same workflow.

I would rather keep the branches separated. This test repo proves that triggering the CI on branches with different workflow definitions works: https://github.com/rsynek/github-actions-test/actions

optaplanner-build-parent/pom.xml
rsynek: Good catch, thanks!

rsynek:
Jenkins build fails as it still runs Java 8. I need to figure out how to change it before we merge this.

rsynek:
jenkins retest this

rsynek:
I love the green color of all checks passing once again.

PR #951, opened 2020-09-30
PLANNER-2181 Encapsulate reflection operations in SolutionDescriptor

triceo:
Merge with squash.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
@@ -347,14 +347,7 @@ public class ConfigUtils {
             Class<?> type, Type genericType,
             Class<? extends Annotation> annotationClass, String memberName) {
         if (!(genericType instanceof ParameterizedType)) {
-            throw new IllegalArgumentException("The " + parentClassConcept + " (" + parentClass + ") has a "
-                    + (annotationClass == null ? "auto discovered" : annotationClass.getSimpleName() + " annotated")
-                    + " member (" + memberName
-                    + ") with a member type (" + type
-                    + ") that returns a " + Collection.class.getSimpleName()
-                    + " which has no generic parameters.\n"
-                    + "Maybe the member (" + memberName + ") should return a typed "
-                    + Collection.class.getSimpleName() + ".");
+            return Object.class;
         }
         ParameterizedType parameterizedType = (ParameterizedType) genericType;
         Type[] typeArguments = parameterizedType.getActualTypeArguments();
@@ -389,6 +382,9 @@ public class ConfigUtils {
             } else {
                 typeArgument = upperBounds[0];
             }
+        }
+        if (typeArgument instanceof ParameterizedType) {
+            return (Class) ((ParameterizedType) typeArgument).getRawType();
         }
         if (!(typeArgument instanceof Class)) {
             throw new IllegalArgumentException("The " + parentClassConcept + " (" + parentClass + ") has a "
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
@@ -383,18 +383,11 @@ public class ConfigUtils {
                 typeArgument = upperBounds[0];
             }
         }
-        if (typeArgument instanceof ParameterizedType) {
+        if (!(typeArgument instanceof Class)) { // Turns SomeGenericType<T> into SomeGenericType.
             return (Class) ((ParameterizedType) typeArgument).getRawType();
+        } else {
+            return ((Class) typeArgument);
         }
-        if (!(typeArgument instanceof Class)) {
-            throw new IllegalArgumentException("The " + parentClassConcept + " (" + parentClass + ") has a "
-                    + (annotationClass == null ? "auto discovered" : annotationClass.getSimpleName() + " annotated")
-                    + " member (" + memberName
-                    + ") with a member type (" + type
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -712,6 +714,23 @@ public ScoreDefinition getScoreDefinition() {
...
+                .values()
+                .stream()
+                .map(MemberAccessor::getType);
+        Stream<Class<?>> factCollectionClassStream = getProblemFactCollectionMemberAccessorMap()
+                .entrySet()
+                .stream()
+                .map(entry -> {
+                    MemberAccessor accessor = entry.getValue();
+                    return ConfigUtils.extractCollectionGenericTypeParameter("solutionClass", getSolutionClass(),
+                            accessor.getType(), accessor.getGenericType(), ProblemFactCollectionProperty.class, entry.getKey());
```
ge0ffrey, line 728: I think the problemFactClassSet should be a cached as a field on the SolutionDescriptor, so this code runs once.
This code to fill in that new field could run at the end of `SolutionDescriptor.processProblemFactPropertyAnnotation()` (or at the end of `SolutionDescriptor.processFactEntityOrScoreAnnotation()` if it relies on entities too).

A basic rule around reflection is to scan each class only once for the same thing.
That rule is violated as this code calls `extractCollectionGenericTypeParameter()` (which uses reflection) for every from() call.
Also, as this code calls that reflection method, thinking Quarkus wise, it's much better if it already runs during SolutinDescriptor creation.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/InnerConstraintFactory.java
```diff
@@ -82,22 +81,11 @@
 
     public <A> void assertValidFromType(Class<A> fromType) {
         SolutionDescriptor<Solution_> solutionDescriptor = getSolutionDescriptor();
-        Stream<Class> entityClassStream = solutionDescriptor.getEntityDescriptors().stream()
+        Stream<Class<?>> entityClassStream = solutionDescriptor.getEntityDescriptors().stream()
                 .map(EntityDescriptor::getEntityClass);
```
ge0ffrey, line 85: Maybe we don't need a field `problemFactClassSet` on SolutionDescriptor, but a `problemFactOrEntityClassSet` (== allAcceptedClassSet) field instead? That would avoid this mapping too and the problemFactOrEntityClassSet is only used here.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -347,14 +347,7 @@ public static int resolvePoolSize(String propertyName, String value, String... m
...
         if (!(genericType instanceof ParameterizedType)) {
-            throw new IllegalArgumentException("The " + parentClassConcept + " (" + parentClass + ") has a "
-                    + (annotationClass == null ? "auto discovered" : annotationClass.getSimpleName() + " annotated")
-                    + " member (" + memberName
-                    + ") with a member type (" + type
-                    + ") that returns a " + Collection.class.getSimpleName()
-                    + " which has no generic parameters.\n"
-                    + "Maybe the member (" + memberName + ") should return a typed "
-                    + Collection.class.getSimpleName() + ".");
+            return Object.class;
```
triceo, line 350: The fail-fast is incorrect here. Raw `Collection` is not an untyped collection. It's `Collection<Object>`. This should not fail.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -390,6 +383,9 @@ public static int resolvePoolSize(String propertyName, String value, String... m
                 typeArgument = upperBounds[0];
             }
         }
+        if (typeArgument instanceof ParameterizedType) {
+            return (Class) ((ParameterizedType) typeArgument).getRawType();
+        }
```
triceo, line 388: Another instance of an over-eager fail-fast. `Collection<Something<T>>` shouldn't fail just because we don't know what `T` is. It's a `Collection<Something>` then.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
triceo: Actually, I could use this argument for the entire method. It should never throw. It should just return `Object` - as the answer is always `Object`, unless we can detect a more specific answer.

ge0ffrey: That's true unless that boolean autoDiscovered is true. Because with @PlanningSolution set to auto discover, IIRC, raw types don't work well.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
ge0ffrey: This the same code as on line 375/365? So that code is already there?

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
ge0ffrey: As discussed, we want to move the auto discover fail fast out of this method, into the SolutionDescriptor. This indeed should return Object.

triceo:
Having resolved all Geoffrey's comments after doing a separate call on the issue, I'm merging it.

PR #956, opened 2020-10-05
PLANNER-2191 Remove Scrabble example

triceo:
The example brings no added value and it was agreed that it would be removed.


rsynek (approved):

rsynek:
Thanks, feel free to merge once the build is green.

PR #964, opened 2020-10-09
Use maven.compiler.release for javadoc too

ge0ffrey:


optaplanner-build-parent/pom.xml
```diff
@@ -1143,7 +1143,7 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-javadoc-plugin</artifactId>
               <configuration>
-                <source>${maven.compiler.source}</source>
+                <source>${maven.compiler.release}</source>
```
ge0ffrey, line 1146: This probably picked up the property from jboss-parent, so 8 instead of 11 as it should.

ge0ffrey:
From the looks of the CI output, there's more to this change. It looks like javadoc defaults to jigsaw enabled. We should turn that off.

triceo:
I'm going to bring my own version of this PR shorty. Closing this one as outdated.

PR #965, opened 2020-10-11
PLANNER-2203 Add <Solution_> to various classes

triceo:
Please squash when merging.


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/result/SolverBenchmarkResult.java
```diff
@@ -337,7 +337,7 @@ protected void determineTotalsAndAverages() {
         if (!firstNonFailure) {
             int successCount = getSuccessCount();
             averageScore = totalScore.divide(successCount);
-            averageWorstScoreDifferencePercentage = totalWorstScoreDifferencePercentage.divide((double) successCount);
+            averageWorstScoreDifferencePercentage = totalWorstScoreDifferencePercentage.divide(successCount);
```
ge0ffrey, line 340: Nitpick: I don't mind the explicit cast from int to double here. Autocasting for primitives has been shown to be dangerous. For example `double a = 9007199254740993L` becomes `9007199254740992.0`

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/factory/MoveIteratorFactory.java
```diff
@@ -32,7 +32,7 @@
  *
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
-public interface MoveIteratorFactory<Solution_> {
+public interface MoveIteratorFactory<Solution_, Move_ extends Move<Solution_>> {
```
ge0ffrey, line 35: Will affect upgrade recipe and docs

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/ScoreUtils.java
```diff
@@ -118,7 +118,7 @@ public static double calculateTimeGradient(Number[] totalDiffNumbers, Number[] s
                 // timeGradient += 0.0
                 break;
             } else {
-                double levelTimeGradient = (double) scoreDiffLevel / (double) totalDiffLevel;
+                double levelTimeGradient = scoreDiffLevel / totalDiffLevel;
```
ge0ffrey, line 121: -1 veto. This is a bug. Given scorediffLevel 4 and totalDiffLevel 5, the levelTimeGradient changes from 0.8 to 0.0.

ge0ffrey: nevermind, it isn't.

triceo: Not true. Both of those values are already doubles, the casts do nothing.

ge0ffrey (approved):
LGTM.

2 things to note:
- The removal of casts for primitive types etc scare me a bit, because I have checked about 20-30% of this PR in detail. It would be nicer if they were in a different PR to review next time to avoid glossing over them in a monster PR like this one. I trust however that it will be fine.
- Several classes are mentioned in the docs - we need the docs adjusted too and the upgrade recipe adjusted too. I'd rather see an opt-out and opt-in way of determining which classes those are?

triceo:
Upgrade recipe and docs coming in a subsequent PR tomorrow.

PR #970, opened 2020-10-15
[7.x] PLANNER-2204 VariableListener becomes public API

triceo:


triceo:
Opened against the wrong branch.

PR #971, opened 2020-10-15
[7.x] PLANNER-2204 VariableListener becomes public API #970

triceo:


triceo:
CI failures:

- 8 fails due to Maven Central.
- 11 always fails.
- Jenkins is broken all day.

Ready to merge when @ge0ffrey approves.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/VariableListener.java
```diff
@@ -47,36 +51,36 @@ default boolean requiresUniqueEntityEvents() {
      * @param scoreDirector never null
      * @param entity never null
      */
-    void beforeEntityAdded(ScoreDirector scoreDirector, Entity_ entity);
+    void beforeEntityAdded(ScoreDirector<Object> scoreDirector, Entity_ entity);
```
ge0ffrey, line 54: Are we sure this doesn't break the API?

If a user had written:  `method(ScoreDirector<MyRoster> ...)` it wouldn't have compiled
But if a user had written 

```
beforeEntityAdded(ScoreDirector scoreDirector, ...) {
   ((ScoreDirector<MyRoster>) scoreDirector).getMyShiftList()...
   ...
}
```
Would that still compile?

ge0ffrey: Would it compile if the method signature was ScoreDirector<?>

ge0ffrey (approved):
LGTM, if the compile generics question is ok.
Also, none of the examples should be using the old deprecated method any more, because user copy paste that. Yet it's very good to see that if they do use it, it still compiles and run (as it should).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/VariableListener.java
triceo: I've removed `<Solution_>` from the new `VariableListener`. This should remove all potential for generic corner case problems.

PR #975, opened 2020-10-17
Adjust to the effects of MVEL descoping in upstream

triceo:
This fixes the build at the cost of breaking native compilation.
Native compilation will be fixed later this week by Mario's Drools changes.
Several tests in `optaplanner-examples` have also been disabled and new issues filed in the Drools project.


ge0ffrey (approved):

ge0ffrey:
@triceo Did you run `-Dnative` on school-timetabling to confirm the native build still works? We don't have CI for that IIRC.

triceo:
@ge0ffrey At the moment, we have bigger issues - see Zulip channel. When that works, I'll move on to native.

PR #976, opened 2020-10-19
PLANNER-2142 Make OptaPlanner 8 a part of kogito release pipeline

rsynek:
### JIRA

https://issues.redhat.com/browse/PLANNER-2142

### Referenced pull requests
https://github.com/kiegroup/kogito-pipelines/pull/69
https://github.com/kiegroup/kogito-examples/pull/402

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -59,6 +59,7 @@ pipeline {
 
                     if (isRelease()) {
                         assert getProjectVersion() != ''
+                        assert getKogitoVersion() != ''
                     }
                 }
             }
@@ -258,7 +259,7 @@ void runMaven(String goals, String directory, boolean skipTests = false, List pr
 }
 
 void updateKogitoVersion(String newVersion) {
-    sh "mvn versions:set-property -Dproperty=version.org.kie.kogito -DnewVersion=$newVersion -DallowSnapshots=true -DgenerateBackupPoms=false"
+    maven.mvnSetVersionProperty('version.org.kie.kogito', newVersion)
 }
 
 // Getters and Setters of params/properties
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -102,6 +102,9 @@ pipeline {
                     sh "sed -i 's|<repository>|<repository><id>staging</id><name>Staging Repository</name><url>${params.MAVEN_DEPENDENCIES_REPOSITORY}</url><layout>default</layout><snapshots><enabled>true</enabled></snapshots><releases><enabled>true</enabled></releases></repository>\\n        <repository>|g' ${MAVEN_SETTINGS_FILE}"
                     sh "sed -i 's|<pluginRepository>|<pluginRepository><id>staging</id><name>Staging Repository</name><url>${params.MAVEN_DEPENDENCIES_REPOSITORY}</url><layout>default</layout><snapshots><enabled>true</enabled></snapshots><releases><enabled>true</enabled></releases></pluginRepository>\\n        <pluginRepository>|g' ${MAVEN_SETTINGS_FILE}"
                     sh "sed -i 's|external:\\*|central|g' ${MAVEN_SETTINGS_FILE}" // Done to allow Maven to download release artifacts from MAVEN_DEPENDENCIES_REPOSITORY
+
+                    echo 'Resulting maven settings.xml:'
+                    sh "cat ${MAVEN_SETTINGS_FILE}"
                 }
             }
         }
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -256,7 +256,7 @@ void runMaven(String goals, String directory, boolean skipTests = false, List pr
 }
 
 void updateKogitoVersion(String newVersion) {
-    sh "mvn versions:update-property -Dproperty=version.org.kie.kogito -DnewVersion=[$newVersion] -DallowSnapshots=true -DgenerateBackupPoms=false"
+    sh "mvn versions:set-property -Dproperty=version.org.kie.kogito -DnewVersion=$newVersion -DallowSnapshots=true -DgenerateBackupPoms=false"
 }
 
 // Getters and Setters of params/properties
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -121,11 +121,11 @@ pipeline {
                         maven.mvnVersionsSet(getProjectVersion())
                         updateKogitoVersion(getKogitoVersion())
                     }
-                    /*
+                    /* TODO:
                         Install optaplanner artifacts to a local maven repository to make them accessible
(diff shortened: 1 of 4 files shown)


Jenkinsfile.deploy
```diff
@@ -129,6 +247,55 @@ void runMaven(String goals, String directory, boolean skipTests = false, List pr
...
+            // In case of release, we already have the settings.xml
+            maven.runMaven(mvnCmd, skipTests, ['-fae'])
+        } else {
+            maven.runMavenWithSubmarineSettings(mvnCmd, skipTests)
+        }
+    }
+}
+
+void updateKogitoVersion(String newVersion) {
+    sh "mvn versions:set-property -Dproperty=version.org.kie.kogito -DnewVersion=$newVersion -DallowSnapshots=true -DgenerateBackupPoms=false"
```
rsynek, line 260: May be worth moving to a shared library (after parametrization).

pom.xml
```diff
@@ -34,6 +34,8 @@
   </licenses>
 
   <properties>
+    <!-- TODO: remove before merging -->
+    <enforcer.skip>true</enforcer.skip>
```
rsynek, line 38: Required just for testing the pipeline. Enforcer fails the build due to the http:// protocol of the testing Nexus server (https:// required).

Jenkinsfile.deploy
```diff
@@ -42,37 +56,118 @@ pipeline {
...
+                    }
                 }
             }
             post {
                 success {
                     script {
                         setDeployPropertyIfNeeded('git.branch', getBuildBranch())
                         setDeployPropertyIfNeeded('git.author', getGitAuthor())
+                        setDeployPropertyIfNeeded('project.version', getProjectVersion())
+                        setDeployPropertyIfNeeded('release', isRelease())
```
radtriste, line 71: maybe add `KOGITO_VERSION` as well in `deployment.properties`, as you will need it in promote pipeline

Jenkinsfile.deploy
```diff
@@ -42,37 +56,118 @@ pipeline {
...
+        stage('Setup Maven release config'){
+            when {
+                expression { return isRelease() }
+            }
+            steps {
+                echo 'Setup Maven release config'
+                configFileProvider([configFile(fileId: maven.getSubmarineSettingsXmlId(), targetLocation: 'maven-settings.xml', variable: 'MAVEN_SETTINGS_FILE')]){
+                    sh "echo '-B -s ${MAVEN_SETTINGS_FILE}' | tee optaplanner/.mvn/maven.config"
+                    // TODO: figure out why the assert does not work
+                    // assert params.MAVEN_DEPENDENCIES_REPOSITORY != '' // OptaPlanner depends on Kogito Runtimes
```
radtriste, line 101: this parameter should not be mandatory. runtimes could have been released already. thus this pipeline would not need any `MAVEN_DEPENDENCIES_REPOSITORY`

radtriste: just thinking I need to update that as well on examples that we should check `params.MAVEN_DEPENDENCIES_REPOSITORY` is not empty before doing the change ...

Jenkinsfile.deploy
```diff
@@ -42,37 +56,118 @@ pipeline {
...
+            when {
+                expression { return getProjectVersion() != '' }
+            }
+            steps {
+                script {
+                    // To make sure that the project will resolve during version upgrade.
+                    mavenCleanInstall('optaplanner', true, [], '-U')
+                    dir('optaplanner') {
+                        maven.mvnVersionsSet(getProjectVersion())
+                        updateKogitoVersion(getKogitoVersion())
```
radtriste, line 123: what happen if `KOGITO_VERSION` is empty ?

Jenkinsfile.deploy
radtriste: I do agree

Jenkinsfile.promote
```diff
@@ -0,0 +1,282 @@
...
+boolean isRelease() {
+    return params.RELEASE || (getDeployProperty('release') == 'true')
+}
+
+String getProjectVersion() {
+    return getParamOrDeployProperty('PROJECT_VERSION', 'project.version')
+}
+
+String getKogitoVersion() {
+    return getParamOrDeployProperty('KOGITO_VERSION', 'kogito.version')
```
radtriste, line 167: here is the needed `kogito.version` from `deployment.properties`

Jenkinsfile.deploy
radtriste: already implemented in fact on examples.See https://github.com/kiegroup/kogito-examples/blob/master/Jenkinsfile.deploy#L97

Jenkinsfile.deploy
rsynek: I get the version there via a parameter, I suppose. But +1 to make it consistent.

Jenkinsfile.deploy
rsynek: Good point, will remove the assert and add a condition.

ge0ffrey (approved):
versions-maven-plugin isn't in the jboss-parent, so it makes sense we define it

pom.xml
ge0ffrey: Yes, this must not make it to master.

ge0ffrey: Maybe you can use `-Denforcer.skip=true` in the pipeline instead?

radtriste: this would be something similar to: https://github.com/kiegroup/kogito-cloud-operator/pull/618

Jenkinsfile.deploy
rsynek: Already fixed by making the parameter mandatory.

pom.xml
rsynek: That's an option too, thanks.

ge0ffrey: +1 for doing it in the pipeline which also adds the http:// url, instead of doing it in the pom.xml

Jenkinsfile.deploy
```diff
@@ -99,7 +99,7 @@ pipeline {
                 script {
                     echo 'Setup Maven release config'
                     configFileProvider([configFile(fileId: maven.getSubmarineSettingsXmlId(), targetLocation: 'maven-settings.xml', variable: 'MAVEN_SETTINGS_FILE')]) {
-                        sh "echo '-B -s ${MAVEN_SETTINGS_FILE}' | tee optaplanner/.mvn/maven.config"
+                        sh "echo '\n-B -s ${MAVEN_SETTINGS_FILE}' | tee -a optaplanner/.mvn/maven.config"
```
radtriste, line 102: should add also here the `sh "echo '\n-Denforcer.skip=true' | tee -a .mvn/maven.config"` command or ?

radtriste (approved):

PR #982, opened 2020-10-21
PLANNER-2212 Minimize the occurrence of match weight exceptions

triceo:
I know that we spoke about this before, and I'm doing something else than what we agreed on. I believe I have a good reason for why this is the best option, as it is also the only option.

Consider the following:

- The exception about wrong match weight type comes from `AbstractScoreHolder` - I can not be speaking about `penalizeBigDecimal()` there, as the user may see this in DRL as well.
- One level up, in `impactScore()` in Constraint Streams, I do not know which numeric types `AbstractScoreHolder` will or will not accept, and so I can not add that check there.

Therefore, I only have three options:

1. Modify `AbstractScoreHolder` to return the type of match weight it requires. I'd argue this is needless complexity.
1. In CS, catch the exception from `AbstractScoreHolder` and throw a new one, with a better exception message. I'd argue this is ugly.
1. Make sure the user never sees this exception, assuming they're not doing anything crazy like multiplying `SimpleScore` with a `BigDecimal`.

I am doing the latter. Here's why this is not a problem:

- Multiplying `LongScore` with `int` is safe. 
- Multiplying `BigDecimalScore` with `long` or `int` is safe.
- Multiplying `IntScore` with `long` or `BigDecimal` continues to throw an exception.
- Multiplying `LongScore` with `BigDecimal` continues to throw an exception.

Therefore, this is perfectly safe and a better situation than before.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simple/SimpleScoreHolderImpl.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simple/SimpleScoreHolderImpl.java
@@ -16,6 +16,7 @@
 
 package org.optaplanner.core.impl.score.buildin.simple;
 
+import java.math.BigDecimal;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.function.BiConsumer;
@@ -102,6 +103,22 @@ public final class SimpleScoreHolderImpl extends AbstractScoreHolder<SimpleScore
                     + ConstraintConfiguration.class.getSimpleName() + " annotated class.");
         }
         matchExecutor.accept(kcontext, weightMultiplier);
+    }
+
+    @Override
+    public void impactScore(RuleContext kcontext, long weightMultiplier) {
+        throw new UnsupportedOperationException("In the rule (" + kcontext.getRule().getName()
+                + "), the scoreHolder class (" + getClass()
+                + ") does not support a long weightMultiplier (" + weightMultiplier + ").\n"
+                + "If you're using constraint streams, maybe switch from penalizeLong() to penalize()?");
+    }
+
+    @Override
+    public void impactScore(RuleContext kcontext, BigDecimal weightMultiplier) {
+        throw new UnsupportedOperationException("In the rule (" + kcontext.getRule().getName()
+                + "), the scoreHolder class (" + getClass()
+                + ") does not support a BigDecimal weightMultiplier (" + weightMultiplier + ").\n"
+                + "If you're using constraint streams, maybe switch from penalizeBigDecimal() to penalize()?");
     }
 
     // ************************************************************************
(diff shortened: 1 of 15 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/rules/AbstractRuleAssembler.java
```diff
@@ -105,7 +105,11 @@ protected static void impactScore(DroolsConstraint constraint, Drools drools, Ab
             BigDecimal impact) {
         RuleContext kcontext = (RuleContext) drools;
         constraint.assertCorrectImpact(impact);
-        scoreHolder.impactScore(kcontext, impact);
+        try {
+            scoreHolder.impactScore(kcontext, impact);
+        } catch (UnsupportedOperationException exception) {
+
+        }
```
ge0ffrey, line 112: If we want to do this (and the jury is out on that), then we 'd want to comment here clearly why we're sweeping this under the mat.

ge0ffrey:
A couple of notes:

1) "The exception about wrong match weight type comes from AbstractScoreHolder - I can not be speaking about penalizeBigDecimal() there, as the user may see this in DRL as well."

I have had that kind of problem before, so I did something like: "<orginal error message that says the types don't match>\nMaybe, if you're using a ConstraintProvider, you want to call penalizeLong()" instead.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/rules/AbstractRuleAssembler.java
triceo: Yeah, sorry, this catch shouldn't be there at all. :-)

ge0ffrey:
2) What happens if I use HardSoftScore and do penalizeLong(2L)? What if I do penalizeBigDecimal(new BigDecimal("1.2")) ?

triceo:
> I have had that kind of problem before, so I did something like: "<orginal error message that says the types don't match>\nMaybe, if you're using a ConstraintProvider, you want to call penalizeLong()" instead.

Correct, although I really don't like it. Pushing unknown concepts on a DRL user.


> What happens if I use HardSoftScore and do penalizeLong(2L)? What if I do penalizeBigDecimal(new BigDecimal("1.2")) ?

I edited the original comment. Both of those continue to throw an exception, as this operation would not be safe.

rsynek:
@ge0ffrey would you kindly approve and merge if all your concerns have been addressed?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/simple/SimpleScoreHolderImpl.java
```diff
@@ -104,6 +105,22 @@ public void impactScore(RuleContext kcontext, int weightMultiplier) {
...
+                + ") does not support a long weightMultiplier (" + weightMultiplier + ").\n"
+                + "If you're using constraint streams, maybe switch from penalizeLong() to penalize()?");
+    }
+
+    @Override
+    public void impactScore(RuleContext kcontext, BigDecimal weightMultiplier) {
+        throw new UnsupportedOperationException("In the rule (" + kcontext.getRule().getName()
+                + "), the scoreHolder class (" + getClass()
+                + ") does not support a BigDecimal weightMultiplier (" + weightMultiplier + ").\n"
+                + "If you're using constraint streams, maybe switch from penalizeBigDecimal() to penalize()?");
```
ge0ffrey, line 121: nitpick: Typically we don't end the Maybe sentences with a question mark (?), but with a dot (.).

ge0ffrey (approved):

triceo:
The test fail on Jenkins is expected, unrelated to this PR and will disappear when https://github.com/kiegroup/optaplanner/pull/1000 is merged.

PR #990, opened 2020-10-26
PLANNER-2223: Move ide-config into its own module

Christopher-Chianelli:
Before, ide-config was stored in a directory, which the
formatter plugin needed to reference in order to format the
files. This means for child modules one level down from the project
root, you need to set format.directory to "../../ide-config",
for two levels down: "../../../ide-config". This is error-prone
and not developer friendly. To fix this, we move the ide-config
into its own maven module, which can be included as a dependency
for the formatter plugin, and thus we don't need to set
format.directory. Additionally, this allows us to reuse the
ide-config in other projects without copying the files.

### JIRA
https://issues.redhat.com/browse/PLANNER-2223

### Checklist

<details>
<summary>
</summary>

</details>


rsynek (approved):
Looks good, thanks @Christopher-Chianelli ! Please rebase the PR and resolve the conflicts so that we can merge this.

build/optaplanner-ide-config/pom.xml
```diff
@@ -0,0 +1,12 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.optaplanner</groupId>
+    <artifactId>optaplanner-parent</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>optaplanner-ide-config</artifactId>
+  <name>OptaPlanner IDE Configuration</name>
```
ge0ffrey, line 11: Typically we add `<description>` and `<url>` too.
The description section is mostly duplicate indeed, so that's an issue in itself.
And the url is because maven used to have some crazy rules that made it append the artifactId to inherited url's for maven site, which is long since death, in the manifest.mf of the resulting jar IIRC. I am not sure if that's still the case. 

Either way, if we drop description and/or url overwritting here, we should do that on all child poms at the same time, so it's a separate issue.

ge0ffrey (approved):
@Christopher-Chianelli Good merge, so I 'll already do that. Great work! Please deal with the description/url inconsistency with other child poms (see comment) in a separate PR.

ge0ffrey:
@Christopher-Chianelli Ah, there's a merge conflict, so I can't merge yet, please take a look when you have a moment.

Christopher-Chianelli:
I was able to rebase onto lastest master without conflicts; file move + edit to file that was moved probably caused it.

build/optaplanner-ide-config/pom.xml
rsynek: Let's add the description and url for now to keep it consistent with other modules.

rsynek:
The Jenkins build fails on contacting github API => unrelated.

PR #1001, opened 2020-11-02
Clean up for format.directory property

ge0ffrey:
@Christopher-Chianelli when doing refactors like this, it's recommended to also do a "find in path" search (ctrl-shift-f in intellij) for the removed property (or method name, class name or field name).


Christopher-Chianelli (approved):
LGTM. I only searched for properties; didn't realize some comments reference it.

PR #1016, opened 2020-11-09
PLANNER-2240 Switch OptaPlanner to KieHelper

triceo:
I would argue this does not need to land on 8.0.x.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
@@ -257,7 +257,7 @@ public class ScoreDirectorFactoryFactory<Solution_, Score_ extends Score<Score_>
                 return new DroolsScoreDirectorFactory<>(solutionDescriptor, kieBase);
             }
         } catch (Exception ex) {
-            throw new IllegalStateException("There are problems in your DRL.", ex);
+            throw new IllegalStateException("There is an error in a scoreDrl or scoreDrlFile.", ex);
         }
     }
 
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
@@ -19,25 +19,20 @@ package org.optaplanner.core.impl.score.director;
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 import java.io.File;
-import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Stream;
 
+import org.drools.core.io.impl.ClassPathResource;
+import org.drools.core.io.impl.FileSystemResource;
 import org.drools.modelcompiler.ExecutableModelProject;
 import org.kie.api.KieBase;
 import org.kie.api.KieBaseConfiguration;
 import org.kie.api.KieServices;
-import org.kie.api.builder.KieBuilder;
-import org.kie.api.builder.KieFileSystem;
-import org.kie.api.builder.Message;
-import org.kie.api.builder.Results;
-import org.kie.api.builder.model.KieModuleModel;
-import org.kie.api.io.KieResources;
-import org.kie.api.runtime.KieContainer;
 import org.kie.internal.builder.conf.PropertySpecificOption;
+import org.kie.internal.utils.KieHelper;
 import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.score.calculator.EasyScoreCalculator;
 import org.optaplanner.core.api.score.calculator.IncrementalScoreCalculator;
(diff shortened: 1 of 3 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactoryTest.java
```diff
@@ -140,16 +140,20 @@ void multipleScoreCalculations_throwsException() {
...
     @Test
     void nonExistingDrlFile_throwsException() {
-        ScoreDirectorFactoryConfig config = new ScoreDirectorFactoryConfig().withScoreDrlFiles(new File("nonExisting.drl"));
-        assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> buildTestdataScoreDirectoryFactory(config))
-                .withMessageContaining("scoreDrlFile");
+        ScoreDirectorFactoryConfig config = new ScoreDirectorFactoryConfig()
+                .withScoreDrlFiles(new File("nonExisting.drl"));
+        assertThatExceptionOfType(IllegalStateException.class)
+                .isThrownBy(() -> buildTestdataScoreDirectoryFactory(config))
+                .withMessageContaining("DRL");
```
ge0ffrey, line 156: I 'd argue it needs to have the word "scoreDRL" in the message, because that is the breadcrumb that ties back into the solverConfig's `<scoreDRL>` line.
It's also important for us when we get support tickets, to know that they went through scoreDrl and not though kmodule.xml or CS.

ge0ffrey:
I agree no need to land in 8.0.x.

How does the error message look like after these changes? Does it still mention the DRL line that's incorrect? What happens if there are 2 mistakes in the DRL, does it mention both?

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactoryTest.java
triceo: This code will only be used when they go through scoreDrl/scoreDrlFile.
In fact, I'd argue that saying "DRL" is better than saying "scoreDrl or scoreDrlFile".

triceo:
I caused two errors in DRL. This is what I got:

    12:41:43.952 [main        ] WARN  File 'org/optaplanner/core/impl/score/director/invalidDroolsConstraints.drl' is in folder 'org/optaplanner/core/impl/score/director' but declares package 'org.optaplanner.core.impl.testdata.util'. It is advised to have a correspondance between package and folder names.
12:41:44.093 [main        ] ERROR Unable to build KieBaseModel:defaultKieBase
[20,0]: [ERR 107] Line 20:0 mismatched input 'impot' expecting one of the following tokens: '[package, unit, import, global, declare, function, rule, query]'.
[0,0]: Parser returned a null Package



java.lang.IllegalStateException: There are problems in your DRL.
	...
Caused by: java.lang.RuntimeException: [Message [id=1, kieBase=defaultKieBase, level=ERROR, path=org/optaplanner/core/impl/score/director/invalidDroolsConstraints.drl, line=20, column=0
   text=[ERR 107] Line 20:0 mismatched input 'impot' expecting one of the following tokens: '[package, unit, import, global, declare, function, rule, query]'.], Message [id=2, kieBase=defaultKieBase, level=ERROR, path=org/optaplanner/core/impl/score/director/invalidDroolsConstraints.drl, line=0, column=0
   text=Parser returned a null Package]]
	at org.kie.internal.utils.KieHelper.getKieContainer(KieHelper.java:99)
	at org.kie.internal.utils.KieHelper.build(KieHelper.java:79)
	at org.optaplanner.core.impl.score.director.ScoreDirectorFactoryFactory.buildDroolsScoreDirectorFactory(ScoreDirectorFactoryFactory.java:252)
	... 71 more
">

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -237,76 +232,32 @@ public ScoreDirectorFactoryFactory(ScoreDirectorFactoryConfig config) {
...
+        try {
+            KieBase kieBase = kieHelper.build(ExecutableModelProject.class);
+            if (generateDroolsTestOnError) {
+                return new TestGenDroolsScoreDirectorFactory<>(solutionDescriptor, kieBase, config.getScoreDrlList(),
+                        config.getScoreDrlFileList());
+            } else {
+                return new DroolsScoreDirectorFactory<>(solutionDescriptor, kieBase);
             }
+        } catch (Exception ex) {
+            throw new IllegalStateException("There are problems in your DRL.", ex);
```
ge0ffrey, line 260:
```suggestion
            throw new IllegalStateException("There are problems in a scoreDRL.", ex);
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -237,76 +232,32 @@ public ScoreDirectorFactoryFactory(ScoreDirectorFactoryConfig config) {
...
-                }
-                kieFileSystem.write(kieResources.newFileSystemResource(scoreDrlFile, "UTF-8"));
+                kieHelper.addResource(new FileSystemResource(scoreDrlFile));
+            }
+        }
+
+        try {
+            KieBase kieBase = kieHelper.build(ExecutableModelProject.class);
+            if (generateDroolsTestOnError) {
+                return new TestGenDroolsScoreDirectorFactory<>(solutionDescriptor, kieBase, config.getScoreDrlList(),
```
ge0ffrey, line 254:
```suggestion
                return new TestGenDroolsScoreDirectorFactory<>(solutionDescriptor, kieBase,
```

ge0ffrey (approved):
Good to merge if the suggestion as accepted

PR #1022, opened 2020-11-09
PLANNER-2419 Add ConstraintProvider for Nurse Rostering example

Christopher-Chianelli:
### Checklist

<details>
<summary>
</summary>

</details>


triceo:
So, here's what I think we should have before we next discuss the proposal: `ConsecutiveData` should be a better type. 

- It should provide a set of `Sequence` types - those would have `getLength()` and `getItems()` methods.
- Furthermore, each `Sequence` should point to the next and previous `Sequence`, perhaps utilizing `Optional` for when those sentences are missing. (Alternative: have specialized `FirstSequence` and `LastSequence` types.)
- Each `Sequence` should also directly contain the break information, perhaps with a method `getBreakAfter()`.
- First and last sequence should be accessible directly via a method on `ConsecutiveData`.

Let's leave naming out of this for now - we will get to that during the meeting.

ge0ffrey:
@triceo Are we keeping this open until flatten() and consecutive() CS lands? This was a great PoC.

triceo:
@ge0ffrey Yes, this is very useful and I would like us to reuse it when flattening is ready.
@Christopher-Chianelli  Please see my comment above. Is this something you'd want to pick back up? I expect to get flattening done next sprint.

Christopher-Chianelli:
@triceo sure; I'll improve the `ConsecutiveData` data type.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1966,6 +1968,121 @@ public void remove(Key key, Value value) {
...
+     * Break Lengths: 1 2
+     * Consecutive Items [[Shift slot=1] [Shift slot=2]], [[Shift slot=4]] [[Shift slot=6]]
+     * }
+     * 
+     * @param indexMap Maps the fact to its position in the sequence
+     * @param <A> type of the first mapped fact
+     * @return never null
+     */
+    public static <A> UniConstraintCollector<A, ConsecutiveSetTree<A, Integer, Integer>, ConsecutiveData<A, Integer>>
+            consecutive(ToIntFunction<A> indexMap) {
```
triceo, line 1991: We need to figure out a way to support `LocalDateTime` (and similar types) out of the box. A shift has a startTime and and endTime, and we need to be able to process that "natively" into a sequence.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1966,6 +1968,121 @@ public void remove(Key key, Value value) {
...
+     * @param resultMap Maps both facts to an item in the sequence
+     * @param indexMap Maps the item to its position in the sequence
+     * @param <A> type of the first mapped fact
+     * @param <B> type of the second mapped fact
+     * @param <Result> type of item in the sequence
+     * @return never null
+     */
+    public static <A, B, Result>
+            BiConstraintCollector<A, B, ConsecutiveSetTree<Result, Integer, Integer>, ConsecutiveData<Result, Integer>>
+            consecutive(BiFunction<A, B, Result> resultMap, ToIntFunction<Result> indexMap) {
```
triceo, line 2017: Let's discuss later which of these methods we really want to expose as the public API.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveData.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.util;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class ConsecutiveData<T, D extends Comparable<D>> {
```
triceo, line 23: We need a better name for this. Not necessarily now.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveData.java
```diff
@@ -0,0 +1,57 @@
...
+    private final ConsecutiveSetTree<T, ?, D> sourceTree;
```
triceo, line 24: Please pick descriptive names for `T`, `D`. We may have been able to get away with `A` and `B` in CS, but those are purely positional names - in this case, these types have meaning, and their names should carry it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveData.java
```diff
@@ -0,0 +1,57 @@
...
+
+    protected ConsecutiveData(ConsecutiveSetTree<T, ?, D> sourceTree) {
+        this.sourceTree = sourceTree;
+    }
+
+    public List<Sequence<T>> getConsecutiveSequences() {
+        return sourceTree.getConsecutiveSequences();
+    }
+
+    public List<D> getBreaks() {
```
triceo, line 34: In my opinion, we need `Break<D>` just as we have `Sequence<T>`.
`Break`s need to be navigable just like `Sequence`s are.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,314 @@
...
+    Function<T, I> indexFunction;
+    BiFunction<I, I, D> differenceFunction;
+    D maxDifference;
+    D zeroDifference;
+    TreeMap<T, Sequence<T>> startItemToSequence;
+    Class<? extends T> typeClass;
+    Comparator<T> comparator;
+
+    TreeMapValueList<T, Sequence<T>> sequenceList;
+    ConsecutiveData<T, D> consecutiveData;
```
triceo, line 25: `private` or `private final` where possible, please.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,314 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.TreeMap;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ConsecutiveSetTree<T, I extends Comparable<I>, D extends Comparable<D>> implements Collection<T> {
```
triceo, line 15: Same comment on generic variable naming as above.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,314 @@
...
+                    indexFunction.apply(nextStartItem)));
+        }).collect(Collectors.toList());
+    }
+
+    public List<D> getConsecutiveLengths() {
+        return getConsecutiveSequences().stream()
+                .map(sequence -> differenceFunction.apply(indexFunction.apply(sequence.getItems().first()),
+                        indexFunction.apply(sequence.getItems().last())))
+                .collect(Collectors.toList());
+    }
```
triceo, line 62: Do we need this? Where?
If we need this, it must be fully incremental. If we don't need this now, I suggest we remove it for now.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,314 @@
...
+
+    public List<D> getBreaks() {
+        return startItemToSequence.keySet().stream().flatMap(startItem -> {
+            T nextStartItem = startItemToSequence.higherKey(startItem);
+            if (nextStartItem == null) {
+                return Stream.empty();
+            }
+            return Stream.of(differenceFunction.apply(indexFunction.apply(startItemToSequence.get(startItem).getItems().last()),
+                    indexFunction.apply(nextStartItem)));
+        }).collect(Collectors.toList());
```
triceo, line 54: Needs to be incremental.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,314 @@
...
+
+    @Override
+    public void clear() {
+        startItemToSequence.clear();
+    }
+
+    @Override
+    public int size() {
+        return startItemToSequence.values().stream().map(Sequence::getCountIncludingDuplicates)
+                .reduce(0, Integer::sum);
```
triceo, line 258: Surely we can keep a counter instead.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Sequence.java
```diff
@@ -0,0 +1,124 @@
...
+
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class Sequence<T> {
```
triceo, line 27: This is public API, as it is exported by a public method. I suggest we treat these public types in the following way:

- Interface with the user-facing stuff is in `api`. (`Sequence`)
- Implementation with constructors and all the protected stuff is in `impl`. (`SequenceImpl`)
- Implementations have every method public, but the classes themselves are not.

This way, we make it very clear what is or is not accessible by users, but we get to hide the implementation.
Please apply the same to `Break`, `ConsecutiveData` and any other type that actually reaches the user.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Sequence.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    protected int getCountIncludingDuplicates() {
+        return count.values().stream().reduce(Integer::sum).orElse(0);
+    }
+
+    protected Stream<T> getDuplicatedStream() {
+        return consecutiveItemsSet.stream()
+                .flatMap(item -> IntStream.range(0, count.get(item)).mapToObj(index -> item));
+    }
```
triceo, line 61: I am a bit blurry as to what we need this for. But I really do not like the fact that it is not incremental.
My view of `Sequence` is that is just a dumb no-op view to the underlying collection. (See the idea of a timeline I tried to explain earlier.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/TreeMapValueList.java
```diff
@@ -0,0 +1,277 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.NoSuchElementException;
+import java.util.TreeMap;
+
+public class TreeMapValueList<I, T> implements List<T> {
```
triceo, line 27: Type names.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/TreeMapValueList.java
```diff
@@ -0,0 +1,277 @@
...
+
+    @Override
+    public boolean removeAll(Collection<?> c) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c) {
+        throw new UnsupportedOperationException();
+    }
```
triceo, line 97: So all mutating methods are unsupported? In that case I doubt if the type is good for anything.

What's the benefit of this as opposed to a couple of static util methods?
Or a more tailored class that is not a `List`? It looks to me as if the `List` forces you to implement stuff you don't really need, obfuscating the purpose of the class.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/domain/contract/MinMaxContractLine.java
```diff
@@ -29,6 +29,26 @@
...
+
+    public int getViolationAmount(int count) {
+        if (minimumEnabled && count < minimumValue) {
+            return (minimumValue - count) * minimumWeight;
+        } else if (maximumEnabled && count > maximumValue) {
+            return (count - maximumValue) * maximumWeight;
+        } else {
+            return 0;
+        }
+    }
```
triceo, line 50: Nice to standardize this.
Consider implementing one method with the other, so that the logic is not duplicated.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,509 @@
...
+    Constraint minimumConsecutiveWorkingDays(ConstraintFactory constraintFactory) {
+        return getConsecutiveShifts(constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.CONSECUTIVE_WORKING_DAYS &&
+                        minMaxContractLine.isMinimumEnabled()))
+                                .filter((employeeContractPair,
+                                        shiftList) -> shiftList.getLength() < employeeContractPair.getRight().getMinimumValue())
+                                .penalize("minimumConsecutiveWorkingDays", HardSoftScore.ONE_SOFT,
+                                        (employeeContractPair, shiftList) -> employeeContractPair.getRight()
+                                                .getViolationAmount(shiftList.getLength()));
```
triceo, line 93: I would put the nested stream first, into its own variable. That way, this will become much better to read.
Also, the formatting of the consecutive filters and penalize doesn't seem right here. They should be on the same level.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,509 @@
...
+
+    }
+
+    Constraint maximumConsecutiveWorkingDays(ConstraintFactory constraintFactory) {
+        return getConsecutiveShifts(constraintFactory.from(MinMaxContractLine.class)
```
triceo, line 98: Dtto.

triceo (commented):
Good progress! It's starting to look like it will work out.
I have a whole bunch of comments to evolve this further.

Looking at Sonar, the data structures need much better test coverage. This is complex code and I think it will become even more complex before we finish with it. It needs to be bullet-proof.

Do we have any other examples where we could use it? Ideally, we'd have more than just use case on which to validate the API.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,314 @@
...
+                T firstAfterItem = startItemToSequence.higherKey(item);
+                if (firstAfterItem != null) {
+                    I afterStartIndex = indexFunction.apply(firstAfterItem);
+                    if (isSecondSuccessorOfFirst(itemIndex, endIndex)) {
+                        // We need to extend the first bag
+                        Sequence<T> prevBag = startItemToSequence.get(firstBeforeItem);
+                        if (isSecondSuccessorOfFirst(afterStartIndex, itemIndex)) {
+                            // We need to merge the two bags
+                            Sequence<T> afterBag = startItemToSequence.remove(firstAfterItem);
+                            prevBag.putAll(afterBag);
```
triceo, line 118: The additions and removals from these bags are going to be where this code spends the most time, I think.
It will be imperative to optimize it.

I still have the idea of a timeline in my head.
Where the timeline would essentially be a list of shifts.
And the sequences would essentially just be a tuple of (indexFrom, indexTo), pointing to that list.
To retrieve the shifts, you then just do subList(indexFrom, indexTo), which is a no-op.

But I have no proof that it would actually be faster. (It could be *simpler*, though.)
It'd have to be tried.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/TreeMapValueList.java
Christopher-Chianelli: A `TreeMap` doesn't have a `get(index)` method. I expect most people wants a list of sequences (in the order the sequence occurs in). I also makes testing easier (as I can refer to the sequence by index). Most of this stuff would need to be implemented for any Collection type, leaving `Iterable` as the only interface we can implement so we get iterator functionality (but I expect most people want a Collection).

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
Christopher-Chianelli: Intervals are currently not supported (nor do I think can be supported).
Consider the following example:

[1,2], [2,3], [3,4],[1,3],[3,5]

The following are all consecutive:
[1,2],[2,3],[3,4]
[1,3],[3,4]
[1,3],[3,5]
[2,3],[3,5]

In which case, what should the list of consecutive sequences be? There are duplicate nodes. Currently, what is supported is consecutive points, NOT consecutive intervals.

triceo: I see your point, and that is why I estimated this to take longer than 1 sprint. :-) I believe we need to be able to answer the question, because this use case *will* show up.

In my opinion, overlapping intervals belong in the sequence.
It is the job of another constraint to make sure overlapping shifts do not happen.

triceo: Perhaps `Sequence` is the wrong name here. `Cluster` sounds closer to what I think it should be.
It is all the shifts that happen between two breaks. If some of them overlap, it is no longer a sequence, it is a cluster.

Christopher-Chianelli: Added an interval based implementation for clusters in `IntervalTree`

Christopher-Chianelli:
@triceo I implemented the nurse rostering constraint provider

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Break.java
```diff
@@ -0,0 +1,68 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.util;
+
+import java.util.Objects;
+
+public class Break<I, D> {
```
triceo, line 21: `I` and `D` need to be proper names, typically ending with `_`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Break.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public I getBeforeItem() {
+        return beforeItem;
+    }
+
+    public I getAfterItem() {
+        return afterItem;
+    }
+
+    public D getBreakLength() {
```
triceo, line 40: The class is called `Break`. `getLength()` ought to be enough.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Break.java
```diff
@@ -0,0 +1,68 @@
...
+            return false;
+        Break<?, ?> aBreak = (Break<?, ?>) o;
+        return Objects.equals(beforeItem, aBreak.beforeItem) && Objects.equals(afterItem, aBreak.afterItem)
+                && Objects.equals(breakLength, aBreak.breakLength);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(beforeItem, afterItem, breakLength);
+    }
```
triceo, line 58: Why do we need equality? Shouldn't identity be enough for this use case?

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveIntervalData.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.NavigableMap;
+import java.util.Objects;
+import java.util.TreeMap;
+import java.util.TreeSet;
+
+import org.apache.commons.lang3.ObjectUtils;
+
+public class ConsecutiveIntervalData<_IntervalValue, _PointValue extends Comparable<_PointValue>> {
```
triceo, line 28: The `_` goes at the end of the name. :-) But I like the names themselves.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveIntervalData.java
```diff
@@ -0,0 +1,107 @@
...
+        if (o == null || getClass() != o.getClass())
+            return false;
+        ConsecutiveIntervalData<?, ?> that = (ConsecutiveIntervalData<?, ?>) o;
+        return valueList.equals(that.valueList);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(valueList);
+    }
```
triceo, line 99: Same question on what we need equality for here.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,316 @@
...
+        return startItemToSequence.keySet().stream().flatMap(startItem -> {
+            T nextStartItem = startItemToSequence.higherKey(startItem);
+            if (nextStartItem == null) {
+                return Stream.empty();
+            }
+            T endOfCurrentSequence = startItemToSequence.get(startItem).getItems().last();
+            return Stream.of(new Break<>(nextStartItem, endOfCurrentSequence,
+                    differenceFunction.apply(indexFunction.apply(endOfCurrentSequence),
+                            indexFunction.apply(nextStartItem))));
+        }).collect(Collectors.toList());
```
triceo, line 56: This is likely to be used in CS. Therefore it needs to be incremental.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,316 @@
...
+
+    public Optional<D> getBreakBefore(Sequence<T> sequence) {
+        T startItem = sequence.getItems().first();
+        T prevStartItem = startItemToSequence.lowerKey(startItem);
+        if (prevStartItem == null) {
+            return Optional.empty();
+        }
+        return Optional
+                .of(differenceFunction.apply(indexFunction.apply(startItemToSequence.get(prevStartItem).getItems().last()),
+                        indexFunction.apply(startItem)));
```
triceo, line 74: This is hard to understand. I would put the individual components into their own variables, making the information flow more obvious.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Interval.java
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.util;
+
+import java.util.function.Function;
+
+public class Interval<_IntervalValue, _PointValue extends Comparable<_PointValue>> {
```
triceo, line 21: Same note on `_`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Interval.java
```diff
@@ -0,0 +1,76 @@
...
+        if (o == null || getClass() != o.getClass())
+            return false;
+        Interval<?, ?> that = (Interval<?, ?>) o;
+        return value == that.value;
+    }
+
+    @Override
+    public int hashCode() {
+        return System.identityHashCode(value);
+    }
```
triceo, line 66: Same question on the need for equality.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalCluster.java
```diff
@@ -0,0 +1,172 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.util;
+
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.TreeSet;
+
+public class IntervalCluster<_IntervalValue, _PointValue extends Comparable<_PointValue>> implements Iterable<_IntervalValue> {
```
triceo, line 23: `_` belongs at the end.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalSplitPoint.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.util.Comparator;
+import java.util.IdentityHashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Objects;
+import java.util.TreeSet;
+import java.util.stream.IntStream;
+
+public class IntervalSplitPoint<_IntervalValue, _PointValue extends Comparable<_PointValue>>
```
triceo, line 27: `_` goes at the end.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalSplitPoint.java
```diff
@@ -0,0 +1,131 @@
...
+    public boolean containsIntervalEnding(Interval<_IntervalValue, _PointValue> interval) {
+        return intervalsEndingAtSplitPointSet.contains(interval);
+    }
+
+    public Iterator<_IntervalValue> getValuesStartingFromSplitPointIterator() {
+        return intervalsStartingAtSplitPointSet.stream()
+                .flatMap(interval -> IntStream.range(0, startIntervalToCountMap.get(interval.value))
+                        .mapToObj((index) -> interval.value))
+                .iterator();
+    }
```
triceo, line 97: The iterator here seems unnecessary - streams are already iterable.
I do not see where this is used, perhaps there is some reason for it, but unless it's necessary, I'd prefer removing the indirection.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalSplitPoint.java
```diff
@@ -0,0 +1,131 @@
...
+        if (o == null || getClass() != o.getClass())
+            return false;
+        IntervalSplitPoint<?, ?> that = (IntervalSplitPoint<?, ?>) o;
+        return splitPoint.equals(that.splitPoint);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(splitPoint);
+    }
```
triceo, line 116: Same question on the need for equality.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalTree.java
```diff
@@ -0,0 +1,124 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.util;
+
+import java.util.Iterator;
+import java.util.TreeSet;
+import java.util.function.Function;
+
+public class IntervalTree<_IntervalValue, _PointValue extends Comparable<_PointValue>> {
```
triceo, line 23: `_` goes at the end.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalTree.java
```diff
@@ -0,0 +1,124 @@
...
+    final TreeSet<IntervalSplitPoint<_IntervalValue, _PointValue>> splitPointSet;
+    final Function<_IntervalValue, _PointValue> startMapping;
+    final Function<_IntervalValue, _PointValue> endMapping;
+    final ConsecutiveIntervalData<_IntervalValue, _PointValue> consecutiveIntervalData;
```
triceo, line 27: Instance fields are generally private.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalTreeIterator.java
```diff
@@ -0,0 +1,52 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.util;
+
+import java.util.Iterator;
+
+public class IntervalTreeIterator<_IntervalValue, _PointValue extends Comparable<_PointValue>>
```
triceo, line 21: `_` goes at the end.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalTreeIterator.java
```diff
@@ -0,0 +1,52 @@
...
+        implements Iterator<_IntervalValue> {
+    final Iterator<IntervalSplitPoint<_IntervalValue, _PointValue>> splitPointSetIterator;
+    Iterator<_IntervalValue> splitPointValueIterator;
```
triceo, line 24: `private` please.

triceo (changes requested):
I've done another round of review. Good progress since last time!

Please review all the hot paths for incrementality. I noticed some places which continue to build collections on-demand, which would later be flattened by CS. This is very wasteful.

Please review the naming of generic class parameters. We want descriptive names where it makes sense (not one-letter ones), and we want them to end with `_`, not start with it.

We do not want equals/hashCode implemented, unless absolutely necessary. If you're not putting items into a set or otherwise comparing instances, you do not need equality. No equality = less code = less stuff to test.

Also, many comments from my previous review were seemingly left unnoticed. Please either resolve them or reply.

Looking forward to the meeting later this week!

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+import org.optaplanner.examples.nurserostering.score.drools.EmployeeConsecutiveAssignmentEnd;
+import org.optaplanner.examples.nurserostering.score.drools.EmployeeConsecutiveAssignmentStart;
+
+public class NurseRosteringConstraintProvider implements ConstraintProvider {
+
+    public <C extends ContractLine> BiConstraintStream<ImmutablePair<Employee, C>, Sequence<ShiftDate>>
+            getConsecutiveShifts(UniConstraintStream<C> constraintStream) {
+        return constraintStream.join(Employee.class, Joiners.equal(ContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class, Joiners.equal((contract, employee) -> employee, ShiftAssignment::getEmployee))
+                .groupBy((contract, employee, shift) -> ImmutablePair.of(employee, contract),
```
triceo, line 64: I'd rather we not use `ImmutablePair` here if a suitable `groupBy()` overload is available.
It makes for a code that reads better, and does not introduce users to new types which are not strictly necessary.
It may even perform a bit better, as we do not create needless tuple objects all the time.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+                        ExperimentalConstraintCollectors.consecutive((contract, employee, shift) -> shift.getShiftDate(),
+                                ShiftDate::getDayIndex))
+                .flattenLast(ConsecutiveData::getConsecutiveSequences);
+    }
+
+    public BiConstraintStream<ImmutablePair<Employee, MinMaxContractLine>, Break<ShiftDate, Integer>>
+            getBreaks(UniConstraintStream<MinMaxContractLine> constraintStream) {
+        return constraintStream.join(Employee.class, Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class, Joiners.equal((contract, employee) -> employee, ShiftAssignment::getEmployee))
+                .groupBy((contract, employee, shift) -> ImmutablePair.of(employee, contract),
```
triceo, line 74: Dtto.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+                                ShiftDate::getDayIndex))
+                .flattenLast(ConsecutiveData::getBreaks);
+    }
+
+    public BiConstraintStream<ImmutablePair<Employee, MinMaxContractLine>, Sequence<ShiftDate>>
+            getConsecutiveWorkingWeekends(UniConstraintStream<MinMaxContractLine> constraintStream) {
+        return constraintStream.join(Employee.class, Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class, Joiners.equal((contract, employee) -> employee, ShiftAssignment::getEmployee),
+                        Joiners.filtering((c, e, s) -> s.isWeekend()))
+                .groupBy((contract, employee, shift) -> ImmutablePair.of(employee, contract),
```
triceo, line 85: Dtto.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+    // Hard constraints
+    // ############################################################################
+
+    // A nurse can only work one shift per day, i.e. no two shift can be assigned to the same nurse on a day.
+    Constraint oneShiftPerDay(ConstraintFactory constraintFactory) {
+        return constraintFactory
+                .fromUniquePair(ShiftAssignment.class,
+                        Joiners.equal(ShiftAssignment::getEmployee),
+                        Joiners.equal(ShiftAssignment::getShiftDate))
+                .penalize("org.optaplanner.examples.nurserostering.solver", "oneShiftPerDay", HardSoftScore.ONE_HARD);
```
triceo, line 130: I understand that you're including the package so that FULL_ASSERT works when compared against the DRL.
That said, this does not make the constraint provider easier to understand.
I'd rather change the DRL package than introduce this boilerplate in the constraint provider.

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/Break.java
Christopher-Chianelli: It make it easier to test. It not needed outside of tests, but needed for tests so assertEquals() assertions work

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/ConsecutiveIntervalData.java
Christopher-Chianelli: Same as above; tests

optaplanner-core/src/main/java/org/optaplanner/core/impl/util/IntervalSplitPoint.java
Christopher-Chianelli: `IntervalSplitPoint` implements `Comparable` so it should override equals IMO

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+        return getConsecutiveShifts(constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.CONSECUTIVE_WORKING_DAYS &&
+                        minMaxContractLine.isMinimumEnabled()))
+                                .filter((employeeContractPair,
+                                        shiftList) -> shiftList.getLength() < employeeContractPair.getRight().getMinimumValue())
+                                .penalize("org.optaplanner.examples.nurserostering.solver", "minimumConsecutiveWorkingDays",
+                                        HardSoftScore.ONE_SOFT,
+                                        (employeeContractPair, shiftList) -> employeeContractPair.getRight()
+                                                .getViolationAmount(shiftList.getLength()));
```
triceo, line 177: This is hard to read. The fact that the stream is passed to a function is fine, but I think that the stream should first be declared, and then the variable passed. Easier to understand at first sight.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+                                        shiftList) -> shiftList.getLength() > employeeContractPair.getRight().getMaximumValue())
+                                .penalize("org.optaplanner.examples.nurserostering.solver", "maximumConsecutiveWorkingDays",
+                                        HardSoftScore.ONE_SOFT,
+                                        (employeeContractPair, shiftList) -> employeeContractPair.getRight()
+                                                .getViolationAmount(shiftList.getLength()));
+    }
+
+    // Min/Max consecutive free days
+    Constraint minimumConsecutiveFreeDays(ConstraintFactory constraintFactory) {
+        return getBreaks(constraintFactory.from(MinMaxContractLine.class)
```
triceo, line 195: Dtto.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+                                        breakInfo) -> breakInfo.getBreakLength() - 1 < employeeContractPair.getRight()
+                                                .getMinimumValue())
+                                .penalize("org.optaplanner.examples.nurserostering.solver", "minimumConsecutiveFreeDays",
+                                        HardSoftScore.ONE_SOFT,
+                                        (employeeContractPair, breakInfo) -> employeeContractPair.getRight()
+                                                .getViolationAmount(breakInfo.getBreakLength() - 1));
+    }
+
+    Constraint maximumConsecutiveFreeDays(ConstraintFactory constraintFactory) {
+        return getBreaks(constraintFactory.from(MinMaxContractLine.class)
```
triceo, line 209: Dtto.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class, Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class,
+                        Joiners.equal((contract, employee) -> employee, ShiftAssignment::getEmployee))
+                .groupBy((contract, employee, shiftAssignment) -> new BiTuple<>(contract, employee),
+                        ConstraintCollectors.<MinMaxContractLine, Employee, ShiftAssignment, ShiftAssignment> min(
+                                (contract, employee, shiftAssignment) -> shiftAssignment))
+                .join(NurseRosterParametrization.class)
+                .filter((contractEmployeePair, sa,
+                        nrp) -> nrp.getFirstShiftDate() != sa.getShiftDate() && contractEmployeePair.a
```
triceo, line 251: If you need one reason to not use ImmutablePair here, it is this.
`contractEmployeePair.a` is very cryptic. Especially since it can just be `contract`.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
Christopher-Chianelli: It was ImmutablePair, but unfortunately ImmutablePair implements Comparable, meaning that when a score corruption occurs (eariler when I was debugging), it tries to access it compareTo method, which fails since MinMaxContractLine does not implement Comparable.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+        return constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.CONSECUTIVE_FREE_DAYS &&
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class, Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class,
+                        Joiners.equal((contract, employee) -> employee, ShiftAssignment::getEmployee))
+                .groupBy((contract, employee, shiftAssignment) -> new BiTuple<>(contract, employee),
+                        ConstraintCollectors.<MinMaxContractLine, Employee, ShiftAssignment, ShiftAssignment> min(
+                                (contract, employee, shiftAssignment) -> shiftAssignment))
```
triceo, line 248: Is the cast necessary? It may be some compiler corner-case, but from the looks of it, it appears fairly straight-forward.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
Christopher-Chianelli: I don't think it needed so FULL_ASSERT would work. Rather, it removes extra constraint matches shown in the score corruption overview when a score corruption does occur. I'll remove them.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+    Constraint alternativeSkill(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(BooleanContractLine.class)
+                .filter(booleanContractLine -> booleanContractLine.getContractLineType()
+                        .equals(ContractLineType.ALTERNATIVE_SKILL_CATEGORY))
+                .join(ShiftAssignment.class, Joiners.equal(BooleanContractLine::getContract, ShiftAssignment::getContract))
+                .join(ShiftTypeSkillRequirement.class,
+                        Joiners.equal((contract, sa) -> sa.getShiftType(), ShiftTypeSkillRequirement::getShiftType))
+                .ifNotExists(SkillProficiency.class,
+                        Joiners.equal((contract, sa, stsr) -> sa.getEmployee(), SkillProficiency::getEmployee),
+                        Joiners.equal((contract, sa, stsr) -> stsr.getSkill(), SkillProficiency::getSkill))
```
triceo, line 408: We generally don't use abbreviations like `sa`and `stsr`.
I understand the need to shorten these long names, but `assignment` and `skillRequirement` should be OK.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
Christopher-Chianelli: It was necessary for me; IDEA was giving me compiler error without it.

triceo (changes requested):
For some reason, the constraint provider code was hidden. I have now reviewed that as well.

In general, please stop using `ImmutablePair` unless necessary - variable names such as `t.a` tell the user absolutely nothing.

Also, the constraint package is unnecessary - for consistency with DRL, I'd rather change the DRL package, which is done one a single line in DRL, as opposed to doing it in every constraint in the provider.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+
+    // ############################################################################
+    // Soft constraints
+    // ############################################################################
+    Constraint minimumAndMaximumNumberOfAssignments(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine.getContractLineType() == ContractLineType.TOTAL_ASSIGNMENTS &&
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class,
+                        Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
```
triceo, line 141: It would seem that you do not need the join with the `MinMaxContractLine`.
The same could possibly be done with `ifExists(...)` after `from(Employee.class)`.
This has the potential to speed up things quite a bit.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+        return getConsecutiveShifts(constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.CONSECUTIVE_WORKING_DAYS &&
+                        minMaxContractLine.isMaximumEnabled()))
+                                .filter((employeeContractPair,
+                                        shiftList) -> shiftList.getLength() > employeeContractPair.getRight().getMaximumValue())
+                                .penalize("org.optaplanner.examples.nurserostering.solver", "maximumConsecutiveWorkingDays",
+                                        HardSoftScore.ONE_SOFT,
+                                        (employeeContractPair, shiftList) -> employeeContractPair.getRight()
+                                                .getViolationAmount(shiftList.getLength()));
```
triceo, line 190: Please take care to maintain proper indentation, the filtrers and penalties ought to start at the same column. This is just one instance of multiple of this problem.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+
+    Constraint consecutiveFreeDaysFinalBreak(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.CONSECUTIVE_FREE_DAYS &&
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class, Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class,
+                        Joiners.equal((contract, employee) -> employee, ShiftAssignment::getEmployee))
+                .groupBy((contract, employee, shiftAssignment) -> new BiTuple<>(contract, employee),
```
triceo, line 267: Please remove the use of BiTuple here.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+    // Soft constraints
+    // ############################################################################
+    Constraint minimumAndMaximumNumberOfAssignments(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(MinMaxContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine.getContractLineType() == ContractLineType.TOTAL_ASSIGNMENTS &&
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class,
+                        Joiners.equal(MinMaxContractLine::getContract, Employee::getContract))
+                .join(ShiftAssignment.class,
+                        Joiners.equal((contractLine, employee) -> employee, ShiftAssignment::getEmployee))
```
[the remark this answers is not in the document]
triceo: The improvements to this constraint are probably also applicable to other constraints in this class.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class, Joiners.equal(ContractLine::getContract, Employee::getContract))
+                .join(ShiftDate.class, Joiners.filtering((c, e, date) -> date.getDayOfWeek() == DayOfWeek.SUNDAY))
+                .join(ShiftType.class)
+                .map(QuadTuple::new)
+                .join(ShiftAssignment.class,
+                        Joiners.equal(t -> t.c.getWeekendSundayIndex(), ShiftAssignment::getWeekendSundayIndex),
+                        Joiners.equal(t -> t.b, ShiftAssignment::getEmployee),
+                        Joiners.equal(t -> t.d, ShiftAssignment::getShiftType),
+                        Joiners.filtering((t, shift) -> shift.isWeekend()))
```
triceo, line 356: Instead of the many joins and the mapping, wouldn't it be better to just group by employee+shiftType on the shift assignment directly?

Quad joins will be very expensive on their own, we should try to avoid them as much as possible.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    // Identical shiftTypes during a weekend
+    Constraint identicalShiftTypesDuringWeekend(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(BooleanContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.IDENTICAL_SHIFT_TYPES_DURING_WEEKEND &&
+                        minMaxContractLine.isEnabled())
+                .join(Employee.class, Joiners.equal(ContractLine::getContract, Employee::getContract))
+                .join(ShiftDate.class, Joiners.filtering((c, e, date) -> date.getDayOfWeek() == DayOfWeek.SUNDAY))
```
triceo, line 349: Considering that the filter is only applied to ShiftDate and nothing else, this will perform better as a nested filtered stream. (See my latest blog.)

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+import org.optaplanner.core.api.score.stream.Constraint;
+import org.optaplanner.core.api.score.stream.ConstraintCollectors;
+import org.optaplanner.core.api.score.stream.ConstraintFactory;
+import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.core.api.score.stream.Joiners;
+import org.optaplanner.core.api.score.stream.tri.TriConstraintStream;
+import org.optaplanner.core.api.score.stream.uni.UniConstraintStream;
+import org.optaplanner.core.impl.score.stream.drools.common.QuadTuple;
```
triceo, line 30: This definitely is not a public type and should not be used in user-facing code.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+                        minMaxContractLine.isEnabled())
+                .join(constraintFactory.from(ShiftDate.class)
+                        .filter(date -> date.getDayOfWeek() == DayOfWeek.SUNDAY))
+                .join(constraintFactory.from(ShiftAssignment.class).filter(ShiftAssignment::isWeekend),
+                        Joiners.equal((contract, date) -> date.getWeekendSundayIndex(), ShiftAssignment::getWeekendSundayIndex),
+                        Joiners.equal((contract, date) -> contract.getContract(), ShiftAssignment::getContract))
+                .groupBy((contract, date, sa) -> contract,
+                        (contract, date, sa) -> sa.getEmployee(),
+                        (contract, date, sa) -> ImmutablePair.of(sa.getShiftType(), date), // No 4-key groupBy overload
+                        ConstraintCollectors.countTri())
```
triceo, line 294: This is probably the only use case where using a single-key overload would have been acceptable. :-)

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+                        Joiners.equal((pattern, contractLine, shift) -> shift.getShiftDateDayIndex() + 1,
+                                ShiftAssignment::getShiftDateDayIndex),
+                        Joiners.filtering((pattern, contractLine, shift1, shift2) -> pattern.getDayIndex1ShiftType() == null
+                                || shift2.getShiftType() == pattern.getDayIndex1ShiftType()))
+                .penalize("unwantedPatternShiftType2DaysPattern", HardSoftScore.ONE_SOFT,
+                        (pattern, contractLine, shift1, shift2) -> pattern.getWeight());
+    }
+
+    Constraint unwantedPatternShiftType3DaysPattern(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(ShiftType3DaysPattern.class)
```
triceo, line 392: The `PatternContractLine` apparenty includes the `ShiftType3DaysPattern`, therefore we can start from it directly, avoiding the join, and therefore avoiding the `map()` later.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+                        Joiners.filtering((pattern, contractLine, date, employee,
+                                shift) -> shift.getShiftDateDayIndex() == (date.getDayIndex() + 1)
+                                        || shift.getShiftDateDayIndex() == (date.getDayIndex() + 2)))
+                .penalize("unwantedPatternFreeBefore2DaysWithAWorkDayPattern", HardSoftScore.ONE_SOFT,
+                        (pattern, contractLine, date, employee) -> pattern.getWeight());
+    }
+
+    Constraint unwantedPatternShiftType2DaysPattern(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(ShiftType2DaysPattern.class)
+                .join(PatternContractLine.class, Joiners.equal(p -> p, PatternContractLine::getPattern))
```
triceo, line 376: `ShiftType2DaysPattern` appears avoidable.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+                                SkillProficiency::getEmployee),
+                        Joiners.equal((contract, shiftAssignment, skillRequirement) -> skillRequirement.getSkill(),
+                                SkillProficiency::getSkill))
+                .penalize("alternativeSkill", HardSoftScore.ONE_SOFT,
+                        (contractLine, shiftAssignment, skillRequirement) -> contractLine.getWeight());
+    }
+
+    // Unwanted patterns
+    Constraint unwantedPatternFreeBefore2DaysWithAWorkDayPattern(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(FreeBefore2DaysWithAWorkDayPattern.class)
```
triceo, line 354: `FreeBefore2DaysWithAWorkDayPattern` appears avoidable through `PatternContractLine`.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+
+    // Min/Max consecutive working weekends
+    Constraint consecutiveWorkingWeekends(ConstraintFactory constraintFactory) {
+        return getConsecutiveWorkingWeekends(constraintFactory,
+                constraintFactory.from(MinMaxContractLine.class)
+                        .filter(minMaxContractLine -> minMaxContractLine
+                                .getContractLineType() == ContractLineType.CONSECUTIVE_WORKING_WEEKENDS &&
+                                minMaxContractLine.isEnabled()))
+                                        .filter((employee, contract, shiftList) -> contract.isViolated(shiftList.getLength()))
+                                        .penalize("consecutiveWorkingWeekends", HardSoftScore.ONE_SOFT,
```
triceo, line 245: Formatting.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+    // Complete Weekends
+    Constraint startOnNotFirstDayOfWeekend(ConstraintFactory constraintFactory) {
+        return getConsecutiveShifts(constraintFactory.from(BooleanContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.COMPLETE_WEEKENDS &&
+                        minMaxContractLine.isEnabled()))
+                                .filter((employee, contract, shiftList) -> new EmployeeConsecutiveAssignmentStart(
+                                        employee, shiftList.getItems().first())
+                                                .isWeekendAndNotFirstDayOfWeekend())
+                                .penalize("startOnNotFirstDayOfWeekend", HardSoftScore.ONE_SOFT,
```
triceo, line 259: Formatting.

optaplanner-examples/src/main/java/org/optaplanner/examples/nurserostering/optional/score/NurseRosteringConstraintProvider.java
```diff
@@ -0,0 +1,416 @@
...
+    Constraint endOnNotLastDayOfWeekend(ConstraintFactory constraintFactory) {
+        return getConsecutiveShifts(constraintFactory.from(BooleanContractLine.class)
+                .filter(minMaxContractLine -> minMaxContractLine
+                        .getContractLineType() == ContractLineType.COMPLETE_WEEKENDS &&
+                        minMaxContractLine.isEnabled()))
+                                .filter((employee, contract,
+                                        shiftList) -> new EmployeeConsecutiveAssignmentEnd(employee,
+                                                shiftList.getItems().last())
+                                                        .isWeekendAndNotLastDayOfWeekend())
+                                .penalize("endOnNotLastDayOfWeekend", HardSoftScore.ONE_SOFT,
```
triceo, line 274: Formatting.

triceo (commented):
Thanks for the changes! I went over the PR one last time and noticed some details:

- We moved the code to examples, yet some parts of it still reside in `core/impl`. Please move, incl. tests.
- It would be nice if you could clean up some of the Sonar code smells. I don't mind the iterator "issue", but most of the code smells have an easy fix and I agree we needn't introduce those.

When you're done, please undraft the PR and I will merge it.

triceo (approved):
Great work, @Christopher-Chianelli, and a lot of it!

PR #1029, opened 2020-11-11
PLANNER-625 Move optional example code into an example package

triceo:
For the purposes of this PR, "optional" is everything without which the `ExampleApp` still compiles and runs without throwing exceptions.


ge0ffrey (approved):
LGTM, reading diagonally. Great to see this happening!
I saw this affected tests too. What is the approach there to separate "optional" tests from "non-optional" tests?

triceo:
I made sure all tests are in the proper package.
As far as optionality of the tests themselves, I believe there is no such thing. :-)

PR #1058, opened 2020-12-02
PLANNER-2269: Make CS-B Tuple a plain data carrier

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -48,34 +56,80 @@
...
             constraint.createNodes(buildPolicy, declaredClassToNodeMap, constraintWeight);
         });
-        this.nodeOrderSize = buildPolicy.getNodeOrderMaximum() + 1;
+        nodeIndexSize = buildPolicy.getNodeIndexMaximum() + 1;
         constraintIdToScoringNodeMap = buildPolicy.getConstraintIdToScoringNodeMap();
         effectiveClassToNodeListMap = new HashMap<>(declaredClassToNodeMap.size());
-        nodeOrderedQueueList = new ArrayList<>(nodeOrderSize);
-        for (int i = 0; i < nodeOrderSize; i++) {
-            nodeOrderedQueueList.add(new ArrayDeque<>(1000));
+        nodeIndexToNodeMap = new ArrayList<>(this.nodeIndexSize);
```
triceo, line 81: This is not actually in use yet. The change to completely remove `Node` reference from `Tuple` seemed too invasive for this PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetTupleState.java
```diff
@@ -17,20 +17,37 @@
...
 
 public enum BavetTupleState {
-    NEW,
-    CREATING,
-    UPDATING,
-    OK,
-    DYING,
-    DEAD,
-    ABORTING;
+    NEW(false, true),
```
triceo, line 20: Why is `NEW` not `dirty`?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -48,34 +56,80 @@
     private final ScoreInliner<Score_> scoreInliner;
 
     private final Map<Class<?>, BavetFromUniNode<Object>> declaredClassToNodeMap;
-    private final int nodeOrderSize;
+    private final int nodeIndexSize;
```
ge0ffrey, line 59: I guess this can be `nodeSize`? (the number of nodes)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
ge0ffrey: Yes, thank you

triceo:
Pre-requisite: #1064 
Will cause merge conflicts here, will have to resolve.

triceo:
I have rebased this on top of #1064. When that is merged, I will rebase this again on `master`.

triceo:
Rebased, making this PR much smaller.

triceo:
The performance issue was resolved by reverting some of the changes without which we can live.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -88,6 +89,23 @@ public BavetConstraintSession(boolean constraintMatchEnabled, ScoreDefinition<Sc
         return nodeIndexedNodeMap;
     }
 
+    private static void refreshTuple(BavetAbstractTuple tuple) {
+        ((BavetAbstractNode) tuple.getNode()).refresh(tuple);
```
ge0ffrey, line 93: Why doesn't getNode() return type BavetAbstractNode?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -88,6 +89,23 @@ public BavetConstraintSession(boolean constraintMatchEnabled, ScoreDefinition<Sc
...
+            case UPDATING:
+                tuple.setState(BavetTupleState.OK);
+                return;
+            case DYING:
+            case ABORTING:
+                tuple.setState(BavetTupleState.DEAD);
+                return;
+            case DEAD:
+                throw new IllegalStateException(
+                        "The tuple (" + tuple + ") is already in the dead state (" + tuple.getState() + ").");
```
ge0ffrey, line 105: Let's a default case that throws an exception?

ge0ffrey: Also, this exception can start with "Impossible state:"
```suggestion
                        "Impossible state: The tuple (" + tuple + ") is already in the dead state (" + tuple.getState() + ").");
```

ge0ffrey: Do we also want to print out the node toString in the exception message? Might be useful if someone ever sees it on productions and asks us what could be the cause.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetFilterBiNode.java
```diff
@@ -43,6 +45,11 @@ public void addChildNode(BavetAbstractBiNode<A, B> childNode) {
         childNodeList.add(childNode);
     }
 
+    @Override
+    public List<BavetAbstractBiNode<A, B>> getChildNodes() {
+        return Collections.unmodifiableList(childNodeList);
```
ge0ffrey, line 50: (important IMO)
The entire Bavet implementation is not public API. We don't need to wrap in `unmodifiableList()` etc, as that will only create perf loss and memory usage increase.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiNode.java
```diff
@@ -40,6 +42,11 @@ public void addChildNode(BavetAbstractBiNode<GroupKey_, Result_> childNode) {
         childNodeList.add(childNode);
     }
 
+    @Override
+    public List<BavetAbstractBiNode<GroupKey_, Result_>> getChildNodes() {
```
ge0ffrey, line 46: For consistency sake, please call this method `getChildNodeList()`.

The color of the curtains is pink, I know you hate it, but let's not mix colors of curtains - it's going to look ugly.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiNode.java
```diff
@@ -40,6 +42,11 @@ public void addChildNode(BavetAbstractBiNode<GroupKey_, Result_> childNode) {
...
+        return Collections.unmodifiableList(childNodeList);
```
ge0ffrey, line 47: see above, no need for unmodifiableList

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetJoinBridgeBiTuple.java
```diff
@@ -16,19 +16,14 @@
...
 import org.optaplanner.core.impl.score.stream.bavet.common.BavetJoinBridgeTuple;
-import org.optaplanner.core.impl.score.stream.bavet.common.BavetJoinTuple;
 
 public final class BavetJoinBridgeBiTuple<A, B> extends BavetAbstractBiTuple<A, B>
         implements BavetJoinBridgeTuple {
 
     protected final BavetAbstractBiTuple<A, B> parentTuple;
     private final BavetJoinBridgeBiNode<A, B> node;
 
-    protected Set<BavetJoinTuple> childTupleSet = new LinkedHashSet<>(); // TODO capacity
```
ge0ffrey, line 31: This is intresting. The Set becomes a List (which is a good thing I think.)

Did this bring a performance gain? If not, is it possible it was offset by the unmodifiable wrapping?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/tri/BavetJoinTriNode.java
```diff
@@ -49,6 +50,11 @@ public void addChildNode(BavetAbstractTriNode<A, B, C> childNode) {
         childNodeList.add(childNode);
     }
 
+    @Override
+    public List<BavetAbstractTriNode<A, B, C>> getChildNodes() {
+        return Collections.unmodifiableList(childNodeList);
```
ge0ffrey, line 55: see above: method name + unmodifiableList

ge0ffrey (approved):
LGTM if unmodifiableList are gone

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetJoinBridgeBiTuple.java
triceo: I'll re-benchmark once this is merged. But I don't really think this should result in such a big gain - it's just inside one particular tuple, not inside all of them.

triceo:
Jenkins retest this.

PR #1064, opened 2020-12-03
PLANNER-2269: Proper CS-B node indexing

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirector.java
```diff
@@ -172,4 +172,11 @@ public void afterProblemFactRemoved(Object problemFact) {
         super.afterProblemFactRemoved(problemFact);
     }
 
+    // ************************************************************************
+    // Getters/setters
+    // ************************************************************************
+
+    public ConstraintSession<Solution_, Score_> getSession() {
+        return session;
+    }
```
triceo, line 181: Don't like it, but I needed this for the test.
This is private API, so we should be good.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintStreamNodeOrderingTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.optaplanner.core.impl.score.stream.bavet.uni.BavetAbstractUniNode;
+import org.optaplanner.core.impl.score.stream.bavet.uni.BavetFilterUniNode;
+import org.optaplanner.core.impl.score.stream.bavet.uni.BavetFromUniNode;
+import org.optaplanner.core.impl.score.stream.bavet.uni.BavetJoinBridgeUniNode;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishEntity;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishSolution;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishValue;
+import org.optaplanner.core.impl.testdata.domain.score.lavish.TestdataLavishValueGroup;
+
+public class BavetConstraintStreamNodeOrderingTest {
```
triceo, line 46: This is the test that checks, for a sufficiently complex stream, that the ordering is correct.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetAbstractNode.java
```diff
@@ -21,20 +21,24 @@
...
-    public int getNodeOrder() {
-        return nodeOrder;
+    public int getNodeIndex() {
+        return nodeIndex;
     }
 
+    @Override
+    public int compareTo(BavetNode o) {
+        return Integer.compare(this.nodeIndex, o.getNodeIndex());
+    }
```
triceo, line 43: This is not consistent with `equals()`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -37,18 +37,11 @@ public BavetNodeBuildPolicy(BavetConstraintSession session, int constraintCount)
...
-            nodeOrderMaximum = nodeOrder;
-        }
-    }
-
     public <Node_ extends BavetAbstractNode> Node_ retrieveSharedNode(Node_ node) {
         Node_ sharedNode = (Node_) sharableNodeMap.computeIfAbsent(node, k -> node);
-        if (node.getNodeOrder() != sharedNode.getNodeOrder()) {
-            throw new IllegalStateException("Impossible state: the node (" + node
-                    + ")'s nodeOrder (" + node.getNodeOrder() + ") differs from the sharedNode (" + sharedNode
-                    + ")'s nodeOrder (" + sharedNode.getNodeOrder() + ").");
```
triceo, line 51: All node indexes are unique now. Therefore this fail-fast makes no sense anymore.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/ConstraintStreamScoreDirector.java
ge0ffrey: Ok. Someday we need to talk about merging ConstraintStreamSession and ConstraintStreamScoreDirector to avoid unneeded indirection :)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/BavetConstraintSession.java
```diff
@@ -48,12 +49,12 @@
     private final ScoreInliner<Score_> scoreInliner;
 
     private final Map<Class<?>, BavetFromUniNode<Object>> declaredClassToNodeMap;
-    private final int nodeOrderSize;
+    private final int nodeIndexSize;
```
ge0ffrey, line 52: Actually, this is now just nodeSize, I'd argue. Wdyt?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetAbstractBiConstraintStream.java
```diff
@@ -277,39 +277,35 @@ public final Constraint impactScoreConfigurableBigDecimal(String constraintPacka
...
-        if (sharedNode != node) {
-            // Share node
-            node = sharedNode;
-        } else {
-            if (parentNode != null) { // TODO remove null check and don't go through this code like this for from and joins
-                parentNode.addChildNode(node);
-            }
+        if (sharedNode != node) { // Share node
+            return sharedNode;
+        } else if (parentNode != null) { // TODO remove null check and don't go through this for from and joins
```
ge0ffrey, line 292: I wouldn't write this as an else if because the 2 conditions are unrelated.
Suggestion: replace the else with a line break. - because the sharedNode check is a quick return (special condition for which the main content of this method doesn't apply).

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetFilterBiConstraintStream.java
```diff
@@ -51,8 +51,8 @@ public BavetFilterBiConstraintStream(BavetConstraintFactory<Solution_> constrain
 
     @Override
     protected BavetFilterBiNode<A, B> createNode(BavetNodeBuildPolicy<Solution_> buildPolicy,
-            Score<?> constraintWeight, int nodeOrder, BavetAbstractBiNode<A, B> parentNode) {
-        return new BavetFilterBiNode<>(buildPolicy.getSession(), nodeOrder, parentNode, predicate);
+            Score<?> constraintWeight, BavetAbstractBiNode<A, B> parentNode) {
+        return new BavetFilterBiNode<>(buildPolicy.getSession(), buildPolicy.getNodeIndexMaximum() + 1, parentNode, predicate);
```
ge0ffrey, line 55: (important)
This is very wierd.
I would expect the it take as the next Index from the buildPolicy.
This way the next Index doesn't increment. That's a bug I think?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiConstraintStream.java
```diff
@@ -50,19 +50,20 @@ public BavetGroupBiConstraintStream(BavetConstraintFactory<Solution_> constraint
...
     protected BavetGroupBiNode<GroupKey_, ResultContainer_, Result_> createNode(BavetNodeBuildPolicy<Solution_> buildPolicy,
-            Score<?> constraintWeight, int nodeOrder, BavetAbstractBiNode<GroupKey_, Result_> parentNode) {
+            Score<?> constraintWeight, BavetAbstractBiNode<GroupKey_, Result_> parentNode) {
         if (parentNode != null) {
             throw new IllegalStateException("Impossible state: the stream (" + this
                     + ") cannot have a parentNode (" + parentNode + ").");
         }
-        return new BavetGroupBiNode<>(buildPolicy.getSession(), nodeOrder, finisher);
+        // GroupBridgeNodes are created after GroupNodes, so increment by 2 to create room for the bridge inbetween.
+        return new BavetGroupBiNode<>(buildPolicy.getSession(), buildPolicy.getNodeIndexMaximum() + 2, finisher);
```
ge0ffrey, line 66: Hell no. This appraoch won't work with composition approaches.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -26,7 +26,7 @@
 
     private final BavetConstraintSession session;
 
-    private int nodeOrderMaximum = 0;
+    private int nodeIndexMaximum = -1; // So that the first node starts at 0 when it increments.
```
ge0ffrey, line 29: My suggestion is to remove this field and instead have

```suggestion
    private int nextNodeIndex = 0;
    
    public int nextNodeIndex() {
        return nextNodeIndex++;
    }
```

Every time a new node is created, it call this method and automatically gets the next one.
Remember that join nodes are created after both bridges are created, so a join will never be before it's bridges or anything before it (but we should test those nodeIndexes with a unit test).

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
ge0ffrey: Crap, ideally when a node is replaced by a sharedNode, it doesn't take up an extra nodeIndex.

Because it does, the nodeSize will be actually bigger than the number of actual nodes in place. (so it shouldn't be called nodeSize indeed and a comment should specify that strange behaviour). So the queue will have a number of empty spaces. Actually, that hurts performance...
... so we need to recycle nodeIndexes of nodes that get shared. Maybe upon sharing, call BuildPolicy.undoNextIndex() that does nextNodeIndex--?

ge0ffrey (changes requested):
See nextNodeIndex changes.
The +1 and + 2 tricks won't work in complex node networks

ge0ffrey:
I'll review again when nextNodeIndex vs +1 + 2 issue is fixed. Only reviewed it partially now - I'd rather wait on the nextNodeIndex impact.

triceo:
@ge0ffrey I will have to see examples of those networks that supposedly won't work. I have verified every situation I could think of.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
triceo: This is not an issue. Node indexes are perfectly sequential, they are never duplicated. (The "+1 trick" takes care of that.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetFilterBiConstraintStream.java
triceo: The index is incremented immediately after this, when the node is checked for sharing. There is a test that checks for this.
I am interested in counter-examples.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiConstraintStream.java
triceo: Assuming that Group is always preceded by a GroupBridge, this is perfectly safe.
Otherwise I'll need to see an example of a network where that is not the case.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
triceo: No. The entire algorithm depends on the fact, that `nodeIndex` is only ever actually incremented when node sharing is ruled out. It makes it incredibly simple and effective - no need for any sort of rollback.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -77,4 +86,22 @@ public int getNodeOrderMaximum() {
...
+                }, TreeMap::new));
+        // Ensure there are no gaps in that list.
+        int maxNodeIndex = nodeIndexToNodeMap.lastKey();
+        int expectedMaxNodeIndex = nodeIndexToNodeMap.size() - 1;
+        if (maxNodeIndex != expectedMaxNodeIndex) {
+            throw new IllegalStateException("Impossible state: maximum node index (" + maxNodeIndex +
+                    ") does not match the expected maximum node count (" + expectedMaxNodeIndex + ").");
+        }
+        return Collections.unmodifiableList(new ArrayList<>(nodeIndexToNodeMap.values()));
+    }
```
triceo, line 105: This is heavy. Thankfully it is only ever used once per `BavetConstraintSession`.
Considering that it fails fast on a pretty important thing, I say we need it.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
```diff
@@ -1419,45 +1419,6 @@ public void duplicateConstraintId() {
                 constraintStreamImplType));
     }
 
-    @TestTemplate
-    public void globalNodeOrder() {
```
triceo, line 1423: Not useful anymore. We now have a better test, and that one checks the node indexes directly.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBridgeBiConstraintStream.java
```diff
@@ -57,20 +57,23 @@ public void setGroupStream(BavetGroupBiConstraintStream<Solution_, NewA, ResultC
...
     @Override
-    protected void createChildNodeChains(BavetNodeBuildPolicy<Solution_> buildPolicy, Score<?> constraintWeight, int nodeOrder,
+    protected void createChildNodeChains(BavetNodeBuildPolicy<Solution_> buildPolicy, Score<?> constraintWeight,
             BavetAbstractBiNode<A, B> node) {
         if (!childStreamList.isEmpty()) {
             throw new IllegalStateException("Impossible state: the stream (" + this
                     + ") has an non-empty childStreamList (" + childStreamList + ") but it's a groupBy bridge.");
         }
+        BavetGroupBiNode<NewA, ResultContainer_, NewB> groupNode = groupStream.createNodeChain(buildPolicy,
+                constraintWeight, null);
```
ge0ffrey, line 73: Interesting approach by pushing that here. It's inline with the rest of what Bavet currently does, so +1.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -16,39 +16,39 @@
...
 
     private final BavetConstraintSession session;
 
-    private int nodeOrderMaximum = 0;
+    private int nextNodeIndex = 0;
     private Map<String, BavetScoringNode> constraintIdToScoringNodeMap;
     private Map<BavetJoinConstraintStream<Solution_>, BavetJoinBridgeNode> joinConstraintStreamToJoinBridgeNodeMap =
             new HashMap<>();
-    private Map<BavetAbstractNode, BavetAbstractNode> sharableNodeMap = new HashMap<>();
+    private Map<BavetAbstractNode, BavetAbstractNode> sharableNodeMap = new HashMap<>(0);
```
ge0ffrey, line 41: Adding that zero is strange. Use at least 16 (or better yet the default): the memory cost of 16 is triviial, especially as only a handfully policy elements will exist, while the cost of sharing 9 nodes will be quite significant already.

By the way, due to from() filtering on initialized variables only, there is pretty much always node sharing.

So my advise is to remove this line's change.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -77,4 +77,22 @@ public int getNodeOrderMaximum() {
         return joinConstraintStreamToJoinBridgeNodeMap;
     }
 
+    public List<BavetNode> getCreatedNodes() {
+        // Make a sequential list of unique nodes.
+        SortedMap<Integer, BavetNode> nodeIndexToNodeMap = Stream.concat(sharableNodeMap.keySet().stream(),
+                constraintIdToScoringNodeMap.values().stream())
+                .distinct()
```
ge0ffrey, line 84: Why is the distinct needed on the nodes? I'd argue if they are not distinct, then node sharing is broken and we want it to fail fast.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -77,4 +77,22 @@ public int getNodeOrderMaximum() {
...
+                .collect(Collectors.toMap(k -> k.getNodeIndex(), Function.identity(), (a, b) -> {
+                    throw new IllegalStateException("Impossible state: nodes (" + a + ") and (" + b + ") share index.");
```
ge0ffrey, line 86:
```suggestion
                    throw new IllegalStateException("Impossible state: 2 nodes (" + a + ", " + b + ") share the same index (" + a.getNodeIndex() + ").");
```

ge0ffrey: Motivation: grammar ("the", etc) and show the node index in that error message if we have it available anyway.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
ge0ffrey: 100% agree that is worth it.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
ge0ffrey: By removing it, this test also doesn't run on CS-D any more?
+0 to remove if it's fully duplicate for Bavet, including the actual scoreCalculation verification.
Your call.

If we do keep it, maybe it should be renamed to something like "nodeIndexChallenger" or "joinFilterLeftBridgeVersusJoinFilterRightBridge"? Idunno.

ge0ffrey (approved):

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStreamTest.java
triceo: CS-D never had node indexes, we need to worry about these things there.
Everything else this test does is covered elsewhere.

triceo:
Jenkins please retest this.

PR #1065, opened 2020-12-07
PLANNER-2265 Add constraints-drl property

rsynek:
### JIRA
https://issues.redhat.com/browse/PLANNER-2265

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
@@ -414,6 +414,21 @@ public class ConfigUtils {
         return memberAccessor;
     }
 
+    public static String abbreviate(List<String> list, int limit) {
+        String abbreviation = "";
+        if (list != null) {
+            abbreviation = list.stream().limit(limit).collect(Collectors.joining(", "));
+            if (list.size() > limit) {
+                abbreviation += ", ...";
+            }
+        }
+        return abbreviation;
+    }
+
+    public static String abbreviate(List<String> list) {
+        return abbreviate(list, 3);
+    }
+
     // ************************************************************************
     // Private constructor
     // ************************************************************************
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.drools.core.io.impl.ClassPathResource;
@@ -108,16 +109,27 @@ public class ScoreDirectorFactoryFactory<Solution_, Score_ extends Score<Score_>
                 .filter(Objects::nonNull).count() > 1) {
             List<String> scoreDirectorFactoryPropertyList = new ArrayList<>(4);
             if (easyScoreDirectorFactory != null) {
-                scoreDirectorFactoryPropertyList.add("an easyScoreCalculatorClass");
+                scoreDirectorFactoryPropertyList
+                        .add("an easyScoreCalculatorClass (" + config.getEasyScoreCalculatorClass().getName() + ")");
             }
             if (constraintStreamScoreDirectorFactory != null) {
-                scoreDirectorFactoryPropertyList.add("a constraintProviderClass");
+                scoreDirectorFactoryPropertyList
+                        .add("a constraintProviderClass (" + config.getConstraintProviderClass().getName() + ")");
             }
             if (incrementalScoreDirectorFactory != null) {
-                scoreDirectorFactoryPropertyList.add("an incrementalScoreCalculatorClass");
+                scoreDirectorFactoryPropertyList.add(
+                        "an incrementalScoreCalculatorClass (" + config.getIncrementalScoreCalculatorClass().getName() + ")");
             }
             if (droolsScoreDirectorFactory != null) {
-                scoreDirectorFactoryPropertyList.add("a droolsScoreDirectorFactory");
+                String abbreviatedScoreDrlList = ConfigUtils.abbreviate(config.getScoreDrlList());
+                String abbreviatedScoreDrlFileList = config.getScoreDrlFileList() == null ? ""
+                        : ConfigUtils.abbreviate(config.getScoreDrlFileList()
+                                .stream()
+                                .map(File::getName)
+                                .collect(Collectors.toList()));
+                scoreDirectorFactoryPropertyList
+                        .add("a scoreDrlList (" + abbreviatedScoreDrlList + ") or a scoreDrlFileList ("
+                                + abbreviatedScoreDrlFileList + ")");
             }
             throw new IllegalArgumentException("The scoreDirectorFactory cannot have "
                     + String.join(" and ", scoreDirectorFactoryPropertyList) + " together.");
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
@@ -50,13 +50,10 @@ import org.optaplanner.core.impl.score.director.easy.EasyScoreDirectorFactory;
 import org.optaplanner.core.impl.score.director.incremental.IncrementalScoreDirectorFactory;
 import org.optaplanner.core.impl.score.director.stream.ConstraintStreamScoreDirectorFactory;
 import org.optaplanner.core.impl.score.trend.InitializingScoreTrend;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class ScoreDirectorFactoryFactory<Solution_, Score_ extends Score<Score_>> {
 
     private static final String GENERATE_DROOLS_TEST_ON_ERROR_PROPERTY_NAME = "optaplanner.drools.generateTestOnError";
-    private static final Logger logger = LoggerFactory.getLogger(ScoreDirectorFactoryFactory.class);
 
     private final ScoreDirectorFactoryConfig config;
 
@@ -96,14 +93,39 @@ public class ScoreDirectorFactoryFactory<Solution_, Score_ extends Score<Score_>
 
     protected AbstractScoreDirectorFactory<Solution_, Score_> decideMultipleScoreDirectorFactories(
             ClassLoader classLoader, SolutionDescriptor<Solution_> solutionDescriptor) {
-        AbstractScoreDirectorFactory<Solution_, Score_> easyScoreDirectorFactory =
+        EasyScoreDirectorFactory<Solution_, Score_> easyScoreDirectorFactory =
                 buildEasyScoreDirectorFactory(solutionDescriptor);
-        AbstractScoreDirectorFactory<Solution_, Score_> constraintStreamScoreDirectorFactory =
+        ConstraintStreamScoreDirectorFactory<Solution_, Score_> constraintStreamScoreDirectorFactory =
                 buildConstraintStreamScoreDirectorFactory(solutionDescriptor);
-        AbstractScoreDirectorFactory<Solution_, Score_> incrementalScoreDirectorFactory =
+        IncrementalScoreDirectorFactory<Solution_, Score_> incrementalScoreDirectorFactory =
                 buildIncrementalScoreDirectorFactory(solutionDescriptor);
-        AbstractScoreDirectorFactory<Solution_, Score_> droolsScoreDirectorFactory = buildDroolsScoreDirectorFactory(
+        DroolsScoreDirectorFactory<Solution_, Score_> droolsScoreDirectorFactory = buildDroolsScoreDirectorFactory(
                 classLoader, solutionDescriptor);
+
+        checkMultipleScoreDirectorFactoryTypes(easyScoreDirectorFactory, constraintStreamScoreDirectorFactory,
+                incrementalScoreDirectorFactory, droolsScoreDirectorFactory);
+
+        AbstractScoreDirectorFactory<Solution_, Score_> scoreDirectorFactory;
+        if (easyScoreDirectorFactory != null) {
+            scoreDirectorFactory = easyScoreDirectorFactory;
+        } else if (constraintStreamScoreDirectorFactory != null) {
+            scoreDirectorFactory = constraintStreamScoreDirectorFactory;
+        } else if (incrementalScoreDirectorFactory != null) {
+            scoreDirectorFactory = incrementalScoreDirectorFactory;
+        } else if (droolsScoreDirectorFactory != null) {
+            scoreDirectorFactory = droolsScoreDirectorFactory;
+        } else {
+            throw new IllegalArgumentException("The scoreDirectorFactory lacks a configuration for an "
+                    + "easyScoreCalculatorClass, a constraintProviderClass, an incrementalScoreCalculatorClass or a droolsScoreDirectorFactory.");
+        }
+
+        return scoreDirectorFactory;
+    }
+
+    private void checkMultipleScoreDirectorFactoryTypes(EasyScoreDirectorFactory easyScoreDirectorFactory,
+            ConstraintStreamScoreDirectorFactory constraintStreamScoreDirectorFactory,
+            IncrementalScoreDirectorFactory incrementalScoreDirectorFactory,
+            DroolsScoreDirectorFactory droolsScoreDirectorFactory) {
         if (Stream.of(easyScoreDirectorFactory, constraintStreamScoreDirectorFactory,
                 incrementalScoreDirectorFactory, droolsScoreDirectorFactory)
                 .filter(Objects::nonNull).count() > 1) {
@@ -134,21 +156,6 @@ public class ScoreDirectorFactoryFactory<Solution_, Score_ extends Score<Score_>
             throw new IllegalArgumentException("The scoreDirectorFactory cannot have "
                     + String.join(" and ", scoreDirectorFactoryPropertyList) + " together.");
         }
-        AbstractScoreDirectorFactory<Solution_, Score_> scoreDirectorFactory;
-        if (easyScoreDirectorFactory != null) {
-            scoreDirectorFactory = easyScoreDirectorFactory;
-        } else if (constraintStreamScoreDirectorFactory != null) {
-            scoreDirectorFactory = constraintStreamScoreDirectorFactory;
-        } else if (incrementalScoreDirectorFactory != null) {
-            scoreDirectorFactory = incrementalScoreDirectorFactory;
-        } else if (droolsScoreDirectorFactory != null) {
-            scoreDirectorFactory = droolsScoreDirectorFactory;
-        } else {
-            throw new IllegalArgumentException("The scoreDirectorFactory lacks a configuration for an "
-                    + "easyScoreCalculatorClass, a constraintProviderClass, an incrementalScoreCalculatorClass or a droolsScoreDirectorFactory.");
-        }
-
-        return scoreDirectorFactory;
     }
 
     protected EasyScoreDirectorFactory<Solution_, Score_> buildEasyScoreDirectorFactory(
--- a/optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
+++ b/optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
@@ -508,6 +508,10 @@ optaplanner.solver-config-xml::
 A classpath resource to read the solver configuration XML.
 Defaults to `solverConfig.xml`.
 If this property isn't specified, that file is optional.
+optaplanner.score-drl::
+A classpath resource to read the score DRL.
+Defaults to `constraints.drl`.
+Do not define this property when a `ConstraintProvider`, `EasyScoreCalculator` or `IncrementalScoreCalculator` class exists.
 optaplanner.solver.environment-mode::
 Enable runtime assertions to detect common bugs in your implementation during development.
 optaplanner.solver.daemon::
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
@@ -25,14 +25,6 @@
     <dependency>
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-quarkus</artifactId>
-      <exclusions>
-        <!-- The constraints.drl requires this dependency to be excluded, otherwise the quarkus-maven-plugin
-             downloads it at runtime. -->
-        <exclusion>
-          <groupId>org.kie.kogito</groupId>
-          <artifactId>drools-core-static</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
 
     <dependency>
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
@@ -74,6 +74,11 @@
       </exclusions>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
     <!-- Required to test Quarkus with constraints DRL. -->
     <dependency>
       <groupId>org.kie.kogito</groupId>
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
@@ -25,6 +25,14 @@
     <dependency>
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-quarkus</artifactId>
+      <exclusions>
+        <!-- The constraints.drl requires this dependency to be excluded, otherwise the quarkus-maven-plugin
+             downloads it at runtime. -->
+        <exclusion>
+          <groupId>org.kie.kogito</groupId>
+          <artifactId>drools-core-static</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
@@ -64,6 +72,12 @@
           <artifactId>org.apache.sling.javax.activation</artifactId>
         </exclusion>
       </exclusions>
+      <scope>test</scope>
+    </dependency>
+    <!-- Required to test Quarkus with constraints DRL. -->
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>drools-core-dynamic</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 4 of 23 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -139,6 +144,29 @@ public ScoreDirectorFactoryFactory(ScoreDirectorFactoryConfig config) {
         return scoreDirectorFactory;
     }
 
+    protected String getDrlResourcesAndFilesOverview() {
```
ge0ffrey, line 147: It's not a simple getter and it's for a toString, so how about we call it `composeScoreDrlAndScoreDrlFileToString`?

It might not be clear for everyone that DrlResourcesAndFiles applies to scoreDrl and scoreDrlFile.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -108,16 +109,20 @@ public ScoreDirectorFactoryFactory(ScoreDirectorFactoryConfig config) {
...
             }
             if (incrementalScoreDirectorFactory != null) {
-                scoreDirectorFactoryPropertyList.add("an incrementalScoreCalculatorClass");
+                scoreDirectorFactoryPropertyList.add(
+                        "an incrementalScoreCalculatorClass (" + config.getIncrementalScoreCalculatorClass().getName() + ")");
             }
             if (droolsScoreDirectorFactory != null) {
-                scoreDirectorFactoryPropertyList.add("a droolsScoreDirectorFactory");
+                scoreDirectorFactoryPropertyList
+                        .add("a droolsScoreDirectorFactory (" + getDrlResourcesAndFilesOverview() + ")");
```
ge0ffrey, line 125: It might not be clear for everyone that droolsScoreDirectorFactory applies to scoreDrl and scoreDrlFile in their solverConfig.xml, making it much harder to fix the error.

ge0ffrey: Can we do something like this?

```
                        .add("a scoreDrlList (" + ConfigUtils.limit(solverConfig.getScoreDrlList())
                        + ") or a scoreDrlFileList (" + ConfigUtils.limit(solverConfig.getScoreFileDrlList().stream().map(File::getName).toList()) + "));
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
ge0ffrey: Update: I think we can remove this entire method if we add `ConfigUtils.abbreviate(List<String> list)` which calls `ConfigUtils.abbreviate(list, 3)` for which 3 is the limit size.

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
```diff
@@ -508,6 +508,10 @@ optaplanner.solver-config-xml::
 A classpath resource to read the solver configuration XML.
 Defaults to `solverConfig.xml`.
 If this property isn't specified, that file is optional.
+optaplanner.constraints-drl::
+A classpath resource to read the solver score DRL.
+Defaults to `constraints.drl`.
+This property must NOT be defined in case a `ConstraintProvider` implementation or other score calculation type exists.
```
ge0ffrey, line 514: style: "must not be defined" is too passive tense (= too verbose)

```suggestion
Do note define this property when a `ConstraintProvider` implementation or other score calculation type exists.
```

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
```diff
@@ -508,6 +508,10 @@ optaplanner.solver-config-xml::
 A classpath resource to read the solver configuration XML.
 Defaults to `solverConfig.xml`.
 If this property isn't specified, that file is optional.
+optaplanner.constraints-drl::
+A classpath resource to read the solver score DRL.
```
ge0ffrey, line 512: Very soft suggestion: I can't explain why, but "solver score DRL" doesn't sound right to me, maybe because it implies there is a difference between "score DRL" and "solver score DRL"? (which there is not). 3 lines, above, "Solver configuration XML" is right. 
```suggestion
A classpath resource to read the score DRL.
```

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
```diff
@@ -508,6 +508,10 @@ optaplanner.solver-config-xml::
 A classpath resource to read the solver configuration XML.
 Defaults to `solverConfig.xml`.
 If this property isn't specified, that file is optional.
+optaplanner.constraints-drl::
```
ge0ffrey, line 511: This is documented in the Spring boot section. We need it copy paste into the Quarkus section too.

ge0ffrey: Order wise, I suggest we follow the order of SolverConfig.java
So this property should come after optaplanner.solver.move-thread-count and before optaplanner.solver.termination.spent-limit

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerBuildTimeConfig.java
```diff
@@ -31,6 +32,8 @@
 public class OptaPlannerBuildTimeConfig {
 
     public static final String DEFAULT_SOLVER_CONFIG_URL = "solverConfig.xml";
+    public static final String DEFAULT_CONSTRAINTS_DRL_URL = "constraints.drl";
+    public static final String CONSTRAINTS_DRL_PROPERTY = "quarkus.optaplanner.constraints-drl";
```
ge0ffrey, line 36: None of the other properties have a hard coded string like this in the non-test classes. Can we move it to the test class?

rsynek: I see your point, however, the property name is used also in exception messages when we fail fast.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerBuildTimeConfig.java
```diff
@@ -40,6 +43,14 @@
     @ConfigItem
     Optional<String> solverConfigXml;
 
+    /**
+     * A classpath resource to read the solver score DRL.
+     * Defaults to "{@link #DEFAULT_CONSTRAINTS_DRL_URL}".
+     * This property must NOT be defined in case a {@link ConstraintProvider} implementation exists.
```
ge0ffrey, line 49: See
```suggestion
     * Do note define this property when a {@link ConstraintProvider} class, {@link EasyScoreCalculator} or {@link IncrementalScoreCalculator} exists.
```

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
ge0ffrey: Update: more specific is better:
```suggestion
Do note define this property when a `ConstraintProvider`, `EasyScoreCalculator` or `IncrementalScoreCalculator` class exists.
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerBuildTimeConfig.java
```diff
@@ -40,6 +43,14 @@
...
+     */
+    @ConfigItem
+    Optional<String> constraintsDrl;
```
ge0ffrey, line 52: (important)

Should this be constraintsDrl or scoreDRL? (regardless of what we decided before)

I'd argue there is no point in naming things in application.properties and solverConfig.xml differently if they do the same thing. It just creates unneeded complexity and takes a user's brainpower unnecessarily. Wdyt?

ge0ffrey: On the other hand, it does indeed default to `constraints.drl`, something we changed intentionally from `scoreRules.drl`

ge0ffrey: Worth discussing I think on bluejeans. Let me finish the review first.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -211,31 +214,77 @@ private void applySolverProperties(RecorderContext recorderContext,
...
+        if (scoreDirectorFactoryConfig.getEasyScoreCalculatorClass() == null
+                && scoreDirectorFactoryConfig.getConstraintProviderClass() == null
+                && scoreDirectorFactoryConfig.getIncrementalScoreCalculatorClass() == null
+                && scoreDirectorFactoryConfig.getScoreDrlList() == null) {
+            throw new IllegalStateException("No classes found that implement "
+                    + EasyScoreCalculator.class.getSimpleName() + ", "
+                    + ConstraintProvider.class.getSimpleName() + " or "
+                    + IncrementalScoreCalculator.class.getSimpleName() + ".\n"
+                    + "Neither was a property " + OptaPlannerBuildTimeConfig.CONSTRAINTS_DRL_PROPERTY + " defined, nor a "
+                    + OptaPlannerBuildTimeConfig.DEFAULT_CONSTRAINTS_DRL_URL + " resource found.\n");
```
ge0ffrey, line 285: I'd argue that this fail fast duplicates a fail-fast in SolverFactory.create(solverConfig).

Also, if we ever support SolverConfigCustomizer in Spring Boot, we can do these checks here yet, because the customer might "fix it".

ge0ffrey: So far my philosophy has been that the Spring Boot starter and Quarkus extension should not duplicate fail fasts that
1) are detected as fail-fasts in SolverConfig later itself, once the SolverFActory is created
2) have a good error message.

Now, the 2) is debatable here, indeed.

ge0ffrey: => bluejeans topic?

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
rsynek: That's an option of course; I tried to follow the order of fields in the autoconfiguration.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -211,31 +214,77 @@ private void applySolverProperties(RecorderContext recorderContext,
...
             }
-            if (scoreDirectorFactoryConfig.getEasyScoreCalculatorClass() == null
-                    && scoreDirectorFactoryConfig.getConstraintProviderClass() == null
-                    && scoreDirectorFactoryConfig.getIncrementalScoreCalculatorClass() == null
-                    && scoreDirectorFactoryConfig.getScoreDrlList() == null) {
-                throw new IllegalStateException("No classes found that implement "
-                        + EasyScoreCalculator.class.getSimpleName() + ", "
-                        + ConstraintProvider.class.getSimpleName() + " or "
-                        + IncrementalScoreCalculator.class.getSimpleName() + ", nor a "
-                        + SolverBuildTimeConfig.DEFAULT_SCORE_DRL_URL + " resource.");
```
ge0ffrey, line 235: Ow, it was here already...

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
rsynek: Sure thing; I like the idea of reposrting the `DrlList` and `DrlFileList` separately, which makes this method obsolete.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
rsynek: :) oh yes, but nevermind, it's a good topic to go through.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -211,31 +214,77 @@ private void applySolverProperties(RecorderContext recorderContext,
...
+        if (solverConfig.getScoreDirectorFactoryConfig().getScoreDrlList() != null) {
             ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
-            if (classLoader.getResource(SolverBuildTimeConfig.DEFAULT_SCORE_DRL_URL) != null) {
-                scoreDirectorFactoryConfig.setScoreDrlList(Collections.singletonList(
-                        SolverBuildTimeConfig.DEFAULT_SCORE_DRL_URL));
+            try {
+                Class.forName("org.drools.dynamic.DynamicComponentsSupplier", false, classLoader);
+            } catch (ClassNotFoundException e) {
+                throw new IllegalStateException(
+                        "Constraints DRL has been detected, but drools-core-dynamic is not on classpath."
```
ge0ffrey, line 242: We put a newline between the "error fact" and the "error maybe suggestion":
```suggestion
                        "Using scoreDRL in Quarkus, but the dependency drools-core-dynamic is not on the classpath.\n"
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -211,31 +214,77 @@ private void applySolverProperties(RecorderContext recorderContext,
...
+                                + " Maybe try adding it. Please note that constraints DRL does not work in the native mode.");
```
ge0ffrey, line 243: I tend to avoid "it" references in the "error maybe fix" to the previous line, so users can "just fix it and move on" faster. Especially if the "error fact" is very long and complex to parse (hard to see in a console window, especially if it doesn't wrap and because the error message only starts at column 70 or so)
```suggestion
                                + "\nMaybe add the dependency org.kie.kogito:drools-core-dynamic and exclude the dependency org.kie.kogito:drools-core-static.");
```

ge0ffrey: For the part about not working in native mode: that doesn't belong here, but in the Quarkus integration chapter docs, especially on the constraints-drl property.
Also, if the do try to run it in native mode with a scoreDRL, the error message should be proper - but that's a different check than this one - can we somehow detect that?

ge0ffrey: Also, what's the error message if I add dynamic to the classpath but forget to exclude static?

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerBuildTimeConfig.java
rsynek: Agreed to change the property name to `optaplanner.constraints-drl` later(TODO: file Jira).
`optaplanner.score-drl`
The default stays `constraints.drl`.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
rsynek: Agreed to leave as it is.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
rsynek: Agreed: let's provide fail fast checks for both.

"Maybe" suggestion for native: either use ConstraintProvider or don't use native.

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
rsynek: Decision: respect the order of field in the autoconfiguration.

rsynek: ad the Quarkus section: I was wandering: how could I have missed the quarkus section of our documentation? The answer is quite prosaic: the section is empty so far.
I assigned the https://issues.redhat.com/browse/PLANNER-2284 I filed to myself; not in scope of this PR, though.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerBuildTimeConfig.java
rsynek: https://issues.redhat.com/browse/PLANNER-2285

rsynek:
jenkins retest this

rsynek:
@ge0ffrey please note that the `mvn quarkus:dev` does not work with a constraints DRL. I did some experiments regarding what works which drools-core-static/dynamic and what does not. We should meet and discuss it shortly.

rsynek:
jenkins retest this

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
```diff
@@ -25,6 +25,14 @@
     <dependency>
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-quarkus</artifactId>
+      <exclusions>
+        <!-- The constraints.drl requires this dependency to be excluded, otherwise the quarkus-maven-plugin
+             downloads it at runtime. -->
+        <exclusion>
```
rsynek, line 31: This solves the issue with `drools-core-static` getting on the classpath. A user still has to exclude this dependency from the dependency on `optaplanner-quarkus` and `optaplanner-quarkus-json`, etc.

optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -414,6 +414,21 @@ public static int resolvePoolSize(String propertyName, String value, String... m
         return memberAccessor;
     }
 
+    public static String abbreviate(List<String> list, int limit) {
+        String abbreviation = "";
+        if (list != null) {
```
ge0ffrey, line 419: Nitpick:
I personally think the method should not accept null lists. Just add a javadoc that has "@param list never null"

But if should accept null too, we treat is as an exceptionally situation (so handle it with an if statement before doing the real code)
```

if (list == null) {
   return "";
}
String abbreviation = ...
```

ge0ffrey: ah, I see below it's easier for it to also accept null.

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
```diff
@@ -508,6 +508,10 @@ optaplanner.solver-config-xml::
 A classpath resource to read the solver configuration XML.
 Defaults to `solverConfig.xml`.
 If this property isn't specified, that file is optional.
+optaplanner.score-drl::
+A classpath resource to read the score DRL.
+Defaults to `constraints.drl`.
+Do not define this property when a `ConstraintProvider`, `EasyScoreCalculator` or `IncrementalScoreCalculator` class exists.
```
ge0ffrey, line 514: Great work. Mind if I add this issue on your backlog (not for this PR)? https://issues.redhat.com/browse/PLANNER-2288

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
ge0ffrey: Intresting, let's dicuss indeed

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -210,32 +213,95 @@ private void applySolverProperties(RecorderContext recorderContext,
...
-                throw new IllegalStateException("No classes found that implement "
-                        + EasyScoreCalculator.class.getSimpleName() + ", "
-                        + ConstraintProvider.class.getSimpleName() + " or "
-                        + IncrementalScoreCalculator.class.getSimpleName() + ", nor a "
-                        + SolverBuildTimeConfig.DEFAULT_SCORE_DRL_URL + " resource.");
+        }
+
+        if (solverConfig.getScoreDirectorFactoryConfig().getScoreDrlList() != null) {
+            boolean isDroolsDynamicPresent = isClassDefined("org.drools.dynamic.DynamicServiceRegistrySupplier");
+            boolean isDroolsStaticPresent = isClassDefined("org.drools.statics.StaticServiceRegistrySupplier");
```
ge0ffrey, line 238: Nitpick: Personally, I'd inline these booleans in the if statements, as they are only used once.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -210,32 +213,95 @@ private void applySolverProperties(RecorderContext recorderContext,
...
+
+            if (!isDroolsDynamicPresent) {
+                throw new IllegalStateException(
+                        "Using scoreDRL in Quarkus, but the dependency drools-core-dynamic is not on the classpath.\n"
```
ge0ffrey, line 242: grammar: "Using scoreDRL in Quarkus, but"
```suggestion
                        "If scoreDRL is used in Quarkus, the dependency drools-core-dynamic must be on the classpath.\n"
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -210,32 +213,95 @@ private void applySolverProperties(RecorderContext recorderContext,
...
+                                + "Maybe add the dependency org.kie.kogito:drools-core-dynamic and exclude the dependency"
+                                + " org.kie.kogito:drools-core-static. Use a ConstraintProvider for a native compilation.");
```
ge0ffrey, line 244:
```suggestion
                                + " org.kie.kogito:drools-core-static."
                                + "\nOr instead, maybe use a " + ConstraintProvider.class.getSimpleName() + " instead of scoreDRL.");
```

ge0ffrey: I don't think native compilation should be mentioned here - this is not a check related to native compilation.
The benefits of using ConstraintProvider over scoreDRL should be in the docs instead.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
rsynek: Myself, I don't see the DRY principle as the only motivation for creating methods or, in this case, local variables. The variable name clearly states what's the purpose of the check. In contrast to that, `isClassDefined("org.drools.dynamic.DynamicServiceRegistrySupplier")` is not so clear about it's purpose as I just deliberately picked the class `DynamicServiceRegistrySupplier` as something pointing to the presence of the `drools-core-dynamic`. The local variables serve here rather in the role of comments.

optaplanner-docs/src/main/asciidoc/Integration/Integration-chapter.adoc
rsynek: No problem, in fact I have already created one (PLANNER-2284) - see one of my earlier responses on this PR.

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
```diff
@@ -200,33 +201,69 @@ private void applySolverProperties(SolverConfig solverConfig) {
         return new ArrayList<>(entityClassSet);
     }
 
-    private void applyScoreDirectorFactoryProperties(SolverConfig solverConfig) {
+    protected void applyScoreDirectorFactoryProperties(SolverConfig solverConfig) {
+        Optional<String> constraintsDrlFromProperty = constraintsDrl();
+        Optional<String> defaultConstraintsDrl = defaultConstraintsDrl();
+        Optional<String> effectiveConstraintsDrl = constraintsDrlFromProperty.map(Optional::of).orElse(defaultConstraintsDrl);
```
ge0ffrey, line 207: (non blocking)
Personally, I am not a fan of using Optional outside of Streams - a nullable var works just as well - see Stuart's and Goetz's talks around this topic.

ge0ffrey (approved):

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/pom.xml
ge0ffrey: As discussed, we won't do this change and remove the fail fast if static is not in the classpath with scoreDRL.

PR #1067, opened 2020-12-08
[7.x] RHPAM-3334 ConstraintCollectors.toMap(): Only remove a key when no value remains

triceo:
(cherry picked from commit 200964881ef5adb4a2f84f001757a21a12775528)


dupliaka (approved):

triceo:
Merged, even though we don't have Jenkins.
GHA is green, and this change is absolutely self-contained, no impact on other repos is possible here.

PR #1076, opened 2020-12-23
Docs: Fix the queen moves in the Local Search chapter

vadimeisenbergibm:
The current moves are wrong.


ge0ffrey:
ok to test

ge0ffrey (approved):
Thanks for spotting and fixing this :)

PR #1078, opened 2020-12-28
PLANNER-2163 Use Drools groupBy node throughout CS-D

triceo:
Implements DROOLS-5875 in CS-D. This allows us to remove the vast majority of CS-D code, which until now had to deal with 3 different underlying groupBy implementations. CS-D is finally a thin wrapper around Drools executable model, as it always should have been. 

This change is expected to bring some performance benefits to groupBy, both single and double, although double groupBy will never perform well in CS-D.

This code has been tested on `FULL_ASSERT` with `nqueens`, `cloudbalancing` and `machinereassignment`.
The PR depends on https://github.com/kiegroup/kogito-runtimes/pull/990 to compile and test cleanly.


Jenkinsfile
```diff
@@ -43,7 +43,7 @@ pipeline {
         }
         stage('Build OptaPlanner') {
             steps {
-                mavenCleanInstall('optaplanner', false, ['run-code-coverage'], '-Dfull')
+                mavenCleanInstall('optaplanner', false, ['run-code-coverage'], '-Dfull -Dmaven.test.failure.ignore=true')
```
triceo, line 46: Note to self: remove once Kogito has Drools 7.49.
Until then, the tests will be failing and I would therefore not get to see Sonar.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -268,7 +269,7 @@ private void checkMultipleScoreDirectorFactoryTypes(EasyScoreDirectorFactory eas
         }
 
         try {
-            KieBase kieBase = kieHelper.build(ExecutableModelProject.class);
+            KieBase kieBase = kieHelper.build(ExecutableModelProject.class, KieBaseMutabilityOption.DISABLED);
```
triceo, line 272: Although technically unrelated, this change improves DRL performance by a couple percentage points, and I was doing the same thing for CS-D, so I decided to introduce it here, too.

Jenkinsfile
ge0ffrey: 1) Can we add a "expiration condition" here? So basically a comment

    // TODO Remove when drools dependency is 7.49.0

ge0ffrey: 2) Or will this be removed before we merge this PR?

ge0ffrey (approved):
LGTM, although I didn't do an in-depth review.I do like the fact that we're losing almost 8 000 lines of code :)

The disabling of fail on test cannot be merged, of course.

Jenkinsfile
triceo: It will.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/BiLeftHandSide.java
```diff
@@ -0,0 +1,287 @@
...
+        Variable<BiTuple<A, B>> accumulateSource =
+                (Variable<BiTuple<A, B>>) variableFactory.createVariable(BiTuple.class, "source");
+        PatternVariable<B> newPatternVariableB = patternVariableB.bind(accumulateSource, inputA,
+                (b, a) -> new BiTuple<>(a, b));
+        Variable<BiTuple<NewA, NewB>> groupKey =
+                (Variable<BiTuple<NewA, NewB>>) variableFactory.createVariable(BiTuple.class, "groupKey");
+        Variable<NewC> output = variableFactory.createVariable("output");
+        ViewItem<?> innerGroupByPattern = joinViewItemsWithLogicalAnd(patternVariableA, newPatternVariableB);
+        ViewItem<?> groupByPattern = groupBy(innerGroupByPattern, inputA, inputB, groupKey,
+                (a, b) -> new BiTuple<>(keyMappingA.apply(a, b), keyMappingB.apply(a, b)),
```
rsynek, line 219: I think the trick with joining two columns into a `BiTuple` is worth documenting; mainly the motivation.

rsynek (commented):
Not yet done, will finish the review tomorrow.

rsynek (approved):
Nice work, I like the big reduction of the code base and the fact that the new implementation reads much better.

See some suggestions inline; apart of those, please consider adding comments over methods of `*LeftHandSide` classes that would show equivalent DRL code to clearly illustrate their goal.

triceo:
Will fail CI today, temporarily. Things will settle down when the proper SNAPSHOTs are built and distributed.
There was no way around it.

PR #1087, opened 2021-01-09
[KOGITO-1031] Add Maven local deploy

Kevin-Mok:
See: https://issues.redhat.com/browse/KOGITO-1031

This is an accompanying PR to [the one opened in kogito-pipelines](https://github.com/kiegroup/kogito-pipelines/pull/46). 

## Changes
- for PR checks, check out branches in the user's repository where there are matching branch names to the PR (fallback to `master` otherwise)
  - setup Spock tests to test helper function used for this
- deploys all artifacts locally before zipping them altogether to push to Nexus
  * added to provide authentication to push artifacts to PR artifact repository

## Testing
In my test build, you can see that my bot's PR branch was [merged into its target branch](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/kmok/job/KOGITO-1031-full/job/optaplanner-deploy/36/execution/node/90/log/) and used for the build. However for the other repositories where there is no PR or matching branch name, the [target author's `master` branch](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/kmok/job/KOGITO-1031-full/job/optaplanner-deploy/36/execution/node/171/log/) is used (`development` for optaplanner-quickstarts). I was also able to [deploy the artifacts](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/kmok/job/KOGITO-1031-full/job/optaplanner-deploy/36/execution/node/348/log/) to the PR test repository in this build.

<details>
<summary>
</summary>

</details>


Jenkinsfile.deploy
```diff
@@ -248,13 +252,29 @@ void saveReports(boolean allowEmpty=false){
     junit testResults: '**/target/surefire-reports/**/*.xml, **/target/failsafe-reports/**/*.xml', allowEmptyResults: allowEmpty
 }
 
-void checkoutRepo(String repo, String branch, String dirName=repo) {
+void checkoutRepo(String repo, String dirName=repo) {
     dir(dirName) {
         deleteDir()
-        checkout(githubscm.resolveRepository(repo, getGitAuthor(), branch, false))
+        githubscm.checkoutIfExists(repo, getGitAuthor(), getBuildBranch(), 'kiegroup', getFallbackBranch(repo), true)
```
radtriste, line 258: I guess this should be done only if `PR_TARGET_BRANCH` is given, else "old code" would apply

```suggestion
        if(params.PR_TARGET_BRANCH) {
            githubscm.checkoutIfExists(repo, getGitAuthor(), getBuildBranch(), 'kiegroup', getFallbackBranch(repo), true)
        } else {
            checkout(githubscm.resolveRepository(repo, getGitAuthor(), getBuildBranch(), false))
        }
```

Jenkinsfile.deploy
```diff
@@ -248,13 +252,29 @@ void saveReports(boolean allowEmpty=false){
...
     }
 }
 
+void checkoutQuickstarts(String dirName=quickstartsRepository) {
+    dir(dirName) {
+        deleteDir()
+        githubscm.checkoutIfExists(quickstartsRepository, getGitAuthor(), getQuickStartsBranch(), 'kiegroup', getQuickStartsBranch(), true)
```
radtriste, line 265: same apply here
```suggestion
        if(params.PR_TARGET_BRANCH) {
            githubscm.checkoutIfExists(quickstartsRepository, getGitAuthor(), getQuickStartsBranch(), 'kiegroup', getQuickStartsBranch(), true)
        } else {
            checkout(githubscm.resolveRepository(repo, getGitAuthor(), getQuickStartsBranch(), false))
        }
```

Jenkinsfile.deploy
```diff
@@ -248,13 +252,29 @@ void saveReports(boolean allowEmpty=false){
...
+    }
+}
+
+String getFallbackBranch(String repo) {
+    // if not triggered by PR
+    if (!params.PR_TARGET_BRANCH) {
+        return 'master'
+    }
```
radtriste, line 273: folowing previous change
```suggestion
```

.jenkins/tests/pom.xml
```diff
@@ -0,0 +1,300 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
rsynek, line 2: I don't think this project (OptaPlanner) should own definition of dependencies, and their versions, for testing jenkinsfiles. This `pom.xml` brings a maintenance cost, which will be repeated across any repository that includes such tests.
We've already touched this topic with @radtriste and created https://issues.redhat.com/browse/BXMSPROD-1125.

.jenkins/tests/pom.xml
```diff
@@ -0,0 +1,300 @@
...
+
+        <logback.configration>logback-test.xml</logback.configration>
+        <logdir>${project.build.directory}/log</logdir>
+        <test.loglevel>ERROR</test.loglevel>
+        <log.logback.version>1.2.3</log.logback.version>
+        <log.slf4j.version>1.7.25</log.slf4j.version>
+        <build-helper-maven-plugin.version>1.9.1</build-helper-maven-plugin.version>
+
+        <maven.compiler.source>1.6</maven.compiler.source>
+        <maven.compiler.target>1.6</maven.compiler.target>
```
rsynek, line 58: Really 1.6?

.jenkins/tests/pom.xml
```diff
@@ -0,0 +1,300 @@
...
+            <name>Jenkins Releases</name>
+            <url>https://repo.jenkins-ci.org/releases</url>
+        </repository>
+    </repositories>
+
+    <properties>
+        <!-- Do not deploy this artifact -->
+        <maven.deploy.skip>true</maven.deploy.skip>
+
+        <groovy.core.version>2.4.11</groovy.core.version>
```
rsynek, line 36: The same suggestion as here https://github.com/kiegroup/kogito-pipelines/pull/91#discussion_r537371129.

Jenkinsfile.deploy
```diff
@@ -56,6 +57,9 @@ pipeline {
         // Bot author information. Set as params for easy testing.
         string(name: 'GIT_AUTHOR_BOT', defaultValue: 'bsig-gh-bot', description: 'From which author should the PR be created ?')
         string(name: 'BOT_CREDENTIALS_ID', defaultValue: 'bsig-gh-bot', description: 'Credentials for PR creation')
+
+        // PR parameter to be filled out for PR checks
+        string(name: 'PR_TARGET_BRANCH', defaultValue: '', description: 'What is the target branch of the PR?')
```
rsynek, line 62: What is the purpose of this parameter? This `Jenkinsfile.deploy` is not used in PRs, it's used in the nightly and release pipelines.

.jenkins/tests/pom.xml
rsynek: CC @ge0ffrey

Jenkinsfile.deploy
radtriste: here we do want to reuse the nightly pipelin flow for BDD tests from runtimes testing as it should do the exact same deployment part, but on a specific repository.

The problem here is that `optaplanner` repository is handling deploy/promote for all optaplanner repositories (optaplanner, optaweb[...],optaplanner-quickstarts)

On a PR check, we have a `source` branch and a `target` branch. It first checks if similar source branch exists and if not, it fallbacks to target branch.
That could happen if a change happens on both kogito-runtimes and optaplanner, so with same branch name (but not on optaweb[...])
So , in `GIT_BRANCH` param we sent the source branch. but if that branch does not exist on optaweb[...] repos, it will fail and so we need the target branch to checkout correct version (master or 8.1.x or ... ), hence the `PR_TARGET_BRANCH`  parameter ...

In an ideal world, all optaplanner repositories would have their own deploy/promote specific jenkinsfile/jobs and so we could just do the check if source branch is existing before calling the job. We can do that on runtimes/apps/examples/images/operator as each job is specific to the repository itself.

Else we need to copy that deploy job and create a new one but it would be exactly the same Jenkinsfile apart from that parameter and the correct checkouts. and so could be harder to maintain ...

.jenkins/tests/pom.xml
radtriste: I guess we should remove all the maven related stuff. but leave the test files for now, it will be reused in https://issues.redhat.com/browse/BXMSPROD-1125.
@rsynek wdyt ?

Jenkinsfile.deploy
radtriste: @rsynek is there anything we should change or do you agree with that ?

.jenkins/tests/pom.xml
radtriste: @Kevin-Mok I guess you should remove the tests part. will be handled later

Jenkinsfile.deploy
rsynek: I see, thanks for the clarification @radtriste .

rsynek (approved):

PR #1089, opened 2021-01-11
Add the possibility to export TSP data as SVG.

pilhuhn:
This allows to export TSP data as SVG. 
Part of this is to allow multiple exporters to exist for a certain example.

Bildschirmfoto 2021-01-11 um 17 26 23

In here are two exporters for .path.svg which creates a path, which is great for example for tsp-art (see Hands of god attachment), while the .line.svg creates lines between the locations and marks the locations with dots (see pcb442 output)

Note: attachments have an additional .txt ending, as GH is too stupid to grok .svg
[Hands_of_God_and_Adam2.path.svg.txt](https://github.com/kiegroup/optaplanner/files/5796991/Hands_of_God_and_Adam2.path.svg.txt)
[pcb442.line.svg.txt](https://github.com/kiegroup/optaplanner/files/5796993/pcb442.line.svg.txt)

cc @ge0ffrey

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/common/app/CommonApp.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/common/app/CommonApp.java
@@ -18,6 +18,7 @@ package org.optaplanner.examples.common.app;
 
 import java.awt.Component;
 import java.io.File;
+import java.util.Set;
 import java.util.function.BiConsumer;
 
 import javax.swing.WindowConstants;
@@ -129,7 +130,7 @@ public abstract class CommonApp<Solution_> extends LoggingMain {
         solutionBusiness.setDataDir(determineDataDir(dataDirName));
         solutionBusiness.setSolutionFileIO(createSolutionFileIO());
         solutionBusiness.setImporters(createSolutionImporters());
-        solutionBusiness.setExporter(createSolutionExporter());
+        solutionBusiness.setExporters(createSolutionExporters());
         solutionBusiness.updateDataDirs();
         return solutionBusiness;
     }
@@ -156,7 +157,7 @@ public abstract class CommonApp<Solution_> extends LoggingMain {
         return new AbstractSolutionImporter[0];
     }
 
-    protected AbstractSolutionExporter createSolutionExporter() {
+    protected Set<AbstractSolutionExporter> createSolutionExporters() {
         return null;
     }
 
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/common/app/CommonApp.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/common/app/CommonApp.java
@@ -18,6 +18,7 @@ package org.optaplanner.examples.common.app;
 
 import java.awt.Component;
 import java.io.File;
+import java.util.Collections;
 import java.util.Set;
 import java.util.function.BiConsumer;
 
@@ -158,7 +159,7 @@ public abstract class CommonApp<Solution_> extends LoggingMain {
     }
 
     protected Set<AbstractSolutionExporter> createSolutionExporters() {
-        return null;
+        return Collections.emptySet();
     }
 
     public interface ExtraAction<Solution_> {
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
@@ -20,8 +20,10 @@ import static java.util.stream.Collectors.toList;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.swing.SwingUtilities;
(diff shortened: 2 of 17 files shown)


triceo:
OK to test.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/app/CommonApp.java
```diff
@@ -156,7 +157,7 @@ public void init(Component centerForComponent, boolean exitOnClose) {
         return new AbstractSolutionImporter[0];
     }
 
-    protected AbstractSolutionExporter createSolutionExporter() {
+    protected Set<AbstractSolutionExporter> createSolutionExporters() {
         return null;
```
triceo, line 161: `Collections.emptySet()` works better in this situation, I think.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
```diff
@@ -126,16 +128,30 @@ public void setImporters(AbstractSolutionImporter<Solution_>[] importers) {
...
+            this.exporters = new HashSet<>();
+        }
+        this.exporters.add(exporter);
+    }
+
+    public Set<AbstractSolutionExporter> getExporters() {
+        if (this.exporters == null) {
+            return new HashSet<>();
+        }
+        return this.exporters;
```
triceo, line 146: If `exporters` are never null (see previous comment), the handling of that can disappear from this class.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
```diff
@@ -181,8 +197,12 @@ public File getExportDataDir() {
         return exportDataDir;
     }
 
-    public String getExportFileSuffix() {
-        return exporter.getOutputFileSuffix();
+    public Set<String> getExportFileSuffix() {
+        Set<String> suffixes = new HashSet<>(exporters.size());
```
triceo, line 201: I believe this would throw an NPE if `exporters` are allowed to be `null`.

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/app/TspApp.java
```diff
@@ -66,8 +71,13 @@ protected TspPanel createSolutionPanel() {
     }
 
     @Override
-    protected AbstractSolutionExporter createSolutionExporter() {
-        return new TspExporter();
+    protected Set<AbstractSolutionExporter> createSolutionExporters() {
+        Set<AbstractSolutionExporter> exporters = new HashSet<>();
+        exporters.add(new TspExporter());
+        exporters.add(new SvgTspExporter());
+        exporters.add(new SvgTspExporter2());
```
triceo, line 78: Let's only have 1 SVG exporter.
I understand that you may have some specific use case that requires two slightly different SVG, but I'd rather not expose that in a generic example app.

triceo (changes requested):
Thank you for your contribution! I took the liberty of doing a first quick review - there may be more later when we can assign someone properly.

Also, I don't see any test coverage in the PR. The example module is a bit undertested in general, but I believe that at least the SVG export code should have some basic correctness checks. Not necessarily for contents, but perhaps to make sure that the generated XML is well-formed.

optaplanner-examples/src/main/java/org/optaplanner/examples/tsp/app/TspApp.java
pilhuhn: I don't think this is good, as there is no (simple) way of parametrizing one exporter to do different flavours. And as you can see / guess from the rendered SVGs, there are specific use cases.
It may be possible to use an Accessory view, but I think that is way above simple usages.
I did on purpose use two different file suffixes for the different exporters for these use cases.

triceo: OK, if we can give the different exporter classes some descriptive names (as opposed to `1` and `2`), so be it.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/app/CommonApp.java
pilhuhn: I agree.

pilhuhn:
@triceo Can you please pass me a pointer on how to tackle testing?

triceo:
@pilhuhn The Swing app itself is hard to test, which is why we usually don't bother. The SVG exporter shouldn't be, though. 

1. Load some solution.
1. Export the SVG.
1. Check that the exported SVG is a valid XML.

I'm assuming that the problem comes in step 1. You'll have to export a solution in the app, put it in `src/test/resources/...`, deserialize the solution in your test using `SolutionFileIO` and run it through the SVG exporter.

Now that I think of it, it may be too much to ask of a first-time contributor to the project. :-) Feel free to skip it, I'll add the test myself after we merge.

pilhuhn:
On 12 Jan 2021, at 11:01, Lukáš Petrovický wrote:

> @pilhuhn The Swing app itself is hard to test, which is why we usually 
> don't bother. The SVG exporter shouldn't be, though.

Yes. I was accidentally looking at the optaplanner-test module
and did not see anything that remotely looked like testing TSP,
which is why I was asking :)

But I'd happy to just skip this :-)

triceo (approved):

triceo:
Thanks again for the contribution, @pilhuhn.

PR #1104, opened 2021-01-18
PLANNER-1501: Gizmo Solution Cloner

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-1501

### Referenced pull requests

https://github.com/kiegroup/optaplanner/pull/1059

### Checklist

<details>
<summary>
</summary>

</details>

Depends on https://github.com/kiegroup/optaplanner/pull/1059 . This is a draft solution cloner using Gizmo. The code it generates look like this:

```
     // For a Collection
     Collection original = field;
     Collection clone = new ActualCollectionType();
     Iterator iterator = original.iterator();
     while (iterator.hasNext()) {
         Object nextClone = (result from recursion on iterator.next());
         clone.add(nextClone);
     }
     
     // For a Map
     Map original = field;
     Map clone = new ActualMapType();
     Iterator iterator = original.entrySet().iterator();
     while (iterator.hasNext()) {
          Entry next = iterator.next();
          nextClone = (result from recursion on next.getValue());
          clone.put(next.getKey(), nextClone);
     }
    
     // For an array
     Object[] original = field;
     Object[] clone = new Object[original.length];
    
     for (int i = 0; i < original.length; i++) {
         clone[i] = (result from recursion on original[i]);
     }
    
     // For an entity
     if (original instanceof SubclassOfEntity1) {
         SubclassOfEntity1 original = field;
         SubclassOfEntity1 clone = new SubclassOfEntity1();
    
           // shallowly clone fields using writeShallowCloneInstructions()
          // for any deeply cloned field, do recursion on it
      } else if (original instanceof SubclassOfEntity2) {
        // ...
     }
```

An example decompiled result (with meaningful variable names, and generated getters/setters replaced with getters/setters):

```
// $FF: synthetic class
public class $optaplanner$__TestdataQuarkusSolution implements SolutionCloner {
    public $optaplanner$__TestdataQuarkusSolution() {
    }

    public Object cloneSolution(Object original) {
        TestdataQuarkusSolution clone = new TestdataQuarkusSolution();
        SimpleScore scoreClone = ((TestdataQuarkusSolution)original).getScore();
        clone.setScore(scoreClone);
        List valueListClone = ((TestdataQuarkusSolution)original).getValueList();
        clone.setValueList(valueListClone);
        HashMap existingCloneMap = new HashMap();
        List originalEntityList = ((TestdataQuarkusSolution)original).getEntityList();
        int entityListSize = ((Collection)originalEntityList).size();
        Iterator originalEntityListIterator = ((Iterable)originalEntityList).iterator();

        ArrayList cloneEntityList;
        TestdataQuarkusEntity cloneEntity;
        for(cloneEntityList = new ArrayList(entityListSize); originalEntityListIterator.hasNext(); ((Collection)cloneEntityList).add(cloneEntity)) {
            Object originalEntity = originalEntityListIterator.next();
            if (!(originalEntity instanceof TestdataQuarkusEntity)) {
                cloneEntity = (TestdataQuarkusEntity)originalEntity;
            } else {
                cloneEntity = $cloneorg_optaplanner_quarkus_testdata_normal_domain_TestdataQuarkusEntity((TestdataQuarkusEntity)originalEntity, (Map)existingCloneMap);
            }
        }

        List cloneEntityListResult = (List)cloneEntityList;
        clone.setEntityList(cloneEntityListResult);
        return clone;
    }

    private static TestdataQuarkusEntity $cloneorg_optaplanner_quarkus_testdata_normal_domain_TestdataQuarkusEntity(TestdataQuarkusEntity original, Map existingCloneMap) {
        TestdataQuarkusEntity clone = new TestdataQuarkusEntity();
        if (!existingCloneMap.containsKey(original)) {
            String originalValue = original.getValue();
            clone.setValue(originalValue);
            return clone;
        } else {
            return (TestdataQuarkusEntity)existingCloneMap.get(original);
        }
    }
}
```


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,200 @@
...
+     *        from the field's declaring class (ex: subclass).
+     * @param actualValueClass The class of the value that is currently assigned
+     *        to the field; can be different from the field type
+     *        (ex: for the field "List myList", the actual value
+     *        class might be ArrayList).
+     * @param fieldDeepClonedMemoization Memoization map for (field, owningClass) decisions.
+     * @param actualValueClassDeepClonedMemoization Memoization maps for actualValueClass decisions.
+     * @return True iff the field should be deep cloned with a particular value.
+     */
+    public static boolean getMemorizedDeepCloneDecision(final SolutionDescriptor<?> solutionDescriptor,
```
triceo, line 55: Please be consistent with your `final`s.
In the case of OptaPlanner and method arguments, this means don't use them.
(Please check the rest of the class, too.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,200 @@
...
+     * @param solutionDescriptor The solution descriptor for the solution class
+     * @param field The field to get the deep cloning decision of
+     * @param owningClass The class that owns the field; can be different
+     *        from the field's declaring class (ex: subclass).
+     * @param actualValueClass The class of the value that is currently assigned
+     *        to the field; can be different from the field type
+     *        (ex: for the field "List myList", the actual value
+     *        class might be ArrayList).
+     * @param fieldDeepClonedMemoization Memoization map for (field, owningClass) decisions.
+     * @param actualValueClassDeepClonedMemoization Memoization maps for actualValueClass decisions.
```
triceo, line 52: OptaPlanner Javadoc convention: param doesn't start with an uppercase letter, therefore the whole expression is not a sentence. Therefore no full stop at the end.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.optaplanner.core.api.domain.solution.cloner.DeepPlanningClone;
+import org.optaplanner.core.impl.domain.common.ReflectionHelper;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+public class DeepCloningUtils {
```
triceo, line 32: Can be final.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,200 @@
...
+    private DeepCloningUtils() {
```
triceo, line 33: We usually add a comment like "No external instances.", so that it clear what the intention behind the empty method is.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
triceo: Also, does it have to be public?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
```diff
@@ -0,0 +1,53 @@
+package org.optaplanner.core.impl.domain.solution.cloner.gizmo;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.optaplanner.core.api.domain.common.DomainAccessType;
+import org.optaplanner.core.api.domain.solution.cloner.SolutionCloner;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+public class GizmoSolutionClonerFactory {
```
triceo, line 10: Can be final.
I'd argue `final` is a useful keyword, signifying your intention for this class to not be overridden.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
```diff
@@ -0,0 +1,53 @@
...
+                        ") the classpath or modulepath must contain io.quarkus.gizmo:gizmo.\n" +
+                        "Maybe add a dependency to io.quarkus.gizmo:gizmo.");
+            }
+            SolutionCloner<T> cloner = GizmoSolutionClonerImplementor.createClonerFor(solutionDescriptor);
+            solutionClonerMap.put(gizmoMemberAccessorClassName, cloner);
+            return cloner;
+        }
+    }
+
+    private GizmoSolutionClonerFactory() {
```
triceo, line 51: In one of the previous classes, the constructor is first in the class. Here, it is last.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -0,0 +1,764 @@
...
+                MethodDescriptor.ofMethod(Map.class, "put", Object.class, Object.class, Object.class),
+                createdCloneMap, thisObj, clone);
+
+        for (GizmoMemberDescriptor shallowlyClonedField : solutionInfo.getShallowClonedMemberDescriptors()) {
+            writeShallowCloneInstructions(solutionInfo, noCloneBranch, shallowlyClonedField, thisObj, clone, createdCloneMap);
+        }
+
+        for (Field deeplyClonedField : solutionInfo.getDeepClonedFields()) {
+            GizmoMemberDescriptor gizmoMemberDescriptor = solutionInfo.getMemberDescriptorForField(deeplyClonedField);
+            final ResultHandle[] resultHandleHolder = new ResultHandle[1];
```
triceo, line 201: If the point of this is to simply have an effectively final var to put in the lambdas, `AtomicReference<ResultHandle>` will probably make that point a bit clearer. Extraction of the handling code to a separate method would probably allow to get rid of either.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -0,0 +1,764 @@
...
+                        Thread.currentThread().getContextClassLoader());
+            }
+
+            if (type != null && !isArray) {
+                entitySubclasses = solutionInfo.getSolutionDescriptor().getEntityClassSet().stream()
+                        .filter(type::isAssignableFrom).collect(Collectors.toList());
+            }
+
+            final List<Class<?>> finalEntitySubclasses = entitySubclasses;
+            final Class<?> finalType = type;
```
triceo, line 268: I really don't like the fact that we're mixing final and non-final vars.
I know it's because of the lambdas, but there has to be a nicer way.
How about extracting this stuff to a separate method, where these would be the arguments?

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/TestdataEntity.java
```diff
@@ -37,7 +37,7 @@
         return entityDescriptor.getGenuineVariableDescriptor("value");
     }
 
-    private TestdataValue value;
+    public TestdataValue value;
```
triceo, line 40: I have to say I absolutely hate this. Public fields are an antipattern, and we are now proliferating it throughout the tests.
There has to be a better way of testing this.

triceo (commented):
I have added some comments.
Many of them apply to multiple files, even though they're really only present in one. (Would be too much repetition.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
Christopher-Chianelli: Since `gizmo` is a subpackage of `cloner` that uses it, it must be public. I rather have the classes exclusive Gizmo Solution Cloner in their own subpackage.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: So that what the mutable holder class in java called!

Why it is set up like this is to prevent errors from assuming the descriptor is a field or a method (or accidentally using the wrong one inside an if guard). That method was designed to have no return value (as the MemberAccessor code only use it for generating code, never for returning ResultHandles to the outside.

I could add a compute version of the method that takes two functions that returns ResultHandle (one accepts field descriptor, another accepts method descriptor), which removes the need to store the result in a holder.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: Done via adding read/write methods to GizmoMemberDescriptor

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: Done via adding Read/Write methods to GizmoMemberDescriptor

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
Christopher-Chianelli: Move the other constructor to last to match rest of optaplanner-core

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/TestdataEntity.java
Christopher-Chianelli: I also hate it. If there were some way to shallow clone a Java Object (via a method or bytecode instruction), then in theory we could do the planning cloning using only the member accessors (in which case the old restrictions apply: you can use Gizmo if and only if all planning annotations are on public members).

radtriste:
Jenkins retest this

radtriste:
Jenkins retest this

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/domain/TestdataEntity.java
triceo: I have accepted the limitation that Gizmo would only be useful in Quarkus - we can't change that. But can we do anything about the tests?

If we only run the Gizmo tests in Quarkus, we do not have to make everything public. Alternatively, we could introduce some special Testdata domain that is only use for Gizmo, and keep the rest of the domain as is.

Christopher-Chianelli: The main issue is sharing the test cases between the two implementations of SolutionCloners (FieldAccessingSolutionCloner and GizmoSolutionCloner). It might be possible if we import the test-jar of OptaPlanner Core into OptaPlanner Quarkus. But then we don't have testing for the plain java case (it can work in Java, just with limitations).

triceo: Or we give up on running the same tests in both environments.
Sonds like copy-pasting the test and specializing it for Gizmo may end up being the best solution.

triceo: I think there may be a different solution, but it is a hail mary pass. The only reason why we can not use Gizmo with non-public fields is that we can not redefine classes. They are already loaded!

Well, they aren't, if you use a different classloader. :-) So if we mess around with classloaders and run the test only with the correct classloader with the Gizmo-instrumented classes, it should work.

However, once you start doing black magic with classloaders, you realize you are probably doing something you shouldn't. It tends to be brittle, and hard to understand and test. The previous suggestion is probably better, all things considered.

Christopher-Chianelli: The third option is to reuse the Quarkus "map non-public field descriptors to method descriptors" trick. The reason this isn't done in the code is because we have a guarantee we won't go through getters/setters, although since our test classes have simple getters/setters, we can use them in the test (which leads to the question: should we have another domain access type: GIZMO_RELAXED or GIZMO_DEFFERED etc. which is GIZMO except of instead of failing on non-public fields, it will use getters/setters of the fields)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoMemberDescriptor.java
```diff
@@ -239,7 +282,31 @@ public String getTypeName() {
...
+        whenMetadataIsOnMethod(md -> {
+            try {
+                out[0] = declaringClass.getDeclaredMethod(md.getName()).getGenericReturnType();
+            } catch (NoSuchMethodException e) {
+                throw new IllegalStateException("Cannot find method (" + md.getName() + ") on class (" + declaringClass + ").",
+                        e);
+            }
+        });
+
+        return out[0];
```
triceo, line 309: This code is weird.
The fact that the methods accept lambdas forces you to use the array trick to represent an effectively-final variable.
Plus this return value is quite a nasty side-effect of those lambdas.

I suggest:

- the methods return the value as opposed to storing it as a side-effect.
- the method calls turn into `if (metadataIsOnField()) then return ...`

That way, all the weirdness of this method disappears, and it will be easier to understand.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,203 @@
...
+     *        (ex: for the field "List myList", the actual value
+     *        class might be ArrayList).
+     * @param fieldDeepClonedMemoization memoization map for (field, owningClass) decisions
+     * @param actualValueClassDeepClonedMemoization memoization maps for actualValueClass decisions
+     * @return true iff the field should be deep cloned with a particular value.
+     */
+    public static boolean getMemorizedDeepCloneDecision(SolutionDescriptor<?> solutionDescriptor,
+            Field field, Class<?> owningClass,
+            Class<?> actualValueClass,
+            ConcurrentMap<Pair<Field, Class<?>>, Boolean> fieldDeepClonedMemoization,
```
triceo, line 55: This formatting seems weird. Many of the arguments could be on the same line.
(That said, it is legal, because our formatter config will honor manual line breaks.)

triceo: The "problem" (it's not really a problem) shows up in many places throughout the PR. As if the line length was 80, not 120.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,203 @@
...
+     * @param field The field to get the deep cloning decision of
+     * @param owningClass The class that owns the field; can be different
+     *        from the field's declaring class (ex: subclass).
+     * @return True only if the field is an entity property on the solution class.
+     *         May return false if the field getter/setter is complex.
+     */
+    public static boolean isFieldAnEntityPropertyOnSolution(SolutionDescriptor<?> solutionDescriptor,
+            Field field, Class<?> owningClass) {
+        // field.getDeclaringClass() is a superclass of or equal to the owningClass
+        if (solutionDescriptor.getSolutionClass().isAssignableFrom(owningClass)) {
```
triceo, line 118: Personally, I'd invert the logic here.
Then the most of this method would not need to be nested, making it easier to reason about.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionOrEntityDescriptor.java
```diff
@@ -0,0 +1,117 @@
...
+        deepClonedFields = new HashSet<>();
+        shallowlyClonedFields = new HashSet<>();
+
+        for (Field field : solutionFieldToMemberDescriptorMap.keySet()) {
+            if (DeepCloningUtils.getDeepCloneDecision(solutionDescriptor, field, entityClass, field.getType())) {
+                deepClonedFields.add(field);
+            } else {
+                shallowlyClonedFields.add(field);
+            }
+        }
```
triceo, line 68: This is code duplication from the previous constructor. Consider reusing the code.
Could possibly also include the constructor after this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionOrEntityDescriptor.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public SolutionDescriptor<?> getSolutionDescriptor() {
+        return solutionDescriptor;
+    }
+
+    public Set<GizmoMemberDescriptor> getShallowClonedMemberDescriptors() {
+        return solutionFieldToMemberDescriptorMap.keySet().stream()
+                .filter(field -> shallowlyClonedFields.contains(field))
+                .map(solutionFieldToMemberDescriptorMap::get).collect(Collectors.toSet());
+    }
```
triceo, line 107: This is a whole lot of calculation that should really be cached if it ends up on the hot path.
I assume you've made sure it isn't used on the hot path?

triceo (approved):
Looks much better!
I've made some soft-ish suggestions for things to think about.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionOrEntityDescriptor.java
Christopher-Chianelli: This code is only called once (at build time for Quarkus, on creating the Solver in plain Java). The code is called once per GizmoSolutionOrEntityDescriptor to determine the fields to shallowly clone (i.e. each MemberDescriptor is only used once).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorFactory.java
```diff
@@ -36,8 +39,11 @@
      * @return The generated class name for member
      */
     public static String getGeneratedClassName(Member member) {
-        return member.getDeclaringClass().getPackage().getName() + ".$optaplanner$__"
-                + member.getDeclaringClass().getSimpleName() + "$__" + member.getName();
+        String memberName = Optional.ofNullable(ReflectionHelper.getGetterPropertyName(member))
+                .orElse(member.getName());
```
ge0ffrey, line 43: nitpick: we typical avoid Optional in everything except for streams (which is also why it was designed for in the first place). This code creates an Optional instance every time to throw it away. 
`org.apache.commons.lang3.ObjectUtils.defaultIfNull` accomplishes the same goal, or just a `a == null ? ... : a`. Neither of those create a new object instance

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,196 @@
...
+     *        (ex: for the field "List myList", the actual value
+     *        class might be ArrayList).
+     * @param fieldDeepClonedMemoization memoization map for (field, owningClass) decisions
+     * @param actualValueClassDeepClonedMemoization memoization maps for actualValueClass decisions
+     * @return true iff the field should be deep cloned with a particular value.
+     */
+    public static boolean getMemorizedDeepCloneDecision(SolutionDescriptor<?> solutionDescriptor, Field field,
+            Class<?> owningClass, Class<?> actualValueClass,
+            ConcurrentMap<Pair<Field, Class<?>>, Boolean> fieldDeepClonedMemoization,
+            ConcurrentMap<Class<?>, Boolean> actualValueClassDeepClonedMemoization) {
```
ge0ffrey, line 55: parameter order in order of creation/longlivety
- solutionDescript first (lives longest, was created first)
- the maps next, as those are external state
- the parameters for this specific decision (field, owningclass, etc) next.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.concurrent.ConcurrentMap;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.optaplanner.core.api.domain.solution.cloner.DeepPlanningClone;
+import org.optaplanner.core.impl.domain.common.ReflectionHelper;
+import org.optaplanner.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+public final class DeepCloningUtils {
+    // owningClass != declaringClass
+    // owningClass refers to the actual instance class, which may be a subclass of the declaring class
```
ge0ffrey, line 34: Instead of having all static methods, can this class have a non-static field for the Memoizations maps and have all non-static methods? Is that reasonable for the clients of this class, to remember the class instance (instead of those Memoizations maps) until they're done with it (so done with the Memoizations) at which point it gets garbage collected.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -303,7 +304,11 @@ private void processSolutionCloner(DescriptorPolicy descriptorPolicy, PlanningSo
         if (solutionClonerClass != null) {
             solutionCloner = ConfigUtils.newInstance(this, "solutionClonerClass", solutionClonerClass);
         } else {
-            solutionCloner = new FieldAccessingSolutionCloner<>(this);
+            if (descriptorPolicy.getDomainAccessType() == DomainAccessType.GIZMO) {
+                solutionCloner = GizmoSolutionClonerFactory.build(this);
+            } else {
+                solutionCloner = new FieldAccessingSolutionCloner<>(this);
+            }
```
ge0ffrey, line 311: code style: I 'd rather see a switch here so if we ever introduce a third DomainAccessType this code fails-fast if we forget to implement it here, instead of doing something silently under the covers. I guess the same applies for other usages of `descriptorPolicy.getDomainAccessType()` in previous PR's.

```
            switch (foo) {
                case AAA:
                    ...
                    break;
                case BBB:
                    ...;
                    break;
                default:
                    throw new IllegalStateException("The foo (" + foo + ") is not implemented.");
            }
```

What about null? Well, DescriptorPolicy.getDomainAccessType() never returns null and it's javadoc should state "@return never null".

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -149,14 +173,9 @@ public static String generateFieldAccessor(AnnotationInstance annotationInstance
...
-        if (methodInfo.name().startsWith("get")) { // Case 1: Getter method
-            return methodInfo.name().substring(3, 4).toLowerCase(Locale.ROOT) + methodInfo.name().substring(4);
-        } else if (methodInfo.name().startsWith("is")) { // Case 2: Getter method for boolean
-            return methodInfo.name().substring(2, 3).toLowerCase(Locale.ROOT) + methodInfo.name().substring(3);
-        } else { // Case 3: Read method
-            return methodInfo.name();
-        }
+    private static String getMemberName(Member member) {
+        return Optional.ofNullable(ReflectionHelper.getGetterPropertyName(member))
+                .orElse(member.getName());
```
ge0ffrey, line 178: same here

ge0ffrey (approved):
LGTM, great work Chris!

1) Please look at and resolve or reject my suggestions before merging.

2) Also do a big coverage manual test:
- Run GeneralOptaPlannerBenchmarkApp with GIZMO on (do this while sleeping, it can take a long time).
- Run all quarkus quickstarts for at least 1 minute. 
-- Run one with and without this GIZMO on next to each other (I have optaplanner git clones locally and 2 seperate Intellij projects for this purpose) with DEBUG logging. By running them next to each other you can verify in their logs if LS step 1234 on the left one has the exact same score and selected move as LS step 1234 on the right one. The time spent can differ.
-- Try one with FULL_ASSERT too.

Christopher-Chianelli:
Results (consistent LS steps from random sampling => me checking random steps for equivalence) (IK the additional steps is pointless w/o score calculation speed (was too focus on steps equivalence I recorded the wrong data point), but it is improvements all around compared to REFLECTION; (it around 10 - 20% improvement in quarkus dev mode w/ DEBUG logging):
```
School Timetabling: ~10,000 additional steps w/ DEBUG logging in 60 seconds compared to REFLECTION; consistent LS steps from random sampling

Facility Location: 0 additional steps w/ DEBUG logging in 60 seconds compared to REFLECTION; consistent LS steps from random sampling (both REFLECTION and GIZMO stop picking moves at LS step 44935, with the final step 44936 (inconsistent) being taking at the termination time (10 whole seconds later for GIZMO, 6 seconds later for REFLECTION)

inconsistency:
GIZMO:
LS step (44936), time spent (60000), score (0hard/-1467509soft),     best score (0hard/-1467479soft), accepted/selected move count (0/*242015*), picked move (Consumer 32 (15 dem) {Facility 11 ($46001, 150 cap) -> Facility 29 ($61843, 150 cap)}).

REFLECTION:
LS step (44936), time spent (60000), score (0hard/-1467509soft),     best score (0hard/-1467479soft), accepted/selected move count (0/*162781*), picked move (Consumer 32 (15 dem) {Facility 11 ($46001, 150 cap) -> Facility 29 ($61843, 150 cap)}).

(selected move count differs; 242015 in GIZMO, 162781 in REFLECTION. Both are WAY too high; max value seen is 1000, average under 100).

Factorio Layout: ~300 additional steps w/ DEBUG logging in 60 seconds compared to REFLECTION; consistent LS steps from random sampling

Maintenance Scheduling: ~26000 additional steps w/ DEBUG logging in 60 seconds compared to REFLECTION; consistent LS steps from random sampling

Vaccine Scheduling: ~18000 additional steps w/ DEBUG logging in 60 seconds compared to REFLECTION;  consistent CH steps from random sampling (CH is slow, LS is fast, so using 10 seconds so I have logs in both terminals leads to me only have CH data)
```

TODO: GeneralOptaPlannerBenchmarkApp

Christopher-Chianelli:
No errors from FULL_ASSERT on maintenance-scheduling and vaccination-scheduling after 5 minutes

Christopher-Chianelli:
GeneralOptaPlannerBenchmarkApp cannot run due to the strict requirements of GIZMO (public fields). Will run w/ modified domain (public fields). It probably can run with the workaround I used in the test (as a new DomainAccessType: GIZMO_BEAN, which allows it to use public getters/setters)

Christopher-Chianelli:
The GeneralOptaPlannerBenchmarkApp revealed an issue with the MemberAccessorImplementor code (its assertGoodMethod checks for "boolean" instead of "Z" in the return type, causing it to fail on valid method names). The code is fixed and a test is added.

PR #1106, opened 2021-01-19
PLANNER-1406 Show solver move thread count in INFO logging

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -97,8 +97,8 @@ public DefaultSolverFactory(SolverConfig solverConfig) {
         Termination<Solution_> termination = TerminationFactory.<Solution_> create(terminationConfig_)
                 .buildTermination(configPolicy, basicPlumbingTermination);
         List<Phase<Solution_>> phaseList = buildPhaseList(configPolicy, bestSolutionRecaller, termination);
-        return new DefaultSolver<>(environmentMode_, randomFactory,
-                bestSolutionRecaller, basicPlumbingTermination, termination, phaseList, solverScope);
+        return new DefaultSolver<>(environmentMode_, randomFactory, bestSolutionRecaller, basicPlumbingTermination,
+                termination, phaseList, solverScope, moveThreadCount_ == null ? 1 : moveThreadCount_);
```
rsynek, line 101: If `moveThreadCount` is `null`, is it really a single move thread or rather 0, as the moves are evaluated directly on the solver thread?

triceo: Good question. Rather a philosophical one, I'd say. There will be 1 thread.
Does it make sense to say there would be 0 move threads?

rsynek: None of these options is accurate. How about:
`move thread count (solver thread)` in case the `moveThreadCount == 0` ?

ge0ffrey: when moveThreadCount is null, we use NONE which is != 1.

With NONE, the moves are evaluated in the solver thread. with 1, the moves are evaluated in 1 move thread on which the solver thread is waiting.
move thread count 1 is mainly usefull for debugging purposes, not for in production. NONE is always better.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -179,11 +182,13 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
         super.solvingStarted(solverScope);
         int startingSolverCount = solverScope.getStartingSolverCount() + 1;
         solverScope.setStartingSolverCount(startingSolverCount);
-        logger.info("Solving {}: time spent ({}), best score ({}), environment mode ({}), random ({}).",
+        logger.info("Solving {}: time spent ({}), best score ({}), move thread count ({}), " +
+                "environment mode ({}), random ({}).",
```
ge0ffrey, line 186: 1) We show the environmentMode in both the "solving started" and "solving ended" log.

(soft suggestion) How about we show the move thread count in both too?

ge0ffrey: 2. (very very soft suggestion, wdyt? your call) I might mention environment mode before move thread count because it's more important when they're looking at score calculation speed, as it distorts env mode even more than move thread count.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverFactory.java
ge0ffrey:
```suggestion
                termination, phaseList, solverScope, moveThreadCount_ == null ? SolverConfig.MOVE_THREAD_COUNT_NONE : Integer.toString(moveThreadCount_));
```

ge0ffrey (commented):
Good for me to merge if NONE problem is fixed.

PR #1110, opened 2021-01-21
VariableListener javadoc should clarify its the source var's entity 

ge0ffrey:
+ align maven-exec to run scripts

### DIFF
--- a/optaplanner-examples/pom.xml
+++ b/optaplanner-examples/pom.xml
@@ -68,9 +68,8 @@
             </additionalClasspathElements>
             <arguments>
               <argument>-Xms256m</argument>
-              <!-- Most examples run (potentially slower) with max heap of 128 MB (so -Xmx128m), but 1 example's dataset requires 1.5 GB -->
-              <argument>-Xmx1536m</argument>
-              <argument>-server</argument>
+              <!-- Most examples run (potentially slower) with max heap of 128 MB (so -Xmx128m), but 1 example's dataset requires 2 GB -->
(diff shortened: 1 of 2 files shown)


triceo (approved):

optaplanner-examples/pom.xml
```diff
@@ -68,9 +68,8 @@
             </additionalClasspathElements>
             <arguments>
               <argument>-Xms256m</argument>
-              <!-- Most examples run (potentially slower) with max heap of 128 MB (so -Xmx128m), but 1 example's dataset requires 1.5 GB -->
-              <argument>-Xmx1536m</argument>
-              <argument>-server</argument>
+              <!-- Most examples run (potentially slower) with max heap of 128 MB (so -Xmx128m), but 1 example's dataset requires 2 GB -->
```
Christopher-Chianelli, line 71: Did a problem set for a problem get bigger? Why the 500MB increase in memory usage?

triceo: I believe there are two reasons for that. First of all, some of the problems never really fit, specifically Machine Reassignment. Second, 1.5G is essentially equivalent to 0 these days. :-) We can take the 2G easily, so why not.

Christopher-Chianelli: So the 1.5GB was never actually enough; okay. 2GB isn't really 0 though; it is 1/16th of memory for a computer with 32GB of RAM (about 8.5%). I am okay with these changes.

Christopher-Chianelli (approved):

PR #1111, opened 2021-01-21
Fail to compile if a member annotation is not in a @PlanningEntity

Christopher-Chianelli:
In maven at least, the annotation processor is automatically used, no changes required on the user end. To test, try commenting out `@PlanningEntity` on some classes in `optaplanner-examples`.

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/pom.xml
+++ b/optaplanner-core/pom.xml
@@ -134,6 +134,33 @@
           </ignoredUnusedDeclaredDependencies>
         </configuration>
       </plugin>
+
+      <plugin>
+        <artifactId>maven-resources-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>add-annotation-processor-service</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${basedir}/target/classes/META-INF/services</outputDirectory>
+              <!-- Cannot put this file into resources since it will try to look for
+                   a class that is compiled during the build -->
+              <resources>
+                <resource>
+                  <directory>src/build</directory>
+                  <includes>
+                      <include>javax.annotation.processing.Processor</include>
+                  </includes>
+                </resource>
+              </resources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
       <plugin>
         <groupId>org.revapi</groupId>
         <artifactId>revapi-maven-plugin</artifactId>
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/processor/OptaPlannerAnnotationProcessor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/processor/OptaPlannerAnnotationProcessor.java
@@ -1,0 +1,76 @@
+package org.optaplanner.core.impl.processor;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import javax.annotation.processing.AbstractProcessor;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.annotation.processing.RoundEnvironment;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.TypeElement;
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.AnchorShadowVariable;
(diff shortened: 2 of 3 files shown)


optaplanner-core/pom.xml
```diff
@@ -134,6 +134,33 @@
...
+
+      <plugin>
+        <artifactId>maven-resources-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>add-annotation-processor-service</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
```
ge0ffrey, line 146: I am not a fan of complicating our build.
The solution could be worse than the problem.

optaplanner-core/src/main/java/org/optaplanner/core/impl/processor/OptaPlannerAnnotationProcessor.java
```diff
@@ -0,0 +1,76 @@
...
+            switch (element.getKind()) {
+                case FIELD:
+                case METHOD:
+                    if (element.getEnclosingElement().getAnnotation(PlanningEntity.class) == null) {
+                        throw new IllegalStateException("Member annotation @" + annotation.getSimpleName() +
+                                " was placed in a class without @PlanningEntity ("
+                                + element.getEnclosingElement().getSimpleName()
+                                + "). Maybe add @PlanningEntity to the class (" + element.getEnclosingElement().getSimpleName()
+                                + ")?");
+                    }
```
ge0ffrey, line 67: Does this work for example if the swing optaplanner-examples make this mistake?

optaplanner-core/pom.xml
Christopher-Chianelli: In an ideal world, the annotation processor + annotations would be in a different maven module, and optaplanner-core would use it as a dependency. But that would be a major refactor and split API into two places.

optaplanner-core/src/main/java/org/optaplanner/core/impl/processor/OptaPlannerAnnotationProcessor.java
Christopher-Chianelli: It will always work if `optaplanner-core` is on the classpath (as it contains the annotation processor entry). No additional configuration is required (at least for maven and gradle). If they are compiling with `javac`, some additional configuration might be required (since it doesn't generate any files, only verify proper annotation usage, everything will still work but the error won't be caught). It will also automatically work/be used for spring-boot. It doesn't seem to automatically work for quarkus though (at least in the deployment test), meaning quarkus probably does not automatically run annotation processors.

ge0ffrey:
@Christopher-Chianelli @triceo @rsynek Good POC, good that this has been researched, but I am not convinced the solution is better than the problem. I might suggest closing this PR without merging. Wdyt?

Christopher-Chianelli:
If the API Is gonna be split, there will be no changes to the pom (the reason for the pom change is to put the annotation processor declaration post build (otherwise it will look for a non-existent annotation processor during the build). `mvn` will automatically detect the annotation processor in consumers (no need for changes there); not sure about other build tools (but since this doesn't generate any files, the worst that can happen is it doesn't fail-fast = current behavior)

optaplanner-core/pom.xml
rsynek: How much time does it take to execute this goal? 
After looking into the scope of splitting the entire `api` package into a separate module, I am afraid we would have to keep this workaround in place.

Christopher-Chianelli: If the annotations are moved to the api module, the workaround can be removed and the file directly placed in optaplanner-core/src/main/resources. The goal itself take < 1 second to execute (it is a file copy)

triceo (changes requested):
I vote for not merging this. I do not like the build impact as well - the more things we do like that, the harder it becomes to have the project work in the IDE.

(We already have the XSD situation which bugs me a lot.)

ge0ffrey (changes requested):
I agree. Builds should be simple. Think IDE's, think CI's, ...

ge0ffrey:
As for the XSD: the benefits outweigh the downsides. I 100% loved the XSD the other day when writing a solver config and benchmark config for the vaccination impl. But if there's a potential way to have our cake and eat it too, let's investigate.

triceo:
Closing the PR as we have more than 1 negative review.
Thanks for the effort, @Christopher-Chianelli , still!

PR #1112, opened 2021-01-22
Gizmo member accessors: javadocs suggestions

ge0ffrey:


Christopher-Chianelli (approved):
Note: current Gizmo Solution Cloner requires *all* fields of the Planning Solution/Planning Entities to be public (unless it Quarkus). I probably won't be able to remove that restriction, unless you know a way to shallowly clone an Object without reflection.

PR #1118, opened 2021-01-26
PLANNER-2314: added new Jenkinsfile for updating opta* master to next…

mbiarnes:
… SNAPSHOT

PLANNER-2314

### Checklist

<details>
<summary>
</summary>

</details>


mbiarnes:
@radtriste @rsynek this is not the "ultimate" solution. The reason why this if first a draft. There may be another solution to do this with less changes. @radtriste and me have to look at this.

radtriste:
I am thinking about reusing the `deploy` pipeline existing already and doing similar thing, with `SKIP_TESTS` and `RELEASE` to true (name of `RELEASE` should be changed to something like `FULL_RUN` to be more generic)
We could then just need a small Jenkinsfile in kogito-pipelines instead calling those deploy pipeline (optaplanner and examplex)

Jenkinsfile.update
```diff
@@ -0,0 +1,211 @@
...
+                checkoutRepo(optaplannerRepository, getBuildBranch())
+                checkoutRepo(vehicleRoutingRepository, getBuildBranch())
+                checkoutRepo(employeeRosteringRepository, getBuildBranch())
+                checkoutRepo(kogitoExamplesRepository, getBuildBranch())
+                checkoutRepo(kogitoAppsRepository, getBuildBranch())
+            }
+        }
+        stage('Prepare for PR') {
+            when {
+                expression { return isRelease() }
```
rsynek, line 68: The job should be triggered only when we want to upgrade the versions, thus we don't need any conditional stages.

Jenkinsfile.update
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+        stage('Update to next SNAPSHOT version') {
+            when {
+                expression { return getProjectVersion() != '' }
+            }
+            steps {
+                script {
+                    maven.mvnVersionsSet(getOptaplannerMavenCommand(), getProjectVersion(), true)
+                    maven.mvnSetVersionProperty(getOptaplannerMavenCommand(), 'version.org.kie.kogito', getKogitoVersion())
```
rsynek, line 85: When we are upgrading the master branch to the next minor version, the kogito version stays intact, as it uses the next major version snapshot:
https://github.com/kiegroup/kogito-runtimes/blob/master/pom.xml#L15

Jenkinsfile.update
```diff
@@ -0,0 +1,211 @@
...
+
+                    mavenCleanInstallOptaPlannerParents()
```
rsynek, line 87: The new optaplanner snapshot should be also deployed to JBoss Nexus; otherwise, the CI on PRs to the projects depending on optaplanner will fail.

Jenkinsfile.update
```diff
@@ -0,0 +1,211 @@
...
+                cleanWs()
+            }
+        }
+        stage('Checkout repositories') {
+            steps {
+                checkoutRepo(optaplannerRepository, getBuildBranch())
+                checkoutRepo(vehicleRoutingRepository, getBuildBranch())
+                checkoutRepo(employeeRosteringRepository, getBuildBranch())
+                checkoutRepo(kogitoExamplesRepository, getBuildBranch())
+                checkoutRepo(kogitoAppsRepository, getBuildBranch())
```
radtriste, line 63: as we are impacting and modifying more than just optaplanner projects, I would put that Jenkinsfile in https://github.com/kiegroup/kogito-pipelines instead.

Jenkinsfile.update
radtriste: I would keep that one as it offers more flexibility if that changes in the future ...
anyway, if version is already set correctly, this will be a fast command to be run

Jenkinsfile.update
```diff
@@ -0,0 +1,237 @@
...
+void updateOptaplannerSnapshot(String optaString, String optaVersion, String repo, String dirName=repo) {
+    dir(dirName) {
+        maven.mvnSetVersionProperty(optaString, optaVersion)
+    }
+}
+
+void updateQuickstartsSnapshot(String quickstartVersion, String repo, String dirName=repo) {
+    dir(dirName) {
+        sh "sed -i -E 's/def optaplannerVersion = \"[^\"\\s]+\"/def optaplannerVersion = \"${quickstartVersion}\"/' */build.gradle"
+        sh "git add */build.gradle"
```
dupliaka, line 172: Suggesting to move git add to the place with the other files staging
githubscm.commitChanges(commitMsg, { githubscm.findAndStageNotIgnoredFiles('pom.xml') githubscm.findAndStageNotIgnoredFiles('build.gradle') })

- We will keep it in one place
- We will avoid pushing ignored files if build.gradle once will be moved to git ignore

mbiarnes:
replaced by #1134, therefore closed

PR #1119, opened 2021-01-26
PLANNER-2315 Properly size Bavet child tuple lists

triceo:
Some tuples will only ever have 1 child.
Some tuples will mostly only have 1 expected child.
Some tuples will never have children.

The sizing of the collections should reflect that, which leads to a performance improvement of ~10 % on average.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiTuple.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiTuple.java
@@ -16,12 +16,12 @@
 
 package org.optaplanner.core.impl.score.stream.bavet.bi;
 
-import org.optaplanner.core.impl.score.stream.bavet.common.BavetAbstractTuple;
-import org.optaplanner.core.impl.score.stream.bavet.common.BavetGroupTuple;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
+
+import org.optaplanner.core.impl.score.stream.bavet.common.BavetAbstractTuple;
+import org.optaplanner.core.impl.score.stream.bavet.common.BavetGroupTuple;
 
 public final class BavetGroupBiTuple<GroupKey_, ResultContainer_, Result_> extends BavetAbstractBiTuple<GroupKey_, Result_>
         implements BavetGroupTuple {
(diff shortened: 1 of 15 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiTuple.java
```diff
@@ -29,6 +32,7 @@
     private int parentCount;
     private ResultContainer_ resultContainer;
     private Result_ result;
+    private final List<BavetAbstractTuple> childTupleList = new ArrayList<>();
```
ge0ffrey, line 35: The bridge will have n childs typically, but the group/joins itself - like this one - will have 1 child typically.

```suggestion
    private final List<BavetAbstractTuple> childTupleList = new ArrayList<>(1);
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFromUniTuple.java
```diff
@@ -16,12 +16,19 @@
...
+import org.optaplanner.core.impl.score.stream.bavet.common.BavetAbstractTuple;
+
 public final class BavetFromUniTuple<A> extends BavetAbstractUniTuple<A> {
 
     private final BavetFromUniNode<A> node;
     private final A factA;
+    private final List<BavetAbstractTuple> childTupleList;
 
     public BavetFromUniTuple(BavetFromUniNode<A> node, A factA, int childTupleListSize) {
+        this.childTupleList = new ArrayList<>(childTupleListSize);
```
ge0ffrey, line 31: Idea: as we apparrantly know the maximum childTupleListSize for a fromUni tuple,
we probably know it for the filterUni tuples too?

For example:
- from(A)
-- filter(A.b != null)
---  join(C).penalize
--- join(D).penalize
--- filter(E).penalize
-- filter(F).penalize

the from has max 2 child tuples, the filter has max 3. Even if the turns out less that than, the 3 is still a very good ArrayList capacity.

ge0ffrey (approved):

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/uni/BavetFromUniTuple.java
triceo: You keep assuming that there is node sharing. That is a very poor assumption, IMO, outside of the from nodes.
The bar to actually get a node shared is very high, and people will never meet it.

(Unless they write some complicated and non-straight-forward constraint providers.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetGroupBiTuple.java
triceo: Hmm, I have it switched the other way. Will need rebenchmarking then, could be even better.

PR #1123, opened 2021-01-27
Add location to JAXB validation error

rsynek:
Slightly improves the validation message as requested in https://kie.zulipchat.com/#narrow/stream/232679-optaplanner/topic/XML.20mistake.20.3D.3E.20no.20line.20number.

However, it's not possible to get the line number, as this information is available only if we use SAX as the underlying parser. In our case, we use DOM, so we get the Node where the validation error occurs.

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
@@ -37,7 +37,6 @@ import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Marshaller;
 import javax.xml.bind.Unmarshaller;
-import javax.xml.bind.ValidationEvent;
 import javax.xml.bind.util.ValidationEventCollector;
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/io/jaxb/GenericJaxbIO.java
```diff
@@ -184,6 +184,14 @@ public T readAndValidate(Document document, Schema schema) {
         }
     }
 
+    private String validationEventToString(ValidationEvent validationEvent) {
+        return new StringBuilder()
+                .append(validationEvent.getMessage())
+                .append("\n")
+                .append("Node: ").append(validationEvent.getLocator().getNode().getNodeName())
+                .toString();
```
ge0ffrey, line 192: Use string concatenation, as that's equal or faster than a StringBuilder if no loops are invovled (java compiler actually translates it into a StringBuilder but deals better with capacity iirc)
Also reads better:

return validationEvent.getMessage() + "n" + "Node: " + validationEvent.getLocator().getNode().getNodeName()

=> at which point I'd inline it

ge0ffrey (approved):
LGTM, see suggetion

PR #1133, opened 2021-02-01
PLANNER-2259 Add examples to Joiner Javadocs

triceo:


build/optaplanner-build-parent/pom.xml
```diff
@@ -975,7 +975,7 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-javadoc-plugin</artifactId>
               <configuration>
-                <source>${maven.compiler.source}</source>
+                <release>${maven.compiler.release}</release>
```
triceo, line 978: There was a stale PR doing this. I closed it and put it in this one instead.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/Joiners.java
```diff
@@ -139,9 +136,7 @@
     }
 
     /**
-     * Joins every A and B where a value of property on A is less than or equal to the value of the same property on B.
-     * These are exactly the pairs where {@code mapping.apply(A).compareTo(mapping.apply(B)) <= 0}.
-     * Delegates to {@link #lessThanOrEqual(Function, Function)} with both arguments using the same mapping.
+     * As defined by {@link #lessThanOrEqual(Function, Function)} with both arguments using the same mapping.
```
dupliaka, line 139: Maybe I am a bit confused
In one line I see  
 never null, a joiner that filters the constraint stream to only include elements (A,B) where a property
But there is no B type in the definition 
 public static <A, Property_ extends Comparable<Property_>> BiJoiner<A, A> lessThanOrEqual(

triceo: Hmm, good point. The A and B here are just other ways for saying "left" and "right". 

That said, since the Javadoc is now much more complete, we can probably just remove all the return descriptions, and keep there just the "never null" part.

rsynek (approved):

PR #1135, opened 2021-02-03
PoC: UniConstraintStream.map(...)

triceo:
The fundamentals are there, but it's still far too broken.


triceo:
This is out for now. Will be re-introduced in another PR.

PR #1138, opened 2021-02-04
PLANNER-2322 Allow duplicate constraint matches to coexist

triceo:
Also see https://github.com/kiegroup/optaplanner-website/pull/313

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
@@ -27,6 +27,16 @@ import org.optaplanner.core.impl.domain.lookup.ClassAndPlanningIdComparator;
 /**
  * Retrievable from {@link ConstraintMatchTotal#getConstraintMatchSet()}
  * and {@link Indictment#getConstraintMatchSet()}.
+ *
+ * <p>
+ * Note: this class has a {@link #compareTo(ConstraintMatch)} method which is inconsistent with equals.
+ * (See {@link Comparable}.)
+ * Two different {@link ConstraintMatch} instances ({@code (a == b) == false}) must never {@link #equals(Object)},
+ * or else it would be impossible for two matches with the same justifications to coexist.
+ * (This functionality is being relied upon in the Constraint Streams API.)
+ * Yet two instances may {@link #compareTo(ConstraintMatch)} equal in case they come from the same constraint and their
+ * justifications are equal.
+ * This is so that we get consistent ordering of constraint matches in visualizations.
  * 
  * @param <Score_> the actual score type
  */
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -28,6 +27,16 @@
...
+ *
+ * <p>
+ * Note: this class has a {@link #compareTo(ConstraintMatch)} method which is inconsistent with equals.
+ * (See {@link Comparable}.)
+ * Two different {@link ConstraintMatch} instances ({@code (a == b) == false}) must never {@link #equals(Object)},
+ * or else it would be impossible for two matches with the same justifications to coexist.
+ * (This functionality is being relied upon in the Constraint Streams API.)
+ * Yet two instances may {@link #compareTo(ConstraintMatch)} equal in case they come from the same constraint and their
+ * justifications are equal.
+ * This is so that we get consistent ordering of constraint matches in visualizations.
```
ge0ffrey, line 39:
```suggestion
 * This is for consistent ordering of constraint matches in visualizations.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -28,6 +27,16 @@
 /**
  * Retrievable from {@link ConstraintMatchTotal#getConstraintMatchSet()}
  * and {@link Indictment#getConstraintMatchSet()}.
+ *
+ * <p>
+ * Note: this class has a {@link #compareTo(ConstraintMatch)} method which is inconsistent with equals.
```
ge0ffrey, line 32: (soft) remove "Note", as all javadocs are a notes

Also link "equals" and mention hashcode (also linked).

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -28,6 +27,16 @@
...
+ * (See {@link Comparable}.)
+ * Two different {@link ConstraintMatch} instances ({@code (a == b) == false}) must never {@link #equals(Object)},
+ * or else it would be impossible for two matches with the same justifications to coexist.
```
ge0ffrey, line 35: I'd pull the "with the same justificaiton" list to the front, for clarity, to pin point on the case where this matters.
 I'd remove the "(a == b) == false" because "different" is clear enough.
 I'd avoid "must never ..." in favor of "can not ..." because "must" is requirements talk, while javadocs should be specification talk and clear statements on how it works.

Two different {@link ConstraintMatch} instances with the same justification list, aren't {@link equals(Object) equal}
because some ConstraintStream API methods can result in duplicate facts, which are treated as independent matches.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -91,9 +100,10 @@ public int compareTo(ConstraintMatch<Score_> other) {
             /*
              * TODO Come up with a better cache.
              *
-             * We reuse the comparator from here, since it internally caches some reflection that we don't want to be performing
-             * over and over again. However, there are possibly thousands of instances of this class, and each will get its own
-             * comparator. Therefore, the caching is only partially effective.
+             * We reuse the comparator from here, since it internally caches some reflection that we don't want to be
```
ge0ffrey, line 103: Avoid "we" and "that we don't want to be ...ing".
```suggestion
             * Reuse the comparator to internally caches reflection for performance benefits.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -91,9 +100,10 @@ public int compareTo(ConstraintMatch<Score_> other) {
...
+             * performing over and over again.
+             * However, there are possibly thousands of instances of this class, and each will get its own comparator.
```
ge0ffrey, line 105: Avoid "will"
```suggestion
             * However, there are possibly thousands of instances of this class, and each gets its own comparator.
```

ge0ffrey (approved):

PR #1141, opened 2021-02-04
Upgrade JBoss Parent pom from 36 to 37

ge0ffrey:
jboss parent 36 has a maven compiler version (something ending with -jboss-2) that seemed to have disappeared from Maven Central


triceo (approved):
LGTM.
Although it seems fishy that we're only adding module names to _some_ of the modules.

triceo:
Merged even though Jenkins was red. There is an explanation for that.

PR #1142, opened 2021-02-04
PLANNER-2308: Pick up classes referenced in annotations in native mode

Christopher-Chianelli:
Depends on https://github.com/kiegroup/optaplanner/pull/1104 . Note: I discovered that OptaPlanner doesn't pick up annotations in native mode unless jackson/other JSON framework is used (that issue isn't fixed by this PR; I'll investigate further).

### JIRA

https://issues.redhat.com/browse/PLANNER-2308

### Referenced pull requests

https://github.com/kiegroup/optaplanner/pull/1104

### Checklist

<details>
<summary>
</summary>

</details>


Christopher-Chianelli:
Look like it was caused because my test was in a subpackage of `org.optaplanner`; fixed the issue

optaplanner-quarkus-integration/optaplanner-quarkus/runtime/src/main/java/org/optaplanner/quarkus/nativeimage/Substitute_ConfigUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    public static <T> T newInstance(Object bean, String propertyName, Class<T> clazz) {
+        T out = OptaPlannerGizmoBeanFactory.INSTANCE.getReference().newInstance(clazz);
+        if (out != null) {
+            return out;
+        } else {
+            throw new IllegalArgumentException("The " + bean.getClass().getSimpleName() + "'s " + propertyName + " ("
+                    + clazz.getName() + ") does not have a public no-arg constructor"
+                    // Inner classes include local, anonymous and non-static member classes
+                    + ((clazz.isLocalClass() || clazz.isAnonymousClass() || clazz.isMemberClass())
+                            && !Modifier.isStatic(clazz.getModifiers()) ? " because it is an inner class." : "."));
```
triceo, line 39: What is the user supposed to do with an exception like this? What could the user do?

optaplanner-quarkus-integration/optaplanner-quarkus/integration-test/src/main/java/org/optaplanner/quarkus/it/domain/ITestdataPlanningEntity.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.optaplanner.quarkus.it.domain;
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.variable.CustomShadowVariable;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+import org.optaplanner.core.api.domain.variable.PlanningVariableReference;
+
+@PlanningEntity
+public class ITestdataPlanningEntity {
```
triceo, line 25: Consider reusing stuff from `testdata` in `core`. You can include the test JAR into this Maven module.

triceo (changes requested):
There is room for improvement.
Good to see the test, though.

optaplanner-quarkus-integration/optaplanner-quarkus/runtime/src/main/java/org/optaplanner/quarkus/nativeimage/Substitute_ConfigUtils.java
Christopher-Chianelli: This was to duplication the error behavior of ConfigUtils: https://github.com/kiegroup/optaplanner/blob/fb88c87e1799505b902f1f144235277a7a32f349/optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java#L58-L65 . Should never happen normally (unless ConfigUtils is used to create a class that wasn't used in an annotation/we forget to add the annotation to the detector.

optaplanner-quarkus-integration/optaplanner-quarkus/integration-test/src/main/java/org/optaplanner/quarkus/it/domain/ITestdataPlanningEntity.java
Christopher-Chianelli: It does test a shadow variable to ensure custom variable listeners work (which is used in a constraint provider). Also, can cause issues since there are multiple `@PlanningSolution` classes (using a solverConfig the defines the solution class and entity classes might fix that). I think it be more work + code to reuse classes than creating new classes in this case.

triceo: Ok, two options then:

- Put these to `testdata` in `core` and use them from there.
- Create a `testdata` package here, for consistency sake.

In either case, we do not use the `I` prefix. I understand you used it to say this testdata is for an integration test, but IMO that is not important.

optaplanner-quarkus-integration/optaplanner-quarkus/runtime/src/main/java/org/optaplanner/quarkus/nativeimage/Substitute_ConfigUtils.java
triceo: We usually do one of the two things with exceptions:

- If it is unrecoverable, we usually add the prefix "Impossible state: " to tell the user that this is not their fault, but our bug.
- If it could, however, be their fault, we add a new line to the message, saying "Maybe... " and pointing them to what they could try to fix. See other uses of "maybe" in the codebase.

triceo:
Also, please take a look at Sonar. The 1 bug seems like nonsense, but some of the code smells should be easily addressable.

optaplanner-quarkus-integration/optaplanner-quarkus/runtime/src/main/java/org/optaplanner/quarkus/nativeimage/Substitute_ConfigUtils.java
Christopher-Chianelli: Changed the message to "Impossible State: could not find the " + bean.getClass().getSimpleName() +
                    "'s " + propertyName + " (" + clazz.getName() + ") generated Gizmo supplier."

optaplanner-quarkus-integration/optaplanner-quarkus/runtime/src/main/java/org/optaplanner/quarkus/nativeimage/Substitute_ConfigUtils.java
```diff
@@ -0,0 +1,37 @@
...
+@TargetClass(className = "org.optaplanner.core.config.util.ConfigUtils")
+public final class Substitute_ConfigUtils {
+
+    @Substitute
+    public static <T> T newInstance(Object bean, String propertyName, Class<T> clazz) {
+        T out = OptaPlannerGizmoBeanFactory.INSTANCE.getInstance().newInstance(clazz);
+        if (out != null) {
+            return out;
+        } else {
+            throw new IllegalArgumentException("Impossible State: could not find the " + bean.getClass().getSimpleName() +
```
triceo, line 33:
```suggestion
            throw new IllegalArgumentException("Impossible state: could not find the " + bean.getClass().getSimpleName() +
```

triceo (approved):

PR #1146, opened 2021-02-08
PLANNER-2324 Make CS-D reuse the justifications which are already known

triceo:
I have measured both increases and drops in CS-D and DRL performance, case by case. 
On average, both CS-D and DRL improved by ~2 %.
From that, I conclude that this has a negligible impact on performance, if any.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreHolderImpl.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreHolderImpl.java
@@ -20,7 +20,6 @@ import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.function.BiConsumer;
 
 import org.kie.api.definition.rule.Rule;
 import org.kie.api.runtime.rule.RuleContext;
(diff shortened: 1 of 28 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreHolderImpl.java
```diff
@@ -74,9 +73,9 @@ public int getSoftScore(int softLevel) {
     @Override
     public void configureConstraintWeight(Rule rule, BendableScore constraintWeight) {
         super.configureConstraintWeight(rule, constraintWeight);
-        BiConsumer<RuleContext, Integer> matchExecutor;
+        IntMatchExecutor matchExecutor;
```
triceo, line 76: This (and similar changes in other `ScoreHolder`s) removes *a lot* of auto-boxing.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/holder/AbstractScoreHolder.java
```diff
@@ -106,20 +118,23 @@ public void configureConstraintWeight(Rule rule, Score_ constraintWeight) {
...
+            Supplier<Score_> scoreSupplier, Object... justifications) {
+        AgendaItem<?> agendaItem = (AgendaItem<?>) kcontext.getMatch();
         ConstraintActivationUnMatchListener constraintActivationUnMatchListener = new ConstraintActivationUnMatchListener(
                 constraintUndoListener);
         agendaItem.setCallback(constraintActivationUnMatchListener);
         if (constraintMatchEnabled) {
-            List<Object> justificationList = extractJustificationList(kcontext);
+            // If we have justifications coming from CS-D directly, use them exclusively.
+            List<Object> completeJustificationList =
+                    justifications.length == 0 ? extractJustificationList(kcontext) : Arrays.asList(justifications);
```
triceo, line 132: This change will make constraint matches in CS-D perform about a bazillion times faster.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreHolderImpl.java
```diff
@@ -36,9 +35,9 @@
  */
 public final class BendableScoreHolderImpl extends AbstractScoreHolder<BendableScore> implements BendableScoreHolder {
```
triceo, line 36: The changes to all score holders all do the same thing. It's just a whole bunch of code duplication.
If you've reviewed the changes to one score holder, you've reviewed them all.

rsynek (approved):

PR #1153, opened 2021-02-09
PLANNER-2165, PLANNER-2166, PLANNER-2167 Index higher cardinality joins

triceo:
Requires https://github.com/kiegroup/kogito-runtimes/pull/1052 and the subsequent Kogito SNAPSHOT.
Otherwise the build breaks.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/UniLeftHandSide.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/UniLeftHandSide.java
@@ -16,25 +16,7 @@
 
 package org.optaplanner.core.impl.score.stream.drools.common;
 
-import static java.util.Collections.singletonList;
-import static org.drools.model.DSL.accFunction;
-import static org.drools.model.DSL.accumulate;
-import static org.drools.model.DSL.exists;
-import static org.drools.model.DSL.groupBy;
-import static org.drools.model.DSL.not;
-import static org.drools.model.PatternDSL.betaIndexedBy;
-import static org.drools.model.PatternDSL.pattern;
-
-import java.math.BigDecimal;
-import java.util.Arrays;
-import java.util.List;
-import java.util.function.BiPredicate;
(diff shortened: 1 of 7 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/UniLeftHandSide.java
```diff
@@ -121,51 +121,37 @@ protected UniLeftHandSide(PatternVariable<A> patternVariable, DroolsVariableFact
...
         for (int mappingIndex = 0; mappingIndex < joinerTypes.length; mappingIndex++) {
-            // For each mapping, bind one join variable.
-            Variable<Object> joinVar = variableFactory.createVariable("joinVar");
-            Function<A, Object> leftMapping = joiner.getLeftMapping(mappingIndex);
-            newPatternVariable = newPatternVariable.bind(joinVar, leftMapping);
-            joinVars[mappingIndex] = joinVar;
-        }
-        // Create the B pattern, joining with the new A pattern using its freshly bound join variables.
-        for (int mappingIndex = 0; mappingIndex < joinerTypes.length; mappingIndex++) {
-            // For each mapping, bind a join variable from A to B and index the binding.
```
rsynek, line 144: How was such a simplification possible?

rsynek (approved):

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/UniLeftHandSide.java
triceo: Good question. The original code was a leftover from a much more complicated world of the ancient CS-D implementation.
In this simplified CS-D, many things are different and one of them is that I can use variables directly instead of proxying them.

triceo:
Jenkins retest this now that the upstream PR is merged.

triceo:
Jenkins was green, which means GHA will be green tomorrow, with the proper SNAPSHOTs.

PR #1154, opened 2021-02-09
Fix update-version.sh following Quarkus quickstart removal

triceo:


triceo:
Build will be broken as expected, because it will still be missing the DROOLS-5995 SNAPSHOT. 
Nevertheless, this needs to be merged.

rsynek (approved):

PR #1174, opened 2021-02-22
PLANNER-2331, PLANNER-2344 Remove CS-D reliance on Drools' FromNode

triceo:
The worst-case difference of 36 % compared to DRL gets down to 25 %.
The median difference becomes 3 % in favor of DRL.
The average difference is none.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/DetachedPatternVariable.java
```diff
@@ -0,0 +1,193 @@
...
+ * {@link IndirectPatternVariable}.
+ *
+ * <p>
+ * The pattern that produces the detached variable will be included in prerequisites for another variable.
+ * Therefore {@link #build()} of this class returns an empty result.
+ *
+ * @param <A> generic type of the variable
+ * @param <PatternVar_>> generic type of the pattern variable, has no effect and exists only to satisfy the interface
+ */
+final class DetachedPatternVariable<A, PatternVar_>
```
triceo, line 77: This is what we had been discussing, @rsynek . I think "detached" works pretty well, but I am open to discussion.

rsynek (approved):

rsynek (approved):

triceo:
Jenkins retest this.

triceo:
Github Actions will be green tomorrow, when the new Kogito SNAPSHOT becomes available.

PR #1179, opened 2021-02-24
PLANNER-2326 Make Constraint Verifier fail on entities with the same PlanningID

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/config/util/ConfigUtils.java
```diff
@@ -415,6 +415,33 @@ public static int resolvePoolSize(String propertyName, String value, String... m
...
 
+    /**
+     * Returns true if the application is currently running in the context of Quarkus.
+     * 
+     * @return true if on Quarkus
+     */
+    public static boolean isInQuarkus() {
+        return Arrays.stream(Thread.currentThread().getStackTrace())
+                .anyMatch(clz -> clz.getClassName().contains("io.quarkus"));
+    }
```
triceo, line 426: See the discussion below on why this is here.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -121,7 +121,11 @@
 
     public static <Solution_> SolutionDescriptor<Solution_> buildSolutionDescriptor(Class<Solution_> solutionClass,
             List<Class<?>> entityClassList) {
-        return buildSolutionDescriptor(DomainAccessType.REFLECTION, solutionClass, entityClassList);
+        // Auto-detect if we should be using Gizmo, as the user did not specify a preference.
+        DomainAccessType domainAccessType =
+                (ConfigUtils.isInQuarkus() && ConfigUtils.isGizmoOnClasspath()) ? DomainAccessType.GIZMO
+                        : DomainAccessType.REFLECTION;
+        return buildSolutionDescriptor(domainAccessType, solutionClass, entityClassList);
     }
```
triceo, line 129: In tests, I want to "do the right thing" and use the best member accessor that is available.
That requires me to know, at this stage, if I'm running in Quarkus.

This change is controversial. We can just say "it always uses reflection", as it did before. I am OK with that.

optaplanner-examples/src/main/java/org/optaplanner/examples/coachshuttlegathering/domain/Bus.java
```diff
@@ -112,7 +112,7 @@ public Bus getBus() {
 
     @Override
     public String toString() {
-        return name;
+        return getClass().getSimpleName() + " " + name;
```
triceo, line 115: Otherwise every exception message just says `null`.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultSingleConstraintVerification.java
```diff
@@ -35,8 +46,53 @@ protected DefaultSingleConstraintVerification(
         this.scoreDirectorFactory = scoreDirectorFactory;
     }
 
+    private void assertDistinctPlanningIds(SolutionDescriptor<Solution_> solutionDescriptor, Object... facts) {
```
triceo, line 49: As you can see, the logic is very heavy. This will slow the Constraint Verifier down.
We have to decide if this is even something we want.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultSingleConstraintVerification.java
```diff
@@ -35,8 +46,53 @@ protected DefaultSingleConstraintVerification(
...
+                    .collect(Collectors.groupingBy(fact -> {
+                        Long planningId = (Long) planningIdAccessor.executeGetter(fact);
+                        return Optional.ofNullable(planningId); // Return as reference to prevent null keys.
+                    }));
+            // Eliminate those matches where there are no duplicate IDs and find the first one.
+            Optional<Map.Entry<Optional<Long>, List<Object>>> firstDuplicateIdEntry = entitiesWithSameIdMap.entrySet()
+                    .stream()
+                    .filter(e -> e.getValue().size() > 1) // More than 1 instance means there is a duplicate ID.
+                    .findFirst();
+            firstDuplicateIdEntry.ifPresent(entry -> {
```
triceo, line 79: I chose to only report the first entity class with duplicate instances.
So when the user fixes this failure in tests, they may need to fix another for another entity class.

I do consider this worthy. Otherwise the exception message would become even more massive than it already is.

ge0ffrey (approved):
Remove the quarkus/gizmo stuff as discussed, because we'll inject the ScoreVerifier with that already sorted out, good to merge then

PR #1184, opened 2021-02-24
PLANNER-2348 MoveIteratorFactory: phase start/stop events

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/factory/MoveIteratorFactoryToMoveSelectorBridge.java
```diff
@@ -51,12 +51,14 @@ public boolean supportsPhaseAndSolverCaching() {
 
     @Override
     public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
-        super.phaseStarted(phaseScope);
         scoreDirector = phaseScope.getScoreDirector();
+        super.phaseStarted(phaseScope);
```
rsynek, line 55: What is the reason for changing the order of these operations?

rsynek (approved):

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/factory/MoveIteratorFactoryToMoveSelectorBridge.java
ge0ffrey: It's like pealing an union and then putting it back together: outside in, inside out.
if you do start A, start B, start C with B depending on A, then it's better to do end C, end B, end A, then the other way around

PR #1188, opened 2021-02-26
PLANNER-2352 Benchmarker: show score explanation summary for each single benchmark

ge0ffrey:
https://issues.redhat.com/browse/PLANNER-2352

Selection_574


triceo:
Is this a good thing? The benchmark reports suffer heavily from "information overload" already. Making it even longer and more dense won't help.

ge0ffrey:
@triceo It's not linked in the menu on the left, people have to scroll to see it, so I am not too worried about information overload.
Over the years and especially today, I've often needed this information, especially when a hard constraint is broken.

triceo (approved):

PR #1200, opened 2021-03-06
PLANNER-2357 Better CS-D accumulate function

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulateFunction.java
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Adapts a CS constraint collector to {@link AccumulateFunction}.
+ *
+ * <p>
+ * {@link AccumulateFunction} has two behaviors:
+ *
+ * <ul>
+ * <li>First, using the {@link #accumulate(Object, Object)} method, it returns the accumulated object which is
+ * later consumed by {@link #reverse(AccumulateContext, Object)}.</li>
```
rsynek, line 38: Is it later consumed by `reverse`? This method throws `UnsupportedOperationException`.

rsynek (approved):
LGTM, just one thing in the JavaDoc is not clear.

PR #1210, opened 2021-03-12
PLANNER-2367: Fix attempt order of class cloning in GIZMO

Christopher-Chianelli:
It appears instanceOfComparator was inverted:
a.isAssignableFrom(b) => b is more specific than a => b should be
attempted/checked in an instanceof before a. Before, in this
case, it returned -1, meaning a was before b in the list, hence
the instanceof was in reverse order! Correction was to change it
to 1. Additionally, interfaces and abstract classes are filtered
out in the Gizmo generating code (so Gizmo won't attempt to
create them).

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
@@ -161,7 +161,7 @@ public final class DeepCloningUtils {
 
     public boolean isTypeArgumentDeepCloned(Type genericType) {
         // Check the generic type arguments of the field.
-        // Yes, it is possible for fields and methods, but not instances!
+        // It is possible for fields and methods, but not instances.
         if (genericType instanceof ParameterizedType) {
             ParameterizedType parameterizedType = (ParameterizedType) genericType;
             for (Type actualTypeArgument : parameterizedType.getActualTypeArguments()) {
@@ -193,39 +193,40 @@ public final class DeepCloningUtils {
      */
     public Set<Class<?>> getDeepClonedTypeArguments(Type genericType) {
         // Check the generic type arguments of the field.
-        // Yes, it is possible for fields and methods, but not instances!
-        if (genericType instanceof ParameterizedType) {
-            Set<Class<?>> deepClonedTypeArguments = new HashSet<>();
-            ParameterizedType parameterizedType = (ParameterizedType) genericType;
-            for (Type actualTypeArgument : parameterizedType.getActualTypeArguments()) {
-                if (actualTypeArgument instanceof Class
-                        && isClassDeepCloned((Class) actualTypeArgument)) {
-                    deepClonedTypeArguments.add((Class) actualTypeArgument);
-                }
-                deepClonedTypeArguments.addAll(getDeepClonedTypeArguments(actualTypeArgument));
-            }
-            return deepClonedTypeArguments;
+        // It is possible for fields and methods, but not instances.
+        if (!(genericType instanceof ParameterizedType)) {
+            return Collections.emptySet();
         }
-        return Collections.emptySet();
+
+        Set<Class<?>> deepClonedTypeArguments = new HashSet<>();
+        ParameterizedType parameterizedType = (ParameterizedType) genericType;
+        for (Type actualTypeArgument : parameterizedType.getActualTypeArguments()) {
+            if (actualTypeArgument instanceof Class
+                    && isClassDeepCloned((Class) actualTypeArgument)) {
+                deepClonedTypeArguments.add((Class) actualTypeArgument);
+            }
+            deepClonedTypeArguments.addAll(getDeepClonedTypeArguments(actualTypeArgument));
+        }
+        return deepClonedTypeArguments;
     }
 
     public Set<Class<?>> getDeepClonedClasses(Collection<Class<?>> entitySubclasses) {
         Set<Class<?>> deepClonedClassSet = new HashSet<>();
-        Set<Class<?>> toSearch = new HashSet<>();
 
-        toSearch.add(solutionDescriptor.getSolutionClass());
-        toSearch.addAll(solutionDescriptor.getEntityClassSet());
-        toSearch.addAll(entitySubclasses);
+        Stream.of(Stream.of(solutionDescriptor.getSolutionClass()),
+                solutionDescriptor.getEntityClassSet().stream(),
+                entitySubclasses.stream())
+                .flatMap(classStream -> classStream)
+                .forEach(clazz -> {
+                    deepClonedClassSet.add(clazz);
+                    for (Field field : getAllFields(clazz)) {
+                        deepClonedClassSet.addAll(getDeepClonedTypeArguments(field.getGenericType()));
+                        if (isClassDeepCloned(field.getType())) {
+                            deepClonedClassSet.add(field.getType());
+                        }
+                    }
+                });
 
-        for (Class<?> clazz : toSearch) {
-            deepClonedClassSet.add(clazz);
-            for (Field field : getAllFields(clazz)) {
-                deepClonedClassSet.addAll(getDeepClonedTypeArguments(field.getGenericType()));
-                if (isClassDeepCloned(field.getType())) {
-                    deepClonedClassSet.add(field.getType());
-                }
-            }
-        }
         return deepClonedClassSet;
     }
 
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
@@ -209,11 +209,21 @@ public final class DeepCloningUtils {
         return Collections.emptySet();
     }
 
-    public Set<Class<?>> getDeepClonedClasses(List<Class<?>> solutionClassList) {
+    public Set<Class<?>> getDeepClonedClasses(Collection<Class<?>> entitySubclasses) {
         Set<Class<?>> deepClonedClassSet = new HashSet<>();
-        for (Class<?> clazz : solutionClassList) {
+        Set<Class<?>> toSearch = new HashSet<>();
+
+        toSearch.add(solutionDescriptor.getSolutionClass());
+        toSearch.addAll(solutionDescriptor.getEntityClassSet());
+        toSearch.addAll(entitySubclasses);
+
+        for (Class<?> clazz : toSearch) {
+            deepClonedClassSet.add(clazz);
             for (Field field : getAllFields(clazz)) {
                 deepClonedClassSet.addAll(getDeepClonedTypeArguments(field.getGenericType()));
+                if (isClassDeepCloned(field.getType())) {
+                    deepClonedClassSet.add(field.getType());
+                }
             }
         }
         return deepClonedClassSet;
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
@@ -21,8 +21,14 @@ import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.optaplanner.core.api.domain.solution.cloner.DeepPlanningClone;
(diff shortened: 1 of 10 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -181,4 +187,56 @@ public boolean isFieldADeepCloneProperty(Field field, Class<?> owningClass) {
         }
         return false;
     }
+
+    /**
+     * @return never null
+     */
+    public Set<Class<?>> getDeepClonedTypeArguments(Type genericType) {
```
triceo, line 194: This method (and others in this class) create a whole lot of collections just to throw them away a little while later.
All of that is going to create quite some GC pressure, and that is going to affect performance.
Of course, this is only a problem if these methods are called frequently on the hot path.

If that is the case, I would consider breaking a guideline on stateless methods and use one collection throughout the recursion. Alternatively, you can do this by concatenating streams, but that has performance downsides of its own.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -181,4 +187,56 @@ public boolean isFieldADeepCloneProperty(Field field, Class<?> owningClass) {
...
+        // Check the generic type arguments of the field.
+        // Yes, it is possible for fields and methods, but not instances!
```
triceo, line 196: Let's leave prose and exclamation marks out of this, shall we? :-)

```suggestion
        // It is possible for fields and methods, but not instances.
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -181,4 +187,56 @@ public boolean isFieldADeepCloneProperty(Field field, Class<?> owningClass) {
...
+            for (Type actualTypeArgument : parameterizedType.getActualTypeArguments()) {
+                if (actualTypeArgument instanceof Class
+                        && isClassDeepCloned((Class) actualTypeArgument)) {
+                    deepClonedTypeArguments.add((Class) actualTypeArgument);
+                }
+                deepClonedTypeArguments.addAll(getDeepClonedTypeArguments(actualTypeArgument));
+            }
+            return deepClonedTypeArguments;
+        }
+        return Collections.emptySet();
```
triceo, line 209: Personally, I'd invert the logic here. First get the simple thing out of the way, then deal with the more complex thing.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -181,4 +187,56 @@ public boolean isFieldADeepCloneProperty(Field field, Class<?> owningClass) {
...
+    }
+
+    public Set<Class<?>> getDeepClonedClasses(Collection<Class<?>> entitySubclasses) {
+        Set<Class<?>> deepClonedClassSet = new HashSet<>();
+        Set<Class<?>> toSearch = new HashSet<>();
+
+        toSearch.add(solutionDescriptor.getSolutionClass());
+        toSearch.addAll(solutionDescriptor.getEntityClassSet());
+        toSearch.addAll(entitySubclasses);
```
triceo, line 218: This could actually be a `stream.forEach()` - creating and resizing this collection (twice) is pure waste.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -181,4 +187,56 @@ public boolean isFieldADeepCloneProperty(Field field, Class<?> owningClass) {
...
+        return deepClonedClassSet;
+    }
+
+    private static List<Field> getAllFields(Class<?> baseClass) {
+        Class<?> clazz = baseClass;
+        Stream<Field> memberStream = Stream.empty();
+        while (clazz != null) {
+            Stream<Field> fieldStream = Stream.of(clazz.getDeclaredFields());
+            memberStream = Stream.concat(memberStream, fieldStream);
+            clazz = clazz.getSuperclass();
```
triceo, line 238: Interfaces can also have fields. But they are `static final`, so we do not need to clone them.
Just adding this for completeness.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -102,13 +103,25 @@ public String getName() {
...
-            Map<Class<?>, GizmoSolutionOrEntityDescriptor> memoizedSolutionOrEntityDescriptorMap) {
+            Map<Class<?>, GizmoSolutionOrEntityDescriptor> memoizedSolutionOrEntityDescriptorMap,
+            Set<Class<?>> deepClonedClassSet) {
+        List<Class<?>> deepCloneClassesThatAreNotSolutionList =
+                deepClonedClassSet.stream()
+                        .filter(clazz -> !solutionClassList.contains(clazz))
+                        .filter(clazz -> !clazz.isInterface() && !Modifier.isAbstract(clazz.getModifiers()))
+                        .sorted(instanceOfComparator)
+                        .distinct()
+                        .collect(Collectors.toList());
```
triceo, line 114: If you put it into a `Set`, you won't need the `distinct()`.
But iterating over the Set will be slower than interating over the List would have been.
Was that the intention here?

triceo: If the intention is to have order, you can use `Collectors.toCollection(...)` and provide a `TreeSet`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -607,7 +629,7 @@ private static void writeDeepCloneMapInstructions(BytecodeCreator bytecodeCreato
             throw new IllegalStateException("Cannot infer element type for Map type (" + type + ").");
         }
 
-        List<Class<?>> entitySubclasses = solutionDescriptor.getSolutionDescriptor().getEntityClassSet().stream()
+        List<Class<?>> entitySubclasses = entityClassOrSubclassList.stream()
                 .filter(keyClass::isAssignableFrom).collect(Collectors.toList());
```
triceo, line 633: I don't understand this formatting choice. Why is `collect()` not on its own line?
(Not the first time in this PR.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -731,10 +754,11 @@ private static String getEntityHelperMethodName(Class<?> entityClass) {
     }
 
     // To prevent stack overflow on chained models
-    private static void createEntityHelperMethod(ClassCreator classCreator,
+    private static void createDeepCloneHelperMethod(ClassCreator classCreator,
             Class<?> entityClass,
             SolutionDescriptor<?> solutionDescriptor,
-            Map<Class<?>, GizmoSolutionOrEntityDescriptor> memoizedSolutionOrEntityDescriptorMap) {
+            Map<Class<?>, GizmoSolutionOrEntityDescriptor> memoizedSolutionOrEntityDescriptorMap,
+            List<Class<?>> entityClassOrSubclassList) {
```
triceo, line 761: Again, strange formatting. Many of these arguments could be on the same line.

optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
```diff
@@ -403,6 +407,84 @@ private void assertChainedEntityClone(TestdataChainedEntity originalEntity, Test
...
+        assertChainedShadowingAnchorClone(a0, cloneA0, "a0", cloneA1);
+        assertChainedShadowingEntityClone(a1, cloneA1, "a1", cloneA0, cloneA2);
+        assertChainedShadowingEntityClone(a2, cloneA2, "a2", cloneA1, cloneA3);
+        assertChainedShadowingEntityClone(a3, cloneA3, "a3", cloneA2, null);
+        assertChainedShadowingAnchorClone(b0, cloneB0, "b0", cloneB1);
+        assertChainedShadowingEntityClone(b1, cloneB1, "b1", cloneB0, null);
+
+        a3.setChainedObject(b1);
+        assertCode("b1", a3.getChainedObject());
+        // Clone remains unchanged
```
triceo, line 465: Capital letter, full stop at the end.

```suggestion
        // Clone remains unchanged.
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -294,8 +296,37 @@ public static String generateSolutionCloner(SolutionDescriptor solutionDescripto
                         transformers);
             }
 
+            DeepCloningUtils deepCloningUtils = new DeepCloningUtils(solutionDescriptor);
+
+            Set<Class<?>> solutionAndEntitySubclassSet = new HashSet<>(solutionSubclasses);
```
triceo, line 301: According to the convention, `solutionSubclasses` should be a `solutionSubclassesList`.
Also, not sure why the copy is being made - maybe I missed something in the diff, but it seems like you don't need the original collection after this.

triceo (approved):
Looks good, although there are some small adjustments to be made.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: TBH, the comparator fails transitively: "Animal > Robot, Robot > Zebra", but "Zebra > Animal" (unrelated are sorted by name, related by isInstanceOf). We will need to use something like https://github.com/kiegroup/optaweb-employee-rostering/blob/master/optaweb-employee-rostering-backend/src/main/java/org/optaweb/employeerostering/util/HierarchyTree.java  in order to decide what subclass to create properly. But I rather not create a new data structure for this; we might be able to calculate the "Root" classes (classes which are not subclasses of any other classes in the collection), construct a Map<Class<?>, List<Class<?>> that map each class in the collection to a list of it superclasses in order to define a better comparator).

As for why it needed:
We want:
```
if (x instanceof subclass1) {
...
} else if ((x instanceof subclass2) {
...
}  else if (x instanceof superclass) {
...
}
```
and not
```
if (x instanceof superclass) {
...
} else if ((x instanceof subclass1) {
...
}  else if (x instanceof subclass2) {
...
}
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
Christopher-Chianelli: Original list is used on 328 (and expects the list to contain only solution subclasses)

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: Fixed;
for something like this:
```
    private interface Animal {
    }

    private interface Robot {
    }

    private interface Zebra extends Animal {
    }

    private interface RobotZebra extends Zebra, Robot {
    }
```
the comparator will sort it into
`[RobotZebra (2), Zebra (1), Robot (0), Animal (0)]`

(The number in parenthesis is the subclass level; root classes (which are not subclasses of anything) have a subclass level of 0; all other classes have a subclass level of the max subclass level of it parent classes + 1).

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -181,4 +187,57 @@ public boolean isFieldADeepCloneProperty(Field field, Class<?> owningClass) {
...
+        return deepClonedTypeArguments;
+    }
+
+    public Set<Class<?>> getDeepClonedClasses(Collection<Class<?>> entitySubclasses) {
+        Set<Class<?>> deepClonedClassSet = new HashSet<>();
+
+        Stream.of(Stream.of(solutionDescriptor.getSolutionClass()),
+                solutionDescriptor.getEntityClassSet().stream(),
+                entitySubclasses.stream())
+                .flatMap(classStream -> classStream)
```
triceo, line 219: `Stream.concat()` seems easier to read in this case.

triceo (approved):
One last comment. After that, with all this test coverage, I say it's good to merge. Or are you waiting for someone else, too?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
Christopher-Chianelli: `Stream.concat()` only exists for two arguments, so it would be
`Stream.concat(Stream.of(solutionDescriptor.getSolutionClass()),
Stream.concat(solutionDescriptor.getEntityClassSet().stream(),entitySubclasses.stream()))` (and if for each argument, another `Stream.concat` is needed). `Stream.of(...).flatmap(s -> s)` works for any number of arguments .

Christopher-Chianelli:
@triceo @yurloc already tried the patch on OptaWeb Vehicle Routing and can confirm it fixes the issue. I don't think `Stream.concat()` will be easier to read since it need to be nested when there more than one argument. Note: I did not change the default back to GIZMO in Quarkus yet (although this is the only known issue). Feel free to merge (the change default back to GIZMO can be done in another PR; I'll test optaplanner-examples with GIZMO (w/ some modifications so they'll work)).

PR #1220, opened 2021-03-17
PLANNER-2372 Improve performance of CS-D

triceo:
Requires https://github.com/kiegroup/kogito-runtimes/pull/1171.
Don't be afraid of the 500.000k lines of code, that's just a change in XML of one of the examples.


build/optaplanner-build-parent/pom.xml
```diff
@@ -22,6 +22,7 @@
 
     <!-- SNAPSHOT and KIE dependency versions -->
     <version.org.kie.kogito>2.0.0-SNAPSHOT</version.org.kie.kogito>
+    <version.org.drools>7.52.0-SNAPSHOT</version.org.drools> <!-- Must match version used by Kogito. -->
```
triceo, line 25: I don't really like this, but we need it for the alpha network compiler module, which is only in Drools, not in Kogito.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintFactory.java
```diff
@@ -35,10 +35,6 @@
 import org.optaplanner.core.impl.score.stream.bi.NoneBiJoiner;
 
 /**
- * WARNING: The ConstraintStreams/ConstraintProvider API is TECH PREVIEW.
- * It works but it has many API gaps.
- * Therefore, it is not rich enough yet to handle complex constraints.
- * <p>
```
triceo, line 41: This has somehow managed to survive there for a year.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -281,13 +280,4 @@ private void checkMultipleScoreDirectorFactoryTypes(EasyScoreDirectorFactory eas
...
 
-    protected KieBaseConfiguration buildKieBaseConfiguration(KieServices kieServices) {
-        KieBaseConfiguration kieBaseConfiguration = kieServices.newKieBaseConfiguration();
-        if (config.getKieBaseConfigurationProperties() != null) {
-            for (Map.Entry<String, String> entry : config.getKieBaseConfigurationProperties().entrySet()) {
-                kieBaseConfiguration.setProperty(entry.getKey(), entry.getValue());
-            }
-        }
-        return kieBaseConfiguration;
-    }
```
triceo, line 292: This was only used in a test, testing this method.

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/domain/Examination.java
```diff
@@ -134,29 +143,4 @@ public void setScore(HardSoftScore score) {
...
-                        }
-                    }
-                    if (studentSize > 0) {
-                        topicConflictList.add(new TopicConflict(leftTopic, rightTopic, studentSize));
-                    }
-                }
-            }
-        }
-        return topicConflictList;
-    }
```
triceo, line 160: As I was testing different examples, I found out that examination takes an ungodly amount of time to initialize. Often many many minutes. I fixed it - the student list is now a student set, because `contains()` is just incredibly painful on a list.

build/optaplanner-build-parent/pom.xml
ge0ffrey: The only reason I don't like it is because Drools is part of the KIE 7 release train. The release train architecture is a trainwreck for as long as Drools 8 isn't branched. By making this direct dependency we become a active participant to the situation.

ge0ffrey: Let's standarize the comment on "Keep in sync with"  and add a better breadcrumb to the actual location
```suggestion
    <!-- Keep in sync with version.org.kie7 in kogito-runtimes: https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-build-parent/pom.xml#L185 -->
    <version.org.drools>7.52.0-SNAPSHOT</version.org.drools> <!-- Must match version used by Kogito. -->
```

optaplanner-examples/src/main/java/org/optaplanner/examples/examination/domain/Examination.java
ge0ffrey: I agree the studentList should be a studentSet (and yes a LinkedHashSet for reproducibility which this PR does)

Are we sure we want to make List<TopicConflict> part of the dataset generation (= increase the dataset XML file sizes) and not just calculate it on the fly once when the solver started like we did before?
I suspect your student List to Set change made this method a lot faster so it can stay here instead of going into the generator. Wdyt?

ge0ffrey (approved):
LGTM
do improve the "keep in sync" comment

PR #1230, opened 2021-03-22
PLANNER-2323 Constraint Streams: map() and distinct()

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsVariableFactory.java
```diff
@@ -37,28 +42,106 @@
...
      */
-    <X> Variable<X> createVariable(String baseName, Variable<X> source);
+    <U> Variable<U> createVariable(String baseName, Variable<U> source, boolean flatten);
 
     /**
-     * Declares a new {@link Object}-typed variable, see {@link #createVariable(Class, String)} for details.
+     * Declare a new {@link Variable} with a given name,
+     * where the value of the variable will be computed from the provided source variable.
+     * If the computed value is {@link Iterable},
+     * the final argument to this method controls whether the elements will be treated individually or not.
```
rsynek, line 116: These lines duplicate the javadoc starting on line 98. Could we use a @link like in the others?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * calling {@code map(Person::getAge)} on such stream will produce a stream of {@link Integer}s
+     * {@code [20, 25, 30]},
+     *
+     * <p>
+     * Example with a non-bijective mapping function: assuming a constraint stream of tuple of {@code Person}s
+     * {@code [Ann(age = 20), Beth(age = 25), Cathy(age = 30), David(age = 30), Eric(age = 20)]},
+     * calling {@code map(Person::getAge)} on such stream will produce a stream of {@link Integer}s
+     * {@code [20, 25, 30, 30, 20]}.
+     *
+     * @param mapping never null, function to convert the original tuple into the new tuple
```
rsynek, line 990: I haven't noticed any null check to fail fast if this precondition is not met in the implementation code.

rsynek (approved):
LGTM, just a couple of suggestions for improvement.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/quad/DroolsGroupingQuadConstraintStream.java
```diff
@@ -102,6 +102,78 @@
...
+            DroolsAbstractQuadConstraintStream<Solution_, A, B, C, D> parent,
+            QuadFunction<A, B, C, D, NewA> groupKeyAMapping, QuadFunction<A, B, C, D, NewB> groupKeyBMapping,
+            QuadFunction<A, B, C, D, NewC> groupKeyCMapping, QuadConstraintCollector<A, B, C, D, ?, NewD> collectorD) {
+        super(constraintFactory);
+        this.leftHandSide =
+                parent.getLeftHandSide().andGroupBy(groupKeyAMapping, groupKeyBMapping, groupKeyCMapping, collectorD);
+    }
+
+    public <A> DroolsGroupingQuadConstraintStream(DroolsConstraintFactory<Solution_> constraintFactory,
+            DroolsAbstractUniConstraintStream<Solution_, A> parent, Function<A, NewA> groupKeyAMapping,
```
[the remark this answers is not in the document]
triceo: The ordering is tricky here. I generally follow this convention throughout CS:

Let `C >= 0` be the number of collectors.
Let `M >= 0` be the number of mappings.

Methods will first be sorted by increasing order of `C`. Where `C1 == C2`, methods will be sorted by increasing order of `M`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsVariableFactory.java
triceo: The two are not exactly duplicates, the verbs are different. Therefore I'd rather keep it as is.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractLeftHandSide.java
```diff
@@ -67,4 +69,91 @@ protected AbstractLeftHandSide(DroolsVariableFactory variableFactory) {
...
+    protected static ViewItem<?> buildAccumulate(ViewItem<?> innerAccumulatePattern,
+            AccumulateFunction... accFunctions) {
+        if (accFunctions.length == 0) {
+            throw new IllegalStateException("Impossible state: no accumulate functions provided.");
+        } else if (accFunctions.length == 1) {
+            return DSL.accumulate(innerAccumulatePattern, accFunctions[0]);
+        } else {
+            return DSL.accumulate(innerAccumulatePattern, accFunctions[0],
+                    Arrays.stream(accFunctions)
+                            .skip(1)
```
rsynek, line 154: Would you please explain why we skip the first element?

rsynek (approved):

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractLeftHandSide.java
triceo: Because we've already used it on the previous line.

rsynek: Yes, I've just overlooked it.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -685,6 +685,106 @@
                     BiConstraintCollector<A, B, ResultContainerC_, ResultC_> collectorC,
                     BiConstraintCollector<A, B, ResultContainerD_, ResultD_> collectorD);
 
+    /**
+     * Convert the {@link BiConstraintStream} to a {@link TriConstraintStream}, consisting of unique tuples with three
+     * facts.
+     * <p>
+     * The first fact is the return value of the first group key mapping function, applied on the incoming tuple.
+     * The second fact is the return value of the second group key mapping function, applied on all incoming tuples with
+     * the same first fact.
```
ge0ffrey, line 694: "the first fact is ..." "the second fact is  ... with the same first fact". That "same first fact" seems wrong. Or at least confusing.
The second result fact isn't more dependent on the first result fact than visa versa, this line seems to imply that.

I'd scratch this paragraph with 3 lines and just add an example instead that shows that illustrates when and when not tuples get grouped together.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -685,6 +685,106 @@
...
+     *
+     * @param mapping never null, function to convert the original tuple into the new tuple
+     * @param <ResultA_> the type of the only fact in the resulting {@link UniConstraintStream}'s tuple
+     * @return never null
+     */
+    <ResultA_> UniConstraintStream<ResultA_> map(BiFunction<A, B, ResultA_> mapping);
+
+    /**
+     * Transforms the stream in such a way that all the tuples going through it are distinct.
+     * (No two tuples will {@link Object#equals(Object) equal}.)
```
ge0ffrey, line 776: Avoid future tense:
```suggestion
     * (No two result tuples are {@link Object#equals(Object) equal}.)
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -685,6 +685,106 @@
...
+     *
+     * <p>
+     * By default, tuples going through a constraint stream are distinct.
+     * However, operations such as {@link #map(BiFunction)} may create a stream which breaks that promise.
+     * By calling this method on such a stream,
+     * duplicate copies of the same tuple will be omitted at a performance cost.
```
ge0ffrey, line 782:
```suggestion
     * duplicate copies of the same tuple are omitted at a performance cost.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintStream.java
```diff
@@ -688,6 +688,106 @@
...
+     * @param <GroupKeyD_> the type of the fourth fact in the destination {@link QuadConstraintStream}'s tuple
+     * @return never null
+     */
+    <GroupKeyA_, GroupKeyB_, GroupKeyC_, GroupKeyD_>
+            QuadConstraintStream<GroupKeyA_, GroupKeyB_, GroupKeyC_, GroupKeyD_> groupBy(
+                    TriFunction<A, B, C, GroupKeyA_> groupKeyAMapping, TriFunction<A, B, C, GroupKeyB_> groupKeyBMapping,
+                    TriFunction<A, B, C, GroupKeyC_> groupKeyCMapping, TriFunction<A, B, C, GroupKeyD_> groupKeyDMapping);
+
+    // ************************************************************************
+    // Operations w/ duplicate tuple possibility
```
ge0ffrey, line 765: What does w/ stand for?

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/tri/TriConstraintStream.java
```diff
@@ -688,6 +688,106 @@
...
+     * As defined by {@link UniConstraintStream#map(Function)}.
+     *
+     * @param mapping never null, function to convert the original tuple into the new tuple
+     * @param <ResultA_> the type of the only fact in the resulting {@link UniConstraintStream}'s tuple
+     * @return never null
+     */
+    <ResultA_> UniConstraintStream<ResultA_> map(TriFunction<A, B, C, ResultA_> mapping);
+
+    /**
+     * Transforms the stream in such a way that all the tuples going through it are distinct.
```
ge0ffrey, line 778: "in such a way" seems a bit verbose, how about something like:

```suggestion
     * Removes duplicate tuples from the stream, according to the tuple's fact's {@link Object#equals(Object)equals}/hashcode methods, such that only distinct tuples remain.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+            QuadConstraintStream<GroupKeyA_, GroupKeyB_, GroupKeyC_, GroupKeyD_> groupBy(
+                    Function<A, GroupKeyA_> groupKeyAMapping, Function<A, GroupKeyB_> groupKeyBMapping,
+                    Function<A, GroupKeyC_> groupKeyCMapping, Function<A, GroupKeyD_> groupKeyDMapping);
+
+    // ************************************************************************
+    // Operations w/ duplicate tuple possibility
+    // ************************************************************************
+
+    /**
+     * Transforms the stream in such a way that tuples will be remapped using the given function.
```
ge0ffrey, line 949:
```suggestion
     * Transforms the stream in such a way that tuples areremapped using the given function.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * <ul>
+     * <li>Purity.
+     * The mapping function should only depend on its input.
+     * That is, given the same input, it always returns the same output.
+     * </li>
+     * <li>Simplicity.
+     * The function should only take the argument and return another simple value,
+     * without performing any significant computation.
+     * Not following this recommendation may result in performance loss,
+     * as the function may be called frequently.</li>
```
ge0ffrey, line 964: Why? If the business requirements require a complex, performance intensive function, so be it? I wouldn't recommend them to avoid map() in those cases.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * <li>Bijectivity.
+     * No two input tuples should map to the same output tuple,
+     * or to tuples that are {@link Object#equals(Object) equal}.
+     * Not following this recommendation will create a constraint stream with duplicate tuples,
```
ge0ffrey, line 968:
```suggestion
     * Not following this recommendation creates a constraint stream with duplicate tuples,
```
Please do a search and find on "will" :)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * and may force you to use {@link #distinct()} later.</li>
```
ge0ffrey, line 969:
```suggestion
     * and may force you to use {@link #distinct()} later, which comes with a performance cost.</li>
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * <li>Immutable data carriers.
+     * The tuples returned by the mapping function should be identified by their contents and nothing else.
+     * They should also be immutable, meaning their contents should not be allowed to change.
+     * If two tuples carry objects which {@link Object#equals(Object) equal} one another,
+     * those two tuples should likewise {@link Object#equals(Object) equal} and preferably be the same instance.
+     * Java Records are the ideal tool for this particular use case.</li>
```
ge0ffrey, line 975: Did we test Java Records for this? If it's not proven - we just think it is - we should probably leave it out.
I have done recommendations in the past that I hadn't tried out myself yet (although I was convinced they were a great idea) and people have treated those recommendations as something-that-should-be-questioned because the docs said it.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * <li>Bijectivity.
+     * No two input tuples should map to the same output tuple,
+     * or to tuples that are {@link Object#equals(Object) equal}.
+     * Not following this recommendation will create a constraint stream with duplicate tuples,
+     * and may force you to use {@link #distinct()} later.</li>
+     * <li>Immutable data carriers.
+     * The tuples returned by the mapping function should be identified by their contents and nothing else.
+     * They should also be immutable, meaning their contents should not be allowed to change.
+     * If two tuples carry objects which {@link Object#equals(Object) equal} one another,
+     * those two tuples should likewise {@link Object#equals(Object) equal} and preferably be the same instance.
```
ge0ffrey, line 974: The use of the word "tuple" is confusing here.

Is "tuple" something internally to OptaPlanner constraints, the internal wrapper of <A, B> in a BiStream<A, B>. Or is it the wrapper that a user creates to wrap-map the <A, B> into his/her data carrier?

Sugestion: replace "tuple" with "data carrier" (or something better) for all cases where we don't mean the internal tuple.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/uni/UniConstraintStream.java
```diff
@@ -867,6 +867,146 @@
...
+     * Java Records are the ideal tool for this particular use case.</li>
+     * </ul>
+     *
+     * <p>
+     * Simple example: assuming a constraint stream of tuple of {@code Person}s
+     * {@code [Ann(age = 20), Beth(age = 25), Cathy(age = 30)]},
+     * calling {@code map(Person::getAge)} on such stream will produce a stream of {@link Integer}s
+     * {@code [20, 25, 30]},
```
ge0ffrey, line 982: I love such examples! :) (just hit resolve on this one)

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -176,6 +176,31 @@ The following constraint stream cardinalities are currently supported:
 |4          |  Quad|`QuadConstraintStream<A, B, C, D>`
 |===
 
+[[constraintStreamsHigherCardinalities]]
+=== Achieving higher cardinalities
+
+OptaPlanner currently does not support constraint stream cardinalities higher than 4.
+However, with <<constraintStreamsMappingTuples,tuple mapping>> it is easy to achieve effectively infinite cardinality:
```
ge0ffrey, line 183: docs and blogs should avoid "it is easy" (although I suspect more docs chapter I wrote suffer from this).

"it is easy" is subjective - it depends on previous experience and understanding of context.
Worse, it also implies an intelligence test for the reader. -> If you can't understand this quickly, you're an idiot. => OptaPlanner is too difficult for me, I need to look for a simpler solution.

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -176,6 +176,31 @@ The following constraint stream cardinalities are currently supported:
...
+                .map(Tuple::of)          // UniConstraintStream<Tuple<Shift, Shift, Shift, Shift>>
+                .join(Shift.class)       // BiConstraintStream<Tuple<Shift, Shift, Shift, Shift>, Shift>
+                ...                      // This BiConstraintStream carries 5 Shift elements.
+    }
+----
+
+[NOTE]
+====
+OptaPlanner does not provide any `Tuple` implementations out of the box.
+We recommend using one of the myriad freely available 3rd party implementations.
```
ge0ffrey, line 201: Avoid "we" in docs. Avoid the tense of "using" in favor of active:

```suggestion
It's recommended to use one of the myriad freely available 3rd party implementations.
```

Ps: What does "myriad" mean? Many docs readers aren't native English speakers. It's ok to use difficult words if they describe an property of the behavior (such as "bijective"), as it is explained immediately. But in normal sentences, like this, maybe we should avoid uncommon English words without explanation unless they bring enough value. Wdyt?

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -176,6 +176,31 @@ The following constraint stream cardinalities are currently supported:
...
+
+[source,java,options="nowrap"]
+----
+    private Constraint pentaStreamExample(ConstraintFactory factory) {
+        return factory.from(Shift.class) // UniConstraintStream<Shift>
+                .join(Shift.class)       // BiConstraintStream<Shift, Shift>
+                .join(Shift.class)       // TriConstraintStream<Shift, Shift, Shift>
+                .join(Shift.class)       // QuadConstraintStream<Shift, Shift, Shift, Shift>
+                .map(Tuple::of)          // UniConstraintStream<Tuple<Shift, Shift, Shift, Shift>>
+                .join(Shift.class)       // BiConstraintStream<Tuple<Shift, Shift, Shift, Shift>, Shift>
```
ge0ffrey, line 193: the class Tuple doesn't exist, that makes it a bit clumsy to copy-paste. commons-lang3 has Triple and Pair, but not for 4...

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -585,6 +610,77 @@ Joining should only be used in cases where the other fact is actually required f
...
+
+[NOTE]
+====
+In the example above, the mapping function produces duplicate tuples if two different `CloudProcess`es share a single `CloudComputer`.
+That is, such `CloudComputer` will appear in the resulting constraint stream twice.
+See <<constraintStreamsDistinct,`distinct()`>> for how to deal with duplicate tuples.
+====
+
+[[constraintStreamsMappingFunction]]
+==== Designing the mapping function
```
ge0ffrey, line 636: (soft) the id differs from the title - so far the id, was usually <namespace of out section if needed> + <camelCase of title or at least the first few words of it>
I'd keep the title and the id very similar, so it's clear they are the same section.

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -585,6 +610,77 @@ Joining should only be used in cases where the other fact is actually required f
...
+
+When designing the mapping function, we recommend you follow these guidelines for optimal performance:
```
ge0ffrey, line 638:
```suggestion
When designing the mapping function, follow these guidelines for optimal performance:
```

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -585,6 +610,77 @@ Joining should only be used in cases where the other fact is actually required f
...
+  The mapping function should only depend on its input.
+  That is, given the same input, it always returns the same output.
+* Keep the function simple.
+  The function should only take the argument and return another simple value,
+  without performing any significant computation.
+  Not following this recommendation may result in performance loss,
+  as the function may be called frequently.
+* Keep the function bijective.
+  No two input tuples should map to the same output tuple, or to tuples that are equal.
+  Not following this recommendation will create a constraint stream with duplicate tuples,
```
ge0ffrey, line 650:
```suggestion
  Not following this recommendation creates a constraint stream with duplicate tuples,
```

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -585,6 +610,77 @@ Joining should only be used in cases where the other fact is actually required f
...
+  and may force you to use <<constraintStreamsDistinct,`distinct()`>> later.
+* Use immutable data carriers.
+  The tuples returned by the mapping function should be immutable and identified by their contents and nothing else.
+  If two tuples carry objects which equal one another,
+  those two tuples should likewise equal and preferably be the same instance.
+  https://openjdk.java.net/jeps/395[Java's Records] are an ideal tool for the purpose.
```
ge0ffrey, line 656: Same comment here as above -> only if we tested it and we are sure it's a good recommendation. Otherwise, let's leave it out for now.

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -585,6 +610,77 @@ Joining should only be used in cases where the other fact is actually required f
...
+
+[[constraintStreamsDistinct]]
+==== Dealing with duplicate tuples
```
ge0ffrey, line 659: I'd love to see the word "distinct" in this section header as that's the operation it's talking about.
In case I use it (or I inherit a piece of code that uses distinct()) and I want to fully understand what it does by reading the docs, so I need to find it in the TOC.

optaplanner-docs/src/main/asciidoc/ConstraintStreams/ConstraintStreams-chapter.adoc
```diff
@@ -585,6 +610,77 @@ Joining should only be used in cases where the other fact is actually required f
...
+                .map(CloudProcess::getComputer)           // UniConstraintStream<CloudComputer>
+                .distinct()                               // The same, each CloudComputer just once.
+                ...
+    }
+----
+
+[NOTE]
+====
+There is a performance cost to `distinct()`.
+For optimal performance, make sure that your constraint streams never produce duplicate tuples.
```
ge0ffrey, line 682:
```suggestion
For optimal performance, don't use constraint stream operations that produce duplicate tuples, to avoid the need to call `distinct()`.
```

ge0ffrey (approved):
Looks great, I look forward to play with these operations :)

PR #1231, opened 2021-03-23
PLANNER-2374 Add support for JDK 16

triceo:
XStream has issues with JDK 16. I solve them by using different collection types, which XStream can serialize just fine.
This, unfortunately, required us to regenerate some unsolved example XMLs.


rsynek (approved):
Nice work

PR #1236, opened 2021-03-25
PLANNER-2372 Re-enable ANC when not native

triceo:
This assumes that we can have ANC on the classpath in native. 
If that is not true, then there is currently no solution to bring ANC back.


triceo:
Jenkins run native.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/KieBaseUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import org.drools.ancompiler.KieBaseUpdaterANC;
+import org.drools.core.base.CoreComponentsBuilder;
+import org.kie.api.KieBase;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class KieBaseUtil {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KieBaseUtil.class);
+    private static final String ANC_ENABLED_PROPERTY = "optaplanner.drools.compile_alpha_network";
```
ge0ffrey, line 28: Let's not use a system property, but a field in ScoreDirectorFactoryConfig
That what it's consistent with SolverConfig.domainAccessType
and we can run multiple solver instances with a different setting in the same JVM.

ge0ffrey: It will also need fail fasts:
if the score director is not CS-D, then fail fast if that field is non-null.

ge0ffrey (changes requested):
Don't use a system property please.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -156,6 +174,35 @@ public void setScoreDrlFileList(List<File> scoreDrlFileList) {
...
 
+    public Boolean getCompileDroolsAlphaNetwork() {
+        return compileDroolsAlphaNetwork;
+    }
+
+    public void setCompileDroolsAlphaNetwork(Boolean compileDroolsAlphaNetwork) {
+        this.compileDroolsAlphaNetwork = compileDroolsAlphaNetwork;
+    }
+
+    public boolean isDroolsAlphaNetworkCompilerEnabled() {
```
triceo, line 185: This method is used from Quarkus/Spring boot code injection.
I don't particularly like that it is here, but it is the only reasonable way I could think of to share this code.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -180,7 +188,17 @@ private void checkMultipleScoreDirectorFactoryTypes(EasyScoreDirectorFactory eas
         }
     }
 
-    protected ConstraintStreamScoreDirectorFactory<Solution_, Score_> buildConstraintStreamScoreDirectorFactory(
+    private void enforceNoAlphaNetworkCompiler() {
```
triceo, line 191: Fail-fast.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/AbstractConstraintStreamScoreDirectorFactory.java
```diff
@@ -41,27 +38,16 @@
  * @see ConstraintStreamScoreDirector
  * @see ScoreDirectorFactory
  */
-public class ConstraintStreamScoreDirectorFactory<Solution_, Score_ extends Score<Score_>>
+public abstract class AbstractConstraintStreamScoreDirectorFactory<Solution_, Score_ extends Score<Score_>>
```
triceo, line 41: So that I can only deal with ACN in code that is actually using Drools.

triceo:
@rsynek Please verify the changes to solver config.
@Christopher-Chianelli Please check the changes to SB/Q injection/autowiring.
@ge0ffrey Your final approval, please.

rsynek (approved):
LGTM, but I mostly checked the config code.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -43,12 +46,15 @@
...
         "scoreDrlList",
         "scoreDrlFileList",
+        "compileDroolsAlphaNetwork",
         "kieBaseConfigurationProperties",
         "initializingScoreTrend",
         "assertionScoreDirectorFactory"
 })
 public class ScoreDirectorFactoryConfig extends AbstractConfig<ScoreDirectorFactoryConfig> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(ScoreDirectorFactoryConfig.class);
```
ge0ffrey, line 56: (separate issue)
in SolverManagerConfig we use the variable name `logger`, which violates name conventions, lets' fix this accross the code in a different PR?

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
ge0ffrey: In any case, let's not put it between the getters and setters, but before or after all the "simple getters and setters". These days we put the "simple getters and setters" at the end because it's just Java ceremony.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -123,6 +131,16 @@ public void setConstraintStreamImplType(ConstraintStreamImplType constraintStrea
         this.constraintStreamImplType = constraintStreamImplType;
     }
 
+    private boolean isUsingDrools() {
```
ge0ffrey, line 134: put before or after all simple getters/setters

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -123,6 +131,16 @@ public void setConstraintStreamImplType(ConstraintStreamImplType constraintStrea
...
+        if (scoreDrlList != null || scoreDrlFileList != null) { // We know we're in DRL.
+            return true;
+        } else if (constraintProviderClass == null) { // We know we're neither in DRL nor in CS.
+            return false;
+        }
+        // We know we're in CS.
```
ge0ffrey, line 140: I'd remove this comment as it's obvious

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -156,6 +174,35 @@ public void setScoreDrlFileList(List<File> scoreDrlFileList) {
...
+        return compileDroolsAlphaNetwork;
+    }
+
+    public void setCompileDroolsAlphaNetwork(Boolean compileDroolsAlphaNetwork) {
+        this.compileDroolsAlphaNetwork = compileDroolsAlphaNetwork;
+    }
+
+    public boolean isDroolsAlphaNetworkCompilerEnabled() {
+        if (!isUsingDrools()) {
+            LOGGER.trace("Drools Alpha Network compiler is disabled when not using Drools.");
```
ge0ffrey, line 187: An is***() method should not be doing logs like that. If it's used twice, we log twice.
A solvingStarted can - if desired - if the alpha network is enabled. But probably on a line along with other score director information (for example number of constraints) to avoid a verbose log.

See how much info the solvingSTarted line has

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -244,6 +291,11 @@ public ScoreDirectorFactoryConfig withScoreDrlFiles(File... scoreDrlFiles) {
         return this;
     }
 
+    public ScoreDirectorFactoryConfig withCompileDroolsAlphaNetworkEnabled(boolean compileDroolsAlphaNetwork) {
```
ge0ffrey, line 294: method name doesn't match field name "Enabled)

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -70,6 +76,8 @@
     @XmlElement(name = "scoreDrlFile")
     protected List<File> scoreDrlFileList = null;
 
+    protected Boolean compileDroolsAlphaNetwork = null;
```
ge0ffrey, line 79: Naming: maybe droolsAlphaNetworkCompilationEnabled ?

ge0ffrey: reasoning: start with drools for context + easier to find if you were looking for something like this
end with Enabled to make it clear it's a switch and what true/false actually does (= turn it on/off)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
ge0ffrey: We typically call such methods "validate"

```suggestion
    private void validateNoAlphaNetworkCompiler() {
```

ge0ffrey: Looks like this validation doesn't trigger if BAVET is used and it's true

ge0ffrey: Maybe, you can just create a validateDroolsAlphaNetworkCompilation method that is always called, not just when it needs to be disabled?

triceo:
Jenkins run native.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/AbstractConstraintStreamScoreDirectorFactory.java
ge0ffrey: ConstraintStreamScoreDirectorFactory is not public API. Users don't know it exists.
So why do we need to split it up in Abstract* and subclasses?

ge0ffrey: ah, in the drools version we want to pass that boolean and in bavet it doesn't exist. ok, makes sense...

optaplanner-test/pom.xml
```diff
@@ -64,6 +64,24 @@
         <groupId>org.revapi</groupId>
         <artifactId>revapi-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
```
ge0ffrey, line 69: Is this temporary or permanent?

ge0ffrey: Let's add an expirary comment, something like add "Remove when we depend on drools directly" comment - or even better, with jira url to that task

ge0ffrey: => will be done in next pr anyway apparently, good to leave as is

optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/ConstraintVerifier.java
```diff
@@ -60,6 +60,16 @@
...
             ConstraintStreamImplType constraintStreamImplType);
 
+    /**
+     * Applies only to {@link ConstraintStreamImplType#DROOLS}.
+     * Do not enable when running in a native image.
+     *
+     * @param droolsAlphaNetworkCompilerEnabled true to enable the alpha network compiler
+     * @return this
+     */
+    ConstraintVerifier<ConstraintProvider_, Solution_> withDroolsAlphaNetworkCompiler(
```
ge0ffrey, line 70: if the field is called CompiLATION, this needs to be called ...ation too. And visa versa.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultConstraintVerifier.java
```diff
@@ -56,6 +56,21 @@ public ConstraintStreamImplType getConstraintStreamImplType() {
...
     }
 
+    public boolean isDroolsAlphaNetworkCompilerEnabled() {
+        return droolsAlphaNetworkCompilerEnabled;
+    }
+
+    @Override
+    public ConstraintVerifier<ConstraintProvider_, Solution_>
+            withDroolsAlphaNetworkCompiler(boolean droolsAlphaNetworkCompilerEnabled) {
+        if (getConstraintStreamImplType() != ConstraintStreamImplType.DROOLS && droolsAlphaNetworkCompilerEnabled) {
```
ge0ffrey, line 66: I wouldn't even check the value of droolsAlphaNetworkCompilerEnabled.
If this is called and it's not using drools, let's just fail fast?

ge0ffrey (approved):
LGTM if naming is consistent and starts with drools + ends with Enabled

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
```diff
@@ -179,11 +185,14 @@ public SolverConfig solverConfig() {
...
+                    solverConfig.getScoreDirectorFactoryConfig().isDroolsAlphaNetworkCompilerEnabled();
 
             return (ConstraintVerifier<ConstraintProvider_, SolutionClass_>) ConstraintVerifier.build(constraintProvider,
-                    solverConfig.getSolutionClass(),
-                    entityClasses).withConstraintStreamImplType(
-                            (constraintStreamImplType != null) ? constraintStreamImplType : ConstraintStreamImplType.DROOLS);
+                    solverConfig.getSolutionClass(), entityClasses)
+                    .withConstraintStreamImplType(
+                            (constraintStreamImplType != null) ? constraintStreamImplType : ConstraintStreamImplType.DROOLS)
+                    .withDroolsAlphaNetworkCompiler(droolsAlphaNetworkCompilerEnabled);
```
ge0ffrey, line 195: ah, that's why the validation accepts false.
Maybe it should accept Boolean? Idunno. As is is fine too. This is serious nitpick. Feel free to ignore.

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
```diff
@@ -162,6 +162,12 @@ public SolverConfig solverConfig() {
                         throw new UnsupportedOperationException(noConstraintProviderErrorMsg);
                     }
 
+                    @Override
+                    public ConstraintVerifier<ConstraintProvider_, SolutionClass_> withDroolsAlphaNetworkCompiler(
+                            boolean droolsAlphaNetworkCompilerEnabled) {
+                        throw new UnsupportedOperationException(noConstraintProviderErrorMsg);
```
ge0ffrey, line 168: (separate issue, not part of this PR) cc @Christopher-Chianelli if you could take a look, Chris:

Let's also improve the error message for clarity, so it :
```
final String noConstraintProviderErrorMsg = "ConstraintVerifier is only supported for ConstraintProviders";
```
=>
```
final String noConstraintProviderErrorMsg = "Cannot create a " + ConstraintVerifier.class.getSimpleName() + " because there is no " + ConstraintProvider.class.getSimpleName() + " configured in the " + SolverConfig.class.getSimpleName() + ".";
```

triceo:
Jenkins run native.

Christopher-Chianelli (approved):
LGTM

triceo:
Jenkins run native.

PR #1237, opened 2021-03-25
PLANNER-2249 Increase test timeout to prevent flakiness

michaltomco:


optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseTest.java
```diff
@@ -166,9 +165,8 @@ public void solvingStarted(SolverScope<TestdataSolution> solverScope) {
         assertThat(solutionFuture.get()).isNotNull();
     }
 
-    @Disabled("PLANNER-2249")
     @Test
-    @Timeout(5)
+    @Timeout(15)
```
triceo, line 169: 15 of what? I believe the annotation now has a way of specifying the unit.

optaplanner-examples/src/test/java/org/optaplanner/examples/cloudbalancing/app/CloudBalancingDaemonTest.java
```diff
@@ -51,9 +50,8 @@
     private volatile Throwable solverThreadException = null;
     private volatile CloudBalance currentBestSolution = null;
 
-    @Disabled("PLANNER-2249")
     @Test
-    @Timeout(600)
+    @Timeout(1500)
```
triceo, line 54: 15 above, 1500 here? That is quite the range!
Also, 1500 seconds, seriously? This test needs to become a turtle test and not be run within the default test suite.

triceo (changes requested):
The timeouts proposed are excessive.

dupliaka:
I guess @yurloc added @Timeout based on the performance test results right? 
@michaltomco could you tell if we have the date when it first became unstable?
@triceo could you give as some ideas what could potentially distabilize thouse tests so the timeout needed to successfully finish the build become twice as big?

optaplanner-core/src/test/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhaseTest.java
yurloc: Timeout 15 seconds doesn't mean the test runs for 15 seconds. I'm not sure if the timeouts are necessary at all but I think that if there is a timeout it should be orders of magnitude higher than the regular observed test execution time. For example, if the test is observed to run for 100ms on development machine I wouldn't hesitate to give it a 60 seconds timeout. Why? Because:
1. Under normal conditions it doesn't matter. The test will run for 100ms.
2. If you see a 100ms test hang for 60s it's a clear evidence that there's a race condition in the tested code or in the test itself causing a deadlock. If this happens, timeout increase is out of question. The test needs to be investigated and fixed (or removed if the bug is in the test and there is no reasonable way to make the test stable).
3. Some might argue a 60 seconds timeout is still better than nothing because it avoids the CI job to hang in the broken test until it's aborted. With the timeout, you'll see the test failure in the test report and at the same time, everything after the test will still execute normally.

yurloc: Originally, a timeout of 5 seconds was added to the `shutdownMainThreadAbruptly()` test: https://github.com/kiegroup/optaplanner/pull/340/files and the test appeared to be stable for several years.

Judging by the fact that this test wasn't `@Disabled`, it's still stable and the timeout increase doesn't hurt but what's the point @michaltomco?

dupliaka: It has been disabled 11/11/20. Since this PR suppose to stabilize the test I assume it was successfully launched. In that case, I suppose we do not need to stabilize the test, but create a ticket

ge0ffrey: > Timeout 15 seconds doesn't mean the test runs for 15 seconds. I'm not sure if the timeouts are necessary at all but I think that if there is a timeout it should be orders of magnitude higher than the regular observed test execution time. For example, if the test is observed to run for 100ms on development machine I wouldn't hesitate to give it a 60 seconds timeout. Why? Because:
> 
>     1. Under normal conditions it doesn't matter. The test will run for 100ms.
> 
>     2. If you see a 100ms test hang for 60s it's a clear evidence that there's a race condition in the tested code or in the test itself causing a deadlock. If this happens, timeout increase is out of question. The test needs to be investigated and fixed (or removed if the bug is in the test and there is no reasonable way to make the test stable).
> 
>     3. Some might argue a 60 seconds timeout is still better than nothing because it avoids the CI job to hang in the broken test until it's aborted. With the timeout, you'll see the test failure in the test report and at the same time, everything after the test will still execute normally.

Very clear explanation of how we do it. We have very fast tests (milliseconds), but very high timeouts (often 10 minutes), to avoid infinite builds and to avoid random failures on slow machines.

optaplanner-examples/src/test/java/org/optaplanner/examples/cloudbalancing/app/CloudBalancingDaemonTest.java
ge0ffrey: Most timeouts for our optpalanner-examples tests (not the turtle tests) are 600 seconds, yet they run for 5 seconds on most machines. That's intentional, because we sometimes saw them randomly break if was only 60 seconds because jenkins slaves had unreliable cpu.

triceo:
I see the points made in the comments here. That said, I do not agree.

If you set a timeout to `1500 seconds`, you are saying "I am OK with this test taking 25 minutes".
I am not OK with any test taking 25 minutes. 

I believe that if a test normally runs for 30 seconds but needs a 5000 % safety margin, the test is not worth its weight.
If you wish, override my NAK, @ge0ffrey, but I am unlikely to change my view on this.

michaltomco:
Thank you everyone for your discussion. I'll try to run the tests a bunch of times with different numbers there and reduce the timeouts to the most acceptable number. I've done it a while ago. I would like to double-check the results and pick the best number with maybe having a compromise between the two opinions presented.

triceo:
Closing as it's been open for more than 3 months without action..

PR #1239, opened 2021-03-25
DO NOT MERGE: Experimental GIZMO_MEMBER DomainAccessType

Christopher-Chianelli:
I added GIZMO_MEMBER access type for testing purposes, which will use getters/setters
if the field is not public. This change allow GIZMO to be used on all optaplanner-examples.

As for why GIZMO does not defer to getters/setters currently: users may have
complex getters/setters (validity checks, null checks, additional calculations, etc.)
which would make cloning/accessing fail in odd ways if used, and we rather fail fast and hard
than silently fail in an undetectable way. However, a large number of examples have
simple public getters/setters but have private fields (meaning GIZMO won't work on them
unless in Quarkus, but would work anywhere with GIZMO_MEMBER).

All optaplanner-examples where changed to use GIZMO_MEMBER domain access
type to determine cases where GIZMO fails.

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/build/optaplanner-build-parent/pom.xml
+++ b/build/optaplanner-build-parent/pom.xml
@@ -913,6 +913,11 @@
               <failOnWarning>true</failOnWarning>
               <ignoreNonCompile>true</ignoreNonCompile>
               <verbose>true</verbose>
+              <ignoredDependencies>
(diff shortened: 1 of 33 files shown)


Christopher-Chianelli:
Closing until this is worked on during a sprint; will keep the branch locally

PR #1241, opened 2021-03-29
PLANNER-2377 Depend on Drools directly where possible

triceo:


triceo:
Jenkins run native.
Jenkins run FDB.

triceo:
Jenkins run native.

triceo:
Jenkins run native

optaplanner-quarkus-integration/optaplanner-quarkus/runtime/pom.xml
```diff
@@ -30,15 +30,55 @@
...
+      <artifactId>drools-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>drools-compiler</artifactId>
+    </dependency>
+    <dependency> <!-- Static replaces dynamic. -->
       <groupId>org.kie.kogito</groupId>
       <artifactId>drools-core-static</artifactId>
     </dependency>
```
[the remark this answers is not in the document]
triceo: I can see what happens if I remove it. I suspect the dep plugin will complain, but let's do the experiment.

triceo:
Jenkins run native. Don't fail me, man.

rsynek (approved):
LGTM, thanks!

triceo:
Jenkins run native. One last time, please.

optaplanner-core/pom.xml
```diff
@@ -134,24 +136,6 @@
 
   <build>
     <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-dependency-plugin</artifactId>
```
yurloc, line 139: Why do you remove this?

triceo: Because the overrides are no longer necessary. The build now passes without it.
The dep plugin still runs, because that is defined elsewhere. If that is not the case then, obviously, that is wrong.

yurloc: You're right. Thanks for catching this. It's always nice to see a piece of code to go away because it turns out to be doing nothing :slightly_smiling_face:

PR #1244, opened 2021-03-30
PLANNER-2368 Unify LOGGER naming

triceo:
I have applied the following criteria:

- Abstract classes keep their `protected final transient Logger logger`. Messing with that could alter the performance characteristics of those classes, and that would not be worth it.
- Classes that do not expose their loggers all get converted to `private static final Logger LOGGER`.
- In some cases, this uncovered unused loggers, which were therefore removed.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/DefaultPlannerBenchmark.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/DefaultPlannerBenchmark.java
@@ -49,8 +49,8 @@ import org.slf4j.LoggerFactory;
 
 public class DefaultPlannerBenchmark implements PlannerBenchmark {
 
-    private static final transient Logger LOGGER = LoggerFactory.getLogger(DefaultPlannerBenchmark.class);
-    private static final transient Logger SINGLE_BENCHMARK_RUNNER_EXCEPTION_LOGGER =
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPlannerBenchmark.class);
+    private static final Logger SINGLE_BENCHMARK_RUNNER_EXCEPTION_LOGGER =
             LoggerFactory.getLogger(DefaultPlannerBenchmark.class + ".singleBenchmarkRunnerException");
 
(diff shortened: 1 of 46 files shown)


rsynek (approved):
LGTM

yurloc:
I disagree with this change. As far as I know, upper case is used for constants only. Note that not every static final field is a constant (and a constant doesn't necessarily have to be static).

https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names
> Constants are static final fields whose contents are deeply immutable and whose methods have no detectable side effects. This includes primitives, Strings, immutable types, and immutable collections of immutable types. If any of the instance's observable state can change, it is not a constant.

In my opinion, logger is not a constant because we don't have control over the type returned by `LoggerFactory`. It _might_ have an internal state that changes with calls of its methods. We don't know that, and we don't even need to know that because a logger doesn't play the role of a constant.

Also, I would expect a wider consensus before a change to the existing code style is accepted.

triceo:
In my understanding, a "constant" with respect to code style has never been about what the field does, or if it is deeply mutable.
A "constant" is a shorthand for "constant reference", that is a reference which can never change. The `static` keyword then adds the notion of the field being shared with everyone.

Technically, `private final` are also constant references. But the convention says that instance fields use camelCase. We only ever use the uppercasing for static constant references. One could ask why is there even such a convention to change code style for certain constructs. In my opinion, it makes zero sense to alter code style based on if something is `final`. But it makes *a lot of sense* to highlight that something is `static final`, because that means immortality and shared state, and people should be warned of that at first sight.

All in all, I see nothing wrong here and I consider this in full compliance with the convention, whereas before it was not.

ge0ffrey:
@triceo @yurloc According to the code style we follow, "final static" fields (AKA constants) must be ALL_CAPS and others must not be camelCase.

The real question is: how many times are we using a "transient non-static final" field for a logger (_creating the logger every time an instance of that class is being created_) when we should just use a static logger? Well, in cases that the logger is inherited to a subclass, we probably the transient one...

=> "static final" loggers should be private, always. We should prefer using those, always. Wdyt?

PR #1252, opened 2021-04-06
PLANNER-2391 CS-D: Don't go through constraint session

triceo:
For CS-D, the indirection of SDF -> `ConstraintSession` -> `KieSession` is needless waste of CPU cycles.

For CS-D, the chain becomes SDF -> `KieSession`.
For CS-B, the chain becomes SDF -> `BavetConstraintSession`.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/ConstraintSessionFactory.java
```diff
@@ -28,6 +28,6 @@
      * @param workingSolution if null, uniform synthetic constraint weights will be applied
      * @return never null
      */
-    ConstraintSession<Solution_, Score_> buildSession(boolean constraintMatchEnabled, Solution_ workingSolution);
+    Object buildSession(boolean constraintMatchEnabled, Solution_ workingSolution);
```
triceo, line 31: `KieSession`/`BavetConstraintSession` have nothing in common.
Handling of this happens in `*ConstraintStreamScoreDirectorFactory` where it becomes fully type-safe.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -37,12 +37,13 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  * @see ScoreDirector
  */
-public class ConstraintStreamScoreDirector<Solution_, Score_ extends Score<Score_>>
-        extends AbstractScoreDirector<Solution_, Score_, AbstractConstraintStreamScoreDirectorFactory<Solution_, Score_>> {
+public final class BavetConstraintStreamScoreDirector<Solution_, Score_ extends Score<Score_>>
+        extends AbstractScoreDirector<Solution_, Score_, BavetConstraintStreamScoreDirectorFactory<Solution_, Score_>> {
 
-    protected ConstraintSession<Solution_, Score_> session;
+    protected BavetConstraintSession<Solution_, Score_> session;
```
ge0ffrey, line 43: Can't we roll the BavetConstraintSession also into BavetConstraintStreamScoreDirector?

ge0ffrey: Ok, maybe for a later PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/AbstractConstraintStreamScoreDirectorFactory.java
```diff
@@ -65,21 +63,6 @@ protected AbstractConstraintStreamScoreDirectorFactory(SolutionDescriptor<Soluti
                 (ConstraintSessionFactory<Solution_, Score_>) constraintFactory.buildSessionFactory(constraints);
```
ge0ffrey, line 63: This line should now happen in the drools/bavet subclasses. Those should also no longer doe the fancy supplier trick, just the validation of the constraints can go through a common method.

ge0ffrey (approved):
Approved if buildSessionFactory() doesn't return an Object

triceo:
I agree with the comments, they will be addressed in a follow-up PR as agreed.

PR #1253, opened 2021-04-06
Java 16 deprecations

sudolea:
Modified all occurrences of (deprecated) calls to constructors of wrapper classes into their static valueOf() equivalent

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/lookup/LookUpStrategyImmutableTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/lookup/LookUpStrategyImmutableTest.java
@@ -52,16 +52,16 @@ public class LookUpStrategyImmutableTest {
 
     public static Stream<Arguments> data() {
         return Stream.of(
-                arguments(true, new Boolean(true)),
-                arguments((byte) 1, new Byte((byte) 1)),
-                arguments((short) 1, new Short((short) 1)),
-                arguments(1, new Integer(1)),
-                arguments(1L, new Long(1)),
-                arguments(0.5f, new Float(0.5f)),
-                arguments(0.1d, new Double(0.1d)),
+                arguments(true, Boolean.valueOf(true)),
+                arguments((byte) 1, Byte.valueOf((byte) 1)),
+                arguments((short) 1, Short.valueOf((short) 1)),
+                arguments(1, Integer.valueOf(1)),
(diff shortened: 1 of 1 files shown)


rsynek (approved):
Thank you for this PR, @sudolea. Looks good.

To your question on warning about unnecessary boxing: I think you're right and the test can be simplified. If you want to look into it too, please let me know. Otherwise, the PR is good as it is, as it already brings an improvement by removing the deprecated constructors.

sudolea:
@rsynek : it's not that I didn't want to look at simplifying further, but I could be helped by some clarification.

Because I had asked myself indeed that the former code (the equivalent of which is now complaining about "unnecessary" boxing) might have had it's (test purpose ?) reason. Failing to see that reason myself withheld me from modifying that code too, though.

I _suspect_ there had been a reason, and I would only modify the rest of the code if I have the "go" from someone who confirms it's OK to remove that boxing.

triceo:
OK to test

rsynek:
@sudolea I think the test can be simplified as it focuses on the `ImmutableLookUpStrategy`, which assumes every object it handles is immutable. Under such assumption, its lookup methods just return the very same instance you provide as a parameter.

For this reason, I believe the test does not need to provide pairs of independent (but equal) objects of supported immutable types. It may just create a single instance of each type and check through the `LookupManager` the same instance is retrieved every time.

There is one more related simplification, although out of scope of improving this test. I think we should not need more than a single instance of the `ImmutableLookUpStrategy` here https://github.com/kiegroup/optaplanner/blob/master/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/lookup/LookUpStrategyResolver.java#L59-L84.

rsynek:
@sudolea Please let me know if you want me to merge this PR first. If you want to discuss the other refactoring further, we can solve it over the Zulip chat. Thanks!

sudolea:
@rsynek I should first have a look at how I should proceed, as this is my first PR (i.e. what to do to update my fork, to then update my local code, etc). I could look at it further this weekend. But then again, I fear this will then be holding back this pull request ? I want to leave the decision to checkin this pull request up to you (as I don't know your team's habits). Iow if it is (holding back), then feel free to checkin this pull request, I would then do the rest at a later time (soon). Tomorrow Friday I will barely be able to continue on it, unfortunately.

rsynek:
No problem @sudolea. I am merging the PR as it fixes the original problem.

PR #1257, opened 2021-04-08
PLANNER-2392 Convert CS-D to use ScoreInliner

triceo:
Both CS-D and CS-B now use ScoreInliner.
ScoreInliner gets support for justifications.
Both CS-D and CS-B now use justifications through ScoreInliner.

This leads to a major unification on CS implementations.
Benchmarks are currently running.


optaplanner-core/src/main/java/org/optaplanner/core/api/score/holder/ScoreHolder.java
```diff
@@ -21,7 +21,7 @@
 import org.optaplanner.core.api.score.Score;
 
 /**
- * This is the base interface for all score holder implementations, used exclusively by Drools-based score calculators.
+ * This is the base interface for all score holder implementations.
```
triceo, line 24: All changes you see to `ScoreHolder` in this PR are simplifications now that CS-D is not using it anymore.
(Specifically, the removal of `Object... justifications` method overloads. That was *only used by CS-D*.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreInliner.java
```diff
@@ -55,31 +53,27 @@ public IntWeightedScoreImpacter buildWeightedScoreImpacter(BendableScore constra
...
                     this.hardScores[level] += hardImpact;
-                    if (constraintMatchEnabled) {
-                        matchScoreConsumer
-                                .accept(BendableScore.ofHard(hardScores.length, softScores.length, level, hardImpact));
-                    }
-                    return () -> this.hardScores[level] -= hardImpact;
+                    return buildUndo(constraintPackage, constraintName, constraintWeight,
+                            () -> this.hardScores[level] -= hardImpact,
+                            () -> BendableScore.ofHard(hardScores.length, softScores.length, level, hardImpact),
+                            justifications);
```
triceo, line 62: This is a pattern that will repeat in all `ScoreInliner`s.
All of the code duplication is moved to a new method `buildUndo()`, which deals with constraint matches when they're necessary.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreInliner.java
```diff
@@ -55,31 +53,27 @@ public IntWeightedScoreImpacter buildWeightedScoreImpacter(BendableScore constra
             int levelWeight = constraintWeight.getHardOrSoftScore(singleLevel);
             if (singleLevel < constraintWeight.getHardLevelsSize()) {
                 int level = singleLevel;
-                return (int matchWeight, Consumer<Score<?>> matchScoreConsumer) -> {
+                return (int matchWeight, Supplier<List<Object>> justifications) -> {
```
triceo, line 56: Justification list is a `Supplier` - this way, the collection will only ever be initialized (with the perf impact) when we need justifications. Therefore, when constraint match is disabled, justifications have literally zero impact.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreInliner.java
```diff
@@ -17,30 +17,28 @@
...
 
     @Override
-    public IntWeightedScoreImpacter buildWeightedScoreImpacter(BendableScore constraintWeight) {
-        if (constraintWeight.equals(BendableScore.zero(hardScores.length, softScores.length))) {
-            throw new IllegalArgumentException("The constraintWeight (" + constraintWeight + ") cannot be zero,"
-                    + " this constraint should have been culled during node creation.");
-        }
+    public IntWeightedScoreImpacter buildWeightedScoreImpacter(String constraintPackage, String constraintName,
+            BendableScore constraintWeight) {
+        ensureNonZeroConstraintWeight(constraintWeight);
```
ge0ffrey, line 41: Naming consistency: if that ensureNon... method throws an exception if it's not the case, it should be called assertNon... instead.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardsoft/HardSoftScoreInliner.java
```diff
@@ -16,61 +16,59 @@
...
-                    + " this constraint should have been culled during node creation.");
-        }
+    public IntWeightedScoreImpacter buildWeightedScoreImpacter(String constraintPackage, String constraintName,
+            HardSoftScore constraintWeight) {
+        ensureNonZeroConstraintWeight(constraintWeight);
         int hardConstraintWeight = constraintWeight.getHardScore();
         int softConstraintWeight = constraintWeight.getSoftScore();
         if (softConstraintWeight == 0) {
-            return (int matchWeight, Consumer<Score<?>> matchScoreConsumer) -> {
+            return (int matchWeight, Supplier<List<Object>> justifications) -> {
```
ge0ffrey, line 42: Naming: justifications isn't an array or Collection of justification objects. It's a supplier that returns that list.
=> Rename justifications to justificationsSupplier (across entire PR)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
```diff
@@ -16,18 +16,110 @@
...
+
+    public Map<Object, Indictment<Score_>> getIndictmentMap() {
+        // Unchecked assignment necessary as Indictment and DefaultIndictment incompatible in the Map generics.
+        return (Map) indictmentMap;
+    }
+
+    protected void ensureNonZeroConstraintWeight(Score_ constraintWeight) {
+        if (constraintWeight.equals(zeroScore)) {
+            throw new IllegalArgumentException("The constraintWeight (" + constraintWeight + ") cannot be zero,"
+                    + " this constraint should have been culled during node creation.");
```
ge0ffrey, line 93: I presume this is an "impossible state" (AKA, the user should never see this no matter how much the user screws up - or put differently: it's definitely our fault), so let's have the error message start with "Impossible state: "
(see other exceptions in optaplanner codebase that start with "Impossible state:")

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
```diff
@@ -16,18 +16,110 @@
...
+     *
+     * @param constraintPackage never null
+     * @param constraintName never null
+     * @param constraintWeight never null
+     * @param undoWithoutConstraintMatch never null
+     * @param score never null, lazy creation ensures instance only created if constraint matching enabled
+     * @param justifications never null, lazy creation ensures collection only created if constraint matching enabled
+     * @return never null
+     */
+    protected UndoScoreImpacter buildUndo(String constraintPackage, String constraintName, Score_ constraintWeight,
```
ge0ffrey, line 111: Let's make this method final to give the JVM a hint it can be inlined

triceo: I'll remove it altogether.

This code is on the hot path, and even creating the supplier lambdas is overhead that doesn't need to be there when constraint matching is disabled. There will be a bit more code duplication, but the performance of this is really important.

ge0ffrey (approved):
This is simply brilliant. I absolutely love the way it gets rid of the ugly Consumer and setMatchScore hack, how it simplified the code, reduces duplication and reduces the responsibility of the ScoringNodes. It's brilliant.

Do look at suggestions (mostly about naming)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
ge0ffrey: +1 I agree

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractRuleContext.java
```diff
@@ -88,35 +94,108 @@ protected AbstractRuleContext(ViewItem<?>... viewItems) {
...
-        constraint.assertCorrectImpact(impact);
-        scoreHolder.impactScore((RuleContext) drools, impact, justifications);
+    protected static IntImpactExecutor buildIntImpactExecutor(WeightedScoreImpacter scoreImpacter) {
+        if (scoreImpacter instanceof IntWeightedScoreImpacter) {
+            return ((IntWeightedScoreImpacter) scoreImpacter)::impactScore;
+        } else if (scoreImpacter instanceof LongWeightedScoreImpacter) {
+            return ((LongWeightedScoreImpacter) scoreImpacter)::impactScore;
+        } else if (scoreImpacter instanceof BigDecimalWeightedScoreImpacter) {
+            return (impact, justificationsSupplier) -> ((BigDecimalWeightedScoreImpacter) scoreImpacter)
+                    .impactScore(BigDecimal.valueOf(impact), justificationsSupplier);
```
rsynek, line 128: Why do we care about the `Long` and the `BigDecimal` here?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
```diff
@@ -103,41 +92,98 @@ public boolean isDroolsAlphaNetworkCompilationEnabled() {
...
-                }
-                model.addRule(modelRule);
-            });
-            currentKieBase = buildKieBaseFromModel(model, droolsAlphaNetworkCompilationEnabled);
-            currentlyDisabledConstraintIdSet = disabledConstraintIdSet;
+
+        // Extract constraint weights, excluding constraints where weight is zero.
+        Map<DroolsConstraint<Solution_>, Score_> constraintWeightMap = constraintList.stream()
+                .map(constraint -> {
+                    Object weight = constraint.extractConstraintWeight(workingSolution); // Expensive, only do once.
```
rsynek, line 101: Can this ever return null? I don't see the usual "@return never null" javadoc on the `extractConstraintWeight`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
```diff
@@ -103,41 +92,98 @@ public boolean isDroolsAlphaNetworkCompilationEnabled() {
...
+                    return new Object[] { constraint, weight };
+                })
+                .filter(constraintAndWeight -> !constraintAndWeight[1].equals(zeroScore)) // Exclude zero-weighted.
+                .collect(Collectors.toMap(c -> (DroolsConstraint<Solution_>) c[0], c -> (Score_) c[1]));
```
rsynek, line 105: Naming suggestion: let's keep the `constraintAndWeight` from the previous line. It's longer than `c` but tells what you should expect, which I think is important, as the type is `Object []`.

rsynek (approved):
A couple of naming suggestions and questions, nice work!

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
triceo: Deep down, a validator will throw if the weight is null.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractRuleContext.java
triceo: This is a mess. To explain, consider the following situations:

    penalize(..., (int) 1)
    penalizeLong(..., (int) 1)
    penalizeLong(..., (long) 1)
    penalizeBigDecimal(..., (int) 1)
    penalizeBigDecimal(..., (long) 1)
    penalizeBigDecimal(..., BigDecimal.ONE)

All of these are possible, and all of these require different handling, so that `int` is never auto-boxed into `Integer` etc.

triceo:
Merged without Jenkins, which is FUBAR.
Github Actions was green and I have seen Sonar earlier today, with an earlier iteration of this PR.

PR #812, opened 2020-06-23
Use ScoreVerifier in Spring Boot and Quarkus (wait for 8.x - step 4 - needs changes)

ge0ffrey:
Failing test due to PLANNER-2038
Failing build that is fixed when this is merged https://github.com/kiegroup/optaplanner/pull/811

This proves some of the issues that @yurloc brought up last week.

PR #1262, opened 2021-04-09
add ConstraintVerifier.create(SolverConfig)

ge0ffrey:
* Adds ConstraintVerifier.create(SolverConfig) for convenience
* Salvages this stale PR https://github.com/kiegroup/optaplanner/pull/812/files (which I 'll now close)

### DIFF
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
@@ -221,6 +221,7 @@ class OptaPlannerProcessor {
             final Class<?> constraintProviderClass = solverConfig.getScoreDirectorFactoryConfig().getConstraintProviderClass();
             final Class<?> planningSolutionClass = solverConfig.getSolutionClass();
             final List<Class<?>> planningEntityClasses = solverConfig.getEntityClassList();
+            // TODO Don't duplicate defaults by using ConstraintVerifier.create(solverConfig) instead
             final ConstraintStreamImplType constraintStreamImplType =
                     ObjectUtils.defaultIfNull(solverConfig.getScoreDirectorFactoryConfig().getConstraintStreamImplType(),
                             ConstraintStreamImplType.DROOLS);
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
@@ -267,7 +267,7 @@ class OptaPlannerProcessor {
                                 MethodDescriptor.ofMethod(constraintVerifierClassName,
                                         "withDroolsAlphaNetworkCompilationEnabled",
                                         constraintVerifierClassName,
-                                        boolean.class),
+                                        Boolean.class),
                                 constraintVerifierResultHandle,
                                 methodCreator.load(droolsAlphaNetworkCompilationEnabled));
                         methodCreator.returnValue(constraintVerifierResultHandle);
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
@@ -267,7 +267,7 @@ class OptaPlannerProcessor {
                                 MethodDescriptor.ofMethod(constraintVerifierClassName,
                                         "withDroolsAlphaNetworkCompilationEnabled",
                                         constraintVerifierClassName,
-                                        Boolean.class),
+                                        boolean.class),
                                 constraintVerifierResultHandle,
                                 methodCreator.load(droolsAlphaNetworkCompilationEnabled));
                         methodCreator.returnValue(constraintVerifierResultHandle);
--- a/optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
+++ b/optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
@@ -148,7 +148,7 @@ public class OptaPlannerAutoConfiguration implements BeanClassLoaderAware {
                 // (Cannot create custom condition that checks SolverConfig, since that
                 //  requires OptaPlannerAutoConfiguration to have a no-args constructor)
                 final String noConstraintProviderErrorMsg =
-                        "Cannot provision a ConstraintVerifier because there is ConstraintProvider class.";
+                        "Cannot provision a ConstraintVerifier because there is no ConstraintProvider class.";
                 return new ConstraintVerifier<ConstraintProvider_, SolutionClass_>() {
                     @Override
                     public ConstraintVerifier<ConstraintProvider_, SolutionClass_>
--- a/optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
+++ b/optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
@@ -39,7 +39,6 @@ import org.optaplanner.core.config.score.director.ScoreDirectorFactoryConfig;
 import org.optaplanner.core.config.solver.SolverConfig;
 import org.optaplanner.core.config.solver.SolverManagerConfig;
 import org.optaplanner.core.config.solver.termination.TerminationConfig;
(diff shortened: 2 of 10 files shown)


optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -217,6 +217,7 @@ private void generateConstraintVerifier(SolverConfig solverConfig,
             final Class<?> constraintProviderClass = solverConfig.getScoreDirectorFactoryConfig().getConstraintProviderClass();
             final Class<?> planningSolutionClass = solverConfig.getSolutionClass();
             final List<Class<?>> planningEntityClasses = solverConfig.getEntityClassList();
+            // TODO Don't duplicate defaults by using ConstraintVerifier.create(solverConfig) instead
```
ge0ffrey, line 220: I tried to also use ConstraintVerifier.create(solverConfig) to reduce the amount of code here significantly (including no longer duplicating the default settings (DROOLS etc), but I didn't see how to transfer the solverConfig instance into the generated code.

@Christopher-Chianelli Might you have an idea?

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfigurationTest.java
```diff
@@ -339,7 +336,7 @@ public void constraintVerifierOnDrl() {
                     assertThatCode(() -> {
                         context.getBean(ConstraintVerifier.class).verifyThat();
                     })
-                            .hasMessage("ConstraintVerifier is only supported for ConstraintProviders");
+                            .hasMessage("Cannot provision a ConstraintVerifier because there is ConstraintProvider class.");
```
triceo, line 339: Does this make sense? Shouldn't it be `because there is no`?

optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/ConstraintVerifier.java
```diff
@@ -50,6 +53,34 @@
...
+     *
+     * @param solverConfig never null, must have a {@link PlanningSolution} class, {@link PlanningEntity} classes
+     *        and a {@link ConstraintProvider} configured.
+     * @param <ConstraintProvider_> type of the {@link ConstraintProvider}
+     * @param <Solution_> type of the {@link PlanningSolution}-annotated class
+     * @return never null
+     */
+    static <ConstraintProvider_ extends ConstraintProvider, Solution_> ConstraintVerifier<ConstraintProvider_, Solution_>
+            create(
+                    SolverConfig solverConfig) {
```
triceo, line 68:
```suggestion
            create(SolverConfig solverConfig) {
```

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultConstraintVerifier.java
```diff
@@ -37,8 +39,8 @@
 
     private final ConstraintProvider_ constraintProvider;
     private final SolutionDescriptor<Solution_> solutionDescriptor;
-    private ConstraintStreamImplType constraintStreamImplType = ConstraintStreamImplType.DROOLS;
-    private boolean droolsAlphaNetworkCompilationEnabled = !CoreComponentsBuilder.isNativeImage();
+    private ConstraintStreamImplType constraintStreamImplType;
+    private Boolean droolsAlphaNetworkCompilationEnabled;
```
ge0ffrey, line 43: Because ConstraintVerifier has with methods (withDroolsAlphaNetworkCompilationEnabled() etc), I strongly believe it must behave similar to SolverConfig etc: so by default all fields are null and we don't use any primitive types for fields.
The defaults values (DROOLS, !isNativeImage) only kick when it builds something.

This way, it's possible to do withConstraintStreamImplType(solverConfig.get...().getConstraintStreamImplType()) even if the solverConfig.get...().getConstraintStreamImplType() returns null.

That in turn, means the spring (and hopefully soon the quarkus) code doesn't have to hard code the defaults too.

optaplanner-test/src/main/java/org/optaplanner/test/api/score/stream/ConstraintVerifier.java
```diff
@@ -68,7 +99,7 @@
      * @return this
      */
     ConstraintVerifier<ConstraintProvider_, Solution_> withDroolsAlphaNetworkCompilationEnabled(
-            boolean droolsAlphaNetworkCompilationEnabled);
+            Boolean droolsAlphaNetworkCompilationEnabled);
```
triceo, line 102: I'd rather this not proliferate outside of configs. Nulls are evil.

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
```diff
@@ -180,19 +180,8 @@ public SolverConfig solverConfig() {
...
-            Class<?>[] entityClasses = solverConfig.getEntityClassList().toArray(new Class<?>[0]);
-            ConstraintStreamImplType constraintStreamImplType =
-                    solverConfig.getScoreDirectorFactoryConfig().getConstraintStreamImplType();
-            boolean droolsAlphaNetworkCompilationEnabled =
-                    solverConfig.getScoreDirectorFactoryConfig().isDroolsAlphaNetworkCompilationEnabled();
-
-            return (ConstraintVerifier<ConstraintProvider_, SolutionClass_>) ConstraintVerifier.build(constraintProvider,
-                    solverConfig.getSolutionClass(), entityClasses)
-                    .withConstraintStreamImplType(
-                            (constraintStreamImplType != null) ? constraintStreamImplType : ConstraintStreamImplType.DROOLS)
```
ge0ffrey, line 194: This mention of the default (DROOLS) in the spring starter is what I really wanted to get rid of, as it belongs in the ConstarintVerifier.

triceo (changes requested):
No on `Boolean`.
I understand it in configs, but outside of configs, I don't want to clutter the APIs with the concept of nullity.

optaplanner-test/src/main/java/org/optaplanner/test/impl/score/stream/DefaultConstraintVerifier.java
triceo: I disagree. And especially for boolean. Boolean with 3 values is not a boolean.
OK for configs, but no to proliferation outside of configs.

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
triceo: If you don't want to pick a default, don't call the method - default will be picked for you.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
Christopher-Chianelli: Maybe create a `OptaplannerTestBeanProvider` (needed as optaplanner-test is optional) which construct the constraint verifier (and would have access to the injected `SolverConfig` instance)?

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/test/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfigurationTest.java
Christopher-Chianelli: Agreed; it should be "because there is no ConstraintProvider class"

optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/src/main/java/org/optaplanner/spring/boot/autoconfigure/OptaPlannerAutoConfiguration.java
```diff
@@ -154,7 +153,8 @@ public SolverConfig solverConfig() {
                 // Return a mock ConstraintVerifier so not having ConstraintProvider doesn't crash tests
                 // (Cannot create custom condition that checks SolverConfig, since that
                 //  requires OptaPlannerAutoConfiguration to have a no-args constructor)
-                final String noConstraintProviderErrorMsg = "ConstraintVerifier is only supported for ConstraintProviders";
+                final String noConstraintProviderErrorMsg =
+                        "Cannot provision a ConstraintVerifier because there is ConstraintProvider class.";
```
Christopher-Chianelli, line 157: Should be "Cannot provision a ConstraintVerifier because there is no ConstraintProvider class."

Christopher-Chianelli (changes requested):
Exception message is wrong (should be "there is no ConstraintProvider class" instead of "there is a ConstraintProvider class"). Otherwise LGTM; to simplify injected constraint verifier, consider adding a new BeanProvider class.

ge0ffrey:
@triceo I am reviving htis PR, let's do the Boolean vs boolean discussion on Monday?

triceo (approved):
Good to go when green.

triceo:
Apps failure unrelated, merging.

PR #1265, opened 2021-04-13
PLANNER-2392 Convert CS-D to use ScoreInliner

triceo:
The biggest benefit of this PR is that CS-D and CS-B now share a lot of code, making them more similar and easier to maintain.

Performance impact measured was well within the margin of error.

I have specifically tested `kogito-apps` downstream to make sure that the errors of the previous iteration are fixed in this one.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
```diff
@@ -16,18 +16,122 @@
...
     public abstract Score_ extractScore(int initScore);
 
-    public abstract WeightedScoreImpacter buildWeightedScoreImpacter(Score_ constraintWeight);
+    /**
+     * Create a new instance of {@link WeightedScoreImpacter} for a particular constraint.
+     * 
+     * @param constraint never null
+     * @return never null
+     */
+    protected abstract Impacter_ buildWeightedScoreImpacter(Constraint constraint);
```
triceo, line 62: Major change: ScoreImpacter now extracts constraint weight from the constraint.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
```diff
@@ -16,18 +16,122 @@
...
+     * it returns a new {@link WeightedScoreImpacter} instance.
+     * All subsequent invocations for that particular {@link Constraint}
+     * return the original {@link WeightedScoreImpacter} instance.
+     *
+     * @param constraint never null
+     * @return never null
+     */
+    public Impacter_ buildOrGetWeightedScoreImpacter(Constraint constraint) {
+        return constraintToImpacterMap.computeIfAbsent(constraint, __ -> buildWeightedScoreImpacter(constraint));
+    }
```
triceo, line 85: Major change: ScoreInliner now keeps a cache of ScoreImpacters created from it.
This cache is necessary so that each rule can reuse its own ScoreImpacter.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/UniRuleContext.java
```diff
@@ -37,33 +39,36 @@ public UniRuleContext(Variable<A> variable, ViewItem<?>... viewItems) {
...
     public <Solution_> RuleBuilder<Solution_> newRuleBuilder(ToIntFunction<A> matchWeighter) {
         ConsequenceBuilder<Solution_> consequenceBuilder =
-                (constraint, scoreHolderGlobal) -> DSL.on(scoreHolderGlobal, variable)
-                        .execute((drools, scoreHolder, a) -> impactScore(constraint, drools, scoreHolder,
-                                matchWeighter.applyAsInt(a), a));
+                (constraint, scoreInlinerGlobal) -> DSL.on(scoreInlinerGlobal, variable)
+                        .execute((drools, scoreInliner, a) -> runConsequence(constraint, drools,
+                                buildIntImpactExecutor(scoreInliner.buildOrGetWeightedScoreImpacter(constraint)),
+                                matchWeighter.applyAsInt(a),
+                                () -> singletonList(a)));
```
triceo, line 46: Major change: scoreImpacter is now built and cached directly in the consequence.
This is necessary as we only ever know the constraint weight at rule runtime.

This is the part I don't like. It possibly slows the whole thing down.
I will have benchmark results to prove or disprove soon.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/UniRuleContext.java
```diff
@@ -37,33 +39,36 @@ public UniRuleContext(Variable<A> variable, ViewItem<?>... viewItems) {
...
     public <Solution_> RuleBuilder<Solution_> newRuleBuilder(ToIntFunction<A> matchWeighter) {
         ConsequenceBuilder<Solution_> consequenceBuilder =
-                (constraint, scoreHolderGlobal) -> DSL.on(scoreHolderGlobal, variable)
-                        .execute((drools, scoreHolder, a) -> impactScore(constraint, drools, scoreHolder,
-                                matchWeighter.applyAsInt(a), a));
+                (constraint, scoreInlinerGlobal) -> DSL.on(scoreInlinerGlobal, variable)
+                        .execute((drools, scoreInliner, a) -> runConsequence(constraint, drools,
+                                scoreInliner.buildOrGetWeightedScoreImpacter(constraint),
+                                matchWeighter.applyAsInt(a),
+                                () -> singletonList(a)));
```
triceo, line 46: Major change: scoreImpacter is now built and cached directly in the consequence.
This is necessary as we only ever know the constraint weight at rule runtime.

This is the part I don't like. It possibly slows the whole thing down.
I will have benchmark results to prove or disprove soon.

There is another possible optimization: never export the score impacter. Instead, we keep it inside the ScoreInliner, and we instead call `ScoreInliner.impactScore(Constraint, int impact)`. This way, we will be able to avoid some casting, possibly improving performance a bit. But still, it's not very nice.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/ScoreDirectorSemanticsTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
+import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
+
+public class ScoreDirectorSemanticsTest {
+
+    private final SolutionDescriptor<TestdataSolution> solutionDescriptor = TestdataSolution.buildSolutionDescriptor();
+
+    @EnumSource(ScoreDirectorType.class)
+    @ParameterizedTest
+    void independentScoreDirectors(ScoreDirectorType scoreDirectorType) {
```
triceo, line 37: One of the tests we agreed on.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/WeightedScoreImpacter.java
```diff
@@ -16,6 +16,46 @@
...
+ * <li>{@code .penalizeBigDecimal(..., (long) 1)}</li>
+ * <li>{@code .penalizeBigDecimal(..., BigDecimal.ONE)}</li>
+ * <li>Plus reward variants of the above.</li>
+ * </ul>
+ *
+ * Implementations of this class need to be able to handle all of these, casting to the appropriate type.
+ * In certain cases, such as when BigDecimal is passed into an int impacter, methods are allowed to throw exceptions.
+ * That state should be considered impossible, ruled out by the CS API itself.
+ */
 public interface WeightedScoreImpacter {
```
triceo, line 38: See Javadoc for why these changes are necessary.
(Otherwise there would have been a lot of casting in the consequence. See Bavet scoring nodes for how much this simplified things, at no measurable impact to performance.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
```diff
@@ -103,41 +98,93 @@ public boolean isDroolsAlphaNetworkCompilationEnabled() {
...
+            AtomicInteger idCounter = new AtomicInteger(0);
+            Map<DroolsConstraint<Solution_>, Global<WeightedScoreImpacter>> constraintToGlobalMap =
+                    constraintToWeightMap.keySet().stream()
+                            .collect(Collectors.toMap(Function.identity(),
+                                    c -> globalOf(WeightedScoreImpacter.class, constraintPackageName,
+                                            "scoreImpacter" + idCounter.getAndIncrement())));
+            ModelImpl model = constraintToWeightMap.keySet().stream()
+                    .map(constraint -> constraint.buildRule(constraintToGlobalMap.get(constraint)))
+                    .reduce(new ModelImpl(), ModelImpl::addRule, (m, __) -> m);
+            constraintToGlobalMap.forEach((constraint, global) -> model.addGlobal(global));
```
triceo, line 125: Instead of 1 ScoreHolder global per session, we now have a ScoreImpacter global per constraint per session.
This is necessary to guarantee that constraint are independent across sessions, and that constraint weights are solution-specific.

This has no impact on performance.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetScoringBiConstraintStream.java
```diff
@@ -111,68 +108,37 @@ public boolean guaranteesDistinct() {
...
+            };
+        } else if (noMatchWeigher) {
+            scoreImpacter = (a, b) -> weightedScoreImpacter.impactScore(1, () -> asList(a, b));
         } else {
-            throw new IllegalStateException("Unsupported weightedScoreImpacter (" + weightedScoreImpacter + ").");
+            throw new IllegalStateException("The matchWeigher of " + BiConstraintStream.class.getSimpleName()
+                    + ".penalize(matchWeigher) of the constraint (" + constraint.getConstraintId()
+                    + ") must return " + weightedScoreImpacter.getExpectedMatchWeightType().getSimpleName() + ".\n" +
+                    "Maybe switch to another type-specific overload of penalize() or reward().");
         }
```
triceo, line 139: See the simplification I was able to make because of the changes to `WeightedScoreImpacter` I pointed out above.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/WeightedScoreImpacter.java
```diff
@@ -16,6 +16,80 @@
...
+ * <li>{@code .penalizeBigDecimal(..., (int) 1)}</li>
+ * <li>{@code .penalizeBigDecimal(..., (long) 1)}</li>
+ * <li>{@code .penalizeBigDecimal(..., BigDecimal.ONE)}</li>
+ * <li>Plus reward variants of the above.</li>
+ * </ul>
+ *
+ * Implementations of this class need to be able to handle all of these, casting to the appropriate type.
+ * In certain cases, such as when BigDecimal is passed into an int impacter, methods are allowed to throw exceptions.
+ * That state should be considered impossible, ruled out by the CS API itself.
+ */
```
triceo, line 38: See Javadoc for why these changes are necessary.
(Otherwise there would have been a lot of casting in the consequence. See Bavet scoring streams below for how much this simplified things, at no measurable impact to performance.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/WeightedScoreImpacter.java
```diff
@@ -16,6 +16,80 @@
...
 public interface WeightedScoreImpacter {
 
+    static WeightedScoreImpacter of(IntImpactFunction impactFunction) {
+        return new IntWeightedScoreImpacter(impactFunction);
+    }
```
triceo, line 43: I decided to hide `IntWeightedScoreImpacter` et. al., as they are implementation details.
`WeightedScoreImpacter` is now the only public type.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
```diff
@@ -16,18 +16,96 @@
...
     public abstract Score_ extractScore(int initScore);
 
-    public abstract WeightedScoreImpacter buildWeightedScoreImpacter(Score_ constraintWeight);
+    /**
+     * Create a new instance of {@link WeightedScoreImpacter} for a particular constraint.
+     * 
+     * @param constraint never null
+     * @return never null
+     */
+    public abstract WeightedScoreImpacter buildWeightedScoreImpacter(Constraint constraint);
```
triceo, line 59: ScoreImpacter now is constraint-specific and it is necessary to construct a new one each time the constraint weight changes.
Since yesterday (commit 0a086a2fdb11e732cc66bb427fe073934fbbec28), we now have tests to make sure that this actually happens.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/OptaPlannerRuleEventListener.java
```diff
@@ -19,25 +19,36 @@
...
+        Runnable callback = agendaItem.getCallback();
+        /*
+         * In DRL, it is possible that RHS would not call addConstraintMatch() and do some insertLogical() instead,
+         * and therefore the callback would be null.
+         * In CS-D, that is not possible, as all the rules we generate will create a match.
+         * But null callbacks can still happen.
+         *
+         * If we insert a fact and then immediately delete it without firing any rules inbetween,
+         * a dummy match will be created by Drools and that match will not have our callback in it.
+         * Although this is inefficient, it was decided that the cure would have been worse than the disease.
```
triceo, line 46: FYI this was the cause of the downstream problem in Walter's Kogito task assigning.
(Through a series of variable listeners, something was updated and then immediately retracted, without ever firing the rules.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/bi/BavetScoringBiConstraintStream.java
```diff
@@ -111,68 +107,34 @@ public boolean guaranteesDistinct() {
...
+            scoreImpacter = (a, b) -> {
+                BigDecimal matchWeight = bigDecimalMatchWeigher.apply(a, b);
+                constraint.assertCorrectImpact(matchWeight);
+                return weightedScoreImpacter.impactScore(matchWeight, () -> asList(a, b));
+            };
+        } else if (noMatchWeigher) {
+            scoreImpacter = (a, b) -> weightedScoreImpacter.impactScore(1, () -> asList(a, b));
         } else {
-            throw new IllegalStateException("Unsupported weightedScoreImpacter (" + weightedScoreImpacter + ").");
+            throw new IllegalStateException("Impossible state: neither of the supported match weighers provided.");
```
triceo, line 134: See the simplification I was able to make because of the changes to WeightedScoreImpacter I pointed out above.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
ge0ffrey: "we now have a ScoreImpacter global per constraint per session". What if there are 100 constraints? How - and when - is the score summed up?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreDefinition.java
```diff
@@ -103,8 +105,9 @@ public BendableScore createScoreUninitialized(int initScore, int... scores) {
     }
 
     @Override
-    public BendableScoreInliner buildScoreInliner(boolean constraintMatchEnabled) {
-        return new BendableScoreInliner(constraintMatchEnabled, hardLevelsSize, softLevelsSize);
+    public BendableScoreInliner buildScoreInliner(Map<Constraint, BendableScore> constraintToWeightMap,
+            boolean constraintMatchEnabled) {
```
ge0ffrey, line 109: This constraintToWeightMap looks fishy. I 'll need to grok it first.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardsoft/HardSoftScoreInliner.java
```diff
@@ -16,62 +16,79 @@
...
 
-    protected int hardScore;
-    protected int softScore;
+    private int hardScore;
+    private int softScore;
 
-    protected HardSoftScoreInliner(boolean constraintMatchEnabled) {
-        super(constraintMatchEnabled);
+    protected HardSoftScoreInliner(Map<Constraint, HardSoftScore> constraintIdToWeightMap, boolean constraintMatchEnabled) {
+        super(constraintIdToWeightMap, constraintMatchEnabled, HardSoftScore.ZERO);
```
ge0ffrey, line 34: This `Map<Constraint, HardSoftScore> constraintIdToWeightMap` is very similar to `ConstraintMatchTotal`

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/DroolsConstraintSessionFactory.java
triceo: The same way it always was - there is only one ScoreInliner.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/bendable/BendableScoreDefinition.java
triceo: I can get rid of it. See my answer to your score inliner comment.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/buildin/hardsoft/HardSoftScoreInliner.java
triceo: It does something different . But I can remove it.

Instead, you'd have to do `ScoreInliner.buildScoreImpacter(Solution, Constraint)`, and the weight would be inferred at that point.
They are the same thing, they have the same effect, but maybe the second approach is better, as it is more local.

ge0ffrey: Agreed that it's not CMT

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/WeightedScoreImpacter.java
```diff
@@ -16,6 +16,78 @@
...
+
+    @FunctionalInterface
+    interface LongImpactFunction {
+
+        UndoScoreImpacter impact(long matchWeight, JustificationsSupplier justificationsSupplier);
+
+    }
+
+    @FunctionalInterface
+    interface BigDecimalImpactFunction extends BiFunction<BigDecimal, JustificationsSupplier, UndoScoreImpacter> {
```
ge0ffrey, line 89: Its a bit clunky that the other interfaces use "impact" as a method name and this one uses "apply"

ge0ffrey: Java has ToIntBiFunction, but nothing that takes an <int,A> returns B.
So I'd just not inherit BiFunction here and just declare it similar to the others?

(soft suggestion)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/WeightedScoreImpacter.java
ge0ffrey: Doesn't this add an extra layer of indirection?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/WeightedScoreImpacter.java
```diff
@@ -16,6 +16,78 @@
...
+ * <li>{@code .penalizeLong(..., (long) 1)}</li>
+ * <li>{@code .penalizeBigDecimal(..., (int) 1)}</li>
+ * <li>{@code .penalizeBigDecimal(..., (long) 1)}</li>
+ * <li>{@code .penalizeBigDecimal(..., BigDecimal.ONE)}</li>
+ * <li>Plus reward variants of the above.</li>
+ * </ul>
+ *
+ * Implementations of this class need to be able to handle all of these, casting to the appropriate type.
+ * In certain cases, such as when BigDecimal is passed into an int impacter, methods are allowed to throw exceptions.
+ * That state should be considered impossible, ruled out by the CS API itself.
```
ge0ffrey, line 37: Wording: "should be considered", "need to be able"
what is "these" in "handle all of these"?

If the method need to "handle all of these" - but can also "throw exceptions", isn't that contradicting itself?

```suggestion
 * An implemtation of this inferface can throw an UnsupportedOperationException for the method types it doesn't support.
 * The CS API guarantees no types are mixed. For example, a BigDecimal parameter method won't be called on an instance build with an IntImpactFunction.
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/ScoreInliner.java
```diff
@@ -16,18 +16,96 @@
...
+
+    protected final Runnable addConstraintMatch(Constraint constraint, Score_ constraintWeight, Score_ score,
+            List<Object> justificationList) {
+        String constraintPackage = constraint.getConstraintPackage();
+        String constraintName = constraint.getConstraintName();
+        DefaultConstraintMatchTotal<Score_> constraintMatchTotal = constraintMatchTotalMap.computeIfAbsent(
+                constraint.getConstraintId(),
+                __ -> new DefaultConstraintMatchTotal<>(constraintPackage, constraintName, constraintWeight, zeroScore));
```
ge0ffrey, line 67: What's `__`? Other code typically uses "key" if we don't care about what it actually is. Same below

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/JustificationsSupplier.java
```diff
@@ -14,20 +14,19 @@
...
-     * @param workingSolution if null, uniform synthetic constraint weights will be applied
-     * @return never null
-     */
-    Object buildSession(boolean constraintMatchEnabled, Solution_ workingSolution);
+import java.util.List;
+import java.util.function.Supplier;
 
+/**
+ * This interface exists so that justifications can be created lazily
+ * and for that pattern to naturally spread throughout the codebase.
```
ge0ffrey, line 24: Wording too passive

```suggestion
 * This interface allows to create justifications lazily
 * if and only if constraint matches are enabled.
```

ge0ffrey: Ok, nevermind that missing part, as the next paragraph explains it already.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/JustificationsSupplier.java
```diff
@@ -14,20 +14,19 @@
...
+ *
+ * Justifications must be specifically enabled, and that is usually done outside of the hot path,
+ * as that code is expensive.
+ * Therefore constructing lists of justifications in case they are ever needed is a waste of CPU cycles.
```
ge0ffrey, line 28: To bring it in sync with the changes above:

```suggestion
 * Justifications creation is performance expensive and constraint matches are typically disabled.
 * So justifications are created lazily, outside of the typical hot path.
```

(all soft suggestions of course)

ge0ffrey (approved):
LGTM, just wondering if the Weighter.of(IntFunction) wrapping is necessary.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/inliner/IntWeightedScoreImpacter.java
```diff
@@ -16,19 +16,30 @@
...
-     * @return never null
-     */
-    UndoScoreImpacter impactScore(int matchWeight, Consumer<Score<?>> matchScoreConsumer);
+    public IntWeightedScoreImpacter(IntImpactFunction impactFunction) {
+        this.impactFunction = Objects.requireNonNull(impactFunction);
+    }
+
+    @Override
+    public UndoScoreImpacter impactScore(int matchWeight, JustificationsSupplier justificationsSupplier) {
+        return impactFunction.impact(matchWeight, justificationsSupplier);
```
ge0ffrey, line 32: Level of indirection

PR #1267, opened 2021-04-14
Standardize some score director semantics

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -460,12 +460,20 @@ public void beforeProblemPropertyChanged(Object problemFactOrEntity) {
...
     }
 
     @Override
     public void beforeProblemFactRemoved(Object problemFact) {
-        // Do nothing
+        if (isConstraintConfiguration(problemFact)) {
+            throw new IllegalStateException("Attempted to remove constraint configuration (" + problemFact +
+                    ") from solution (" + workingSolution + ").\n" +
+                    "Maybe use before/afterProblemPropertyChanged(...) instead.");
+        }
```
ge0ffrey, line 476: In theory you could remove a constraint configuration and then add in another one.
however, if we nuke the working memory in between, that's going to cause issues.
Ideally, we should then delay the nuking which is very complex.

Long story short: +1 to do it like this for now, maybe add a todo comment

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -776,6 +784,16 @@ protected String buildScoreCorruptionAnalysis(InnerScoreDirector<Solution_, Scor
         return constraintMatchMap;
     }
 
+    protected boolean isConstraintConfiguration(Object problemFactOrEntity) {
+        SolutionDescriptor<Solution_> solutionDescriptor = scoreDirectorFactory.getSolutionDescriptor();
+        MemberAccessor constraintConfigurationAccessor = solutionDescriptor.getConstraintConfigurationMemberAccessor();
+        if (constraintConfigurationAccessor == null) {
+            return false;
+        }
+        Object constraintConfiguration = constraintConfigurationAccessor.executeGetter(workingSolution);
```
ge0ffrey, line 793: Intresting. I would have just checked if it's an instance of that ConstraintConfiguration class. No idea which is better:

Proposal A) check executeGetter() returns same instance
Pro: they have other problem facts that are instances of that class.
Con: every single after/before call in addPFC (sometimes n per PFC) calls that executeGetter.

Proposal B) check instanceOf
This presumes that normally a dataset doesn't have other problem facts that implement that constraint configuration class. I 've never seen that happen so far.

Proposal C) Do A), but cache the constraintConfiguration instance on scoreDirector.
Con: more work/maintenance/complexity

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/bavet/common/BavetNodeBuildPolicy.java
```diff
@@ -77,6 +78,9 @@ public int nextNodeIndex() {
     }
 
     public List<BavetNode> getCreatedNodes() {
+        if (sharableNodeMap.isEmpty()) { // No constraints have their constraint weights greater than zero.
```
ge0ffrey, line 81:
```suggestion
        if (sharableNodeMap.isEmpty()) { // No constraints have a constraint weight greater than zero.
```

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/DummyConstraintProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.optaplanner.core.impl.score.director;
+
+import org.optaplanner.core.api.score.stream.Constraint;
+import org.optaplanner.core.api.score.stream.ConstraintFactory;
+import org.optaplanner.core.api.score.stream.ConstraintProvider;
+import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
+
+public final class DummyConstraintProvider implements ConstraintProvider {
```
ge0ffrey, line 24: There are already 2 DummyConstraintProvider:
- SolverConfigTest.DummyConstraintProvider
- And one in quarkus.

I think this new one should survive and replace the other two.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/DummyConstraintProvider.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
+        return new Constraint[] {
+                onlyConstraint(constraintFactory)
+        };
+    }
+
+    private Constraint onlyConstraint(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TestdataEntity.class)
+                .rewardConfigurable("First weight");
```
ge0ffrey, line 34: Ow wait, this isn't a dummy, it actually does something.
 This is a TestdataSolutionConstraintProvider as it only works for TestdataSolutio

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/DummyEasyScoreCalculator.java
```diff
@@ -0,0 +1,31 @@
...
+import org.optaplanner.core.api.score.calculator.EasyScoreCalculator;
+import org.optaplanner.core.impl.testdata.domain.constraintconfiguration.TestdataConstraintConfigurationSolution;
+
+public final class DummyEasyScoreCalculator
+        implements EasyScoreCalculator<TestdataConstraintConfigurationSolution, SimpleScore> {
+
+    @Override
+    public SimpleScore calculateScore(TestdataConstraintConfigurationSolution solution) {
+        SimpleScore constraintWeight = solution.getConstraintConfiguration().getFirstWeight();
+        return constraintWeight.multiply(solution.getEntityList().size());
```
ge0ffrey, line 29: Same story here. SolverconfigTest already has a dummy which really does nothing. I'd call it TestdataSolutionConstraintProvider.

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/DummyConstraintProvider.java
ge0ffrey: And it does something for a @ConstraintWeight specifically, so it's naming so mention that.

how about TestdataConstraintWeightConstraintProvider ? Or something better :)

optaplanner-core/src/test/java/org/optaplanner/core/impl/score/director/ScoreDirectorSemanticsTest.java
```diff
@@ -0,0 +1,177 @@
...
+        EASY(new ScoreDirectorFactoryConfig()
+                .withEasyScoreCalculatorClass(DummyEasyScoreCalculator.class)),
+        INCREMENTAL(new ScoreDirectorFactoryConfig()
+                .withIncrementalScoreCalculatorClass(DummyIncrementalScoreCalculator.class)),
+        CS_DROOLS(new ScoreDirectorFactoryConfig()
+                .withConstraintProviderClass(DummyConstraintProvider.class)
+                .withConstraintStreamImplType(ConstraintStreamImplType.DROOLS)),
+        CS_BAVET(new ScoreDirectorFactoryConfig()
+                .withConstraintProviderClass(DummyConstraintProvider.class)
+                .withConstraintStreamImplType(ConstraintStreamImplType.BAVET)),
```
ge0ffrey, line 149: Ordering: CS before INCREMENTAL, following the ordering of ScoreDirectorFactoryConfig's fields (which is enforced in the XSD ordering).

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/util/PlannerTestUtils.java
```diff
@@ -98,7 +98,7 @@
                 solverConfig.getScoreDirectorFactoryConfig();
         scoreDirectorFactoryConfig.setEasyScoreCalculatorClass(null);
         scoreDirectorFactoryConfig.setScoreDrlList(Collections.singletonList(
-                "org/optaplanner/core/impl/score/dummySimpleScoreDroolsConstraints.drl"));
+                "org/optaplanner/core/impl/score/director/dummyDroolsConstraints.drl"));
```
ge0ffrey, line 101: I'd argue that now that drl actually has a constraint, it's no longer a dummy. Not sure if we want to leak that constraint to all test cases that use this method already.

I'd argue that the test case that really needs that "First weight" constraint in there, should add it explicitly

optaplanner-core/src/test/resources/org/optaplanner/core/impl/score/director/dummyDroolsConstraints.drl
```diff
@@ -14,9 +14,17 @@
...
-package org.optaplanner.core.impl.score.director;
+package org.optaplanner.core.impl.testdata.domain.constraintconfiguration;
 dialect "java"
 
 import org.optaplanner.core.api.score.buildin.simple.SimpleScoreHolder;
+import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
 
 global SimpleScoreHolder scoreHolder;
+
+rule "First weight"
```
ge0ffrey, line 25: -1 on adding this to the existing dummy

ge0ffrey (commented):
A dummy should be empty, otherwise this looks good! Thank you :)

PR #1270, opened 2021-04-15
Delay SolutionCloner creation until EntityDescriptors are created 

ge0ffrey:
for assertModelForCloning


Christopher-Chianelli (approved):

triceo (approved):
I do not understand why the `EnvironmentMode` is always hard-coded to `REPRODUCIBLE`.
That said, assuming that there is a good reason, LGTM.

Christopher-Chianelli:
@triceo in this case, the solution descriptor was only being created to get the score definition (i.e. the SolutionDescriptor itself was not used).

PR #1271, opened 2021-04-15
Removed "unnecessary boxing" commit warnings

sudolea:
Adapted the code to avoid the compiler's complaint that there are "unnecessary boxing" occurrences.
Regression tests were run and all were successful.

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/lookup/LookUpStrategyImmutableTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/impl/domain/lookup/LookUpStrategyImmutableTest.java
@@ -52,17 +52,17 @@ public class LookUpStrategyImmutableTest {
 
     public static Stream<Arguments> data() {
         return Stream.of(
-                arguments(true, Boolean.valueOf(true)),
-                arguments((byte) 1, Byte.valueOf((byte) 1)),
-                arguments((short) 1, Short.valueOf((short) 1)),
-                arguments(1, Integer.valueOf(1)),
-                arguments(1L, Long.valueOf(1)),
-                arguments(0.5f, Float.valueOf(0.5f)),
-                arguments(0.1d, Double.valueOf(0.1d)),
+                arguments(true, true),
+                arguments((byte) 1, (byte) 1),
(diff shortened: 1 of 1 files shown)


sudolea:
Even if this code is ok according to the regression tests, I request someone else will be verifying this code, because :
1. I am unsure as of where this code (the data() method of class LookUpStrategyImmutableTest) is used : in IntelliJ, upon finding the usages of this code, it showed no location; however, I'm sure that code IS used (forcing an error really _does_ result into a regression test error. So I'm somehow puzzled, and thus uncertain of my setup (and of the proper execution of the regression tests)
2. I'm not entirely confident in understanding this code.
If someone would be wanting to enlighten me on any of these issues : I'll be watching Zulip (more specifically the Java deprecations thread, where this pull request is a consequence of)

triceo:
ok to test

rsynek:
@sudolea thanks for this PR. The "unnecessary boxing" is there most likely to make sure a distinct object is created each time. As a matter of optimization, Java reuses the same boxing objects for some frequently used primitive values, see [1] for more details.

This test focuses on `ImmutableLookupStrategy`, which should always return the same instance (only for the supported immutable types). Thus, I am thinking of simplifying the test in the following way:
- let's remove argument tuples and use just a single value for each argument
- in the `lookUpImmutable` test method, let's compare the objects for identity, not equality

@triceo am I missing something?

[1] https://www.baeldung.com/java-compare-long-values

triceo:
No, I think this should be enough.

That said, in my recent commits, I have discovered many more places where we construct new instances of `BigDecimal` etc. I have fixed many of them already, but the fact that they existed leads me to think that the original PR was not thorough enough.

rsynek:
Thanks @triceo. Slightly related to this, I don't think we need more than a single instance of the `ImmutableLookupStrategy` here:
https://github.com/kiegroup/optaplanner/blob/master/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/lookup/LookUpStrategyResolver.java#L59-L84.

triceo (approved):
I'll merge this to preserve the original authorship and do some follow-up changes in another PR.

triceo:
Thanks for the contribution, @sudolea!

PR #1256, opened 2021-04-08
Bump kie7 to 7.52.0.Final

evacchi:
### Referenced pull requests
https://github.com/kiegroup/kogito-runtimes/pull/1208

### Checklist
- [ ] Documentation updated if applicable.
- [ ] Upgrade recipe provided if applicable.

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* for a <b>pull request build</b> please add comment: <b>Jenkins retest this</b>
* for a <b>full downstream build</b> please add comment: <b>Jenkins run fdb</b>
* for a <b>compile downstream build</b> please add comment: <b>Jenkins run cdb</b>
* for a <b>full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>
* for an <b>upstream build</b> please add comment: <b>Jenkins run upstream</b>
* for a <b>Quarkus LTS check</b> please add comment: <b>Jenkins run LTS</b>
* for a <b>Native check</b> please add comment: <b>Jenkins run native</b>
</details>

PR #1276, opened 2021-04-21
PLANNER-2399 Improve performance of CS-D groupBy

triceo:
Brings improvements of a couple of percentage points.
Depends on https://github.com/kiegroup/optaplanner/pull/1256.


build/optaplanner-build-parent/pom.xml
```diff
@@ -23,7 +23,7 @@
     <!-- SNAPSHOT and KIE dependency versions -->
     <version.org.kie.kogito>2.0.0-SNAPSHOT</version.org.kie.kogito>
     <!-- Keep in sync with version.org.kie7 in kogito-runtimes: https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-build-parent/pom.xml#L185 -->
-    <version.org.kie7>7.52.0.Beta1</version.org.kie7>
+    <version.org.kie7>7.52.0.Final</version.org.kie7>
```
triceo, line 26: This will go aways before merging. (When the bump PR is merged.)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractRuleContext.java
```diff
@@ -36,52 +36,6 @@
...
- * }
- * </pre>
- *
- * In this case, the extra pattern variable "newA" is required,
- * because we want to have a pattern on which to apply the filter.
- * The same goes for joining etc.
- * Whenever the variable is not passed directly into a consequence,
- * the use of this class needs to be replaced by the use of {@link PatternVariable}.
  */
 abstract class AbstractRuleContext {
```
triceo, line 40: The explanation was no longer necessary as the context was hidden from people and slightly repurposed, so I removed it.

rsynek (approved):
LGTM, but let's wait for the CI and SonarCloud.

PR #1283, opened 2021-04-23
PLANNER-2400: Add Micrometer Metric for Solver

Christopher-Chianelli:
Metrics added:
- Score Calculation Speed (calculations/seconds)
- Score Calculation Count (calculations)

### JIRA

https://issues.redhat.com/browse/PLANNER-2400

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -173,16 +173,14 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
             try {
                 solvingStarted(solverScope);
                 runPhases(solverScope);
-
                 solvingEnded(solverScope);
-                sample.stop();
-
-                restartSolver = checkProblemFactChanges();
             } catch (Exception e) {
-                sample.stop();
                 errorCounter.increment();
                 throw e;
+            } finally {
+                sample.stop();
             }
+            restartSolver = checkProblemFactChanges();
         }
         outerSolvingEnded(solverScope);
         return solverScope.getBestSolution();
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -34,6 +34,7 @@ import org.optaplanner.core.impl.solver.scope.SolverScope;
 import org.optaplanner.core.impl.solver.termination.BasicPlumbingTermination;
 import org.optaplanner.core.impl.solver.termination.Termination;
 
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.LongTaskTimer;
 import io.micrometer.core.instrument.Metrics;
 
@@ -59,6 +60,7 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
 
     // Metrics
     private LongTaskTimer solveLengthTimer;
+    private Counter errorCounter;
 
     // ************************************************************************
     // Constructors and simple getters/setters
@@ -75,6 +77,7 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
         this.solverScope = solverScope;
         this.moveThreadCountDescription = moveThreadCountDescription;
         this.solveLengthTimer = Metrics.more().longTaskTimer("optaplanner.solver.solve-length");
+        this.errorCounter = Metrics.counter("optaplanner.solver.errors");
     }
 
     public EnvironmentMode getEnvironmentMode() {
@@ -166,9 +169,9 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
         outerSolvingStarted(solverScope);
         boolean restartSolver = true;
         while (restartSolver) {
-            solvingStarted(solverScope);
             LongTaskTimer.Sample sample = solveLengthTimer.start();
             try {
+                solvingStarted(solverScope);
                 runPhases(solverScope);
 
                 solvingEnded(solverScope);
@@ -176,13 +179,8 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
 
                 restartSolver = checkProblemFactChanges();
             } catch (Exception e) {
-                // Can't store this in a local field; don't know in advance
-                // if we'll have an exception, and what kind of exception
-                // it is
                 sample.stop();
-                Metrics.counter("optaplanner.solver.errors",
-                        "exceptionClass", e.getClass().getSimpleName(),
-                        "message", e.getMessage());
+                errorCounter.increment();
                 throw e;
             }
         }
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -34,6 +34,9 @@ import org.optaplanner.core.impl.solver.scope.SolverScope;
 import org.optaplanner.core.impl.solver.termination.BasicPlumbingTermination;
 import org.optaplanner.core.impl.solver.termination.Termination;
 
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Metrics;
+
 /**
  * Default implementation for {@link Solver}.
  *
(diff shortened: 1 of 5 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/stat/OptaPlannerMetrics.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        counter(registry, "optaplanner.solver.calculation-count", "Total calculations performed by the solver",
+                scoreCalculationSpeedSubSingleStatisticListener,
+                ScoreCalculationSpeedSubSingleStatisticListener::getScoreCalculationCount);
+        gauge(registry, "optaplanner.solver.calculation-speed", "The score calculation speed of the solver",
+                scoreCalculationSpeedSubSingleStatisticListener,
+                ScoreCalculationSpeedSubSingleStatisticListener::getScoreCalculationSpeed);
+    }
```
triceo, line 87: How do I disable these if I don't want some of them?

These will have a non-trivial performance impact, when enabled. The users should be able to chose just the ones they need. (Alternatively, they should be able to exclude the ones they do not want.)

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/ScoreCalculationSpeedSubSingleStatisticListener.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package org.optaplanner.core.api.stat;
+
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListenerAdapter;
+import org.optaplanner.core.impl.phase.scope.AbstractPhaseScope;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
+import org.optaplanner.core.impl.solver.scope.SolverScope;
+
+public class ScoreCalculationSpeedSubSingleStatisticListener<Solution_> extends PhaseLifecycleListenerAdapter<Solution_> {
```
triceo, line 24: I see that this code was taken directly from the benchmarker. That is understandable. But the `SubSingleStatistic` name is not, in this new context. There are no `SubSingle`s in core, and therefore this name needs to change.

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/ScoreCalculationSpeedSubSingleStatisticListener.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public void stepEnded(AbstractStepScope<Solution_> stepScope) {
+        long timeMillisSpent = stepScope.getPhaseScope().calculateSolverTimeMillisSpentUpToNow();
+        if (timeMillisSpent >= nextTimeMillisThreshold) {
+            SolverScope<Solution_> solverScope = stepScope.getPhaseScope().getSolverScope();
+            scoreCalculationCount = solverScope.getScoreCalculationCount();
+            long calculationCountInterval = scoreCalculationCount - lastCalculationCount;
+            long timeMillisSpentInterval = timeMillisSpent - lastTimeMillisSpent;
```
triceo, line 50: Can't micrometer do this for us? To simply take the metric and average it over time?
I would expect this to be the case.

triceo (commented):
Some initial observations.

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/OptaPlannerMetrics.java
Christopher-Chianelli: Micrometer provides this functionality for us: https://micrometer.io/docs/concepts#_meter_filters

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/ScoreCalculationSpeedSubSingleStatisticListener.java
Christopher-Chianelli: Given https://micrometer.io/docs/concepts#rate-aggregation and https://micrometer.io/docs/concepts#_counters probably, but it seems to be an operation that needed to be done on the measurement consumer.

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/ScoreCalculationSpeedSubSingleStatisticListener.java
Christopher-Chianelli: Hibernate does use Statistics for its Stat tracker: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-micrometer/src/main/java/org/hibernate/stat/HibernateMetrics.java

So perhaps `ScoreCalculationCountStatistic`?

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/OptaPlannerMetrics.java
triceo: I assumed it would. This solves the problem of disabling it by the users of core.
That said, we still need a way for benchmarker to do this, and if that was the same way, that would be nice.
I still think that solver config would be the way to go here - expose it on our level, as opposed to exposing Micrometer.

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/ScoreCalculationSpeedSubSingleStatisticListener.java
triceo: Works for me.

optaplanner-core/src/main/java/org/optaplanner/core/api/stat/ScoreCalculationSpeedSubSingleStatisticListener.java
triceo: For the moment, it will be easier to just keep it as is. (Especially if we want to consume these metrics from benchmarker.) But over time, I'd rather see us use micrometer as opposed to doing things like timing ourselves.

ge0ffrey:
We can meet up with Erin (MicroMeter expert) to assert our approach

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -160,9 +167,24 @@ public final Solution_ solve(Solution_ problem) {
...
+
+                restartSolver = checkProblemFactChanges();
+            } catch (Exception e) {
+                // Can't store this in a local field; don't know in advance
+                // if we'll have an exception, and what kind of exception
+                // it is
+                sample.stop();
+                Metrics.counter("optaplanner.solver.errors",
+                        "exceptionClass", e.getClass().getSimpleName(),
+                        "message", e.getMessage());
```
triceo, line 185: First of all, does this actually measure anything? I see the counter being created, but not used.
Second, do we need to tag by exception and message? Is that important information?

I think we need just error count. And in that case, you can keep it in a field.

triceo (commented):
Some questions here:

- Did you benchmark the impacts of this on the solver performance? Let's confirm the hypothesis that, without and metric provider, this has no impact.
- I vaguely remember we areed we'd enable configuration via the solver config. That is not available in this PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -160,9 +167,24 @@ public final Solution_ solve(Solution_ problem) {
...
         while (restartSolver) {
             solvingStarted(solverScope);
-            runPhases(solverScope);
-            solvingEnded(solverScope);
-            restartSolver = checkProblemFactChanges();
+            LongTaskTimer.Sample sample = solveLengthTimer.start();
+            try {
+                runPhases(solverScope);
+
+                solvingEnded(solverScope);
```
triceo, line 174: If we're measuring solvingEnded(), we should also measure solvingStarted().

Christopher-Chianelli:
@triceo Configuration of meters is under phase 2; this is phase 1, which is just the `# of active solver, length of solve, error count` metrics.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -159,10 +169,20 @@ public final Solution_ solve(Solution_ problem) {
...
+                sample.stop();
+
+                restartSolver = checkProblemFactChanges();
+            } catch (Exception e) {
+                sample.stop();
+                errorCounter.increment();
+                throw e;
+            }
```
triceo, line 185:
```suggestion
            try {
                solvingStarted(solverScope);
                runPhases(solverScope);
                solvingEnded(solverScope);
            } catch (Exception e) {
                errorCounter.increment();
                throw e;
            } finally {
                sample.stop();
            }
            restartSolver = checkProblemFactChanges();
```

triceo (approved):

PR #1289, opened 2021-04-27
PLANNER-2404 Avoid defensive copies of collections in ConstraintCollectors

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -974,7 +977,10 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
      * Creates constraint collector that returns {@link Set} of the same element type as the {@link ConstraintStream}.
      * Makes no guarantees on iteration order.
      * For stable iteration order, use {@link #toSortedSet()}.
-     *
+     * <p>
+     * The contents of the collection will be changing during {@link org.optaplanner.core.api.solver.Solver} lifetime,
```
ge0ffrey, line 981:
```suggestion
     * The contents of the collection will change during {@link org.optaplanner.core.api.solver.Solver} lifetime,
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -985,18 +991,24 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
     /**
      * Creates constraint collector that returns {@link SortedSet} of the same element type as the
      * {@link ConstraintStream}.
+     * <p>
+     * The contents of the collection will be changing during {@link org.optaplanner.core.api.solver.Solver} lifetime,
```
ge0ffrey, line 995:
```suggestion
     * The contents of the collection will change during {@link org.optaplanner.core.api.solver.Solver} lifetime,
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -985,18 +991,24 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
...
-        return toCollection(Function.identity(), i -> new TreeSet<>());
+        return toSortedSet(a -> a);
     }
 
     /**
      * Creates constraint collector that returns {@link List} of the same element type as the {@link ConstraintStream}.
      * Makes no guarantees on iteration order.
      * For stable iteration order, use {@link #toSortedSet()}.
+     * <p>
+     * The contents of the collection will be changing during {@link org.optaplanner.core.api.solver.Solver} lifetime,
```
ge0ffrey, line 1010:
```suggestion
     * The contents of the collection will change during {@link org.optaplanner.core.api.solver.Solver} lifetime,
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -985,18 +991,24 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
     /**
      * Creates constraint collector that returns {@link SortedSet} of the same element type as the
      * {@link ConstraintStream}.
+     * <p>
+     * The contents of the collection will be changing during {@link org.optaplanner.core.api.solver.Solver} lifetime,
+     * they must therefore never be modified nor stored without making a defensive copy first.
```
ge0ffrey, line 996: Can the users modify or store elements of that list? No, they can't.
In a vanilla CS, can the users modify or store the tuple facts? No, they can't.

I don't think this needs to be explicitly stated.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1406,6 +1524,8 @@ public void remove(Key key, Value value) {
      * <p>
      * Makes no guarantees on iteration order for map entries.
      * For stable iteration order, use {@link #toSortedMap(Function, Function, BinaryOperator)}.
+     * <p>
+     * This constraint collector is slow and causes considerable garbage collector pressure.
```
ge0ffrey, line 1528: If we do this change for toList and toSet and toSortedSet, we should do it for toMap etc too, for consistency (principle of least surprise).

And yes, I think we need to do this change :)

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1225,13 +1331,18 @@ private static Runnable innerCountDistinctLong(CountDistinctLongResultContainer
      */
     public static <A, B, C, D, Mapped extends Comparable<Mapped>> QuadConstraintCollector<A, B, C, D, ?, SortedSet<Mapped>>
             toSortedSet(QuadFunction<A, B, C, D, Mapped> groupValueMapping) {
```
ge0ffrey, line 1333: Separate issue: as we support toSortedSet(mapping), we should probably also support toSortedSet(mapping, Comparator). Not having it is an API gap.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: Making `toMap()` incremental in this way is likely impossible.
The interface of the method requires post-processing on the internal map before the collector returns.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: Making toMap() incremental in this way is likely impossible.
The interface of the method requires post-processing on the internal map before the collector returns.

ge0ffrey (approved):
Great improvement!

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: https://issues.redhat.com/browse/PLANNER-2405

optaplanner-core/src/main/java/org/optaplanner/core/api/score/stream/ConstraintCollectors.java
triceo: https://issues.redhat.com/browse/PLANNER-2406

PR #1294, opened 2021-04-28
Only run full profile on Jenkins

triceo:


.github/workflows/os_java_ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
       - name: Build with Maven
-        run: mvn -B --fail-at-end clean install -Dfull
+        run: mvn -B --fail-at-end clean install
 
   # Early feedback on a compatibility with the newest Java version.
   # TODO: update each time a new major Java version is released.
```
rsynek, line 43:
```suggestion
  # TODO: Update each time a new major Java version is released. Make sure you build OptaPlanner with the full profile locally with the new Java version first.
```

rsynek (approved):
One suggestion to increase the chance we won't forget to test the full profile with the new JDK.

ge0ffrey (changes requested):
I 'd rather stop relying on Jenkins. It's not reliable.
Many important checks only happen in the full profile.

triceo:
There is no consensus on this.

PR #1296, opened 2021-04-30
generalPhaseSequence: use Late Acceptance because that's the default now

ge0ffrey:
This can be merged earlier. It's only the blog PR that needs to wait till 5-May.


yurloc (approved):
Makes sense.

PR #1308, opened 2021-05-07
PLANNER-2413: Make classes @ClassReactive by default in Quarkus

Christopher-Chianelli:
This able disables drools alpha compilation when DRL is used
in Quarkus (which cause rules to not fire) and make the DRL integration
test more robust.

### JIRA

https://issues.redhat.com/browse/PLANNER-2413

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
@@ -53,12 +53,15 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
+import org.kie.api.definition.type.ClassReactive;
+import org.kie.api.definition.type.PropertyReactive;
 import org.kie.kogito.rules.KieRuntimeBuilder;
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Type;
 import org.optaplanner.core.api.domain.solution.cloner.SolutionCloner;
 import org.optaplanner.core.config.score.director.ScoreDirectorFactoryConfig;
+import org.optaplanner.core.config.solver.SolverConfig;
 import org.optaplanner.core.config.util.ConfigUtils;
 import org.optaplanner.core.impl.domain.common.ReflectionHelper;
 import org.optaplanner.core.impl.domain.common.accessor.MemberAccessor;
@@ -441,8 +444,9 @@ public class GizmoMemberAccessorEntityEnhancer {
         return generatedClassName;
     }
 
-    public static String generateKieRuntimeBuilder(ClassOutput classOutput, ScoreDirectorFactoryConfig config,
-            BuildProducer<UnremovableBeanBuildItem> unremovableBeans) {
+    public static String generateKieRuntimeBuilder(ClassOutput classOutput, SolverConfig config,
+            BuildProducer<UnremovableBeanBuildItem> unremovableBeans,
+            BuildProducer<BytecodeTransformerBuildItem> transformers) {
         String generatedClassName = DROOLS_INITIALIZER_CLASS_NAME;
         try (ClassCreator classCreator = ClassCreator
                 .builder()
@@ -454,8 +458,8 @@ public class GizmoMemberAccessorEntityEnhancer {
             classCreator.addAnnotation(ApplicationScoped.class);
 
             MethodCreator methodCreator;
-            if (!ConfigUtils.isEmptyCollection(config.getScoreDrlList()) ||
-                    !ConfigUtils.isEmptyCollection(config.getScoreDrlFileList())) {
+            if (!ConfigUtils.isEmptyCollection(config.getScoreDirectorFactoryConfig().getScoreDrlList()) ||
+                    !ConfigUtils.isEmptyCollection(config.getScoreDirectorFactoryConfig().getScoreDrlFileList())) {
                 unremovableBeans.produce(UnremovableBeanBuildItem.beanTypes(KieRuntimeBuilder.class));
                 methodCreator = classCreator.getMethodCreator(MethodDescriptor.ofMethod(OptaPlannerDroolsInitializer.class,
                         "setup", void.class, ScoreDirectorFactoryConfig.class));
@@ -482,6 +486,13 @@ public class GizmoMemberAccessorEntityEnhancer {
                                 KieBaseExtractor.class),
                         methodCreator.getMethodParam(0), kieBaseExtractor);
 
+                transformers.produce(new BytecodeTransformerBuildItem(config.getSolutionClass().getName(),
+                        (className, classVisitor) -> new OptaPlannerDroolsReactiveClassVisitor(config.getSolutionClass(),
+                                classVisitor)));
+                for (Class<?> entityClass : config.getEntityClassList()) {
+                    transformers.produce(new BytecodeTransformerBuildItem(entityClass.getName(),
+                            (className, classVisitor) -> new OptaPlannerDroolsReactiveClassVisitor(entityClass, classVisitor)));
+                }
             } else {
                 // No additional setup needed; Drools isn't used
                 methodCreator = classCreator.getMethodCreator(MethodDescriptor.ofMethod(OptaPlannerDroolsInitializer.class,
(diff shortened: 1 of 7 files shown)


optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -432,6 +432,8 @@ protected void applyScoreDirectorFactoryProperties(IndexView indexView, SolverCo
 
         if (solverConfig.getScoreDirectorFactoryConfig().getScoreDrlList() != null) {
             boolean isKogitoExtensionPresent = capabilities.isPresent("kogito-rules");
+            // Rules do not fire when Drools Alpha Network Compilation is enabled
+            solverConfig.getScoreDirectorFactoryConfig().setDroolsAlphaNetworkCompilationEnabled(false);
```
triceo, line 436: I am not sure why this should happen with DRL and not with CS-D. Both use ANC.
ANC has to be disabled in native mode, because there it doesn't work. But I would really like to understand why it works for non-native CS-D and apparently not for non-native DRL.

Christopher-Chianelli: I would like to know to. A rule like
```
rule "always fire"
    eval(true)
then
    ...
end
```
will always fire whereas
```
rule "never fires"
    Object()
then
    ...
end
```
never fires (despite the fact it should fire for every fact). It does this silently; no errors -- it can only be caught using an assertion score director or looking at the solution.

triceo: Weird indeed. Is ANC on the classpath? Could be useful to hook up debugger to Drools and check what happens when facts are inserted into the working memory. Maybe nothing happens because nothing in inserted?

Although I think that it is more likely that there is some classloader magic somewhere and this `Object` isn't the same as the `Object` being inserted.

triceo: Could the ANC be happening with a different classloader than what Quarkus uses? (Just guessing.)

Christopher-Chianelli: It might; it uses the root class loader of the InternalKnowledgeBase.

Christopher-Chianelli: `KieBaseUpdaterANC.generateAndSetInMemoryANC(kieBase)` works with `kogito-quarkus-rules` with Drools by itself.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -496,6 +507,25 @@ public static String generateKieRuntimeBuilder(ClassOutput classOutput, ScoreDir
...
+        public OptaPlannerDroolsReactiveClassVisitor(Class<?> clazz, ClassVisitor outputClassVisitor) {
+            super(Gizmo.ASM_API_VERSION, outputClassVisitor);
+            this.clazz = clazz;
+        }
+
+        @Override
+        public void visitEnd() {
+            super.visitEnd();
+
+            if (clazz.getAnnotation(ClassReactive.class) == null && clazz.getAnnotation(PropertyReactive.class) == null) {
```
triceo, line 522: For the record, there are three possible values for property reactivity in Drools:

1. DISABLED ("not possible")
2. ALLOWED ("possible")
3. ENABLED ("actually enabled")

CS-D is always 1 - it's a green field project, so we have the power the disable it, and we do.
DRL is always 2, for backwards compatibility reasons.

We should not change this behavior in Quarkus.

Christopher-Chianelli: It is effectively the same behavior. If a class is annotated with `@ClassReactive` or `@PropertyReactive`, nothing happens. If it is not annotated with either of those annotations, it is annotated with `@ClassReactive`. This means types are Class Reactive unless is it annotated with `@PropertyReactive`, in which case they are property reactive. Drools doesn't allow setting the propertyReactive property yet in Quarkus (see https://issues.redhat.com/browse/KOGITO-5101 )

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -482,6 +486,13 @@ public static String generateKieRuntimeBuilder(ClassOutput classOutput, ScoreDir
                                 KieBaseExtractor.class),
                         methodCreator.getMethodParam(0), kieBaseExtractor);
 
+                transformers.produce(new BytecodeTransformerBuildItem(config.getSolutionClass().getName(),
+                        (className, classVisitor) -> new OptaPlannerDroolsReactiveClassVisitor(config.getSolutionClass(),
+                                classVisitor)));
+                for (Class<?> entityClass : config.getEntityClassList()) {
+                    transformers.produce(new BytecodeTransformerBuildItem(entityClass.getName(),
+                            (className, classVisitor) -> new OptaPlannerDroolsReactiveClassVisitor(entityClass, classVisitor)));
+                }
```
triceo, line 495: I'm afraid this is not enough. Property reactivity could be declared on any problem fact as well.
I think we should just allow prop reactive, no questions asked. (Just like it is done on DRL outside of Quarkus.)

optaplanner-quarkus-integration/optaplanner-quarkus/drl-integration-test/src/test/java/org/optaplanner/quarkus/drl/it/OptaPlannerTestResourceTest.java
```diff
@@ -33,14 +36,21 @@
 
     @Test
     @Timeout(600)
+    @Disabled("There an issue where enabling ANC twice on same KieBase causes rules not to fire" +
+            " which will cause this test to fail. Enable this test once that issue is resolved.")
```
triceo, line 40: I'd add a pointer to the JIRA in question. Otherwise this will be hard to look up in the future.

triceo (changes requested):
I think we need to alter our approach slightly.

optaplanner-quarkus-integration/optaplanner-quarkus/drl-integration-test/src/test/java/org/optaplanner/quarkus/drl/it/OptaPlannerTestResourceTest.java
Christopher-Chianelli: It did not exist at the time; created it now: https://issues.redhat.com/browse/DROOLS-6336

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
Christopher-Chianelli: (can problem facts change during a solver run? My understanding is the solver is restarted after a PFC, in which case the problem fact is removed and re-inserted (vs notifying via changed property like planning entity)

triceo: Is it worth it? I am not sure if we can do this in a foolproof way. Literally any class on the classpath could be configured to be property reactive - you can't scan it all, can you?

Christopher-Chianelli: OptaPlanner only touches Planning Entities in the DRL. And we know the list of planning entities classes. Additional, classes are configured to be property reactive via annotations, which we can (and do) read.

triceo: Ok, so let me put it another way - let's write a test that shows that prop reactivity works as expected.
That means that there should be a DRL that has some rules fired (and some not) based on the fields of objects that changed.

I really do not think we should mess with this. It is not our business - prop reactive is a Drools feature, not OP feature. But if we decide to mess with this, let's make sure we do it correctly.

PR #1321, opened 2021-05-14
PLANNER-2425: Move @PlanningScore annotation processing to ScoreDescriptor

yurloc:
### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
@@ -1,0 +1,238 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.core.impl.domain.score.descriptor;
+
+import static org.optaplanner.core.impl.domain.common.accessor.MemberAccessorFactory.MemberAccessorType.FIELD_OR_GETTER_METHOD_WITH_SETTER;
+
+import java.lang.reflect.Member;
+
+import org.optaplanner.core.api.domain.solution.PlanningScore;
+import org.optaplanner.core.api.score.AbstractBendableScore;
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.api.score.buildin.bendable.BendableScore;
+import org.optaplanner.core.api.score.buildin.bendablebigdecimal.BendableBigDecimalScore;
+import org.optaplanner.core.api.score.buildin.bendablelong.BendableLongScore;
+import org.optaplanner.core.api.score.buildin.hardmediumsoft.HardMediumSoftScore;
+import org.optaplanner.core.api.score.buildin.hardmediumsoftbigdecimal.HardMediumSoftBigDecimalScore;
+import org.optaplanner.core.api.score.buildin.hardmediumsoftlong.HardMediumSoftLongScore;
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+import org.optaplanner.core.api.score.buildin.hardsoftbigdecimal.HardSoftBigDecimalScore;
+import org.optaplanner.core.api.score.buildin.hardsoftlong.HardSoftLongScore;
+import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
+import org.optaplanner.core.api.score.buildin.simplebigdecimal.SimpleBigDecimalScore;
+import org.optaplanner.core.api.score.buildin.simplelong.SimpleLongScore;
+import org.optaplanner.core.config.util.ConfigUtils;
+import org.optaplanner.core.impl.domain.common.accessor.MemberAccessor;
+import org.optaplanner.core.impl.domain.common.accessor.MemberAccessorFactory;
+import org.optaplanner.core.impl.domain.policy.DescriptorPolicy;
+import org.optaplanner.core.impl.score.buildin.bendable.BendableScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.bendablebigdecimal.BendableBigDecimalScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.bendablelong.BendableLongScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardmediumsoft.HardMediumSoftScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardmediumsoftbigdecimal.HardMediumSoftBigDecimalScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardmediumsoftlong.HardMediumSoftLongScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardsoft.HardSoftScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardsoftbigdecimal.HardSoftBigDecimalScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardsoftlong.HardSoftLongScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.simple.SimpleScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.simplebigdecimal.SimpleBigDecimalScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.simplelong.SimpleLongScoreDefinition;
+import org.optaplanner.core.impl.score.definition.ScoreDefinition;
+
+public class ScoreDescriptor {
+
+    // Used to obtain default @PlanningScore attribute values from a score member that was auto-discovered,
+    // as if it had an empty @PlanningScore annotation on it.
+    @PlanningScore
+    private static final Object PLANNING_SCORE = new Object();
+
+    private final MemberAccessor scoreMemberAccessor;
+    private final ScoreDefinition<?> scoreDefinition;
+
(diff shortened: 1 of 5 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -319,7 +304,13 @@ private void processFactEntityOrScoreAnnotation(DescriptorPolicy descriptorPolic
                 || annotationClass.equals(PlanningEntityCollectionProperty.class)) {
             processPlanningEntityPropertyAnnotation(descriptorPolicy, member, annotationClass);
         } else if (annotationClass.equals(PlanningScore.class)) {
-            processScoreAnnotation(descriptorPolicy, member, annotationClass);
+            if (scoreDescriptor == null) {
+                // Bottom class wins. Bottom classes are parsed first due to ConfigUtil.getAllAnnotatedLineageClasses().
+                scoreDescriptor = ScoreDescriptor.buildScoreDescriptor(descriptorPolicy, member, solutionClass);
+            } else {
+                // If an ancestor class has a score member, it must have the same type and name.
```
ge0ffrey, line 311: Comment softly implies that there is not other case that invalidates, but there is one: 1 solution has 2 score members

Maybe something like this:
```suggestion
                // If an ancestor class has a score member (or a solution has 2 score members), it must have the same type and name.
```

ge0ffrey (approved):
LGTM. Do bring annotation class back as an inner class or just a field as discussed

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
```diff
@@ -0,0 +1,221 @@
...
+import org.optaplanner.core.impl.score.buildin.hardmediumsoftlong.HardMediumSoftLongScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardsoft.HardSoftScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardsoftbigdecimal.HardSoftBigDecimalScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.hardsoftlong.HardSoftLongScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.simple.SimpleScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.simplebigdecimal.SimpleBigDecimalScoreDefinition;
+import org.optaplanner.core.impl.score.buildin.simplelong.SimpleLongScoreDefinition;
+import org.optaplanner.core.impl.score.definition.ScoreDefinition;
+
+public class ScoreDescriptor {
```
triceo, line 56: I'd make this class generic in `<Score_ extends Score<Score_>>`. That way, this class can be internally type-safe, even if from the outside, we always use it raw.

This way,  you'll get to avoid a lot of question marks in the source code.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
```diff
@@ -0,0 +1,221 @@
...
+    private static MemberAccessor buildScoreMemberAccessor(DescriptorPolicy descriptorPolicy, Member member) {
+        return MemberAccessorFactory.buildMemberAccessor(
+                member,
+                FIELD_OR_GETTER_METHOD_WITH_SETTER,
+                PlanningScore.class,
+                descriptorPolicy.getDomainAccessType(),
+                descriptorPolicy.getGeneratedMemberAccessorMap());
+    }
+
+    private static Class<? extends Score<?>> extractScoreType(MemberAccessor scoreMemberAccessor, Class<?> solutionClass) {
```
triceo, line 89: If these methods were not static and the class was generified as suggested, a lot of `? extends Score<?>` would go away for code that is easier to read.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
```diff
@@ -0,0 +1,221 @@
...
+
+    public ScoreDefinition<?> getScoreDefinition() {
+        return scoreDefinition;
+    }
+
+    public Class<? extends Score<?>> getScoreClass() {
+        return scoreDefinition.getScoreClass();
+    }
+
+    public <Solution_> Score<?> getScore(Solution_ solution) {
```
triceo, line 202: Consider perhaps generifying this class in <Solution_> too. There are more places than just this method here where it could be useful.

(Many classes throghout the solver are generic both in score and in solution.)

triceo (approved):
I like this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
yurloc: `<Solution_>` here is an automatic refactoring error that I have overlooked.

My first choice would be simply `Object`. That doesn't hurt type safety neither inside nor outside of the `ScoreDescriptor`.

On second thought I could imagine we generify `ScoreDescriptor<Solution_>` just because of the convention that everything that deals with solution is generic. Would you prefer that knowing that `<Solution_>` parameter doesn't add any type safety?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
yurloc: This class is internally type safe even when using question marks. But I would agree that using `Score_` could avoid repetition and improve readability. I'm just not sure if generics were intended for this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
yurloc: I gave it a try and I definitely downvote `ScoreDescriptor<Solution_>`. I found zero benefit.

The reason is that the majority of solution occurrences in this class is `solutionClass` that is used for constructing exception messages, so `Class<?> solutionClass` is perfectly OK, I think.

The rest of occurrences is in `getScore()/setScore()` where the solution type is irrelevant so I can't see the benefit of `Solution_` here either.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
yurloc: Gave it a try. Unlike `<Solution_>`, `<Score_ extends Score<Score_>>` seems to yield some benefit but it also enforces a lot of unchecked casting. I'm puzzled.

I suggest to merge this PR without generic `ScoreDescriptor` and revise `Score_` later.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
yurloc: @ge0ffrey You have a good point but the suggested change doesn't make sense. I've removed the comment and renamed the method to `failFastOnDuplicateMember()` to better reflect the purpose. Exactly which other score members are allowed is not important on this level.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
```diff
@@ -0,0 +1,238 @@
...
+    public void setScore(Object solution, Score<?> score) {
+        scoreMemberAccessor.executeSetter(solution, score);
+    }
+
+    public void failFastOnDuplicateMember(DescriptorPolicy descriptorPolicy, Member member, Class<?> solutionClass) {
+        MemberAccessor memberAccessor = buildScoreMemberAccessor(descriptorPolicy, member);
+        // A solution class cannot have more than one score field or bean property (name check), and the @PlanningScore
+        // annotation cannot appear on both the score field and its getter (member accessor class check).
+        if (!scoreMemberAccessor.getName().equals(memberAccessor.getName())
+                || !scoreMemberAccessor.getClass().equals(memberAccessor.getClass())) {
```
yurloc, line 230: @ge0ffrey Here is the updated comment.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
triceo: Agree on `Solution_`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
triceo: I had to try this code myself to find out why `Score_` is a problem. Java type system doesn't really help here.
Anyway, ignore the suggestion.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/score/descriptor/ScoreDescriptor.java
```diff
@@ -0,0 +1,238 @@
...
+                return new BendableBigDecimalScoreDefinition(bendableHardLevelsSize, bendableSoftLevelsSize);
+            } else {
+                throw new IllegalArgumentException("The solutionClass (" + solutionClass
+                        + ") has a @" + PlanningScore.class.getSimpleName()
+                        + " annotated member (" + scoreMemberAccessor
+                        + ") that returns a bendable scoreType (" + scoreType
+                        + ") that is not recognized as a default " + Score.class.getSimpleName() + " implementation.\n"
+                        + "  If you intend to use a custom implementation,"
+                        + " maybe set a scoreDefinition in the annotation.");
+            }
```
triceo, line 205: Personally, I believe this belongs on `ScoreDefinition`. Something like `public static Optional<ScoreDefinition> findFor(Class<Score> scoreClass)`.

The exception thrown can then be customized at the call site which receives the empty `Optional`.

yurloc: The idea is nice but `ScoreDefinition` construction is not that simple because 1) bendable score definitions require `hardLevelSize` and `softLevelSize`, 2) the exception messages require additional information that is not related to score defintion (solution class, score member).

PR #1324, opened 2021-05-17
PLANNER-2405 Constraint Collector toSortedSet() overloads with Comparator

triceo:


optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -1813,8 +1812,8 @@ public void toSetBi() {
 
     @Test
     public void toSortedSetBi() {
-        BiConstraintCollector<Integer, Integer, ?, SortedSet<Integer>> collector = ConstraintCollectors
-                .toSortedSet(Integer::sum);
+        BiConstraintCollector<Integer, Integer, ?, SortedSet<Integer>> collector =
+                ConstraintCollectors.toSortedSet((BiFunction<Integer, Integer, Integer>) Integer::sum);
```
triceo, line 1816: There is a problem here. The addition of `ConstraintCollectors.toSortedSet(Comparator)` has caused signature conflicts with `ConstraintCollectors.toSortedSet(BiFunction)`. `Comparator` *is* `BiFunction`.

If we allow this, we effectively break compilation of some existing projects, as evidenced by this test.

The solution?

1. Do nothing. This is a corner-case.
2. Not introduce the `toSortedSet()` overload in this case, leading to an unfortunate unfixable API gap.
3. Call the new overload something other than `toSortedSet()`. This creates a nasty inconsistency in the API.

I am leaning towards 1.

rsynek: I agree that in the list of these bad options this looks like the best one. Let's properly document it in the upgrade recipe.

rsynek (approved):

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
triceo: @ge0ffrey Opinions?

ge0ffrey: -1 on 2) Do not introduce. API gap indeed.

How many users use sortedSet with Bi/Tri/Quad vs how many use sorted set with a comperator?
I guesstimate more people use sortedSet with Bi/Tri/Quad (and I think Uni should be in that list too), as the ones who use it with a comperator are a subset. I'd guess one out of 4 uses non-natural comparison.

1) Do nothing. => Everyone that uses toSortedSet on Bi/Tri/Quad (with or without comparator) or Uni+comparator feels the pain: they need to cast to tell the compiler which method they mean.
3A) Keep toSortedSet(BiFunction) and create toSortedSetWithComperator(Comperator). Similar for Uni/Tri/Quad's toSortedSetWithComperator.
Only the comperator users feel the pain.
3B) Keep toSortedSet(Comperator) and create toSortedSetBi(BiFunction). Please god, now, this is inconsistent with all other API Collector methods.

ge0ffrey: I favor 3A), but that would mean that other methods that accept a comperator in the Collectors API would also need to be renamed as such.

I can live with 1) too. Your call, Lukas.

ge0ffrey (approved):
Approved whether 1) or 3A) is taken, but see my motivation in that comment though.

triceo:
OK, I am merging as is. `toSortedSet()` is already pretty niche, and the change in overloads only required one line in our entire codebase to change.

PR #1328, opened 2021-05-17
PLANNER-2426 Support duplicate tuples in score corruption analysis

triceo:


ge0ffrey (approved):
I can't say I grok the code (I especially wonder about that `limit(.limit(corruptedMatchCount - uncorruptedMatchCount)`), but if you've tested it (so it's still correct), it's good for me to merge.

Performance here doesn't matter, only correctness, because this code only executes after score corruption is detected.

PR #1329, opened 2021-05-18
PLANNER-2423 Benchmark returns path to its directory

triceo:


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/api/PlannerBenchmark.java
```diff
@@ -25,13 +27,17 @@
...
+    Path benchmark();
 
     /**
      * Run all the single benchmarks, create an overview report
      * and show it in the default browser.
+     *
+     * @return never null, path to the directory in which the benchmark results are stored
      */
-    void benchmarkAndShowReportInBrowser();
+    Path benchmarkAndShowReportInBrowser();
```
triceo, line 41: Technically, this is backwards incompatible. Practically, it will not do any harm.
It should still be noted in the upgrade recipe, which I will do after this PR is approved.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/api/PlannerBenchmark.java
```diff
@@ -25,13 +27,17 @@
 
     /**
      * Run all the single benchmarks and create an overview report.
+     *
+     * @return never null, path to the directory in which the benchmark results are stored
      */
-    void benchmark();
+    Path benchmark();
```
ge0ffrey, line 33: Why a Path and not a File? Because Path is the new thing.
But the entire optaplanner-benchmark API uses File currently.
Let's **not** mix them.

Proposal A) Return Path (mix them)
Proposal B) Return Path + refactor all usages of File on the benchmark api to use Path instead. This breaks BC. Not possible for 8.X
Proposal C) Return File for now, create an issue linked to PLANNER-2295 to refactor optaplanner-benchmark to use Path instead of File.

ge0ffrey: I think we need to do C). Wdyt?

ge0ffrey (approved):
Good for me if it returns File instead of Path.
Otherwise, happy to discuss on bluejeans.

PR #1304, opened 2021-05-06
KOGITO-5070 Setup multijob pipeline

radtriste:
https://issues.redhat.com/browse/KOGITO-5070

This will allow to run Jenkins pr check, if `multijob-pr` label is set, in separate jobs.
Also, each project check is responsible to set correct Jenkinsfile and this avoids to rewrite the commands to run in each repo for the downstream projects.
Also, this add `kogito-apps` and `kogito-examples` checks to Optaplanner (but can be removed if needed).

_For now, this does not replace the current global PR check. It is only activated with the `multijob-pr` label._

Depends on https://github.com/kiegroup/kogito-pipelines/pull/148

Related PRs:
- https://github.com/kiegroup/kogito-runtimes/pull/1284
- https://github.com/kiegroup/optaplanner/pull/1304
- https://github.com/kiegroup/kogito-apps/pull/805
- https://github.com/kiegroup/kogito-examples/pull/671

_Final note: Current run PR checks for this PR are from my custom folder on Jenkins and will be moved to Kogito folder once all the PRs are merged. Please do not use before ..._

PR #1331, opened 2021-05-19
Kogito-apps/Kogito-examples downstream build

danielezonca:
### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/.github/workflows/kogito-apps-downstream-pr.yml
+++ b/.github/workflows/kogito-apps-downstream-pr.yml
@@ -1,0 +1,59 @@
+name: Kogito Apps Downstream Build
+
+on: [pull_request]
+
+jobs:
+  apps-build:
+    timeout-minutes: 120
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        java-version: [11]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    name: Maven Build
+    steps:
+      - name: Disk space report before modification
+        shell: bash
+        run: |
+          echo "Available storage:"
+          df -h
+      # Inspired to maximize-build-space action https://github.com/easimon/maximize-build-space
+      - name: Free disk space (remove dotnet, android and haskell)
+        shell: bash
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+      - name: Disk space report after modification
+        shell: bash
+        run: |
+          echo "Available storage:"
+          df -h
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java-version }}
+      # See https://docs.github.com/en/actions/guides/building-and-testing-java-with-maven#caching-dependencies
+      - name: Cache Maven packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build Chain ${{ matrix.java-version }}
+        id: build-chain
+        uses: kiegroup/github-action-build-chain@v2.5.7
+        with:
+          definition-file: https://raw.githubusercontent.com/${GROUP}/kogito-pipelines/${BRANCH}/.ci/pull-request-config.yaml
+          starting-project: kiegroup/kogito-apps
+        env:
+          BUILD_MVN_OPTS: "-DskipUI"
+      - name: Publish Test Report
+        if: ${{ always() }}
+        uses: ginxo/action-surefire-report@BXMSPROD-1311
+        with:
+          fail_on_test_failures: true
+          fail_if_no_tests: false
+          skip_check_run_creation: true
+          report_paths: '**/*-reports/TEST-*.xml'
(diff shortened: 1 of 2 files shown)


.github/workflows/kogito-apps-downstream-pr.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build Chain ${{ matrix.java-version }}
+        id: build-chain
+        uses: kiegroup/github-action-build-chain@v2.5.7
+        with:
+          definition-file: https://raw.githubusercontent.com/${GROUP}/kogito-pipelines/${BRANCH}/.ci/pull-request-config.yaml
```
triceo, line 48: I am not an expert at CI, but I have a suspicion that this will not work. While both Kogito and Optaplanner have `master` branches (for now), our release branches differ. Where Kogito uses 1.x, we use 8.x. Reusing `${BRANCH}` therefore can not work, I think. (Unless there is some magic there that I am ignorant of.)

danielezonca:
@rsynek @triceo @ge0ffrey 
I realised that we can use as downstream build the same `kogito-apps` and `kogito-examples` GHA workflow from [`kogito-runtimes`](https://github.com/kiegroup/kogito-runtimes/tree/master/.github/workflows). This PR introduces 2 additional jobs for each PR to verify if it breaks downstream usages in Kogito.

It is a full build of kogito-apps and kogito-examples and each of them takes ~1 hour.

We can make them a simple compile downstream build using a `-DskipTests` as additional [`BUILD_MAVEN_OPTS`](https://github.com/kiegroup/optaplanner/pull/1331/files#diff-1f1f15c02e8ebc813d33db144d9c9fb8d62eaee4d9684f309519124ec4f7911eR51) value but it will be more risky.

If we want to reduce the size of these additional builds, it is probably possible to create an additional local [`pull-request-config`](https://github.com/kiegroup/kogito-pipelines/blob/master/.ci/pull-request-config.yaml) that extends and override the common one similar to what we do for Kogito downstream build in [drools repo](https://github.com/kiegroup/drools/blob/master/.ci/kogito-pull-request-config.yaml#L3). In this additional conf we can specify a build command with the list of modules you want to test (like `mvn clean install -pl module1,module2`). Of course, a similar approach will require to pay more attention when a new OptaPlanner usage is added

Wdyt?

.github/workflows/kogito-apps-downstream-pr.yml
danielezonca: You are right, for `master` it works well while for branches it will require some additional work: it is already possible to provide a static remapping (like we [do for 7.x](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/d3aa20ac2176b34c0a5f62dbeb00ba75b3f3de28/.ci/project-dependencies.yaml#L36)).
I already asked @Ginxo to think about an improvement for `github-action-build-chain` where the branch mapping policies is more advanced. I expect it will take some days/weeks to have it solved.
For now this integration works only if the PR is targeting `master` (so not for backports)

triceo: Makes sense, thanks for the explanation.

Ginxo: @triceo @danielezonca this is the issue to improve it https://github.com/kiegroup/github-action-build-chain/issues/148
I will work on it as soon as I adapt kogito-tooling nightly.

rsynek:
@danielezonca 
I would definitely prefer limiting the downstream builds to only relevant modules. Even now we sometimes wait for GHA as the number of jobs being executed in the kiegroup has grown. This trend, if it continues, may soon render GHA unusable as we will spend more time waiting than building.

rsynek:
@danielezonca One more thing we touched upon with Tristan in https://github.com/kiegroup/optaplanner/pull/1304, which brings an on-demand downstream build on Jenkins: does it make sense to use the Build Chain also on-demand, simply as a backup in case Jenkins is not in the best shape?

danielezonca:
Fine for me to limit downstream to relevant modules.
If we want to limit the number of concurrent jobs we can also add these additional steps to one of the existing PR job that OptaPlanner has but of course in this case it will increase the execution time, I don't know what is better 🤔 

About Jenkins or GHA as far as I can see our Jenkins env is particular instable with integration test (like kafka, testcontainer, end to end integration test) and every Kogito repo has integration test so it is quite hard to have a green build with Jenkins since probably one/two months.
Especially with kogito-runtimes where the build takes 3 hours the risk of getting a networking issue is close to 1 :)
At the moment I prefer to keep both configuration active by default, if networking issues will disapper we can revisit the configuration of course.

rsynek:
I would rather keep the simple, quick builds separated from any FDB, but limiting the FDB to relevant modules is in my opinion a must, since each of the downstream repos takes ~1h to build.

I didn't know the situation with Jenkins was THAT bad. In that case, it makes sense to use the Build Chain by default and maybe open a discussion about a paid GitHub plan.

ge0ffrey:
@danielezonca Radovan is our CI expert, I defer this decision to him.

triceo:
I'm closing this as I believe we already merged the same thing in another PR.
Feel free to reopen @danielezonca if that is not the case.

PR #1334, opened 2021-05-20
PLANNER-2431 Register custom classes in SolverConfig for Quarkus Native Build

Christopher-Chianelli:
Should fix https://stackoverflow.com/questions/67618863/optaplanner-custom-phase-command-fails-in-native-build-cant-find-generated-giz/67620375#67620375

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/AbstractConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/AbstractConfig.java
@@ -16,7 +16,6 @@
 
 package org.optaplanner.core.config;
 
-import java.util.Set;
 import java.util.function.Consumer;
 
 import javax.xml.bind.annotation.XmlAccessType;
@@ -55,11 +54,12 @@ public abstract class AbstractConfig<Config_ extends AbstractConfig<Config_>> {
     public abstract Config_ copyConfig();
 
     /**
-     * Call the class visitor on each (possibly null) referenced class of this config (including those referenced by children
-     * config).
+     * Call the class visitor on each (possibly null) Class instance provided to this config by the user
+     * (including those provided in child configs).
      * Required to create the bean factory in Quarkus.
      *
-     * @return Set of classes referenced by this config
+     * @param classVisitor The visitor of classes, never null. Can accept null instances
+     *        of Class
      */
     public abstract void visitReferencedClasses(Consumer<Class<?>> classVisitor);
 
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/AbstractConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/AbstractConfig.java
@@ -16,6 +16,9 @@
 
 package org.optaplanner.core.config;
 
+import java.util.Set;
+import java.util.function.Consumer;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 
@@ -50,6 +53,15 @@ public abstract class AbstractConfig<Config_ extends AbstractConfig<Config_>> {
      * @return new instance
      */
     public abstract Config_ copyConfig();
+
+    /**
+     * Call the class visitor on each (possibly null) referenced class of this config (including those referenced by children
+     * config).
+     * Required to create the bean factory in Quarkus.
+     *
+     * @return Set of classes referenced by this config
+     */
+    public abstract void visitReferencedClasses(Consumer<Class<?>> classVisitor);
 
     @Override
     public String toString() {
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/LocalSearchAcceptorConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/LocalSearchAcceptorConfig.java
@@ -17,6 +17,7 @@
 package org.optaplanner.core.config.localsearch.decider.acceptor;
 
 import java.util.List;
+import java.util.function.Consumer;
 
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlType;
@@ -378,6 +379,11 @@ public class LocalSearchAcceptorConfig extends AbstractConfig<LocalSearchAccepto
     @Override
     public LocalSearchAcceptorConfig copyConfig() {
         return new LocalSearchAcceptorConfig().inherit(this);
+    }
+
+    @Override
+    public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
+        // No referenced classes
     }
 
 }
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
@@ -18,6 +18,7 @@ package org.optaplanner.core.config.solver.termination;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.function.Consumer;
 
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlType;
@@ -560,6 +561,14 @@ public class TerminationConfig extends AbstractConfig<TerminationConfig> {
     @Override
     public TerminationConfig copyConfig() {
         return new TerminationConfig().inherit(this);
+    }
+
+    @Override
+    public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
+        classVisitor.accept(terminationClass);
+        if (terminationConfigList != null) {
+            terminationConfigList.forEach(tc -> tc.visitReferencedClasses(classVisitor));
+        }
     }
 
 }
--- a/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
+++ b/optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
@@ -30,7 +30,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.Stack;
 import java.util.stream.Collectors;
 
 import javax.inject.Singleton;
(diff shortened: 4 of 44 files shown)


optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -625,4 +643,231 @@ private boolean shouldIgnoreMember(ClassInfo declaringClass) {
         return declaringClass.name().toString().startsWith(SolutionDescriptor.class.getName());
     }
 
+    private void registerCustomClassesFromSolverConfig(SolverConfig solverConfig, Set<Class<?>> reflectiveClassSet) {
+        ScoreDirectorFactoryConfig scoreDirectorFactoryConfig = solverConfig.getScoreDirectorFactoryConfig();
+        while (scoreDirectorFactoryConfig != null) {
+            registerClassForReflection(scoreDirectorFactoryConfig.getEasyScoreCalculatorClass(), reflectiveClassSet);
+            registerClassForReflection(scoreDirectorFactoryConfig.getConstraintProviderClass(), reflectiveClassSet);
+            registerClassForReflection(scoreDirectorFactoryConfig.getIncrementalScoreCalculatorClass(), reflectiveClassSet);
+            scoreDirectorFactoryConfig = scoreDirectorFactoryConfig.getAssertionScoreDirectorFactory();
```
ge0ffrey, line 652: I believe there's a fail fast if a non-null getAssertionScoreDirectorFactory has itself a non-null getAssertionScoreDirectorFactory, so this can just be an if instead of a while.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -625,4 +643,231 @@ private boolean shouldIgnoreMember(ClassInfo declaringClass) {
...
+    }
+
+    private void registerClassForReflection(Class<?> clazz, Set<Class<?>> reflectiveClassSet) {
+        if (clazz != null) {
+            reflectiveClassSet.add(clazz);
+        }
+    }
+
+    private void registerPhaseConfigForReflection(PhaseConfig<?> phaseConfig, Set<Class<?>> reflectiveClassSet) {
+        if (phaseConfig == null) {
```
ge0ffrey, line 672: getPhaseConfigList can not contain null elements

ge0ffrey: But I doubt that that validation triggers before this code is run, so this is valid.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -625,4 +643,231 @@ private boolean shouldIgnoreMember(ClassInfo declaringClass) {
...
+        } else if (phaseConfig instanceof PartitionedSearchPhaseConfig) {
+            PartitionedSearchPhaseConfig psPhaseConfig = (PartitionedSearchPhaseConfig) phaseConfig;
+            registerClassForReflection(psPhaseConfig.getSolutionPartitionerClass(), reflectiveClassSet);
+            if (psPhaseConfig.getPhaseConfigList() != null) {
+                psPhaseConfig.getPhaseConfigList().forEach(pc -> registerPhaseConfigForReflection(pc, reflectiveClassSet));
+            }
+        } else if (phaseConfig instanceof NoChangePhaseConfig) {
+            // Need to do nothing, since termination config is registered above
+        } else {
+            throw new IllegalStateException("Unhandled phase config class (" + phaseConfig.getClass() + ").");
```
ge0ffrey, line 704:
```suggestion
            throw new IllegalStateException("Unsupported phase config class (" + phaseConfig.getClass() + ").");
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -625,4 +643,231 @@ private boolean shouldIgnoreMember(ClassInfo declaringClass) {
...
+        }
+    }
+
+    private void registerPhaseConfigForReflection(PhaseConfig<?> phaseConfig, Set<Class<?>> reflectiveClassSet) {
+        if (phaseConfig == null) {
+            return;
+        }
+
+        registerTerminationConfig(phaseConfig.getTerminationConfig(), reflectiveClassSet);
+        if (phaseConfig instanceof CustomPhaseConfig) {
```
ge0ffrey, line 677: Nitpick: the order of the if else instanceof checks, should be the same as PhaseConfig's @XmlSeeAlso,
that way it's easier to spot if one is missing.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -625,4 +643,231 @@ private boolean shouldIgnoreMember(ClassInfo declaringClass) {
...
+            return;
+        }
+
+        registerClassForReflection(moveSelectorConfig.getFilterClass(), reflectiveClassSet);
+        registerClassForReflection(moveSelectorConfig.getSorterClass(), reflectiveClassSet);
+        registerClassForReflection(moveSelectorConfig.getSorterComparatorClass(), reflectiveClassSet);
+        registerClassForReflection(moveSelectorConfig.getSorterWeightFactoryClass(), reflectiveClassSet);
+        registerClassForReflection(moveSelectorConfig.getProbabilityWeightFactoryClass(), reflectiveClassSet);
+
+        if (moveSelectorConfig instanceof AbstractPillarMoveSelectorConfig) {
```
ge0ffrey, line 719: Ok, this is code in optaplanner-quarkus is getting know way too much from MoveSelectorConfig (it's internal structure etc) than is good for it. When MoveSelectorConfig refactors, this code needs to refactor along. When a new MoveSelectorConfig is create, this code needs to add it too (and it's easy to forget it).

What if we add a method on MoveSelectorConfig called `retrieveCustomClassesForReflection()`? In fact, we should add it on PhaseConfig. No, on SolverConfig which then calls it as needed down

triceo (changes requested):
I agree with Geoffrey's comments here.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
Christopher-Chianelli: Created method `void visitReferencedClasses(Consumer<Class<?>> classVisitor)` in `AbstractConfig`. Decided to make it a visitor since it make the code cleaner (do not need to do null checks on classes)

optaplanner-core/src/main/java/org/optaplanner/core/config/AbstractConfig.java
```diff
@@ -51,6 +54,15 @@
...
     public abstract Config_ copyConfig();
 
+    /**
+     * Call the class visitor on each (possibly null) referenced class of this config (including those referenced by children
+     * config).
+     * Required to create the bean factory in Quarkus.
+     *
+     * @return Set of classes referenced by this config
+     */
+    public abstract void visitReferencedClasses(Consumer<Class<?>> classVisitor);
```
triceo, line 64: Nice solution to the problem!
I'd be a bit more descriptive in the Javadoc, though - what does it mean for a class to be "referenced"? Say something to the tune of "each Class instance provided to this config by the user".

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/LocalSearchAcceptorConfig.java
```diff
@@ -380,4 +381,9 @@ public LocalSearchAcceptorConfig copyConfig() {
         return new LocalSearchAcceptorConfig().inherit(this);
     }
 
+    @Override
+    public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
+        // No referenced classes
+    }
```
triceo, line 387: Maybe we provide a default empty impl to replace the abstract method? Only people who need to override would override, removing the need for these empty overrides.

But thinking about it again, maybe this way is better. Author of each new config will be forced to think about implementing this method, and that's probably a good thing.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
```diff
@@ -562,4 +563,12 @@ public TerminationConfig copyConfig() {
         return new TerminationConfig().inherit(this);
     }
 
+    @Override
+    public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
+        classVisitor.accept(terminationClass);
+        if (terminationConfigList != null) {
+            terminationConfigList.forEach(tc -> tc.visitReferencedClasses(classVisitor));
+        }
+    }
```
triceo, line 572: The more I think about this API, the more I come to the conclusion that this method should be hidden from users.
Even though `core.config` technically isn't public API, we still treat it like one, as people will be using it. In my opinion, they should not be able to access this method.

triceo (approved):
Very nice! Some more things to consider, but otherwise LGTM.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
Christopher-Chianelli: It does need to be accessible from optaplanner-quarkus deployment though. Making it protected will not work (different classloader + different package => not accessible). The only way I see to make this possible is to make this protected (cannot be private since it need to be overriden), and use reflection to access it from Quarkus.

optaplanner-core/src/main/java/org/optaplanner/core/config/localsearch/decider/acceptor/LocalSearchAcceptorConfig.java
Christopher-Chianelli: That the idea (and also why I didn't create a default version in PhaseConfig; I want to force subclasses to implement these methods so writers known they must add the classes to reduce probability of bugs).

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
Christopher-Chianelli: It also need to be accessible from benchmark since some benchmark config override AbstractConfig

triceo: Yeah, let's keep it as is. Anything else is too much hassle for too little benefit.

triceo:
@Christopher-Chianelli Please file a JIRA for this one, so that it's properly tracked. I'll merge later.

Christopher-Chianelli:
JIRA: https://issues.redhat.com/browse/PLANNER-2431

PR #1337, opened 2021-05-21
Update turtle test trigger as dsl deprecated

radtriste:
Due to migration to CSB, with Jenkins version, direct triggers are deprecated and need to be replaced

depends on https://github.com/kiegroup/kogito-pipelines/pull/172

### DIFF
--- a/.jenkins/dsl/jobs.groovy
+++ b/.jenkins/dsl/jobs.groovy
@@ -217,8 +217,14 @@ void setupOptaPlannerTurtleTestsJob(String jobFolder) {
     def jobParams = getJobParams('optaplanner-turtle-tests', jobFolder, 'Jenkinsfile.turtle',
             'Run OptaPlanner turtle tests on a weekly basis.')
     KogitoJobTemplate.createPipelineJob(this, jobParams).with {
-        triggers {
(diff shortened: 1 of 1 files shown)


.jenkins/dsl/jobs.groovy
```diff
@@ -217,8 +217,14 @@ void setupOptaPlannerTurtleTestsJob(String jobFolder) {
     def jobParams = getJobParams('optaplanner-turtle-tests', jobFolder, 'Jenkinsfile.turtle',
             'Run OptaPlanner turtle tests on a weekly basis.')
     KogitoJobTemplate.createPipelineJob(this, jobParams).with {
-        triggers {
```
rsynek, line 220: I trust it's ok. Is there a new DSL reference I could check? All I found seems to refer to the previous syntax:
https://www.jenkins.io/doc/book/pipeline/syntax/#triggers

rsynek (approved):

.jenkins/dsl/jobs.groovy
rsynek: Thanks for the link, I was not aware of this "api-viewer".

PR #1351, opened 2021-06-01
Change default domain access type to GIZMO in Quarkus

Christopher-Chianelli:
GIZMO has been fixed for a while. The only commit content missing in master from the GIZMO_MEMBER PR is the content that adds GIZMO_MEMBER (And the GIZMO_MEMBER PR works on all of optaplanner-examples => GIZMO on master would work on all of optaplanner-examples if they were written in Quarkus). This PR change the default back to GIZMO.

Side note: `optaplanner-core/src/test/resources/org/optaplanner/core/config/solver/testSolverConfigWithoutNamespace.xml` appears to be an invalid SolverConfig; it references `TestdataAnnotatedExtendedEntity`, which references a valueRangeProvider not in the solution class `TestdataSolution`. For GIZMO, this fails the build (as it need to build a solution descriptor). In REFLECTION, the build won't fail since the SolverDescriptor is not created until a Solver is (and in the test, we don't inject SolverFactory, so it isn't ever created).

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigTest.java
+++ b/optaplanner-core/src/test/java/org/optaplanner/core/config/solver/SolverConfigTest.java
@@ -54,6 +54,7 @@ import org.optaplanner.core.impl.testdata.domain.TestdataEntity;
 import org.optaplanner.core.impl.testdata.domain.TestdataSolution;
 import org.optaplanner.core.impl.testdata.domain.TestdataValue;
 import org.optaplanner.core.impl.testdata.domain.extended.TestdataAnnotatedExtendedEntity;
+import org.optaplanner.core.impl.testdata.domain.extended.TestdataAnnotatedExtendedSolution;
 
 public class SolverConfigTest {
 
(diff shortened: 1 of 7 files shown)


triceo:
I believe that we need to fix either the solver config, or the solution. The solver config probably isn't meant to be wrong, and fixing it may be simple. I'm not sure why we would be using `TestdataAnnotatedExtendedEntity` with `TestdataSolution`. We should be using `TestdataAnnotatedExtendedSolution` for that.

Christopher-Chianelli:
From what I gather, the solverConfig is basically a kitchen sink solverConfig (doing basically every possible combination) and is filled with Dummy classes that cannot be created as they are abstract. I can try changing it to `TestdataAnnotatedExtendedSolution` solution, but the Dummy classes may cause an issue since they can't be created.

Christopher-Chianelli:
@triceo can confirm using `TestdataAnnotatedExtendedSolution` fixes the issue on the Quarkus end (although it might cause a problem if `testSolverConfigWithoutNamespace.xml` is used elsewhere for some reason and it expects the Solution class to be TestdataSolution)

triceo (approved):

PR #1352, opened 2021-06-02
PLANNER-2450 Optimize ConsecutiveSetTree

Christopher-Chianelli:
This should make it faster in series; didn't do benchmark yet

### Checklist

<details>
<summary>
</summary>

</details>


triceo (commented):
I like this. I'll see if I can get to it tomorrow to measure the impact.
While you're touching this, please make the interfaces happen - we need them for the types that reach end users, so Sequence, Break, ConsecutiveData, ...

Christopher-Chianelli:
@triceo I got a NPE on ShiftDate on `consecutiveWorkingWeekends` when benchmarking long_hint01.xml and long_hint02.xml. It doesn't happen on master. It appears to happen after the construction heuristic is finish (which would mean all the shifts are assigned), so I have no idea what is happening

Christopher-Chianelli:
(solving long_hint01.xml OUTSIDE of benchmarker doesn't cause a NPE)

Christopher-Chianelli:
(solving long_hint02.xml OUTSIDE of benchmarker doesn't cause a NPE, so consider me doubly bewildered)

Christopher-Chianelli:
(can confirm there is a score corruption; investigating)

optaplanner-examples/src/test/java/org/optaplanner/examples/common/experimental/ConsecutiveSetTreeTest.java
```diff
@@ -286,6 +287,46 @@ public void testRandomSequences() {
         }
     }
 
+    @Test
+    public void testRandomSequencesWithDuplicates() {
```
triceo, line 291: I think that tests like these should be parameterized. This is not one test - this is a 1000 different tests, each with 2 assertions.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
```diff
@@ -30,14 +30,19 @@
...
+import org.optaplanner.examples.common.experimental.impl.IntervalTree;
 
 /**
  * A collection of experimental constraint collectors
  * subject to change in future versions.
  */
 public class ExperimentalConstraintCollectors {
     /**
-     * Creates a constraint collector that returns {@link ConsecutiveData} about the first
+     * Creates a constraint collector that returns {@link ConsecutiveDataImpl} about the first
```
triceo, line 45: We don't want this. Public types must not be `Impl`s.

triceo: Also, it seems this returns `ConsecutiveInfo`, not `ConsecutiveData`.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
```diff
@@ -158,7 +166,7 @@
      * @param <C> type of the fact endpoints
      * @return never null
      */
-    public static <A, C extends Comparable<C>> UniConstraintCollector<A, IntervalTree<A, C>, ConsecutiveIntervalData<A, C>>
+    public static <A, C extends Comparable<C>> UniConstraintCollector<A, IntervalTree<A, C>, ConsecutiveIntervalDataImpl<A, C>>
             consecutiveIntervals(Function<A, C> startMap, Function<A, C> endMap) {
```
triceo, line 170: Returning `Impl`s defeats the point of separating the API. Cast the return if you need to.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface Break<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>> {
```
triceo, line 19: These public types deserve explanatory javadocs on the type and on methods.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/ConsecutiveInfo.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface ConsecutiveInfo<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>> {
```
triceo, line 19: These public types deserve explanatory javadocs on the type and on methods.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Sequence.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface Sequence<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>> {
+    /**
+     * Returns the first item in the sequence
+     * 
+     * @return Never null, the first item in the sequence
```
triceo, line 23: I hate this nitpicking, but this is not a sentence, so it should not start with a capital letter.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Sequence.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface Sequence<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>> {
+    /**
+     * Returns the first item in the sequence
```
triceo, line 21: Probably unnecessary, since the @return says the same thing. If you keep it, this is a sentence, and therefore it ends with a full stop.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Sequence.java
triceo: Both these comments apply to the entire interface.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/BreakImpl.java
```diff
@@ -14,27 +14,33 @@
  * limitations under the License.
  */
 
-package org.optaplanner.examples.common.experimental;
+package org.optaplanner.examples.common.experimental.impl;
 
-public class Break<ValueType_, DifferenceType_> {
+import org.optaplanner.examples.common.experimental.api.Break;
+
+public class BreakImpl<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>>
```
triceo, line 21: Any chance we could hide this, not make it public?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveDataImpl.java
```diff
@@ -0,0 +1,53 @@
...
+package org.optaplanner.examples.common.experimental.impl;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.optaplanner.examples.common.experimental.api.Break;
+import org.optaplanner.examples.common.experimental.api.ConsecutiveInfo;
+import org.optaplanner.examples.common.experimental.api.Sequence;
+
+public class ConsecutiveDataImpl<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>> implements
```
triceo, line 26: Dtto. I'd try to hide.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalDataImpl.java
```diff
@@ -24,12 +24,12 @@
 
 import org.apache.commons.lang3.ObjectUtils;
 
-public class ConsecutiveIntervalData<IntervalValue_, PointValue_ extends Comparable<PointValue_>> {
+public class ConsecutiveIntervalDataImpl<IntervalValue_, PointValue_ extends Comparable<PointValue_>> {
```
triceo, line 27: Dtto, let's try to hide.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
```diff
@@ -39,24 +41,28 @@
 public class ConsecutiveSetTree<ValueType_, PointType_ extends Comparable<PointType_>, DifferenceType_ extends Comparable<DifferenceType_>> {
```
triceo, line 41: Dtto, let's try to hide.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public ValueType_ getFirstItem() {
+        return firstItem;
+    }
+
+    @Override
+    public ValueType_ getLastItem() {
+        return lastItem;
+    }
```
triceo, line 55: Good that we have this. Now let's use it consistently. :-) I noticed places in the code where it's not used.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+package org.optaplanner.examples.common.experimental.impl;
+
+import java.util.NavigableSet;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.optaplanner.examples.common.experimental.api.Sequence;
+
+public class SequenceImpl<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>>
```
triceo, line 25: Let's try to hide this.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        implements Sequence<ValueType_, DifferenceType_> {
+    private NavigableSet<ValueType_> consecutiveItemsSet;
+    private ValueType_ firstItem;
+    private ValueType_ lastItem;
+    private final ConsecutiveSetTree<ValueType_, ?, DifferenceType_> sourceTree;
+
+    protected SequenceImpl(ConsecutiveSetTree<ValueType_, ?, DifferenceType_> sourceTree, ValueType_ item) {
```
triceo, line 32: If the type is hidden, I don't see the need to hide the methods. 

I generally try to hide the type, and make public methods public. This is an API within an API - those who have access to the type should be able to use the type, except for methods you want to keep hidden even from them.

optaplanner-examples/src/test/java/org/optaplanner/examples/common/experimental/impl/IntervalTreeTest.java
```diff
@@ -252,8 +252,9 @@ public void testRandomIntervals() {
             int from = random.nextInt(50);
             int to = from + random.nextInt(50);
             Interval interval = intervalToInstanceMap.computeIfAbsent(new Interval(from, to), Function.identity());
-            org.optaplanner.examples.common.experimental.Interval<Interval, Integer> treeInterval =
-                    new org.optaplanner.examples.common.experimental.Interval<>(interval, Interval::getStart, Interval::getEnd);
+            org.optaplanner.examples.common.experimental.impl.Interval<Interval, Integer> treeInterval =
+                    new org.optaplanner.examples.common.experimental.impl.Interval<>(interval, Interval::getStart,
```
triceo, line 256: Is there any other `Interval`? Why are these FQNs necessary?

triceo (commented):
Much better! Some more comments. 

Let's also address the difference between CS and DRL, unless it turns out it's very hard to do. We still want this to serve as an example - readability comes first.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,329 @@
...
+
+    private boolean isFirstSuccessorOfSecond(PointType_ first, ValueType_ firstValue, PointType_ second,
+            ValueType_ secondValue) {
+        DifferenceType_ difference = differenceFunction.apply(second, first);
+        int differenceComparedTo0 = difference.compareTo(zeroDifference);
+        if (differenceComparedTo0 > 0) {
+            return difference.compareTo(maxDifference) <= 0;
+        } else {
+            return differenceComparedTo0 == 0 && System.identityHashCode(secondValue) < System.identityHashCode(firstValue);
+        }
```
triceo, line 125: Important: all comparisons *must* happen in a shared piece of code. Especially because of this identity trick.
This code all be put in one place, and be well documented there - so that if/when we're touching this in the future, we understand the trick and can rely on the fact that this trick is not exposed anywhere else.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
```diff
@@ -0,0 +1,329 @@
...
+            return true;
+        }
+        itemSet.add(item);
+        ValueType_ firstBeforeItem = startItemToSequence.floorKey(item);
+        PointType_ itemIndex = indexFunction.apply(item);
+        if (firstBeforeItem != null) {
+            ValueType_ endOfBeforeSequenceItem = getEndItem(firstBeforeItem);
+            PointType_ endOfBeforeSequenceIndex = indexFunction.apply(endOfBeforeSequenceItem);
+            if (itemIndex.compareTo(endOfBeforeSequenceIndex) < 0 || (itemIndex.compareTo(endOfBeforeSequenceIndex) == 0
+                    && System.identityHashCode(endOfBeforeSequenceItem) > System.identityHashCode(item))) {
```
triceo, line 181: Dtto. on the identity trick.

triceo (changes requested):
Chris, I've pushed into your branch; minor perf improvements, nothing to worry about.
However, as I was reviewing the code in IDE, I noticed a very important problem. See my comments on identity.

Personally, I think that the comparison operations should be externalized into a `Comparator` and reused that way.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public DifferenceType_ getLength() {
+        return sourceTree.getSequenceLength(this);
+    }
+
+    protected boolean isEmpty() {
+        return consecutiveItemsSet.isEmpty();
+    }
```
triceo, line 74: This is really bothering me. In the big scheme of things, this is irrelevant - but if you take a look at how much this takes out of all collector operations, this is about ~10 %. I really do not understand how this could be costing anything.

Anyway... we know the first item, we know the last item. Do we need to delegate to the collection if we know that we have a first and a last item?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
Christopher-Chianelli: ConsecutiveInfo is the api type of `ConsecutiveDataImpl`; javadoc got stale

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
Christopher-Chianelli: The interval data types I'll do later; they are unused in the points consecutive collector

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    // Called when start or end are removed; since the set
+    // is backed by the underlying set that already have the
+    // point removed, consecutiveItemsSet.first()/consecutiveItemsSet.last()
+    // change to reflect that, and we update the bounds
+    protected void updateEndpoints() {
+        firstItem = consecutiveItemsSet.first();
+        lastItem = consecutiveItemsSet.last();
+        consecutiveItemsSet = consecutiveItemsSet.subSet(firstItem, true, lastItem, true);
+    }
```
triceo, line 94: How does this do anything? It first takes first/last of the set. And then creates a subset of that set, with the exact same start and end. Therefore, the subset should be the same set as original.

Also, I believe this throws if the set is empty. (Because first/last throws.)

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
Christopher-Chianelli: Can't be put into a Comparator; it uses difference types, and the items aren't available within the type; did it within a generic function that takes four parameters

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
Christopher-Chianelli: inlined and changed to getFirstItem() == getLastItem()

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
```diff
@@ -16,10 +16,34 @@
...
+     * @return never null, the item this break is directly after
+     */
     ValueType_ getAfterItem();
 
+    /**
+     * Return the item this break is directly before. For the
+     * break between 6 and 10, this will return 10.
+     * @return never null,  the item this break is directly before
+     */
     ValueType_ getBeforeItem();
```
triceo: Hmm, interesting. I was thinking about the API in the exact oposite way - getAfterItem() returns 10, because it is after the break. Maybe we need a better name for these two methods, one that is not ambiguous?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
```diff
@@ -252,8 +255,8 @@ public boolean remove(ValueType_ item) {
         ValueType_ endItem = bag.getLastItem();
         itemSet.remove(item);
 
-        // Count of item in bag is 0
-        if (bag.isEmpty()) {
+        // Bag is empty if first item = last item
+        if (bag.getFirstItem() == bag.getLastItem()) {
```
triceo, line 259: Strictly speaking - if the bag's first and last item is the same item, the bag has 1 item inside.
If the first and last item are null, then I could see how the bag is empty.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
```diff
@@ -56,62 +58,62 @@ public ValueType_ getLastItem() {
...
-    protected void updateEndpoints() {
-        firstItem = consecutiveItemsSet.first();
-        lastItem = consecutiveItemsSet.last();
-        consecutiveItemsSet = consecutiveItemsSet.subSet(firstItem, true, lastItem, true);
+    // Called when start or end are removed; length
+    // need to be invalidated
+    protected void invalidate() {
+        length = null;
+        items = null;
     }
```
triceo: We still have this problem that changing state of the sequence requires two operations.
Shouldn't we be able to invalidate automatically when start/end changes?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
Christopher-Chianelli: This is called in the remove method, after an item is removed. 1 - 1 = 0.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/SequenceImpl.java
Christopher-Chianelli: set start and set end do call invalidate; I think I need to remove this extra call from remove

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
Christopher-Chianelli: `getStart()` and `getEnd()`?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
```diff
@@ -0,0 +1,52 @@
...
+     * @return never null, the item this break is directly before
+     */
+    ValueType_ getBeforeItem();
+
+    /**
+     * Return the length of the break, which is the difference
+     * between {@link #getBeforeItem()} and {@link #getAfterItem()}. For the
+     * break between 6 and 10, this will return 4.
+     * 
+     * @return never null, the length of this break.
```
triceo, line 49:
```suggestion
     * @return never null, the length of this break
```

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
triceo: Well, it's technically not start or end. It's the thing before start, and thing after end. So the original naming was correct, but ambiguous.

getSequenceBeforeStart()/getSequenceAfterEnd() ?
(Note I replaced "item" with "sequence". It may be too specific, and we may need a more generic word, but I don't like that "item" is so generic that it doesn't say anything.)

Christopher-Chianelli: Then we would need to change the return type; I would expect `getSequenceBeforeStart()` be a sequence not an item. Start and End are still applicable (the last item of the sequence before is the start of the break, the first item of the sequence after is the end of the break).

triceo: Good point! I did not notice that this actually referenced items, not sequences.
Yes, I think that this should point to preceding and succeeding sequence, not the actual item.
The items are accessible from the sequence. From the item, you can not get to the sequence.

triceo:
Please change `Break` to have `getPreviousSequence()` and `getNextSequence()`.
After that, we're done with this PR and I'm merging it, @Christopher-Chianelli .

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param <ValueType_> The type of value in the sequence
+ * @param <DifferenceType_> The type of difference between values in the sequence
+ */
+public interface Break<ValueType_, DifferenceType_ extends Comparable<DifferenceType_>> {
+    /**
+     * Return the end of the sequence before this break. For the
+     * break between 6 and 10, this will return 6.
+     *
+     * @return never null, the item this break is directly after
```
triceo, line 32:
```suggestion
     * @return never null, the sequence leading directly into this
```

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/Break.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    Sequence<ValueType_, DifferenceType_> getPreviousSequence();
+
+    /**
+     * Return the start of the sequence after this break. For the
+     * break between 6 and 10, this will return 10.
+     *
+     * @return never null, the item this break is directly before
```
triceo, line 40:
```suggestion
    /**
     * @return never null, the sequence immediately following this
```

PR #1354, opened 2021-06-03
New diagram: vaccinationSchedulingPartitioning

ge0ffrey:


yurloc (approved):
The image looks good. Why is it not referenced from anywhere?

PR #1370, opened 2021-06-09
BAQE-1904 - Change revapi to check against 7.52.0.Final

MarianMacik:
**JIRA**: [BAQE-1904](https://issues.redhat.com/browse/BAQE-1904)

depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1665

### DIFF
--- a/optaplanner-benchmark/src/build/revapi-config.json
+++ b/optaplanner-benchmark/src/build/revapi-config.json
@@ -25,7 +25,7 @@
   },
   "ignores": {
     "revapi": {
-      "_comment": "Changes between 7.48.0.Final and the current branch. These changes are desired and thus ignored.",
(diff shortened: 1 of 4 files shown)


MarianMacik:
Jenkins retest this

triceo (approved):

MarianMacik:
@triceo Next time please don't merge an individual revapi PR because they have to be merged at once. The droolsjbpm-build-bootstrap (kie-parent) contains the version which the check is run against and at the same time other PRs contain possible ignores files which are related to the checked version, so it is a bi-directional dependency. Now it won't matter because there were simply no changes but in case there were some, OptaPlanner 7.x would be failing as it would be checking empty ignores against the old version. But maybe you realized that and merged this in advance, then it's OK.

PR #1374, opened 2021-06-11
Rename KieBaseExtractor to KieRuntimeBuilderWrapper + prefix gizmo property with gizmo

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -76,10 +76,9 @@
     @XmlElement(name = "scoreDrlFile")
     protected List<File> scoreDrlFileList = null;
     @XmlTransient
-    protected KieBaseExtractor kieBaseExtractor = null;
+    protected KieRuntimeBuilderWrapper gizmoKieRuntimeBuilderWrapper = null;
```
ge0ffrey, line 79: I strongly believe we need to prefix this property with "gizmo", such that users that use code completion on this class  (when using programmatic API instead of XML) don't mistakenly believe that it's a typical property that they can modify to suit their needs.
By prefixing it with "gizmo" they clearly see that it's for gizmo's usage. As for other usages outside gizmo? YAGTNI.

ge0ffrey: I was thinking we might want to make a GizmoConfig, as a property on SolverConfig and put all the gizmo related properties in there. But that would probably be a bad idea, as we want the score director related properties to sit close to the other code for that, so in ScoreDirectorFactoryConfig.
So instead, let's just all prefix them with gizmo (which we already do for the memberaccessor and solutioncloner)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -314,12 +309,7 @@ private void validateNoDroolsAlphaNetworkCompilation() {
                         kieHelper.addResource(new FileSystemResource(scoreDrlFile));
                     }
                 }
-
-                try {
-                    kieBase = kieHelper.build(ExecutableModelProject.class, KieBaseMutabilityOption.DISABLED);
-                } catch (Exception ex) {
```
ge0ffrey, line 320: exception is already caught in the exact same way, see below

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/KieRuntimeBuilderWrapper.java
```diff
@@ -19,15 +19,19 @@
 import org.kie.api.KieBase;
 import org.kie.kogito.rules.KieRuntimeBuilder;
 
-public class KieBaseExtractor {
+/**
+ * Wraps {@link KieRuntimeBuilder} so the dependency on kogito-api is optional.
+ */
+public class KieRuntimeBuilderWrapper {
```
ge0ffrey, line 25: Motivation for rename: We would have used KieRuntimeBuilder directly if it wasn't for the optional dependency requirement, so we should stay as close to that as possible. Other option as KieRuntimeBuilderProxy

Christopher-Chianelli (approved):
LGTM

triceo (approved):

PR #1380, opened 2021-06-15
PLANNER-2445 Introduce a looping phase

triceo:


optaplanner-core/src/test/java/org/optaplanner/core/impl/phase/LoopPhaseTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        DefaultSolver<TestdataSolution> solver =
+                (DefaultSolver<TestdataSolution>) SolverFactory.<TestdataSolution> create(solverConfig)
+                        .buildSolver();
+        TestdataSolution solution =
+                solver.solve(TestdataSolution.generateSolution(1, 1)); // No entities = score of 1.
+        assertThat(solution.getScore()).isEqualTo(SimpleScore.of(4)); // We stopped when we were supposed to.
+
+        PhaseCounter<TestdataSolution> phaseCounter = solver.getSolverScope().getPhaseCounter();
+        assertThat(phaseCounter.getPhasesEnded()).isEqualTo(4); // 1 loop + 3 iterations.
```
triceo, line 65: @ge0ffrey Note: loopPhase counts as a phase. Therefore the phase indexes here are:

- Loop Phase: 0
- Inside Phase A: 1
- Inside Phase B: 2
- Inside Phase A, second loop: 3

optaplanner-core/src/test/resources/org/optaplanner/core/config/solver/testSolverConfigWithoutNamespace.xml
```diff
@@ -156,4 +156,12 @@
...
     </localSearch>
   </partitionedSearch>
+  <loop>
+    <constructionHeuristic />
+    <localSearch />
+    <customPhase />
+    <exhaustiveSearch />
+    <noChangePhase/>
+    <partitionedSearch />
+  </loop>
```
triceo, line 166: @ge0ffrey This is what the solver config looks like. `<loop>` is a phase like any other - can go anywhere where other phases can go, can have terminations, can be part of other phases.

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
```diff
@@ -127,6 +135,8 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
 
     @Override
     public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
+        phaseIndex = phaseCounter.getPhasesStarted();
+        phaseCounter.phaseStarted(phaseScope);
```
triceo, line 139: Because phases can now be reused, the phase index is only assigned at phase start.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -176,13 +177,12 @@ private ExecutorService createThreadPoolExecutor(int partCount) {
         Termination<Solution_> partTermination = new OrCompositeTermination<>(childThreadPlumbingTermination,
                 termination.createChildThreadTermination(solverScope, ChildThreadType.PART_THREAD));
         List<Phase<Solution_>> phaseList = new ArrayList<>(phaseConfigList.size());
-        int partPhaseIndex = 0;
+        PhaseCounter<Solution_> partPhaseCounter = new PhaseCounter<>(); // Phase count independent of the outer solver.
         for (PhaseConfig phaseConfig : phaseConfigList) {
```
triceo, line 181: When you're inside the partitioned search, the phase index starts from 0 again.
I am keeping this behavior.

optaplanner-core/src/test/resources/org/optaplanner/core/config/solver/testSolverConfigWithoutNamespace.xml
rsynek: How does the termination of a loop phase compose with terminations configured on individual phases inside the loop?

optaplanner-core/src/test/java/org/optaplanner/core/impl/phase/LoopPhaseTest.java
rsynek: Should the loop phase show any index, if it's just a wrapper over the individual phases?

triceo: We can do that. We can decide that loop phases don't count. Geoffrey's choice, I don't mind either way.

optaplanner-core/src/test/resources/org/optaplanner/core/config/solver/testSolverConfigWithoutNamespace.xml
triceo: It doesn't. The nested phase needs to terminate according to its own termination, at which point the loop phase termination's evaluated. This is likely a place where we want some real-world examples of things to try and test.

optaplanner-core/src/main/java/org/optaplanner/core/config/phase/loop/LoopPhaseConfig.java
```diff
@@ -0,0 +1,102 @@
...
+        return new LoopPhaseConfig().inherit(this);
+    }
+
+    @Override
+    public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
+        if (getTerminationConfig() != null) {
+            getTerminationConfig().visitReferencedClasses(classVisitor);
+        }
+        if (phaseConfigList != null) {
+            phaseConfigList.forEach(c -> c.visitReferencedClasses(classVisitor));
```
ge0ffrey, line 98: Nitpick: "c" is typically char. Or sometimes class. This makes it easier to read what's going on:
```suggestion
            phaseConfigList.forEach(config -> config.visitReferencedClasses(classVisitor));
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
ge0ffrey: Yes, because the logging with show that they are on another thread. (each log line also includes the thread name) => context

Partition threads each need to have a separate phase counter because they run in parallel. So they all start from zero.

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/PhaseCounter.java
```diff
@@ -0,0 +1,76 @@
...
+    private final AtomicInteger phaseEndedCounter = new AtomicInteger();
+
+    @Override
+    public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
+        this.phaseStartedCounter.incrementAndGet();
+    }
+
+    @Override
+    public void stepStarted(AbstractStepScope<Solution_> stepScope) {
+        // No-op.
```
ge0ffrey, line 44: You can extend PhaseLifecycleListenerAdapter to avoid these (not sure if you want to though)

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/loop/DefaultLoopPhase.java
```diff
@@ -0,0 +1,164 @@
...
+                stepEnded(stepScope);
+                phaseScope.setLastCompletedStepScope(stepScope);
+
+                solverScope.checkYielding();
+                terminated = termination.isPhaseTerminated(phaseScope);
+                if (terminated) { // The termination condition may be met inbetween the nested phases.
+                    break;
+                }
+            }
+        } while (!terminated);
```
ge0ffrey, line 77: All other phases do a normal loop, not a do-while loop. They check the termination condition before looping even.

For example DefaultLocalSearchPhase:
```
        while (!termination.isPhaseTerminated(phaseScope)) {
˘˘˘

File: optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/loop/DefaultLoopPhase.java [hunk 8]
```diff
@@ -0,0 +1,164 @@
...
+        do { // Loop the nested phases until termination is reached.
+            for (AbstractPhase<Solution_> phase : phaseList) {
+                LoopStepScope<Solution_> stepScope = new LoopStepScope<>(phaseScope);
+                stepStarted(stepScope);
+                doStep(solverScope, stepScope, phase);
+                stepEnded(stepScope);
+                phaseScope.setLastCompletedStepScope(stepScope);
+
+                solverScope.checkYielding();
+                terminated = termination.isPhaseTerminated(phaseScope);
```
ge0ffrey, line 72: Will this work in both these scenarios?

```
<ch/>
<loop>
  <termination>timeSpent=100seconds</>
  <localSearch/>
  <ruin/>
</loop>
```

```
<ch/>
<loop>
  <localSearch>
    <termination>timeSpent=100seconds</>
  </localSearch>
  <ruin/>
</loop>
```

ge0ffrey: I suspect B) will misbehave. Here's how it should behave:

run CH,
run LS. stop after 100 seconds (I think vs starting solving time, but that's a termination detail)
run ruin
run LS  stop after 100 seconds (I think vs starting solving time, but that's a termination detail)
run ruin
and so forth (run forever)

triceo: B) Works.
A) Does not. The loop termination is not propagated to the local search, and therefore the local search never works.

Not sure what to do here. Simply push the top-level termination down?

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/PhaseCounter.java
```diff
@@ -0,0 +1,76 @@
...
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
+import org.optaplanner.core.impl.solver.scope.SolverScope;
+
+/**
+ * Keeps track of solver phase lifecycle, so that it can provide phase indexes for upcoming phases and phase total when
+ * the solver ends.
+ *
+ * @param <Solution_> generic type of the Solution
+ */
+public final class PhaseCounter<Solution_> implements PhaseLifecycleListener<Solution_> {
```
ge0ffrey, line 32: How about SolverScope.phaseCount instead?

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/PhaseCounter.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private final AtomicInteger phaseStartedCounter = new AtomicInteger();
+    private final AtomicInteger phaseEndedCounter = new AtomicInteger();
```
ge0ffrey, line 35: This is single-threaded code, can just be a int

triceo:
Will be reintroduced later.

PR #1417, opened 2021-07-12
PLANNER-2478: Consecutive Intervals Constraint Collector: Javadoc + Seperation of API, Breaks, Performance Improvement

Christopher-Chianelli:
Separated the API and added an untested Break support; will undraft once tests + javadocs are added

### JIRA

https://issues.redhat.com/browse/PLANNER-2478

### Checklist

<details>
<summary>
</summary>

</details>


optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
```diff
@@ -162,15 +162,17 @@
      * @param startMap Maps the fact to its start
      * @param endMap Maps the fact to its end
      * @param <A> type of the first mapped fact
-     * @param <C> type of the fact endpoints
+     * @param <PointType_> type of the fact endpoints
```
triceo, line 165: `differenceFunction` is undocumented. Focus on the arguments to that function. Is the first one expected to be smaller than the other? What should happen if that's not the case?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
```diff
@@ -162,15 +162,17 @@
...
      * @return never null
      */
-    public static <A, C extends Comparable<C>> UniConstraintCollector<A, IntervalTree<A, C>, ConsecutiveIntervalDataImpl<A, C>>
-            consecutiveIntervals(Function<A, C> startMap, Function<A, C> endMap) {
+    public static <A, PointType_ extends Comparable<PointType_>, DifferenceType_ extends Comparable<DifferenceType_>>
+            UniConstraintCollector<A, IntervalTree<A, PointType_, DifferenceType_>, ConsecutiveIntervalInfo<A, PointType_, DifferenceType_>>
+            consecutiveIntervals(Function<A, PointType_> startMap, Function<A, PointType_> endMap,
+                    BiFunction<PointType_, PointType_, DifferenceType_> differenceFunction) {
```
triceo, line 171: I understand why the `differenceFunction` exists. That said, I'm thinking that perhaps we can have an overload where `PointType_` is guaranteed to be a number, and therefore the difference function is not necessary?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/ConsecutiveIntervalInfo.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface ConsecutiveIntervalInfo<IntervalType_, PointType_ extends Comparable<PointType_>, DifferenceType_ extends Comparable<DifferenceType_>> {
```
triceo, line 19: I'm thinking maybe we use the same interface for sequences and clusters? With different implementations for each.
It would bring a nice conceptual integrity to the design, I think.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/IntervalBreak.java
```diff
@@ -0,0 +1,65 @@
...
+package org.optaplanner.examples.common.experimental.api;
+
+/**
+ * An IntervalBreak is a gap between two consecutive interval clusters. For instance,
+ * the list [(1,3),(2,4),(3,5),(7,8)] has a break of length 2 between 5 and 7.
+ *
+ * @param <IntervalType_> The type of value in the sequence
+ * @param <DifferenceType_> The type of difference between values in the sequence
+ */
+public interface IntervalBreak<IntervalType_, PointType_ extends Comparable<PointType_>, DifferenceType_ extends Comparable<DifferenceType_>> {
```
triceo, line 26: Dtto. Maybe we have the same break interface for both the sequence and the cluster?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/IntervalCluster.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface IntervalCluster<IntervalType_, PointType_ extends Comparable<PointType_>, DifferenceType_ extends Comparable<DifferenceType_>>
```
triceo, line 19: Dtto. Maybe we have the same interface for both the sequence and the cluster?
We definitely need a list of items in the cluster.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,204 @@
...
+                clusterStartSplitPointToNextBreak.put(intervalCluster.getStartSplitPoint(), finalBreak);
+            }
+        }
+        if (oldStart.compareTo(intervalCluster.getStartSplitPoint()) > 0) {
+            clusterStartSplitPointToCluster.remove(oldStart);
+            clusterStartSplitPointToCluster.put(intervalCluster.getStartSplitPoint(), intervalCluster);
+        }
+    }
+
+    protected void removalInterval(Interval<IntervalType_, PointType_> interval) {
```
triceo, line 148: Shouldn't this be `removeInterval(...)`?

triceo (commented):
Good first step!

- Missing the Javadoc updates for ranges.
- Missing test coverage for breaks.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/ExperimentalConstraintCollectors.java
Christopher-Chianelli: I thinking of two overloads: one for numbers, one for temporal (localdatetime etc.).

triceo:
If you want an example to play with, I can give you my cheaptime work, @Christopher-Chianelli. The constraint provider is done, but does not yet match DRL 1-to-1.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/IntervalBreak.java
Christopher-Chianelli: My main issue with this proposal is the naming of the `nextSequence`/`previousSequence` and `nextCluster/previousCluster`. Strictly speaking, Sequences are a specialized version of Clusters where each interval end point is a constant different away from the start point, but they aren't "viewed" as intervals by the users, but rather points. If they both share the same interface, they need to have the same name.

triceo: That is a good point - but wouldn't that be visible from the generic type? In one instance, `getNext()` would return `Sequence`, in another it would return `Cluster`.

That said, I don't think this is an absolute necessity.

Christopher-Chianelli:
@triceo I'm fine for now; I still need to do tests to ensure breaks actually work. The example would more serve as guidance for the API, but the core logic of breaks doesn't depend on the API.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/IntervalBreak.java
Christopher-Chianelli: `getNext()` for me, would mean "get next break" not "get the thing after the break"

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+
+import org.apache.commons.lang3.ObjectUtils;
+import org.optaplanner.examples.common.experimental.api.ConsecutiveIntervalInfo;
+import org.optaplanner.examples.common.experimental.api.IntervalBreak;
+import org.optaplanner.examples.common.experimental.api.IntervalCluster;
+
+public class ConsecutiveIntervalInfoImpl<IntervalType_, PointType_ extends Comparable<PointType_>, DifferenceType_ extends Comparable<DifferenceType_>>
+        implements
+        ConsecutiveIntervalInfo<IntervalType_, PointType_, DifferenceType_> {
+    private final NavigableMap<IntervalSplitPoint<IntervalType_, PointType_>, IntervalClusterImpl<IntervalType_, PointType_, DifferenceType_>> clusterStartSplitPointToCluster;
```
triceo, line 34: The generics are definitely getting to an insane level here. That said, I don't think there's much we can do. Perhaps we remove the `Type_` suffixes, because it is obvious that they're going to be types?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        NavigableMap<IntervalSplitPoint<IntervalType_, PointType_>, IntervalClusterImpl<IntervalType_, PointType_, DifferenceType_>> intersectedIntervalClusterMap =
+                clusterStartSplitPointToCluster.subMap(
+                        ObjectUtils.defaultIfNull(clusterStartSplitPointToCluster.floorKey(interval.getStartSplitPoint()),
+                                interval.getStartSplitPoint()),
+                        true, interval.getEndSplitPoint(), true);
+
+        // Case: the interval cluster before this interval does not intersect this interval
+        if (!intersectedIntervalClusterMap.isEmpty()
+                && intersectedIntervalClusterMap.get(intersectedIntervalClusterMap.firstKey()).getEndSplitPoint()
+                        .compareTo(interval.getStartSplitPoint()) < 0) {
```
triceo, line 61: With these long names and long chains of getters, it becomes hard to track what's going on here. In such cases, it doesn't hurt to help humans a bit. 

There are many ways how to do that. For example, this could be its own boolean variable, with a well chosen name:

     intersectedIntervalClusterMap.get(intersectedIntervalClusterMap.firstKey()).getEndSplitPoint()
                        .compareTo(interval.getStartSplitPoint()) < 0

That will then make the condition a nicely readable one-liner. Staying with the same code, I would consider:

- extracting `intersectedIntervalClusterMap.get(intersectedIntervalClusterMap.firstKey())` to its own well-named variable;
- introducing `isAfter(..)`/`isBefore(...)` on `SplitPoint`, so that all the nasty comparisons are hidden away.

The goal of all of that is to reduce the complexity of the code, so that it reads better. Remember: code is written once, and then read a hundred times over its lifetime.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            intersectedIntervalClusterMap = intersectedIntervalClusterMap.subMap(intersectedIntervalClusterMap.firstKey(),
```
triceo, line 62: `firstKey` is used several times in this method. For non-trivial methods, you should really consider caching the result.
(Technically this should be a trivial method, but for some reason, it really is not, as my profiling showed when we were doing the consecutive sequences.)

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            IntervalClusterImpl<IntervalType_, PointType_, DifferenceType_> newCluster =
+                    new IntervalClusterImpl<>(splitPointSet, differenceFunction, start);
+            clusterStartSplitPointToCluster.put(start, newCluster);
+            Map.Entry<IntervalSplitPoint<IntervalType_, PointType_>, IntervalClusterImpl<IntervalType_, PointType_, DifferenceType_>> nextClusterEntry =
+                    clusterStartSplitPointToCluster.higherEntry(start);
+            if (nextClusterEntry != null) {
+                clusterStartSplitPointToNextBreak.put(start,
+                        new IntervalBreakImpl<IntervalType_, PointType_, DifferenceType_>(newCluster,
+                                nextClusterEntry.getValue(), differenceFunction.apply(
+                                        newCluster.getEnd(), nextClusterEntry.getValue().getStart())));
```
triceo, line 77: This is another piece of code that would benefit from declaring a bit more variables.
Build the statement incrementally, to make it easier to read.



    ... nextClusterEntryValue = nextClusterEntry.getValue();
    DIfferenceType_ difference = differenceFunction.apply(newCluster.getEnd(), nextClusterEntryValue.getStart());
    IntervalBreakImpl<...> newBreak = new IntervalBreakImpl<IntervalType_, PointType_, DifferenceType_>(newCluster, nextClusterEntryValue, difference);
    clusterStartSplitPointToNextBreak.put(start, newBreak);

See how much easier it is to read?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+            Map.Entry<IntervalSplitPoint<IntervalType_, PointType_>, IntervalClusterImpl<IntervalType_, PointType_, DifferenceType_>> previousClusterEntry =
+                    clusterStartSplitPointToCluster.lowerEntry(start);
+            if (previousClusterEntry != null) {
+                clusterStartSplitPointToNextBreak.put(previousClusterEntry.getKey(),
+                        new IntervalBreakImpl<>(previousClusterEntry.getValue(), newCluster, differenceFunction.apply(
+                                previousClusterEntry.getValue().getEnd(), newCluster.getStart())));
```
triceo, line 84: Another case.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+            return;
+        }
+        NavigableMap<IntervalSplitPoint<IntervalType_, PointType_>, IntervalBreakImpl<IntervalType_, PointType_, DifferenceType_>> intersectedIntervalBreakMap =
+                clusterStartSplitPointToNextBreak.subMap(
+                        ObjectUtils.defaultIfNull(clusterStartSplitPointToNextBreak.floorKey(interval.getStartSplitPoint()),
+                                interval.getStartSplitPoint()),
+                        true, interval.getEndSplitPoint(), true);
```
triceo, line 92: Another case.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        intersectedIntervalClusterMap.tailMap(intersectedIntervalClusterMap.firstKey(), false).values()
+                .forEach(intervalCluster::mergeIntervalCluster);
+        intersectedIntervalClusterMap.tailMap(intersectedIntervalClusterMap.firstKey(), false).clear();
+
+        if (intersectedIntervalBreakMap.size() > 0) {
+            if (interval.getStartSplitPoint().splitPoint
+                    .compareTo(intersectedIntervalBreakMap.firstEntry().getValue().getPreviousIntervalClusterEnd()) <= 0) {
+                if (interval.getEndSplitPoint().splitPoint
+                        .compareTo(intersectedIntervalBreakMap.lastEntry().getValue().getNextIntervalClusterStart()) >= 0) {
+                    intersectedIntervalBreakMap.clear();
```
triceo, line 106: Three nested ifs with non-trivial conditions.
In this case, I would consider making the bodies of those statements their own methods. Break the flow, encapsulate, help people to not have to think about so many things at once.

triceo (commented):
The logic here is becoming heavy, and with the long names and long generics, this code is a nightmare to try to read and understand.

I suggest you think a bit about how to make this code flow better. How to help people understand. Many of those changes are just about extracting things into properly named variables and functions. See for yourself.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
Christopher-Chianelli: Can't do extract `intersectedIntervalClusterMap.get(intersectedIntervalClusterMap.firstKey())` into its own variable since `intersectedIntervalClusterMap` in the first if statement may be empty. It is extracted into intervalCluster (now `firstIntersectedIntervalCluster`) variable later on.

Christopher-Chianelli: (isAfter/isBefore methods are added on SplitPoint)

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
Christopher-Chianelli: Done (technically already cached as oldStart, but I wasn't using it)

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,282 @@
...
+    private final Iterable<IntervalCluster<Interval_, Point_, Difference_>> intervalClusterIterable;
+    private final BiFunction<Point_, Point_, Difference_> differenceFunction;
+    private final Iterable<IntervalBreak<Interval_, Point_, Difference_>> breaksIterable;
+
+    public ConsecutiveIntervalInfoImpl(TreeSet<IntervalSplitPoint<Interval_, Point_>> splitPointSet,
+            BiFunction<Point_, Point_, Difference_> differenceFunction) {
+        clusterStartSplitPointToCluster = new TreeMap<>();
+        clusterStartSplitPointToNextBreak = new TreeMap<>();
+        intervalClusterIterable = new MapValuesIterable<>(clusterStartSplitPointToCluster);
+        breaksIterable = new MapValuesIterable<>(clusterStartSplitPointToNextBreak);
```
triceo, line 45: For the sake of consistency, I'd use `this` in the whole method, not just where it is necessary below.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,282 @@
...
+            return;
+        }
+
+        // Interval intersect at least one cluster
+        // Ex:
+        //      -----------------
+        //  ------  ------  ---   ----
+        var firstIntersectedIntervalCluster =
+                intersectedIntervalClusterMap.firstEntry().getValue();
+        IntervalSplitPoint<Interval_, Point_> oldStart = firstIntersectedIntervalCluster.getStartSplitPoint();
```
triceo, line 81: While `var` has its benefits, I have to say I am not too excited when it's applied inconsistently. Especially if the inconsistent usage is co-located as it is here.

I understand that the compiler sometimes can not infer the type and this explicit type is necessary. In those situations though, I'm not sure if `var` actually brings anything other than inconsistency.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        var intervalCluster = intervalClusterEntry.getValue();
+        clusterStartSplitPointToCluster.remove(intervalClusterEntry.getKey());
+        var previousBreakEntry =
+                clusterStartSplitPointToNextBreak.lowerEntry(intervalClusterEntry.getKey());
+        var nextIntervalClusterEntry =
+                clusterStartSplitPointToCluster.higherEntry(intervalClusterEntry.getKey());
+        clusterStartSplitPointToNextBreak.remove(intervalClusterEntry.getKey());
+
+        IntervalClusterImpl<Interval_, Point_, Difference_> previousIntervalCluster = null;
+        IntervalBreakImpl<Interval_, Point_, Difference_> previousBreak = null;
```
triceo, line 229: You may get away without this type, if you do `var previousIntervalCluster = previousBreakEntry != null ? ... : null`.
Hopefully the Java compiler in my head is not misbehaving. :-)

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,282 @@
...
+                //   -----------
+                //----  ------ -----
+                intersectedIntervalBreakMap.clear();
+            } else {
+                // Case: interval span first break, but does not span the final break
+                // Ex:
+                //   -----------
+                //----  ------   -----
+                IntervalBreakImpl<Interval_, Point_, Difference_> finalBreak =
+                        intersectedIntervalBreakMap.lastEntry().getValue();
```
triceo, line 170: Did you by any chance omit `var` here? Because it seems to me that the compiler should be able to infer this.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+
+    private void removeSpannedBreaksAndUpdateIntersectedBreaks(Interval<Interval_, Point_> interval,
+            IntervalClusterImpl<Interval_, Point_, Difference_> intervalCluster) {
+        var intersectedIntervalBreakMap =
+                clusterStartSplitPointToNextBreak.subMap(
+                        ObjectUtils.defaultIfNull(clusterStartSplitPointToNextBreak.floorKey(interval.getStartSplitPoint()),
+                                interval.getStartSplitPoint()),
+                        true, interval.getEndSplitPoint(), true);
```
triceo, line 141: Another one of those statements that would read better when broken down into component parts.

triceo (commented):
Some notes on the use of `var`. You may also be interested in the guidelines on LVTI written by the Java architects themselves:
http://openjdk.java.net/projects/amber/LVTIstyle.html

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveIntervalInfoImpl.java
Christopher-Chianelli: I didn't see it; it can be a var

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/IntervalCluster.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.optaplanner.examples.common.experimental.api;
+
+public interface IntervalCluster<Interval_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>>
+        extends Iterable<Interval_> {
```
triceo, line 20: We need this to be able to give us the items in the cluster.
(Just like we can retrieve the items in a sequence.)

triceo:
Chris, we have some outstanding comments here. Let's solve them, address the Javadocs, and get this merged. I still need to merge the cheaptime work this sprint, for which I need this PR merged. Thanks!

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/api/IntervalCluster.java
Christopher-Chianelli: You can; it is an `Iterable` of `Interval_` (where `Interval_` is the type of the item, different from the Interval datatype)

triceo (approved):
I am going to make some minor changes in my follow-up PR.

PR #1419, opened 2021-07-13
PLANNER-2470: Add Benchmarker Metrics to Solver via Micrometer

Christopher-Chianelli:
Note: new implementation in benchmark is poll (i.e. measure every 5 seconds) based instead of step (i.e. measure at end of every step) based. Still need to add tests for new metrics.

### JIRA

https://issues.redhat.com/browse/PLANNER-2470

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
@@ -100,12 +100,15 @@ public class StatisticRegistry extends SimpleMeterRegistry implements PhaseLifec
         scoreConsumer.accept(scoreDefinition.fromLevelNumbers(0, levelNumbers));
     }
 
-    public Number getGaugeValue(SolverMetric metric, Tags runId) {
-        return getGaugeValue(metric.getMeterId(), runId);
+    public void getGaugeValue(SolverMetric metric, Tags runId, Consumer<Number> gaugeConsumer) {
+        getGaugeValue(metric.getMeterId(), runId, gaugeConsumer);
     }
 
-    public Number getGaugeValue(String meterId, Tags runId) {
-        return this.find(meterId).tags(runId).gauge().value();
+    public void getGaugeValue(String meterId, Tags runId, Consumer<Number> gaugeConsumer) {
+        Gauge gauge = this.find(meterId).tags(runId).gauge();
+        if (gauge != null) {
+            gaugeConsumer.accept(gauge.value());
+        }
     }
 
     @Override
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
@@ -30,7 +30,7 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.optaplanner.core.api.score.Score;
-import org.optaplanner.core.config.solver.metric.SolverMetric;
+import org.optaplanner.core.config.solver.monitoring.SolverMetric;
 import org.optaplanner.core.impl.phase.event.PhaseLifecycleListener;
 import org.optaplanner.core.impl.phase.scope.AbstractPhaseScope;
 import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
@@ -16,7 +16,6 @@
 
 package org.optaplanner.benchmark.impl.statistic;
 
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
@@ -28,51 +27,27 @@ import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.solver.event.BestSolutionChangedEvent;
 import org.optaplanner.core.api.solver.event.SolverEventListener;
 import org.optaplanner.core.config.solver.metric.SolverMetric;
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListener;
+import org.optaplanner.core.impl.phase.scope.AbstractPhaseScope;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
 import org.optaplanner.core.impl.score.definition.ScoreDefinition;
 import org.optaplanner.core.impl.solver.DefaultSolver;
+import org.optaplanner.core.impl.solver.scope.SolverScope;
 
-import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.search.Search;
-import io.micrometer.core.instrument.step.StepMeterRegistry;
-import io.micrometer.core.instrument.step.StepRegistryConfig;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 
-public class StatisticRegistry extends StepMeterRegistry implements SolverEventListener {
+public class StatisticRegistry extends SimpleMeterRegistry implements PhaseLifecycleListener, SolverEventListener {
 
     List<Consumer<Long>> stepMeterListenerList = new ArrayList<>();
     List<Consumer<Long>> bestSolutionMeterListenerList = new ArrayList<>();
-    long lastUpdatedTimestamp = Long.MIN_VALUE;
     long bestSolutionChangedTimestamp = Long.MIN_VALUE;
     ScoreDefinition scoreDefinition;
 
     public StatisticRegistry(DefaultSolver solver) {
-        super(new StepRegistryConfig() {
-            @Override
-            public Duration step() {
-                return Duration.ofMillis(10L);
-            }
-
-            @Override
-            public String prefix() {
-                return "optaplanner-benchmark-statistic";
-            }
-
-            @Override
-            public String get(String key) {
-                return null;
-            }
-        }, new Clock() {
-            @Override
-            public long wallTime() {
-                return solver.getTimeMillisSpent();
-            }
-
-            @Override
-            public long monotonicTime() {
-                return System.nanoTime();
-            }
-        });
         scoreDefinition = solver.getSolverScope().getScoreDefinition();
     }
 
@@ -111,13 +86,18 @@ public class StatisticRegistry extends StepMeterRegistry implements SolverEventL
                 .collect(Collectors.toSet());
     }
 
-    public Score extractScoreFromMeters(SolverMetric metric, Tags runId) {
+    public void extractScoreFromMeters(SolverMetric metric, Tags runId, Consumer<Score> scoreConsumer) {
         String[] labelNames = scoreDefinition.getLevelLabels();
         Number[] levelNumbers = new Number[labelNames.length];
         for (int i = 0; i < labelNames.length; i++) {
-            levelNumbers[i] = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge().value();
+            Gauge scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
+            if (scoreLevelGauge != null) {
+                levelNumbers[i] = scoreLevelGauge.value();
+            } else {
+                return;
+            }
         }
-        return scoreDefinition.fromLevelNumbers(0, levelNumbers);
+        scoreConsumer.accept(scoreDefinition.fromLevelNumbers(0, levelNumbers));
     }
 
     public Number getGaugeValue(SolverMetric metric, Tags runId) {
@@ -129,21 +109,47 @@ public class StatisticRegistry extends StepMeterRegistry implements SolverEventL
     }
 
     @Override
-    protected void publish() {
-        stepMeterListenerList.forEach(listener -> listener.accept(clock.wallTime()));
-        if (lastUpdatedTimestamp < bestSolutionChangedTimestamp) {
-            bestSolutionMeterListenerList.forEach(listener -> listener.accept(bestSolutionChangedTimestamp));
-            lastUpdatedTimestamp = bestSolutionChangedTimestamp;
-        }
-    }
-
-    @Override
     protected TimeUnit getBaseTimeUnit() {
         return TimeUnit.MILLISECONDS;
     }
 
     @Override
     public void bestSolutionChanged(BestSolutionChangedEvent event) {
+        if (bestSolutionChangedTimestamp != Long.MIN_VALUE) {
+            bestSolutionMeterListenerList.forEach(listener -> listener.accept(bestSolutionChangedTimestamp));
+        }
         bestSolutionChangedTimestamp = event.getTimeMillisSpent();
+    }
+
+    @Override
+    public void stepEnded(AbstractStepScope stepScope) {
+        final long timestamp =
+                System.currentTimeMillis() - stepScope.getPhaseScope().getSolverScope().getStartingSystemTimeMillis();
+        stepMeterListenerList.forEach(listener -> listener.accept(timestamp));
+    }
+
+    @Override
+    public void phaseStarted(AbstractPhaseScope phaseScope) {
+        // intentional empty
+    }
+
+    @Override
+    public void stepStarted(AbstractStepScope stepScope) {
+        // intentional empty
+    }
+
+    @Override
+    public void phaseEnded(AbstractPhaseScope phaseScope) {
+        // intentional empty
+    }
+
+    @Override
+    public void solvingStarted(SolverScope solverScope) {
+        // intentional empty
+    }
+
+    @Override
+    public void solvingEnded(SolverScope solverScope) {
+        // intentional empty
     }
 }
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
@@ -100,6 +100,9 @@ public class StatisticRegistry<Solution_> extends SimpleMeterRegistry
 
     public void extractScoreFromMeters(SolverMetric metric, Tags runId, Consumer<Score<?>> scoreConsumer) {
         String[] labelNames = scoreDefinition.getLevelLabels();
+        for (int i = 0; i < labelNames.length; i++) {
+            labelNames[i] = labelNames[i].replace(' ', '.');
+        }
         Number[] levelNumbers = new Number[labelNames.length];
         for (int i = 0; i < labelNames.length; i++) {
             Gauge scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
@@ -106,7 +106,7 @@ public class StatisticRegistry<Solution_> extends SimpleMeterRegistry
         Number[] levelNumbers = new Number[labelNames.length];
         for (int i = 0; i < labelNames.length; i++) {
             Gauge scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
-            if (scoreLevelGauge != null) {
+            if (scoreLevelGauge != null && Double.isFinite(scoreLevelGauge.value())) {
                 levelNumbers[i] = scoreLevelNumberConverter.apply(scoreLevelGauge.value());
             } else {
                 return;
@@ -144,7 +144,7 @@ public class StatisticRegistry<Solution_> extends SimpleMeterRegistry
 
     public void getGaugeValue(String meterId, Tags runId, Consumer<Number> gaugeConsumer) {
         Gauge gauge = this.find(meterId).tags(runId).gauge();
-        if (gauge != null) {
+        if (gauge != null && Double.isFinite(gauge.value())) {
             gaugeConsumer.accept(gauge.value());
         }
     }
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
@@ -16,6 +16,8 @@
 
 package org.optaplanner.benchmark.impl.statistic;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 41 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+import org.optaplanner.core.impl.solver.DefaultSolver;
+import org.optaplanner.core.impl.solver.scope.SolverScope;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.search.Search;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+
+public class StatisticRegistry extends SimpleMeterRegistry implements PhaseLifecycleListener, SolverEventListener {
```
triceo, line 43: Let's make this generic in `Solution_` and get rid of all the raw type warnings in this class.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+
+    List<Consumer<Long>> stepMeterListenerList = new ArrayList<>();
+    List<Consumer<Long>> bestSolutionMeterListenerList = new ArrayList<>();
+    long bestSolutionChangedTimestamp = Long.MIN_VALUE;
+    ScoreDefinition scoreDefinition;
```
triceo, line 48: I often point this out in your PRs. Please make sure fields are **always `private` and always `final`**, unless there is a reason for them not to be.

This should apply not only to this PR, but to all future PRs, please.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public void addListener(SolverMetric metric, Consumer<Long> listener) {
+        if (isMetricBestSolutionBased(metric)) {
+            bestSolutionMeterListenerList.add(listener);
+        } else {
+            stepMeterListenerList.add(listener);
+        }
+    }
+
+    private boolean isMetricBestSolutionBased(SolverMetric metric) {
```
triceo, line 62: This works better as a property of the `SolverMetric` directly.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+                return false;
+            default:
+                throw new IllegalStateException("SolverMetric (" + metric + ") does not have a case");
+        }
+    }
+
+    public Set<Meter.Id> getMeterIds(SolverMetric metric, Tags runId) {
+        return Search.in(this).name(name -> name.startsWith(metric.getMeterId())).tags(runId)
+                .meters().stream().map(meter -> meter.getId())
+                .collect(Collectors.toSet());
```
triceo, line 86: In this fluent chain, I find it hard to see where micrometer ends and streams start.
May be worth splitting former into its own variable.
Also, `Meter::getId`.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+        Number[] levelNumbers = new Number[labelNames.length];
+        for (int i = 0; i < labelNames.length; i++) {
+            Gauge scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
+            if (scoreLevelGauge != null) {
+                levelNumbers[i] = scoreLevelGauge.value();
+            } else {
+                return;
+            }
+        }
+        scoreConsumer.accept(scoreDefinition.fromLevelNumbers(0, levelNumbers));
```
triceo, line 100: This is potentially a very expensive operation, and it is being used often.
I'd consider a form of cache, at the very least for the `find()` result.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public void getGaugeValue(SolverMetric metric, Tags runId, Consumer<Number> gaugeConsumer) {
+        getGaugeValue(metric.getMeterId(), runId, gaugeConsumer);
+    }
+
+    public void getGaugeValue(String meterId, Tags runId, Consumer<Number> gaugeConsumer) {
+        Gauge gauge = this.find(meterId).tags(runId).gauge();
```
triceo, line 108: Dtto. We do not need to look them up all the time, we can gradually build a cache.
(I'm of course assuming that the references do not change.)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/bestsolutionmutation/BestSolutionMutationSubSingleStatistic.java
```diff
@@ -21,71 +21,35 @@
...
-                mutationCount = 0;
-            } else {
-                mutationCount = mutationCounter.countMutations(oldBestSolution, newBestSolution);
-            }
-            pointList.add(new BestSolutionMutationStatisticPoint(
-                    event.getTimeMillisSpent(), mutationCount));
-            oldBestSolution = newBestSolution;
-        }
-
+    public void close(StatisticRegistry registry, Tags runTag, Solver<Solution_> solver) {
```
triceo, line 52: Shouldn't we unregister the listeners here? If so, this needs to happen in all the statistics.
Or is that something that is not supported by micrometer?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/movecountperstep/MoveCountPerStepSubSingleStatistic.java
```diff
@@ -21,52 +21,38 @@
...
     // Lifecycle methods
     // ************************************************************************
 
     @Override
-    public void open(Solver<Solution_> solver) {
-        ((AbstractSolver<Solution_>) solver).addPhaseLifecycleListener(listener);
+    public void open(StatisticRegistry registry, Tags runTag, Solver<Solution_> solver) {
+        registry.addListener(SolverMetric.MOVE_COUNT_PER_STEP, timeMillisSpent -> {
+            registry.getGaugeValue(SolverMetric.MOVE_COUNT_PER_STEP + ".accepted", runTag, accepted -> {
+                registry.getGaugeValue(SolverMetric.MOVE_COUNT_PER_STEP + ".accepted", runTag, selected -> {
```
triceo, line 46: I don't understand. Why the nesting? The gauge is the same, shouldn't therefore `accepted` and `selected` do the same thing?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/scorecalculationspeed/ScoreCalculationSpeedSubSingleStatistic.java
```diff
@@ -21,20 +21,18 @@
...
 import org.optaplanner.core.impl.score.definition.ScoreDefinition;
-import org.optaplanner.core.impl.solver.AbstractSolver;
-import org.optaplanner.core.impl.solver.scope.SolverScope;
+
+import io.micrometer.core.instrument.Tags;
 
 public class ScoreCalculationSpeedSubSingleStatistic<Solution_>
         extends ProblemBasedSubSingleStatistic<Solution_, ScoreCalculationSpeedStatisticPoint> {
 
     private final long timeMillisThresholdInterval;
```
triceo, line 34: This appears unused. Maybe some code is missing?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/subsingle/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreSubSingleStatistic.java
```diff
@@ -79,54 +76,40 @@ public ConstraintMatchTotalBestScoreSubSingleStatistic(SubSingleBenchmarkResult
...
+                                    count -> {
+                                        pointList.add(new ConstraintMatchTotalBestScoreStatisticPoint(
+                                                timeMillisSpent,
+                                                constraintPackage,
+                                                constraintName,
+                                                count.intValue(),
+                                                score));
+                                    });
+                        });
+            });
```
triceo, line 107: This is complex. It deserves some comments on what does it do and why.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/subsingle/constraintmatchtotalstepscore/ConstraintMatchTotalStepScoreSubSingleStatistic.java
```diff
@@ -79,54 +76,40 @@ public ConstraintMatchTotalStepScoreSubSingleStatistic(SubSingleBenchmarkResult
...
+                                            .and("constraint.name", constraintPackageNamePair.right),
+                                    count -> {
+                                        pointList.add(new ConstraintMatchTotalStepScoreStatisticPoint(
+                                                timeMillisSpent,
+                                                constraintPackage,
+                                                constraintName,
+                                                count.intValue(),
+                                                score));
+                                    });
+                        });
```
triceo, line 106: Dtto.

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/DefaultLocalSearchPhase.java
```diff
@@ -144,6 +151,57 @@ public void stepEnded(LocalSearchStepScope<Solution_> stepScope) {
         }
     }
 
+    private void collectMetrics(LocalSearchStepScope<Solution_> stepScope) {
```
triceo, line 154: Nothing for construction heuristics?
That may be why I was seeing the empty graphs.

We need to figure out how to make this happen for every phase.

optaplanner-core/src/main/java/org/optaplanner/core/impl/statistic/BestScoreStatistic.java
```diff
@@ -0,0 +1,36 @@
...
+package org.optaplanner.core.impl.statistic;
+
+import java.util.function.Consumer;
+
+import org.optaplanner.core.api.solver.Solver;
+import org.optaplanner.core.config.solver.metric.SolverMetric;
+import org.optaplanner.core.impl.score.definition.ScoreDefinition;
+import org.optaplanner.core.impl.solver.DefaultSolver;
+
+public class BestScoreStatistic implements Consumer<Solver> {
```
triceo, line 26: Instead of implementing `Consumer<Solver>`, which is a generic interface, I'd create our own interface.
This is a very specific functionality and we want to see where it is used in the code. `Consumer<Solver>` doesn't let us do this, but `SolverStatistic<Solver>` would.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/memoryuse/MemoryUseSubSingleStatistic.java
```diff
@@ -45,32 +45,39 @@ public MemoryUseSubSingleStatistic(SubSingleBenchmarkResult subSingleBenchmarkRe
...
 
         @Override
-        public void stepEnded(AbstractStepScope<Solution_> stepScope) {
-            long timeMillisSpent = stepScope.getPhaseScope().calculateSolverTimeMillisSpentUpToNow();
+        public void accept(Long timeMillisSpent) {
             if (timeMillisSpent >= nextTimeMillisThreshold) {
-                pointList.add(new MemoryUseStatisticPoint(timeMillisSpent, MemoryUseMeasurement.create()));
+                registry.getGaugeValue(SolverMetric.MEMORY_USE, tags, memoryUse -> {
+                    pointList.add(new MemoryUseStatisticPoint(timeMillisSpent, new MemoryUseMeasurement(memoryUse.longValue(),
+                            (long) registry.find("jvm.memory.max").tags(tags).gauge().value())));
```
triceo, line 79: Is `max` what we want? I'd expect we want the current usage, not the most it's even been.

triceo (changes requested):
Good first pass. See DM on Zulip for a serious issue.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -82,7 +84,8 @@
         "domainAccessType",
         "scoreDirectorFactoryConfig",
         "terminationConfig",
-        "phaseConfigList"
+        "phaseConfigList",
+        "metricConfig",
```
triceo, line 88: Let's not expose it in the solver config just yet. Let's give it a while to settle down.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/bestsolutionmutation/BestSolutionMutationSubSingleStatistic.java
Christopher-Chianelli: The `StatisticRegistry` itself is unregistered/removed from `Metrics`. See `SubSingleBenchmarkRunner`:

```
        StatisticRegistry statisticRegistry = new StatisticRegistry((DefaultSolver) solver);
        Metrics.addRegistry(statisticRegistry);
        solver.addEventListener(statisticRegistry);
        solver.addPhaseLifecycleListener(statisticRegistry);

        Tags runTag = Tags.of("optaplanner.benchmark.run", runId);
        for (SubSingleStatistic<Solution_, ?> subSingleStatistic : subSingleBenchmarkResult.getEffectiveSubSingleStatisticMap()
                .values()) {
            subSingleStatistic.open(statisticRegistry, runTag, solver);
            subSingleStatistic.initPointList();
        }
        Solution_ solution = solver.solve(problem);

        solver.removePhaseLifecycleListener(statisticRegistry);
        solver.removeEventListener(statisticRegistry);
        Metrics.removeRegistry(statisticRegistry);
```

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/movecountperstep/MoveCountPerStepSubSingleStatistic.java
Christopher-Chianelli: The guage should be different (it should be the selected gauge); good catch

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/bestsolutionmutation/BestSolutionMutationSubSingleStatistic.java
triceo: In that case, I wouldn't override these methods. Make it default, add a comment in there.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/memoryuse/MemoryUseSubSingleStatistic.java
Christopher-Chianelli: MemoryUseStatisticPoint use both current and max

optaplanner-core/src/main/java/org/optaplanner/core/impl/localsearch/DefaultLocalSearchPhase.java
Christopher-Chianelli: A bunch of metrics in the original benchmarker code is set up for local search only (but data collection is unrelated to it; should be able to make it every step)

Christopher-Chianelli: Looking at the class it on, it for the statistics that are local search specific

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -46,9 +47,27 @@
     List<Consumer<Long>> bestSolutionMeterListenerList = new ArrayList<>();
     long bestSolutionChangedTimestamp = Long.MIN_VALUE;
     ScoreDefinition scoreDefinition;
+    final Function<Number, Number> scoreLevelNumberConverter;
 
     public StatisticRegistry(DefaultSolver solver) {
         scoreDefinition = solver.getSolverScope().getScoreDefinition();
+        Number zeroScoreLevel0 = scoreDefinition.getZeroScore().toLevelNumbers()[0];
+        if (zeroScoreLevel0 instanceof Double) {
```
triceo, line 55: Aren't you missing BigDecimal here?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -92,7 +111,7 @@ public void extractScoreFromMeters(SolverMetric metric, Tags runId, Consumer<Sco
         for (int i = 0; i < labelNames.length; i++) {
             Gauge scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
             if (scoreLevelGauge != null) {
-                levelNumbers[i] = scoreLevelGauge.value();
+                levelNumbers[i] = scoreLevelNumberConverter.apply(scoreLevelGauge.value());
```
triceo, line 114: I'm thinking... shouldn't you be able to just build the score from all its levels, and then retrieve the level numbers from it?

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/movecountperstep/MoveCountPerStepSubSingleStatistic.java
```diff
@@ -42,8 +42,8 @@ public MoveCountPerStepSubSingleStatistic(SubSingleBenchmarkResult subSingleBenc
     @Override
     public void open(StatisticRegistry registry, Tags runTag, Solver<Solution_> solver) {
         registry.addListener(SolverMetric.MOVE_COUNT_PER_STEP, timeMillisSpent -> {
-            registry.getGaugeValue(SolverMetric.MOVE_COUNT_PER_STEP + ".accepted", runTag, accepted -> {
-                registry.getGaugeValue(SolverMetric.MOVE_COUNT_PER_STEP + ".accepted", runTag, selected -> {
+            registry.getGaugeValue(SolverMetric.MOVE_COUNT_PER_STEP.getMeterId() + ".accepted", runTag, accepted -> {
+                registry.getGaugeValue(SolverMetric.MOVE_COUNT_PER_STEP.getMeterId() + ".selected", runTag, selected -> {
```
triceo, line 46: Since you control the registry API, I suggest you continue accepting the enum and do the `getMeterId()` call internally. Less repetition.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
```diff
@@ -647,6 +648,7 @@ public SolverConfig inherit(SolverConfig inheritedConfig) {
                 inheritedConfig.getScoreDirectorFactoryConfig());
         terminationConfig = ConfigUtils.inheritConfig(terminationConfig, inheritedConfig.getTerminationConfig());
         phaseConfigList = ConfigUtils.inheritMergeableListConfig(phaseConfigList, inheritedConfig.getPhaseConfigList());
+        metricConfig = ConfigUtils.inheritConfig(metricConfig, inheritedConfig.getMetricConfig());
```
triceo, line 651: How do we make sure that this is not exposed to the users?
(Geoffrey wants to keep in control of the solver config. But he will be gone for 3 weeks. I'd like to merge this earlier, but I won't be able to, if it impacts public API.)

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticRegistry.java
Christopher-Chianelli: Score definitions expect the score levels to be a particular type (Exactly integer, exactly long, etc.). Micrometer gauges can only return doubles. So in order to not get a typecast exception in the score definition, I need to convert it to the "expected" type. (There are micrometer gauges for each level)

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/SolverConfig.java
Christopher-Chianelli: Currently, it XML transitive (i.e. can be accessed via Java API, but cannot via XML API. It need to be in the config so it can be read by the SolverFactory (which uses it to set up metrics)

triceo: The problem with it being in the Java API is that it is a public API that needs to be kept stable.
Therefore I can not merge this PR until discussed with and approved by Geoffrey.

optaplanner-core/src/main/java/org/optaplanner/core/impl/statistic/SolverStatistic.java
```diff
@@ -0,0 +1,7 @@
+package org.optaplanner.core.impl.statistic;
+
+import org.optaplanner.core.api.solver.Solver;
+
+public interface SolverStatistic {
```
triceo, line 5: This could be generic in `Solution_`, therefore eliminating some of the unchecked warnings you are suppressing.

triceo:
Since we'll be doing the API review as part of this PR, please reintroduce the solver config changes to affect the XML, @Christopher-Chianelli .

Christopher-Chianelli:
`MetricConfig -> MonitoringConfig`, tag map now part of internal DefaultSolver API with default tags set in SolverManager and Benchmarker (no default id tag set in default solver factory)

triceo:
What's missing to mark this PR as ready for review, @Christopher-Chianelli ?

Christopher-Chianelli:
@triceo strictly speaking, no new test coverage, although how to test metrics beside "they have a value" is beyond me. However, it does have about the same test coverage for metrics as the benchmarker (no tests were removed/modified).

triceo:
Well, if we spend some extra time, there are things we could test that are useful. If the scenarios are defined well enough, with a step count termination, you should be able to know the expected values for some of these metrics, such as score calc count. You can test solve lengths and solver counts.

I think it would be worth to invest that time. We definitely don't need to test all the metrics; for example, mutations per step seems excessive. But I worry that if we say we do not test metrics in principle, that's just going to bring technical debt straight into core. 

The fact that benchmarker is largely untested is a problem that we've been lamenting for years, and it makes the code hard to touch. And that makes people generally avoid it.

Christopher-Chianelli:
Solver count and error count is already tested (they were added w/ tests in a previous PR). If you can give me configs with expected metric values, I can write tests for them (although I can only really test "final" values for some metrics since some metrics use EventListeners, and I cannot guarantee the order of event listeners.

triceo:
Case in point: best score metrics. You can create a solver with a single entity, single variable and only a handful of values. You can set the solver env mode to reproducible, therefore you will know what's the expected best score at each step. And you can write a test to check that this is reflected in the metrics.

This approach is not feasible for some of the other metrics, but for best score, it should be relatively simple, I think.

Also, there appear to be conflicts now. I do not want to hold up merging the PR anymore; if you don't see how you can do a test like I describe, we can discuss later.

Christopher-Chianelli:
@triceo added test for best score metric

triceo (approved):
Personally, I think that this step-based checking makes the test relatively unclear, without bringing too much benefit. How about we only check the metric at the end of solving?

After that, I consider this PR good to merge.

Christopher-Chianelli:
@triceo I disagree; checking only at the end of solving won't catch bugs relating to only updating once / only updating at end, which would make the benchmark best score graph a lot less valuable.

triceo:
@Christopher-Chianelli There appears to be a Windows-specific issue in the test.
I see your point of needing to check after every step.

Christopher-Chianelli:
@triceo fixed the Windows issue; it seems a gauge existed but returned NaN, so I added `Double.isFinite()` checks to the registries when the values are recorded

PR #1433, opened 2021-07-19
PLANNER-1831 Add ConstraintProvider for Cheaptime example

triceo:


optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/optional/score/CheapTimeConstraintProvider.java
```diff
@@ -0,0 +1,144 @@
...
+                        Machine::getSpinUpDownCostMicros);
+    }
+
+    protected Constraint idleCosts(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(TaskAssignment.class)
+                .groupBy(TaskAssignment::getMachine,
+                        consecutiveIntervals(TaskAssignment::getStartPeriod, TaskAssignment::getEndPeriod, (a, b) -> b - a))
+                .flattenLast(ConsecutiveIntervalInfo::getBreaks)
+                .join(Period.class,
+                        lessThanOrEqual((machine, brk) -> brk.getPreviousIntervalClusterEnd(), Period::getIndex),
```
rsynek, line 115: Stylistic note: let's turn `brk` to `break`; now we save just 2 letters and its meaning is less obvious.

rsynek (approved):
Just a small naming suggestion, otherwise looks good. Thanks!

optaplanner-examples/src/main/java/org/optaplanner/examples/cheaptime/optional/score/CheapTimeConstraintProvider.java
triceo: Cannot, `break` is a keyword in Java.

PR #1438, opened 2021-07-19
First draft of optaplanner-quarkus-bom

rsynek:
Draft PR to start a discussion.

- Introduces a new `optaplanner-quarkus-bom` that contains compile and runtime 3rd party dependencies of OptaPlanner.
- Imports the `quarkus-bom` effectively for every OptaPlanner module to keep common dependencies versions aligned.

pom_structure

### JIRA
https://issues.redhat.com/browse/PLANNER-2439

### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/build/optaplanner-build-parent/pom.xml
+++ b/build/optaplanner-build-parent/pom.xml
@@ -5,9 +5,9 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.optaplanner</groupId>
-    <artifactId>optaplanner-parent</artifactId>
+    <artifactId>optaplanner-quarkus-bom</artifactId>
     <version>8.10.0-SNAPSHOT</version>
-    <relativePath>../../pom.xml</relativePath>
+    <relativePath>../../optaplanner-quarkus-bom/pom.xml</relativePath>
   </parent>
 
   <artifactId>optaplanner-build-parent</artifactId>
@@ -20,35 +20,11 @@
     <!-- Dependencies -->
     <!-- ************************************************************************ -->
 
-    <!-- SNAPSHOT and KIE dependency versions -->
-    <version.org.kie.kogito>2.0.0-SNAPSHOT</version.org.kie.kogito>
-
     <!-- Normal dependency versions -->
-    <!-- TODO Do we want import the quarkus-bom instead of defining these ourselves? -->
     <version.ch.qos.logback>1.2.3</version.ch.qos.logback>
-    <version.commons-io>2.8.0</version.commons-io>
-    <version.com.fasterxml.jackson.core>2.12.3</version.com.fasterxml.jackson.core>
-    <version.com.h2database>1.4.197</version.com.h2database>
-    <version.com.thoughtworks.xstream>1.4.17</version.com.thoughtworks.xstream>
-    <version.io.micrometer>1.6.6</version.io.micrometer>
-    <version.io.quarkus>2.0.1.Final</version.io.quarkus>
-    <version.io.quarkus.gizmo>1.0.9.Final</version.io.quarkus.gizmo>
-    <version.org.ow2.asm>9.1</version.org.ow2.asm>
-    <version.jakarta.json.bind>1.0.2</version.jakarta.json.bind>
-    <version.jakarta.xml.bind>2.3.3</version.jakarta.xml.bind>
-    <version.org.apache.commons.lang3>3.12.0</version.org.apache.commons.lang3>
-    <version.org.apache.commons.math3>3.6.1</version.org.apache.commons.math3>
     <version.org.apache.poi>4.1.2</version.org.apache.poi>
-    <version.org.assertj>3.19.0</version.org.assertj>
-    <version.org.freemarker>2.3.31</version.org.freemarker>
-    <version.org.hibernate>5.5.3.Final</version.org.hibernate>
-    <version.org.jboss.jandex>2.3.0.Final</version.org.jboss.jandex>
     <version.org.jdom>1.1.3</version.org.jdom>
     <version.org.jfree.jfreechart>1.5.3</version.org.jfree.jfreechart>
-    <version.org.junit>5.7.1</version.org.junit>
-    <version.org.mockito>3.10.0</version.org.mockito>
-    <version.org.postgresql>42.2.22</version.org.postgresql>
-    <version.org.slf4j>1.7.30</version.org.slf4j>
     <version.org.springframework>5.3.6</version.org.springframework>
     <version.org.springframework.boot>2.4.5</version.org.springframework.boot>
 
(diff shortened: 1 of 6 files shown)


triceo (approved):
Love it! Sooo much less stuff to worry about.

Christopher-Chianelli:
I like it, but unsure of the name. `optaplanner-bom`contains only optaplanner dependencies, so I would think `optaplanner-quarkus-bom` to contain only `optaplanner-quarkus` dependencies (i.e. all the optaplanner quarkus extensions). That said, I do not know a good name replacement for `optaplanner-quarkus-bom`. (it basically optaplanner + quarkus + kogito BOM combined into one)

build/optaplanner-build-parent/pom.xml
```diff
@@ -20,35 +20,11 @@
     <!-- Dependencies -->
     <!-- ************************************************************************ -->
 
-    <!-- SNAPSHOT and KIE dependency versions -->
-    <version.org.kie.kogito>2.0.0-SNAPSHOT</version.org.kie.kogito>
```
rsynek, line 24: Only "internal" dependencies stay in the `optaplanner-build-parent`; typically test dependencies or dependencies of `optaplanner-examples`. Every dependency (compile or runtime) of `optaplanner-quarkus-integration` moves to the new BOM, which becomes a BOM with 3rd party dependencies.

optaplanner-persistence/optaplanner-persistence-jpa/pom.xml
```diff
@@ -171,6 +165,18 @@
       </testResource>
     </testResources>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
```
rsynek, line 170: For some reason, the dependency plugin does not recognize the `jakarta.transaction-api` as a replacement of the `jboss-transaction-api_1.2_spec`.

optaplanner-quarkus-bom/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.optaplanner</groupId>
+    <artifactId>optaplanner-parent</artifactId>
+    <version>8.10.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>optaplanner-quarkus-bom</artifactId>
```
rsynek, line 12: Naming is not perfect; it's definitely not just OptaPlanner and Quarkus dependencies. However, it's what the Quarkus platform member BOM should be.

optaplanner-quarkus-bom/pom.xml
```diff
@@ -0,0 +1,132 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rsynek, line 1: Where do we place this new module:
1/ optaplanner-quarkus-bom (current location)
2/ build/optaplanner-quarkus-bom
3/ optaplanner-quarkus-integration/optaplanner-quarkus-bom => it's not quarkus-specific, we use it in the `optaplanner-build-parent`

For me it's either 1/ or 2/.

optaplanner-quarkus-bom/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+      <dependency>
+        <groupId>org.optaplanner</groupId>
+        <artifactId>optaplanner-bom</artifactId>
+        <version>${project.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-bom</artifactId>
```
rsynek, line 36: This is the trick to avoid aligning with versions coming from Quarkus over and over. We reuse the `quarkus-bom` and use this new `optaplanner-quarkus-bom` as a parent in our `optaplanner-build-parent`.

optaplanner-quarkus-integration/pom.xml
```diff
@@ -35,17 +35,6 @@
     <module>optaplanner-quarkus-jsonb</module>
   </modules>
 
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>io.quarkus</groupId>
-        <artifactId>quarkus-bom</artifactId>
```
rsynek, line 42: No longer needed as it's coming from the parent pom.

optaplanner-distribution/pom.xml
```diff
@@ -45,33 +45,6 @@
                 <dependencySourceInclude>org.optaplanner:optaplanner-test</dependencySourceInclude>
                 <!-- TODO apparently optaplanner-examples gets included too somehow? -->
               </dependencySourceIncludes>
-              <additionalDependencies>
```
rsynek, line 48: Some of the version properties are no longer defined as we rely on the imported `quarkus-bom`. Unfortunately, the `additionalDependencies` do not get the versions. Moving these to project dependencies.

rsynek:
Hi @aloubyansky, this is a PR introducing the `optaplanner-quarkus-bom` we discussed last week.

If you take a look at the picture at the description of the PR, there is an overview of the pom structure in OptaPlanner.

One question that @ge0ffrey has raised: what if we override some dependency version in `optaplanner-build-parent` that is contained already in the `quarkus-bom`? We end up using (and testing) a different version of that dependency than we report to the `quarkus-platform`.

optaplanner-quarkus-bom/pom.xml
aloubyansky: IMO, the naming is fine. In addition, you are importing the `quarkus-bom` which includes everything necessary for building and testing applications. We could say, it's in the line with how it's done in Quarkus itself.

aloubyansky:
> One question that @ge0ffrey has raised: what if we override some dependency version in `optaplanner-build-parent` that is contained already in the `quarkus-bom`? We end up using (and testing) a different version of that dependency than we report to the `quarkus-platform`.

Right, which is why you should not do it :) This won't be possible in the platform though, where we will use the dependency version constraints you provide and run your tests against those version constraints.

rsynek:
> > One question that @ge0ffrey has raised: what if we override some dependency version in `optaplanner-build-parent` that is contained already in the `quarkus-bom`? We end up using (and testing) a different version of that dependency than we report to the `quarkus-platform`.
> 
> Right, which is why you should not do it :) This won't be possible in the platform though, where we will use the dependency version constraints you provide and run your tests against those version constraints.

It won't be so easy to make sure we never override; think if a new dependency is added to the `quarkus-bom` we already use in the `optaplanner-build-parent` (in a different version). There is no way we could spot such an override.

aloubyansky:
It depends on how you add the dependency. Are you thinking of adding an actual dependency (that all your modules inheriting from the parent should actually depend on) or a dependency version constraint (in the dependencyManagement)?
I guess it's the later. In which case you will not override the one from the quarkus-bom.

aloubyansky:
Even if you add a common actual dependency in the parent. Just make sure you define all the versions of the direct dependencies in the dependencyManagement and you'll be fine.

rsynek:
Below is a proposal that compensates for the issue with overriding `dependencyManagement`:
pom_structure_internal_bom

triceo:
This is complicated. I understand the need, and have nothing against it. I'm just wondering how we're going to enforce this structure going forward.

rsynek:
> This is complicated. I understand the need, and have nothing against it. I'm just wondering how we're going to enforce this structure going forward.

Both variants have their own strengths and weaknesses; I've just sent an invite to a meeting so that we can discuss them.

ge0ffrey:
I believe this PR does 2 things:

1) make the optaplanner build import the quarkus-bom, to avoid defining the version of hibernate, resteasy, jackson, etc. This guarantees that optaplanner's hibernate version is always up to date with the one in qaurkus. I would +1 a PR that does just this.

2) Introduce a new parent pom between optaplanner-parent and optaplanner-build-parent called optaplanner-quarkus-bom to accomodate for the quarkus CI testing. I believe this is greatly problematic in its current form and look forward to that meeting to fix it:
- The quarkus CI testing should use the union of the actual dependencies in our 4 optaplanner quarkus extensions. If any of those overwrite the hibernate version, or if the optaplanner-build-parent overwrite the jackson version, the CI testing should not miss that. The fact that the spring version is inherited into the optaplanner quarkus extensions is not a problem, because they don't add any spring dependencies.
- The addition of the complexity of yet another parent layer in our maven build is not worth the profit.
- It's not always obvious which dependencies should be declared in optaplanner-quarkus-bom or in optaplanner-build-parent. It's easy to put them in the wrong place. Right now, all our dependency versions are declared in one place (optaplanner-build-parent). That should stay that way.
- The naming of optaplanner-quarkus-bom should never reach Maven Central, because it will cause too much doubt on optaplanner-bom. This is not a bom, it is a parent and should be named accordingly.

ge0ffrey (changes requested):
See my last comment.

aloubyansky:
> * The naming of optaplanner-quarkus-bom should never reach Maven Central, because it will cause too much doubt on optaplanner-bom. This is not a bom, it is a parent and should be named accordingly.

It is used exactly as a BOM though when importing in the quarkus-platform, which is the original reason to introduce this artifact, iiuc.

aloubyansky:
Just to clarify, I am not insisting on any specific artifactId here. As long the content is ok from the quarkus-platform perspective, it's gonna be ok.

rsynek:
>     2. Introduce a new parent pom between optaplanner-parent and optaplanner-build-parent called optaplanner-quarkus-bom to accomodate for the quarkus CI testing. 
It's included in the pom structure as a parent of the `optaplanner-build-parent` only to avoid duplication of `dependencyManagement`. The real purpose of the new `optaplanner-quarkus-bom` is to replace the `optaplanner-bom` in the `quarkus-platform` here:
https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L194
so that the `quarkus-platform` takes OptaPlanner dependencies into account.

>     * It's not always obvious which dependencies should be declared in optaplanner-quarkus-bom or in optaplanner-build-parent. It's easy to put them in the wrong place. Right now, all our dependency versions are declared in one place (optaplanner-build-parent). That should stay that way.
That implies the `optaplanner-build-parent` should serve as the member BOM in the `quarkus-platform`. 
@aloubyansky has been against that, as some of the `dependencyManagement` entries of the `optaplanner-build-parent` are unwanted in the platform.

aloubyansky:
Right, when on the platform side we receive a set of dependency constraints, we assume these are runtime and build time requirements of your extensions on *application classpaths*. If the BOM we import in the platform you provide more than that, we can't differentiate between the constraints that represent your extension requirements and those used for your internal builds.

rsynek:
Closing in favor of https://github.com/kiegroup/optaplanner/pull/1464.

PR #1446, opened 2021-07-21
PLANNER-2448 Restore the possibility to configure KieBase options

triceo:
@Christopher-Chianelli Can you please check that there is nothing here to do wrt. Quarkus? I only re-enable this functionality outside of Quarkus and I'm not sure if/how we support that with Quarkus.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
@@ -21,6 +21,7 @@ import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ScoreDirectorFactoryFactory.java
```diff
@@ -309,7 +312,9 @@ private void validateNoDroolsAlphaNetworkCompilation() {
                         kieHelper.addResource(new FileSystemResource(scoreDrlFile));
                     }
                 }
-                kieBase = kieHelper.build(ExecutableModelProject.class, KieBaseMutabilityOption.DISABLED);
+                KieBaseConfiguration kieBaseConfiguration = buildKieBaseConfiguration(KieServices.get());
+                kieBaseConfiguration.setOption(KieBaseMutabilityOption.DISABLED); // Performance improvement.
+                kieBase = kieHelper.build(ExecutableModelProject.class, kieBaseConfiguration);
```
rsynek, line 317: Cannot resolve method 'build(java.lang.Class<org.drools.modelcompiler.ExecutableModelProject>, org.kie.api.KieBaseConfiguration)'

Does this PR depend on some latest changes in Drools?

triceo: Yes, merged into Kogito today. Will be available tomorrow.

rsynek (approved):
Good to merge as soon as the new Kogito artifacts are available.

Christopher-Chianelli:
@triceo Quarkus does not go on that code path for DRL. If `drl-integration-test` passes, then everything probably good.

PR #1462, opened 2021-07-26
PLANNER-2476 Improve performance of SubChainSwapMove

triceo:
Improves TSP GREECE 9882 score calculation count 80 times, assuming only subchain swap moves are enabled.


rsynek (approved):
Very nice improvement.

PR #1474, opened 2021-07-29
Docs update: quarkus/spring instead of WildFly/Tomcat, CS instead of DRL, workbench/ES is dead, ...

ge0ffrey:


optaplanner-docs/src/main/asciidoc/OptimizationAlgorithms/OptimizationAlgorithms-chapter.adoc
```diff
@@ -76,10 +76,17 @@ Do not assess quality on small data sets (unless production encounters only such
...
+with score calculation by a rule engine (such as Drools).
+This combination is very efficient, because:
+
+* A rule engine, such as Drools, is *great for calculating the score* of a solution of a planning problem.
+It makes it easy and scalable to add additional soft or hard constraints.
+It does <<incrementalScoreCalculation,incremental score calculation (deltas)>> without any extra code.
+However it tends to be not suitable to actually find new solutions.
+* An optimization algorithm is *great at finding new improving solutions* for a planning problem,
+without necessarily brute-forcing every possibility.
+However, it needs to know the score of a solution and offers no support in calculating that score efficiently.
```
triceo, line 89: I'm not sure here. Do we need to mention Drools? In CS, it is an abstraction. It is a concept they do not need to know about. If things go well, users will never even notice it's there.

optaplanner-docs/src/main/asciidoc/PlannerIntroduction/DownloadAndRunTheExamples/DownloadAndRunTheExamples-section.adoc
```diff
@@ -63,7 +63,7 @@ To run the examples in your favorite IDE:
 
 
 [[useWithMavenGradleEtc]]
-== Use OptaPlanner with Maven, Gradle, Ivy, Buildr, or ANT
+== Use OptaPlanner with Maven, Gradle, or ANT
```
triceo, line 66: Are we keeping ANT then? I didn't notice any `build.xml` in our projects.

triceo (changes requested):
Same comment about BAVET as on the other PR. Please remove the reference.

optaplanner-docs/src/main/asciidoc/OptimizationAlgorithms/OptimizationAlgorithms-chapter.adoc
ge0ffrey: Good point, I fully agree. This is the old text that I just cleaned up a bit and moved, I didn't refactor it along that far.

I'll make those changes.

optaplanner-docs/src/main/asciidoc/PlannerIntroduction/DownloadAndRunTheExamples/DownloadAndRunTheExamples-section.adoc
ge0ffrey: I was thinking of removing ANT too - nobody uses that any more - but we still need to talk about plain jar files that students drop in their IDE without maven or gradle.

ge0ffrey:
Discussed, and agreed to merge this already

PR #1493, opened 2021-08-05
PLANNER-2451 Conditional constraint collectors

triceo:


optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
```diff
@@ -17,6 +17,7 @@
 package org.optaplanner.core.api.score.stream;
 
 import static java.util.Arrays.asList;
+import static java.util.Collections.*;
```
rsynek, line 20: Probably unintentionally changed by the IDE.

Check the settings of star imports; there is usually a threshold from which the IDE uses it. I set mine to 999.

rsynek (approved):
Good to merge after CI is greed and star imports replaced.

optaplanner-core/src/test/java/org/optaplanner/core/api/score/stream/ConstraintCollectorsTest.java
triceo: Thanks. If you don't mind, I'll fix it in a follow-up PR, which deals with the same class.

PR #1498, opened 2021-08-05
PLANNER-2493, PLANNER-2494 Composite and average constraint collector

triceo:


rsynek (approved):

rsynek:
Looks good, but since the tests focus on average collectors only, the underlying composite collectors that take 3 and 4 sub-collectors are not covered at all.

rsynek (approved):

PR #1513, opened 2021-08-19
Improve performance of the experimental collectors

triceo:
Seeing ~10 % performance improvement. Not great, not terrible.


optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
```diff
@@ -40,22 +39,20 @@
...
+public final class ConsecutiveSetTree<Value_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>>
+        implements ConsecutiveInfo<Value_, Difference_> {
     private final Function<Value_, Point_> indexFunction;
     private final BiFunction<Point_, Point_, Difference_> differenceFunction;
     private final BiFunction<Difference_, Difference_, Difference_> sumFunction;
     private final Difference_ maxDifference;
     private final Difference_ zeroDifference;
-    private final Map<Value_, Integer> itemToCountMap;
-    private final NavigableSet<Value_> itemSet;
+    private final NavigableMap<Value_, Integer> itemToCountMap;
```
triceo, line 49: We do not need `itemSet` if you have `itemToCountMap`.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
```diff
@@ -40,22 +39,20 @@
...
     private final NavigableMap<Value_, SequenceImpl<Value_, Difference_>> startItemToSequence;
     private final NavigableMap<Value_, BreakImpl<Value_, Difference_>> startItemToPreviousBreak;
 
     private final MapValuesIterable<Value_, SequenceImpl<Value_, Difference_>> sequenceList;
     private final MapValuesIterable<Value_, BreakImpl<Value_, Difference_>> breakList;
 
-    private final ConsecutiveDataImpl<Value_, Difference_> consecutiveData;
```
triceo, line 57: Instead of copying the data to a new type, I made this type implement the same interface.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/IntervalClusterImpl.java
```diff
@@ -217,4 +169,52 @@ public String toString() {
                 ", hasOverlap=" + hasOverlap +
                 '}';
     }
+
+    // TODO: Make this incremental by only checking between the interval's start and end points
```
triceo, line 173: Maybe we should fix that?

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/IntervalClusterImpl.java
```diff
@@ -102,58 +100,11 @@ public void addInterval(Interval<Interval_, Point_> interval) {
...
-                    end = splitPointSet.lower(current);
-                    current = getStart(current);
-                } else {
-                    end = splitPointSet.last();
-                }
-
-                return new IntervalClusterImpl<>(splitPointSet, differenceFunction, start, end, count, hasOverlap);
-            }
-        };
+    Iterable<IntervalClusterImpl<Interval_, Point_, Difference_>> removeInterval(Interval<Interval_, Point_> interval) {
```
triceo, line 103: Interesting, the `interval` argument is unused.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/ConsecutiveSetTree.java
Christopher-Chianelli: True; I think I added it as a concern for performance, but keySet probably does not take too much time

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/IntervalClusterImpl.java
Christopher-Chianelli: Easier said then done. An interval start point/end point may be under an interval, like so:

------------     --------------
        ------------

which make activeIntervals > 0 at the interval start point. We would either need to memoize the active interval count at all start points, or determine how many active intervals there are before/after it.

Christopher-Chianelli (approved):
LGTM

optaplanner-examples/src/main/java/org/optaplanner/examples/common/experimental/impl/IntervalClusterImpl.java
Christopher-Chianelli: The entire cluster is recomputed since it not easy to get active interval count given a start point (It doesn't recompute all the datapoints, only the datapoints in the same cluster).

PR #1521, opened 2021-08-23
PLANNER-2486: Build KieBase at build time in Quarkus

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-2486

### Checklist

<details>
<summary>
</summary>

</details>


optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -169,14 +173,22 @@ public Boolean getDroolsAlphaNetworkCompilationEnabled() {
...
         this.gizmoKieRuntimeBuilderWrapper = kieRuntimeBuilderWrapper;
     }
 
+    public Supplier<KieBaseDescriptor> getGizmoKieBaseDescriptorWrapper() {
+        return gizmoKieBaseDescriptorWrapper;
+    }
+
+    public void setGizmoKieBaseDescriptorWrapper(Supplier<KieBaseDescriptor> gizmoKieBaseDescriptorWrapper) {
+        this.gizmoKieBaseDescriptorWrapper = gizmoKieBaseDescriptorWrapper;
+    }
```
triceo, line 190: This is a break in public API, we can not have those unfortunately.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/KieBaseDescriptor.java
```diff
@@ -8,7 +8,7 @@
 import org.optaplanner.core.impl.score.inliner.WeightedScoreImpacter;
 import org.optaplanner.core.impl.score.stream.drools.DroolsConstraint;
 
-final class KieBaseDescriptor<Solution_> {
+final public class KieBaseDescriptor<Solution_> {
```
triceo, line 11: I am not prepared to make this a public API. Furthermore, the way you share this type with DRL now makes it kinda ugly - `constraintToGlobalMap` has no meaning there. I think the two wrappers will have to stay separate.

triceo (changes requested):
We can't do it this way, see comments inside. Let's figure out how we can do this.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
Christopher-Chianelli: The break in the public API can be reversed, since the path the code takes is different and they use different fields (one use a supplier of KieBase, another use a supplier of KieBaseDescriptor)

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/KieBaseDescriptor.java
Christopher-Chianelli: The wrappers are separate (one uses `Supplier<KieBase>`, another use `Supplier<KieBaseDescriptor>`. I can try making it `Supplier<Object>` and doing typecasting.

triceo: `KieBaseDescriptor` exposes things that I do not want users to access. The globals are an implementation detail which I absolutely do not want to leak into a public API.

I'm thinking we create some sort of internal config which subtype which is not public and allows us to pass this sort of things. But right now, I'm not sure that's the best thing to do. Will have to think about it a bit more.

triceo:
So, I think I got it:

- We do not change the config API, we keep it as it is.
- We have `KieBaseDescriptor` extend `KieRuntimeBuilderWrapper`, providing the globals on top of it.
- `KieBaseDescriptor` stays where it is, therefore safely outside of any API commitments.
- We modify the config/gizmo code to expect `KieBaseDescriptor` if and only if CS is used, fail fast otherwise.
- And we add a whole bunch of tests, as this is going to be a critical piece of code.

WDYT, @Christopher-Chianelli ?

Christopher-Chianelli:
@triceo I personally not a fan of having `KieBaseDescriptor` extend `KieRuntimeBuilderWrapper`, because it does not wrap a `KieRuntimeBuilder`. It could just set it to null in the constructor and override `getKieBase` to not use it, but ultimately it just a workaround. 

This is what I prefer to do (and just finished testing locally)
```
1. revert KieRuntimeBuilderWrapper changes (i.e. the field/method stays and behave like before)
2.  add a field `Supplier<?>  kieBaseDescriptorSupplier` in the ScoreDirectorConfig
3.   Keep `KieBaseDescriptor` package-private, and refer to it elsewhere as Object
4. Add a new constructor DroolsConstraintFactory(SolutionDescriptor, Supplier, boolean)
```

triceo:
OK, let's try a compromise, which I think may even be slightly cleaner:

- We deprecate the existing config method. We can not remove it, but we can live with a deprecation.
- We introduce a new method that accepts `Supplier<KieBase>`. The old method delegates to this one.
- Both the wrapper and the descriptor implement `Supplier<KieBase>`, which allows them both to be consumed by the new config method, no magic necessary, no `<?>` necessary.

This way, at some point in time in the future, we'll even be able to hide the wrapper type, or rather treat it as an implementation detail.

Christopher-Chianelli:
@triceo done

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -169,12 +171,22 @@ public Boolean getDroolsAlphaNetworkCompilationEnabled() {
         return droolsAlphaNetworkCompilationEnabled;
     }
 
+    @Deprecated
```
triceo, line 174:
```suggestion
        @Deprecated(forRemoval = true)
```

Also please add Javadoc along the lines of "@deprecated in favor of ...".

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -169,12 +171,22 @@ public Boolean getDroolsAlphaNetworkCompilationEnabled() {
...
+        setKieBaseSupplier(kieRuntimeBuilderWrapper);
+    }
+
+    public Supplier<KieBase> getKieBaseSupplier() {
+        return kieBaseSupplier;
+    }
+
+    public void setKieBaseSupplier(Supplier<KieBase> kieBaseSupplier) {
+        this.kieBaseSupplier = kieBaseSupplier;
     }
```
triceo, line 190: This could use a Javadoc describing its purpose. 
Specifically, that people really don't need to use this at all.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -269,8 +281,9 @@ public ScoreDirectorFactoryConfig withScoreDrlFiles(File... scoreDrlFiles) {
         return this;
     }
 
+    @Deprecated
```
triceo, line 284: Same deprecation comment as above.

triceo: Also, I do not see a `with...` variant for the supplier.

triceo (commented):
Much better! See comments.

Also I'm missing fail fasts for situations:
- when the supplier is provided, but the scoring function is neither CS-D nor DRL.
- when the supplier is provided, but we are outside of Quarkus.

We do not want people to use this. And that means preventing them from doing it.

optaplanner-core/src/main/java/org/optaplanner/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -169,12 +171,22 @@ public Boolean getDroolsAlphaNetworkCompilationEnabled() {
...
+        return (KieRuntimeBuilderWrapper) getKieBaseSupplier();
     }
 
+    @Deprecated
     public void setGizmoKieRuntimeBuilderWrapper(KieRuntimeBuilderWrapper kieRuntimeBuilderWrapper) {
-        this.gizmoKieRuntimeBuilderWrapper = kieRuntimeBuilderWrapper;
+        setKieBaseSupplier(kieRuntimeBuilderWrapper);
+    }
+
+    public Supplier<KieBase> getKieBaseSupplier() {
```
triceo, line 184: Maybe we rename to `...GizmoKieBaseSupplier`, so that the point becomes more obvious?

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -545,6 +556,40 @@ public static String generateKieRuntimeBuilder(ClassOutput classOutput, SolverCo
...
+                        methodCreator.load(DomainAccessType.REFLECTION),
+                        methodCreator.loadClass(config.getSolutionClass()),
+                        methodCreator.loadNull(),
+                        methodCreator.loadNull(),
+                        entityClassList);
+                ResultHandle isDroolsAlphaNetworkCompilationEnabled =
+                        methodCreator.load(config.getScoreDirectorFactoryConfig().isDroolsAlphaNetworkCompilationEnabled());
+                ResultHandle kieBaseDescriptor = methodCreator.invokeStaticMethod(
+                        MethodDescriptor.ofMethod(DroolsConstraintStreamScoreDirectorFactory.class, "buildKieBase",
+                                "org.optaplanner.core.impl.score.director.stream.KieBaseDescriptor",
```
triceo, line 585: I think we can avoid this if we make `KieBaseDescriptor` public.
We can do that, because it is in a package which we consider internal, and therefore not give backwards compatibility guarantees.

Christopher-Chianelli:
@triceo how am I suppose to check we are in Quarkus in the ScoreDirector (w/o doing forName on Quarkus classes)?

triceo:
Good point. I thought I remembered that there was a way, but that was actually for native image; never mind then.

triceo (approved):

triceo:
jenkins run native tests

Christopher-Chianelli:
@triceo it seems the KieBase in the recorder throws an error in native mode (from the java parser?????). I can disable this path in native.

triceo:
I think I have an idea of what's happening. The ANC. 

It is not supported in native, but when we are in Quarkus, we are creating a ANC-compiled KieBase regardless of whether we will or will not run native. Therefore the failure.

It's an unconfirmed theory for now. And even if it were true, I'm not sure what we could do about it - other than disabling ANC whenever we are compiling a KieBase with Quarkus.

triceo:
I think I know what to do.

In `KieBaseDescriptor`, we store both KieBases. One with ANC, the other one without.
Then when we run in native (which we already have a check for), we will choose the one without ANC.

That should hopefully allow us to use ANC in Quarkus, but not in Quarkus native.

Christopher-Chianelli:
@triceo or.... simply check if we are in a native build, and if so, disable anc?

triceo:
If that works, even better. :-) I'd maybe just log a warning that ANC is disabled in that case, if it would have otherwise been enabled.

Christopher-Chianelli:
jenkins run native tests

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/drools/KieRuntimeBuilderWrapper.java
```diff
@@ -34,4 +36,8 @@ public KieBase extractKieBase() {
         return kieRuntimeBuilder.getKieBase();
```
triceo, line 36: Please inline this method.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/KieBaseDescriptor.java
```diff
@@ -27,4 +28,9 @@ public KieBase getKieBase() {
         return kieBase;
```
triceo, line 28: Please inline this method.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -231,7 +254,15 @@ SolverConfigBuildStep recordAndRegisterBeans(OptaPlannerRecorder recorder, Recor
...
         additionalBeans.produce(new AdditionalBeanBuildItem(OptaPlannerBeanProvider.class));
         unremovableBeans.produce(UnremovableBeanBuildItem.beanTypes(OptaPlannerRuntimeConfig.class));
-        return new SolverConfigBuildStep(solverConfig);
+        return new SolverConfigBuildItem(solverConfig);
+    }
+
+    private void disableANC(SolverConfig solverConfig) {
+        if (solverConfig.getScoreDirectorFactoryConfig().getDroolsAlphaNetworkCompilationEnabled() != null
+                && solverConfig.getScoreDirectorFactoryConfig().getDroolsAlphaNetworkCompilationEnabled()) {
+            log.warn("Drools ANC is disabled since this is a native build");
```
triceo, line 263:
```suggestion
            log.warn("Disabling Drools Alpha Network Compiler since this is a native build.");
```

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
```diff
@@ -231,7 +254,15 @@ SolverConfigBuildStep recordAndRegisterBeans(OptaPlannerRecorder recorder, Recor
 
         additionalBeans.produce(new AdditionalBeanBuildItem(OptaPlannerBeanProvider.class));
         unremovableBeans.produce(UnremovableBeanBuildItem.beanTypes(OptaPlannerRuntimeConfig.class));
-        return new SolverConfigBuildStep(solverConfig);
+        return new SolverConfigBuildItem(solverConfig);
+    }
+
+    private void disableANC(SolverConfig solverConfig) {
+        if (solverConfig.getScoreDirectorFactoryConfig().getDroolsAlphaNetworkCompilationEnabled() != null
+                && solverConfig.getScoreDirectorFactoryConfig().getDroolsAlphaNetworkCompilationEnabled()) {
```
triceo, line 262: Shouldn't it be enough to check for `== true`?

triceo (approved):
Some final comments.
Also, if you could increase coverage over the 80 % threshold, that'd be appreciated.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/main/java/org/optaplanner/quarkus/deployment/OptaPlannerProcessor.java
Christopher-Chianelli: NPE if getDroolsAlphaNetworkCompilationEnabled is null (it Boolean, not boolean)

triceo: Interesting, I honestly thought that `null == true` is `false`, not an NPE. Well, I learn every day.

Christopher-Chianelli: It might be; I know `if(Boolean)` can lead to a NPE if the boolean is null, but that might be handled differently then `if(Boolean == true)`

Christopher-Chianelli: It the same for the `Boolean == true` case (auto unboxing): https://stackoverflow.com/a/25824276

Christopher-Chianelli:
jenkins run native tests

PR #1523, opened 2021-08-25
Improve performance of Constraint Streams on select use cases

triceo:
Improves performance of checking for disabled constraints by caching the constraint ID.
Improves performance of checking if an entity is initialized, by using a `List` instead of `Map.values()`.
The latter change has the potential to speed up many other parts of OptaPlanner as well, although that was not benchmarked.


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/stream/DroolsConstraintStreamScoreDirectorFactory.java
```diff
@@ -180,7 +181,14 @@ public boolean accept(Match match) {
                 return true;
             }
             Rule rule = match.getRule();
-            String constraintId = ConstraintMatchTotal.composeConstraintId(rule.getPackageName(), rule.getName());
+            // We identify the rule by its constraint ID, which we pre-calculated during rule creation.
+            // The alternative is to pay string concat penalty (packageName + name) to calculate the ID on every rule evaluation.
+            // Since this code is on the hot path, this optimization was confirmed to bring considerable benefits.
```
rsynek, line 186: Maybe a multi-line comment is more appropriate here (not important).

rsynek (approved):

triceo:
Failures unrelated, therefore merged.

PR #1569, opened 2021-09-14
PLANNER-2459 Control solver lifecycle from within the solver

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -36,6 +36,8 @@
 public class DefaultConstructionHeuristicPhase<Solution_> extends AbstractPhase<Solution_>
         implements ConstructionHeuristicPhase<Solution_> {
 
+    private final BestSolutionRecaller<Solution_> bestSolutionRecaller;
```
triceo, line 39: Some phases still need the recaller, because they call their own special methods on it.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -184,10 +185,8 @@ private ExecutorService createThreadPoolExecutor(int partCount) {
             phaseList.add(phase);
             partPhaseIndex++;
         }
-        // TODO create PartitionSolverScope alternative to deal with 3 layer terminations
-        SolverScope<Solution_> partSolverScope = solverScope.createChildThreadSolverScope(ChildThreadType.PART_THREAD);
-        partSolverScope.setRunnableThreadSemaphore(runnablePartThreadSemaphore);
-        return new PartitionSolver<>(bestSolutionRecaller, partTermination, phaseList, partSolverScope);
+        return new PartitionSolver<>(bestSolutionRecaller, partTermination, phaseList, solverScope,
+                runnablePartThreadSemaphore);
```
triceo, line 189: `SolverScope` now carries a reference to the `Solver`, therefore it is created within the solvers, not externally.

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/scope/AbstractMoveScope.java
```diff
@@ -48,11 +48,11 @@ public int getMoveIndex() {
...
-    public <Score_ extends Score<Score_>> Score_ getScore() {
-        return (Score_) score;
+    public Score getScore() {
+        return score;
     }
 
-    public <Score_ extends Score<Score_>> void setScore(Score_ score) {
+    public void setScore(Score score) {
         this.score = score;
     }
```
triceo, line 57: This is unrelated, but these generics were pointless and were causing unnecessary warnings in some tests.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/AbstractSolver.java
```diff
@@ -67,9 +67,6 @@ public AbstractSolver(BestSolutionRecaller<Solution_> bestSolutionRecaller, Term
         this.termination = termination;
         bestSolutionRecaller.setSolverEventSupport(solverEventSupport);
         this.phaseList = phaseList;
-        for (Phase<Solution_> phase : phaseList) {
-            phase.setSolverPhaseLifecycleSupport(phaseLifecycleSupport);
-        }
```
triceo, line 72: No longer needed, as the solver now calls its phaselifecycle support itself.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/scope/SolverScope.java
```diff
@@ -57,6 +61,18 @@
     // Constructors and simple getters/setters
     // ************************************************************************
 
+    public SolverScope() { // For testing purposes.
+        this.solver = null;
+    }
```
triceo, line 66: Otherwise I'd have to change a lot of tests, adding the `null` there.

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/DefaultPartitionedSearchPhase.java
```diff
@@ -184,10 +185,8 @@ private ExecutorService createThreadPoolExecutor(int partCount) {
             phaseList.add(phase);
             partPhaseIndex++;
         }
-        // TODO create PartitionSolverScope alternative to deal with 3 layer terminations
-        SolverScope<Solution_> partSolverScope = solverScope.createChildThreadSolverScope(ChildThreadType.PART_THREAD);
-        partSolverScope.setRunnableThreadSemaphore(runnablePartThreadSemaphore);
```
ge0ffrey, line 189: nitpick: Here setRunnableThreadSemaphore() is replaced for a constructor parameter. Which is usually a good thing. 1/2

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
```diff
@@ -129,16 +113,21 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
...
+        phaseScope.getSolverScope().getSolver().phaseStarted(phaseScope);
         termination.phaseStarted(phaseScope);
         phaseLifecycleSupport.firePhaseStarted(phaseScope);
     }
 
+    @Override
+    public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
+        phaseScope.getSolverScope().getSolver().phaseEnded(phaseScope);
+        termination.phaseEnded(phaseScope);
+        phaseLifecycleSupport.firePhaseEnded(phaseScope);
```
ge0ffrey, line 125: (important) Shouldn't this be inside-out? so reversed order as in the original code.

Or maybe not? The phase lifecycle vents might want to use the solving end time. Other the hand, the solving end time might need to include the execution of the ending events?

Take a look at other phaseStarted vs phaseEnded implementations. Does the end phase do things in the same order or in the reverse order of start events?

Something to discuss! We need a general guideline for this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/AbstractSolver.java
```diff
@@ -114,6 +111,54 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
         phaseLifecycleSupport.fireSolvingEnded(solverScope);
     }
 
+    /**
+     * Used by a phase to notify the solver of its start.
+     * The event should not propagate any further, as that is the job of the phase.
```
ge0ffrey, line 116: Confusing. They do propagate  further to termination etc.
I'd just add a comment inside the method at the end, something like:
`// Do not propagate to phases, because the active phase does that for itself and they should not propagate to other phases`

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/AbstractSolver.java
```diff
@@ -114,6 +111,54 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
...
+     *
+     * @param phaseScope never null
+     */
+    public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
+        bestSolutionRecaller.phaseStarted(phaseScope);
+        phaseLifecycleSupport.firePhaseStarted(phaseScope);
+        termination.phaseStarted(phaseScope);
```
ge0ffrey, line 123: (important) Order, shouldn't termination be called before phaseLifecycleSupport? See original code.

Given that phaseLifecycleSupport is abstract/unknown, it's unlikely that termination would depend on anything in there. But visa versa, it's far more likely that that depends on termination.

ge0ffrey: Applies to other 3 methods too.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/scope/SolverScope.java
```diff
@@ -35,6 +37,8 @@
  */
 public class SolverScope<Solution_> {
 
+    private final AbstractSolver<Solution_> solver;
+
```
ge0ffrey, line 41: veto -1 (hard)

Scopes are data objects and cannot reference services. Oh wait, it does now ScoreDirector and Random.

Well, still Solver knows SolverScope, but not visa versa. Let's discuss?

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/PartitionSolver.java
```diff
@@ -38,9 +40,11 @@
     // ************************************************************************
 
     public PartitionSolver(BestSolutionRecaller<Solution_> bestSolutionRecaller, Termination<Solution_> termination,
-            List<Phase<Solution_>> phaseList, SolverScope<Solution_> solverScope) {
+            List<Phase<Solution_>> phaseList, SolverScope<Solution_> solverScope, Semaphore runnablePartThreadSemaphore) {
         super(bestSolutionRecaller, termination, phaseList);
-        this.solverScope = solverScope;
+        // TODO create PartitionSolverScope alternative to deal with 3 layer terminations
+        this.solverScope = solverScope.createChildThreadSolverScope(this, ChildThreadType.PART_THREAD);
+        this.solverScope.setRunnableThreadSemaphore(runnablePartThreadSemaphore);
```
ge0ffrey, line 47: I don't understand where this code is being refactored from (and why).

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
ge0ffrey: For my understanding: which phases call which special methods on it?

ge0ffrey: I guess the CH because it doesn't do best solution events for every step?

ge0ffrey (approved):

ge0ffrey:
Looks good, great work :)

PR #1585, opened 2021-09-22
PLANNER-2526 Build compatibility with JDK 17

triceo:


optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/LinearDistributionNearbyRandomTest.java
```diff
@@ -34,33 +33,33 @@ public void sizeMaximumTooLow() {
...
     @Test
     public void nextInt() {
-        Random random = mock(Random.class);
+        Random random = new TestRandom(
+                0.0,
+                2.0 / 100.0,
+                2.0 / 100.0 + 2.0 / 100.0 + 2.0 / 10000.0,
+                2.0 / 100.0 + 2.0 / 100.0 + 2.0 / 10000.0 + 2.0 / 100.0 + 4.0 / 10000.0,
+                0.0,
+                2.0 / 10.0);
```
ge0ffrey, line 42: (nitpick) any reason why we're doing these all at ones and not per method like in the other code?

This might be easier to read and maintain:
```
assertThat(nearbyRandom.nextInt(new TestRandom(0.0), 500)).isEqualTo(0);
assertThat(nearbyRandom.nextInt(new TestRandom(2.0 / 100.0), 500)).isEqualTo(1);
...
```

optaplanner-core/src/test/java/org/optaplanner/core/impl/util/ScopeUtils.java
```diff
@@ -0,0 +1,45 @@
+package org.optaplanner.core.impl.util;
+
+import org.optaplanner.core.impl.phase.scope.AbstractPhaseScope;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
+import org.optaplanner.core.impl.solver.scope.SolverScope;
+
+public final class ScopeUtils {
```
ge0ffrey, line 7: A) For clarity, can we rename it to ScopeTestUtils?

B) Or just put these methods in PlannerTestUtils instead?

ge0ffrey: I prefer B), so it's easier to find all these test helper methods, but I don't mind A).

optaplanner-core/src/test/java/org/optaplanner/core/impl/util/TestRandom.java
```diff
@@ -0,0 +1,223 @@
...
+
+import org.junit.jupiter.api.Assertions;
+
+/**
+ * On the later JDKs, it is no longer possible to mock {@link Random} to return custom sequences.
+ * Therefore we introduce this class to allow for that use case.
+ *
+ * It allows to provide a sequence of pre-defined "random" values.
+ * It maintains an internal count of values already returned and if larger than the sequence provided,
+ * the last element in the sequence is returned.
```
ge0ffrey, line 19: Shouldn't it fail fast if it runs out?

ge0ffrey (approved):
LGTM (see comments though)

optaplanner-core/src/test/java/org/optaplanner/core/impl/util/TestRandom.java
ge0ffrey: If a test uses more random values than were provided in the JUnit class, we probably want that to fail fast. Let's not take any chances?

PR #1586, opened 2021-09-22
Unregister the old meters in DefaultSolver when changing tags

Christopher-Chianelli:
### Checklist

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
@@ -40,6 +40,7 @@ import org.optaplanner.core.impl.solver.termination.Termination;
 
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.Tags;
 
@@ -64,7 +65,6 @@ public class DefaultSolver<Solution_> extends AbstractSolver<Solution_> {
     private final String moveThreadCountDescription;
 
     // Metrics
-    private Tags monitoringTags;
     private LongTaskTimer solveLengthTimer;
     private Counter errorCounter;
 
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
```diff
@@ -163,15 +162,22 @@ public boolean isEveryProblemFactChangeProcessed() {
...
 
     public void setMonitorTagMap(Map<String, String> monitorTagMap) {
-        monitoringTags = ObjectUtils.defaultIfNull(monitorTagMap, Collections.<String, String> emptyMap())
+        Tags monitoringTags = ObjectUtils.defaultIfNull(monitorTagMap, Collections.<String, String> emptyMap())
                 .entrySet().stream().map(entry -> Tags.of(entry.getKey(), entry.getValue()))
                 .reduce(Tags.empty(), Tags::and);
-        solverScope.setMonitoringTags(monitoringTags);
         unregisterMetrics();
+        solverScope.setMonitoringTags(monitoringTags);
         registerMetrics();
```
triceo, line 170: This is not particularly clean. 

First you unregister tags here by querying solver scope for them.
Then you set some other tags to the solver scope.
And then you one again read them to register them with this class.

I'd argue that all these operations should be happening from one place, and the data should be encapsulated there.
The more you "leak" internals to other classes, the harder it becomes over time to figure out who does what with them where.

triceo (approved):
Good to go as it is a bugfix, but in general, I think the architecture could be improved.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolver.java
Christopher-Chianelli: solverScope is the single source of tags (since these tags also need to be accessed in phase listeners + other places in the code). The metrics in `Solver` are unique in that they are created before `solving` (for performance reasons, Micrometer recommends storing meters in fields if possible), so they are the only meters that can change after they are created. These operations probably won't be needed if we make the fields local to the `solve` call (at a very sight performance penalty).

triceo: Doing it in `solve` probably isn't such a bad idea, provided that these operations are really cheap.

Christopher-Chianelli:
Alternate fix: #1588 ; separate PR since it does make tests more complicated and changes semantics sightly

PR #1592, opened 2021-09-30
PLANNER-1940 Update best solution during construction heuristics

triceo:


optaplanner-docs/src/modules/ROOT/pages/optimization-algorithms/optimization-algorithms.adoc
```diff
@@ -459,8 +459,7 @@ Alternatively to a `java.util.Duration` in ISO 8601 format, you can also use:
 
 Just like <<timeMillisSpentTermination,time spent termination>>, combinations are summed up.
 
-This termination should not be applied to Construction Heuristics as they only update the best solution at the end.
-Configuring it on a specific `Phase` (such as ``<localSearch>``), instead of on the `Solver` itself is often a better option.
+Configuring this termination on a specific `Phase` (such as ``<localSearch>``) instead of on the `Solver` itself is often a better option.
```
ge0ffrey, line 462: the word "better" doesn't work any more, because the previous sentence is gone, to compare with.

ge0ffrey, line 462: the word "better" doesn't work any more, because the previous sentence is gone, to compare with.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -108,19 +117,26 @@ public void processWorkingSolutionDuringMove(Score score, AbstractStepScope<Solu
...
         Solution_ newBestSolution = solverScope.getScoreDirector().cloneWorkingSolution();
         Score newBestScore = solverScope.getSolutionDescriptor().getScore(newBestSolution);
-        updateBestSolution(solverScope, newBestScore, newBestSolution);
+        updateBestSolutionAndFire(solverScope, newBestScore, newBestSolution);
     }
 
-    protected void updateBestSolution(SolverScope<Solution_> solverScope, Score bestScore, Solution_ bestSolution) {
+    private void updateBestSolutionAndFire(SolverScope<Solution_> solverScope, Score bestScore, Solution_ bestSolution) {
+        // We do not want construction heuristics to fire intermediate best solution changed events.
+        // We still want best solution and score updated though, so that unimproved* terminations work correctly.
```
ge0ffrey, line 134: I'd move these 2 comments in

Code style: comments shouldn't use "we" and should be written more directive (= shorter):
"We do not want construction heuristics to fire intermediate best solution changed events."
=>  "Construction heuristics don't fire intermediate best solution changed events."
"We still want best solution and score updated though, so that unimproved* terminations work correctly."
=> "But the best solution and score are updated, so that unimproved* terminations work correctly."

_Speak to peer coders as an Englishman, write code comments as a German._

ge0ffrey (approved):
LGTM, consider moving those comments into the CH code.

PR #1594, opened 2021-10-01
Make KieModuleMarshaller run time initialized

galderz:
Closes https://github.com/quarkusio/quarkus/issues/20491

A simple way to fix the issue is to make `KieModuleMarshaller` run-time initialized.

@Sanne mentioned [here](https://github.com/quarkusio/quarkus/issues/11563#issuecomment-926479069):
> I had a quick look at java.awt.datatransfer.DataFlavor.. I could be wrong but it looks like a perfect candidate to be optimized via build time initialization.

Not sure exactly what he means, but at a first glance, everything awt related/dependant should really be runtime initialized to start with. Then, you can always optimize further things as needed/required.


triceo:
ok to test

triceo:
jenkins run native tests

triceo (approved):
Good when green, @galderz.
If something fails, and it likely will, I'll look into it - most likely, it will be some flaky test in one of the downstream repos.

PR #1612, opened 2021-10-14
Fix docs upload rsync failure

ge0ffrey:


ge0ffrey:
The fix was merged for 8.12.x: https://github.com/kiegroup/optaplanner/pull/1613
so when that works for the release, we can merge it here too.

PR #1643, opened 2021-10-30
[KOGITO-2391] introducing ReteEvaluator

mariofusco:
Related pull request:
https://github.com/kiegroup/drools/pull/3961
https://github.com/kiegroup/kogito-runtimes/pull/1688
https://github.com/kiegroup/optaplanner/pull/1643


triceo (approved):
Considering that we never actually use the argument, I wonder whether there should be an overload where you don't need to send it. Anyway, LGTM.

triceo:
(You can ignore the JPA CI failure; it does not use build chain and therefore does not see the upstream API changes.)

PR #1645, opened 2021-11-03
List variable: construction heuristic

yurloc:
### JIRA
- https://issues.redhat.com/browse/PLANNER-2515 (moves)
- https://issues.redhat.com/browse/PLANNER-2513 (initScore)
- https://issues.redhat.com/browse/PLANNER-2543 (config)

### TODOs
- [x] Tests.
- [x] Fail fast if CH type is specified?
- [x] Decide what to do with the performance degradation → [PLANNER-2582](https://issues.redhat.com/browse/PLANNER-2582).

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/placer/QueuedValuePlacerFactory.java
```diff
@@ -53,7 +53,7 @@ public QueuedValuePlacerFactory(QueuedValuePlacerConfig placerConfig) {
         ValueSelectorConfig valueSelectorConfig_ = buildValueSelectorConfig(configPolicy, entityDescriptor);
         ValueSelector<Solution_> valueSelector = ValueSelectorFactory.<Solution_> create(valueSelectorConfig_)
                 .buildValueSelector(configPolicy, entityDescriptor, SelectionCacheType.PHASE, SelectionOrder.ORIGINAL,
-                        false);
+                        false, true);
```
yurloc, line 56: I hate to do this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/IndexVariableListener.java
```diff
@@ -48,9 +48,7 @@ public void afterEntityAdded(ScoreDirector<Solution_> scoreDirector, Object enti
 
     @Override
     public void beforeVariableChanged(ScoreDirector<Solution_> scoreDirector, Object entity) {
-        // TODO maybe set all indexes to null
-        // When we support over-constrained planning (e.g. there is not enough Employees to assign all tasks), then
-        // an unassigned planning value should have a null index.
+        retract((InnerScoreDirector<Solution_, ?>) scoreDirector, entity);
```
yurloc, line 51: This has a big performance impact but is necessary to avoid undo score corruption during CH.

TODO benchmark this.

yurloc:
What about `<constructionHeuristicType>` - fail fast?

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/IndexVariableListener.java
yurloc: Disaster.






_ | init-24tasks-8employees | init-50tasks-5employees | init-100tasks-5employees | init-500tasks-20employees
-- | -- | -- | -- | --
Before this PR | 58,635/s | 31,746/s | 21,352/s | 12,865/s
Now | 33,098/s | 10,762/s | 7,035/s | 3,766/s

triceo: Although I see that these numbers are indeed disastrous, does it matter what they were before, if before was wrong?
How do the numbers look comparing to the current implementation?

yurloc: Updated. The list variable was already slightly worse than chained variable before this PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/placer/QueuedValuePlacerFactory.java
yurloc: Added a couple of tests for bugs that I ran into during development. With the extra coverage I can look if this can be refactored without the fear of breaking something.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/EntityIndependentValueSelector.java
```diff
@@ -24,4 +26,14 @@
...
+    /**
+     * If {@link #isNeverEnding()} is true, then {@link #iterator()} will never end.
+     * This returns an ending {@link Iterator}, that tries to match {@link #iterator()} as much as possible,
+     * but return each distinct element only once
+     * and therefore it might not respect the configuration of this {@link ValueSelector} entirely.
+     *
+     * @return never null
+     * @see #iterator()
+     */
+    Iterator<Object> endingIterator();
```
yurloc, line 38: This allows me to easily build an ending iterator for the `ReassignValueToListVariableSelector`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/placer/QueuedValuePlacerFactory.java
yurloc: Not going to improve this in this PR.

# How filtering works

The `false, true` flags are overrides of the reinitializable/reassignable filtering behavior that is an optional feature of the **ValueSelector** (keep this in mind).

The reinitializable filtering is specified by the `configPolicy`, which is an approach that covers most situations.

The reassign filtering is only used by the list variable and the Allocate to value from queue algorithm so there's not much point in adding it to the `configPolicy`. This line is the only place where the reassignable filtering needs to be enabled or disabled.

## 1. Local search
In the Local Search phase, reinitializable/reassignable filtering is not needed because the solution is already initialized. The configPolicy doesn't enable it and its setting is honored.

## 2. Construction heuristics
In the Consturction Heuristic (and Exhaustive Search) phase this type of filtering is needed so that only uninitialized entities and unassigned values are touched. The configPolicy enables the filtering but in some cases (covered by this line) it cannot be honored. See below.

### 2.1. Allocate entity from queue
The Allocate entity from queue algorithm (including the six specialized "Fit" algorithms) uses a single ValueSelector. The configPolicy enables the filtering and the setting is honored. This is the simple case.

### 2.2. Allocate to value from queue
The Allocate to value from queue algorithm uses ValueSelector mimicking so it has two ValueSelectors. Only one of them must use the filtering.

#### 2.2.1. Basic variable
The recording (leading) ValueSelector doesn't use filtering, so the configPolicy setting is being overridden here using the `false` flag. The replaying ValueSelector honors the configPolicy setting and uses reinitializable filtering.

#### 2.2.2. List variable
The recording ValueSelector uses reassignable filtering to avoid selecting values that are already assigned. The filtering is enabled here by the `true` flag. Otherwise, it's not used. That includes the replaying ValueSelector and any ValueSelectors created in the Local Search phase.

# Conclusion
The decision whether to use which filtering is rather complex and is quiet difficult to track down. The code doesn't make it apparent. If the situation can be improved, it won't be without substantial refactoring. This is out of scope of this PR.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/IndexVariableListener.java
yurloc: Vehicle routing example, main (chained) vs. list variable modifications on top of this branch:


_ | cvrp-n181-k23 | cvrp-n344-k43 | cvrp-n627-k43 | cvrp-n1001-k43
-- | -- | -- | -- | --
main (chained) | 52,377/s | 47,495/s | 37,899/s | 33,887/s
list variable | 27,904/s | 25,516/s | 14,772/s | 10,705/s

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListAssignMove.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(variableDescriptor, planningValue, destinationEntity, destinationIndex);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%s {unassigned->%s[%d]}",
```
ge0ffrey, line 107: 1) We want to carefully design these toStrings to be consistent with the other build-in move toStrings,
because they escape into the logs and are a bit therefore something that users get used to being consistent (semi API).
That being said, this looks good, except for a space:

```suggestion
        return String.format("%s {unassigned -> %s[%d]}",
```

2) I like the use of format to make it easier to read. We probably want to use this for older moves too.
OTH, is there a format parsing cost to be paid? So maybe we shouldn't use it here?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelector.java
```diff
@@ -73,7 +75,11 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
     public long getSize() {
         long entityCount = entitySelector.getSize();
         long valueCount = valueSelector.getSize();
-        return valueCount * (valueCount + entityCount);
+        int assignedValueCount = StreamSupport
+                .stream(Spliterators.spliterator(entitySelector.endingIterator(), entitySelector.getSize(), 0), false)
```
ge0ffrey, line 79: Would you mind walking me through this in our Friday meeting?

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListUnassignMove.java
```diff
@@ -0,0 +1,95 @@
...
+            ListVariableDescriptor<Solution_> variableDescriptor,
+            Object sourceEntity, int sourceIndex) {
+        this.variableDescriptor = variableDescriptor;
+        this.sourceEntity = sourceEntity;
+        this.sourceIndex = sourceIndex;
+    }
+
+    @Override
+    protected AbstractMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
+        return null;
```
ge0ffrey, line 42: We better implement this. For undo tabu search maybe.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/OriginalListChangeIterator.java
```diff
@@ -80,6 +81,14 @@ public OriginalListChangeIterator(
             destinationIndexIterator = listIndexIterator(upcomingDestinationEntity);
         }
 
+        if (upcomingSourceEntity == null && upcomingSourceIndex == null) {
+            return new ListAssignMove<>(
+                    listVariableDescriptor,
+                    upcomingValue,
+                    upcomingDestinationEntity,
+                    destinationIndexIterator.nextInt());
+        }
```
ge0ffrey, line 90: It's interesting that there are no unassigned move created in this iterator. It makes sense (unless nullable=true which we don't support yet), so this is good.
Maybe we can add a comment about that. I took me a while to realize this is not a bug.

```suggestion
        }
        // No need to generate ListUnassignMove because they are only used as undo moves.
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListAssignMove.java
ge0ffrey: Actually, in ChangeMove logs, unassigned is written as null IIRC: "Mondayshift { null -> Ann }".
Let's follow that? It's less likely that a (dummy?) shift.toString returns "null" than it does "unassigned", which would lead to confusion.

```suggestion
        return String.format("%s {null -> %s[%d]}",
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListUnassignMove.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(variableDescriptor, sourceEntity, sourceIndex);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%s {%s[%d]->unassigned}",
```
ge0ffrey, line 92:
```suggestion
        return String.format("%s {%s[%d] -> null}",
```

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning_listvariable/domain/solver/TaskDifficultyComparator.java
```diff
@@ -29,6 +29,6 @@
 
     @Override
     public int compare(Task a, Task b) {
-        return COMPARATOR.compare(a, b);
+        return -COMPARATOR.compare(a, b);
```
yurloc, line 32: Both difficulty and strength comparators must sort elements in ascending order.

The original `TaskDifficultyComparator` is now used to sort _values_ by _strength_. Stronger means "less difficult", hence the negation that reverses the original ascending difficulty order and makes it an ascending strength order.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListAssignMove.java
yurloc: > I like the use of format to make it easier to read. We probably want to use this for older moves too.
OTH, is there a format parsing cost to be paid? So maybe we shouldn't use it here?

I don't think the performance of moves' `toString()` methods is an issue. AFAIK, moves are not printed at all on `INFO` level, where you care about the performance.

On `DEBUG` a move is printed once every step, which seems like a good compromise.

Only on `TRACE` level, every move is printed, but that level shouldn't be used in performance-sensitive environments.

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/variable/PlanningCollectionVariable.java
```diff
@@ -57,5 +59,37 @@
...
-    // TODO value strength
+    /**
+     * Allows a collection of planning values for this variable to be sorted by strength.
+     * A strengthWeight estimates how strong a planning value is.
+     * Some algorithms benefit from planning on weaker planning values first or from focusing on them.
+     * <p>
+     * The {@link Comparator} should sort in ascending strength.
+     * For example: sorting 3 computers on strength based on their RAM capacity:
+     * Computer B (1GB RAM), Computer A (2GB RAM), Computer C (7GB RAM),
+     * <p>
```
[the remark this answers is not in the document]
yurloc: Done.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListAssignMove.java
yurloc: Proposed format changes incorporated.

optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/placer/QueuedValuePlacerFactory.java
yurloc: TODO add a TODO.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/decorator/ReinitializeVariableValueSelector.java
```diff
@@ -82,26 +78,30 @@ public boolean isNeverEnding() {
...
-            return Collections.emptyIterator();
+        if (isReinitializable(entity)) {
+            return childValueSelector.endingIterator(entity);
         }
-        return childValueSelector.endingIterator(entity);
+        return Collections.emptyIterator();
+    }
+
+    private boolean isReinitializable(Object entity) {
+        return childValueSelector.getVariableDescriptor().isReinitializable(scoreDirector, entity);
```
yurloc, line 104: Let's inline reinit filter (inside genuine variable descriptor).

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/decorator/ReassignValueToListVariableSelector.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Prevents reassigning of values that are already assigned to a list variable during Construction Heuristics.
+ * <p>
+ * Returns only values that are not yet assigned.
+ * <p>
+ * Does implement {@link EntityIndependentValueSelector} because the question whether a value is assigned or not does not depend
+ * on a specific entity.
+ */
+public class ReassignValueToListVariableSelector<Solution_> extends AbstractValueSelector<Solution_>
```
yurloc, line 39: UnassignedValueSelector

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/decorator/ReinitializeVariableValueSelector.java
```diff
@@ -38,14 +37,11 @@
 public class ReinitializeVariableValueSelector<Solution_> extends AbstractValueSelector<Solution_> {
```
yurloc, line 37: reinitializable

yurloc:
TODO file a JIRA for eliminating creation of undo moves of undo moves.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/EntityIndependentValueSelector.java
yurloc: Actually not needed. The `ReassignValueToListVariableSelector` requires that its `childValueSelector` is not never-ending, so it can use the good old `iterator()`.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/decorator/ReinitializeVariableValueSelector.java
yurloc: https://issues.redhat.com/browse/PLANNER-2590.

optaplanner-core/src/test/java/org/optaplanner/core/impl/testdata/util/CodeAssertable.java
```diff
@@ -53,6 +54,12 @@ static CodeAssertable convert(Object o) {
             }
             final String code = codeBuilder.substring(1);
             return () -> code;
+        } else if (o instanceof ListAssignMove) {
+            ListAssignMove<?> listAssignMove = (ListAssignMove<?>) o;
+            return () -> convert(listAssignMove.getMovedValue())
+                    + " {unassigned->"
```
yurloc, line 60: FIXME.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/value/decorator/ReinitializeVariableValueSelector.java
yurloc: Deferred to avoid merge conflicts: https://issues.redhat.com/browse/PLANNER-2592.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/list/ListUnassignMove.java
yurloc: -1.

Assign and unassign moves are currently only used in CH, so they cannot be used in LS with undo move tabu.

I'll add a comment explaining that an undo move is never being undone.

> Undo move tabu makes the undo move of recent steps tabu.

So what we really need is to be able to say whether a `ListAssignMove` and a `ListUnassignMove` are equal when assign and unassign moves become used in LS.

yurloc:
All comments addressed. I'm ready to merge.

PR #1646, opened 2021-11-03
PLANNER-2564 Reduce the use of internal APIs in examples

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
```diff
@@ -126,6 +126,21 @@
      */
     Score_ negate();
 
+    /**
+     * Returns a Score, all levels of which are zero.
+     *
+     * @return never null
+     */
+    Score_ zero();
```
triceo, line 134: I think that this is a useful addition to the public API, and it helps examples avoid `ScoreDefinition`.

optaplanner-examples/src/main/java/org/optaplanner/examples/common/business/SolutionBusiness.java
```diff
@@ -401,31 +397,4 @@ public void doChangeMove(Object entity, String variableName, Object toPlanningVa
...
-                }
-                inverseVariableSupplyList.add(inverseVariableSupply);
-            }
-            return new ChainedSwapMove<>(variableDescriptorList, inverseVariableSupplyList, leftEntity, rightEntity);
-        } else {
-            return new SwapMove<>(variableDescriptorList, leftEntity, rightEntity);
-        }
-    }
-
-    public void doSwapMove(Object leftEntity, Object rightEntity) {
```
triceo, line 427: This was unused.

optaplanner-core/src/main/java/org/optaplanner/core/api/score/Score.java
ge0ffrey: If it works for BendableScore (haven't got there yet): yes, good idea!

optaplanner-examples/src/main/java/org/optaplanner/examples/investment/solver/move/factory/InvestmentBiQuantityTransferMoveIteratorFactory.java
```diff
@@ -107,17 +106,23 @@ public boolean hasNext() {
...
+            long secondTransferMillis = nextLong(workingRandom, secondFrom.getQuantityMillis()
                     - (firstFrom == secondFrom ? firstTransferMillis : 0L)) + 1L;
             return CompositeMove.buildMove(new InvestmentQuantityTransferMove(firstFrom, firstTo, firstTransferMillis),
                     new InvestmentQuantityTransferMove(secondFrom, secondTo, secondTransferMillis));
         }
 
+        private static long nextLong(Random random, long n) {
+            return random.longs(0, n)
+                    .findAny()
+                    .orElseThrow();
```
ge0ffrey, line 123: (important)
This is very different from the original implementation. Are we sure that this is still correct? Uniformly distributed etc?

Original code:
{code}
    public static long nextLong(Random random, long n) {
        // This code is based on java.util.Random#nextInt(int)'s javadoc.
        if (n <= 0L) {
            throw new IllegalArgumentException("n must be positive");
        }
        if (n < Integer.MAX_VALUE) {
            return random.nextInt((int) n);
        }

        long bits;
        long val;
        do {
            bits = (random.nextLong() << 1) >>> 1;
            val = bits % n;
        } while (bits - val + (n - 1L) < 0L);
        return val;
    }
{code}

optaplanner-examples/src/main/java/org/optaplanner/examples/investment/solver/move/factory/InvestmentQuantityTransferMoveIteratorFactory.java
```diff
@@ -92,7 +91,9 @@ public boolean hasNext() {
 
         @Override
         public InvestmentQuantityTransferMove next() {
-            long transferMillis = RandomUtils.nextLong(workingRandom, InvestmentNumericUtil.MAXIMUM_QUANTITY_MILLIS) + 1L;
+            long transferMillis = workingRandom.longs(0, InvestmentNumericUtil.MAXIMUM_QUANTITY_MILLIS)
+                    .findAny()
+                    .orElseThrow() + 1L;
```
ge0ffrey, line 96: idem

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/app/NQueensApp.java
```diff
@@ -62,7 +62,6 @@ public NQueensApp() {
                 NQueensPanel.LOGO_PATH);
     }
 
-    @Override
```
ge0ffrey, line 65: BUG (important)

This doesn't work any more, because the super method was removed and it no longer passed through this.
So changing this method to switch to createSolverFactoryByApi() won't actually do anything.

Do a find-in-path where createSolverFactoryByApi() and createSolverFactoryByXml() is mentioned in the docs, comments, etc.

Proposals:
A) Revert these changes.
B) Introduce a createSolverConfig() method that can be overwritten?
C) Stop supporting createSolverFactoryByApi() in NQueensApp in bring it inline with the other examples. Adjust all docs accordingly.

Either way, look at NQueensApp.createSolverFactoryByApi() and compare that with TimeTableApp in our optaplanner-quickstarts hello-world:

```
        SolverFactory<TimeTable> solverFactory = SolverFactory.create(new SolverConfig()
                .withSolutionClass(TimeTable.class)
                .withEntityClasses(Lesson.class)
                .withConstraintProviderClass(TimeTableConstraintProvider.class)
                // The solver runs only for 5 seconds on this small dataset.
                // It's recommended to run for at least 5 minutes ("5m") otherwise.
                .withTerminationSpentLimit(Duration.ofSeconds(5)));
```

That is much better, so maybe C) is the path to go?

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/swingui/VehicleRoutingPanel.java
```diff
@@ -120,7 +114,9 @@ protected Customer createCustomer(VehicleRoutingSolution solution, Location newL
             TimeWindowedCustomer newTimeWindowedCustomer = new TimeWindowedCustomer();
             TimeWindowedDepot timeWindowedDepot = (TimeWindowedDepot) solution.getDepotList().get(0);
             long windowTime = (timeWindowedDepot.getDueTime() - timeWindowedDepot.getReadyTime()) / 4L;
-            long readyTime = RandomUtils.nextLong(demandRandom, windowTime * 3L);
+            long readyTime = demandRandom.longs(0, windowTime * 3L)
+                    .findAny()
+                    .orElseThrow();
```
ge0ffrey, line 119: idem

optaplanner-examples/src/test/java/org/optaplanner/examples/common/app/SolverPerformanceTest.java
```diff
@@ -94,65 +95,60 @@ public void setUp() {
...
+                    scoreExplanation.getConstraintMatchTotalMap();
+            assertThat(constraintMatchTotals).isNotNull();
+            assertThat(constraintMatchTotals.values().stream()
+                    .map(ConstraintMatchTotal::getScore)
+                    .reduce(Score::add)
+                    .orElse(bestScore.zero()))
+                            .isEqualTo(scoreExplanation.getScore());
+            assertThat(scoreExplanation.getIndictmentMap()).isNotNull();
+        } catch (IllegalStateException ex) {
+            logger.info("Constraint match likely not supported.", ex);
```
ge0ffrey, line 141: important

This eats an exception! Let's not do that.

ge0ffrey (changes requested):
See my comments, looks good otherwise :)
I'd like to rereview in this case.

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/app/NQueensApp.java
triceo: This change was not necessary and honestly I didn't even know I made it.
I reverted to the original state, keeping the status quo.

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/swingui/VehicleRoutingPanel.java
triceo: I restored the random impl everywhere else to the original, maintaining status quo. (Since when do we question the correctness of JDK methods?)

This particular use is only for the creation of a new customer in the UI, therefore even if the random sequence changed, it would not matter.

triceo:
@ge0ffrey Made changes, please re-review.

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/app/NQueensApp.java
ge0ffrey: Agreed, the optaplanner-quickstart hello-world example covers this much better.

optaplanner-examples/src/test/java/org/optaplanner/examples/common/app/SolverPerformanceTest.java
```diff
@@ -94,65 +92,60 @@ public void setUp() {
...
+                    scoreExplanation.getConstraintMatchTotalMap();
+            assertThat(constraintMatchTotals).isNotNull();
+            assertThat(constraintMatchTotals.values().stream()
+                    .map(ConstraintMatchTotal::getScore)
+                    .reduce(Score::add)
+                    .orElse(bestScore.zero()))
+                            .isEqualTo(scoreExplanation.getScore());
+            assertThat(scoreExplanation.getIndictmentMap()).isNotNull();
+        } catch (IllegalStateException ex) {
+            throw new IllegalStateException("Constraint match likely not supported.", ex);
```
ge0ffrey, line 138: (nitpick, not important)

Do we need that catch that original IllegalStateException and wrap it?
The user is calling assertScoreAndConstraintMatches() and the call to ScoreManager.explainScore() has this javadoc:

```
     * @throws IllegalStateException when constraint matching is disabled or not supported by the underlying score
     *         calculator, such as {@link EasyScoreCalculator}.
     */
    ScoreExplanation<Solution_, Score_> explainScore(Solution_ solution);
```

I would expect that the IllegalStateException thrown by ScoreManager clearly says if Constraint matches are not supported - and there's little value in wrapping it here with "likely not supported". Wdyt?

optaplanner-examples/src/main/java/org/optaplanner/examples/vehiclerouting/swingui/VehicleRoutingPanel.java
ge0ffrey: Agreed on para 2 - even though that changing the random numbers might affect the behavior when doing live demos (as it might generate different random numbers), and theorically some random numbers can be a pain to demo well (for example multiple locations that are pretty much on the same pixel or almost the same pixel). Less surprises during live demos is good. But as it's still seeded, it's still reproducible, so this is probably 100% fine and if not, it will show up during the demo preperation.

As for para 1 - intresting argument, let's discuss. Thanks for taking it out of this PR.

ge0ffrey (approved):

PR #1652, opened 2021-11-05
[8.13.x] Fix broken links in quarkus docs

ge0ffrey:
@radtriste @mbiarnes If this PR is in time to make the 8.13.0.Final release (so we fix the broken links SEO influence sooner rather than later), that would be nice, but it is not a necessity.


ge0ffrey:
the PR to main was approved and merged, so this one can be merged too (regardless if it makes the release or not)

PR #1653, opened 2021-11-08
PLANNER-2557 Thread-safe groupBy in CS-D

triceo:
I ran the reproducer code in a loop several thousand times and the issue did not show.
Before the fix, it would usually show within the first 20 repetitions.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
@@ -46,14 +46,10 @@ abstract class AbstractAccumulator<ResultContainer_, Result_> implements Accumul
         this.finisher = Objects.requireNonNull(finisher);
     }
 
-    protected static <Value_> Value_ extractValue(Declaration declaration, Tuple extractedTuple) {
-        return (Value_) declaration.getValue(null, extractedTuple.getFactHandle().getObject());
-    }
-
-    protected static UnaryOperator<Tuple> getTupleExtractor(Declaration declaration, Tuple tuple) {
+    protected static UnaryOperator<Tuple> getTupleExtractor(Declaration declaration, Tuple leftTuple) {
         int offset = 0;
-        while (tuple.getIndex() != declaration.getTupleIndex()) {
-            tuple = tuple.getParent();
+        while (leftTuple.getIndex() != declaration.getTupleIndex()) {
+            leftTuple = leftTuple.getParent();
             offset++;
         }
         switch (offset) {
@@ -68,6 +64,14 @@ abstract class AbstractAccumulator<ResultContainer_, Result_> implements Accumul
             default:
                 throw new UnsupportedOperationException("Impossible state: tuple delta offset (" + offset + ").");
         }
+    }
+
+    protected static <X> Function<Tuple, X> getValueExtractor(Declaration declaration, Tuple leftTuple) {
+        UnaryOperator<Tuple> tupleExtractor = getTupleExtractor(declaration, leftTuple);
+        return tuple -> {
+            Tuple extractedTuple = tupleExtractor.apply(tuple);
+            return (X) declaration.getValue(null, extractedTuple.getFactHandle().getObject());
+        };
     }
 
     @Override
@@ -100,8 +104,8 @@ abstract class AbstractAccumulator<ResultContainer_, Result_> implements Accumul
     }
 
     @Override
-    public Result_ getResult(Object workingMemoryContext, Object context, Tuple leftTuple, Declaration[] declarations,
-            ReteEvaluator reteEvaluator) {
+    public final Result_ getResult(Object workingMemoryContext, Object context, Tuple leftTuple,
+            Declaration[] declarations, ReteEvaluator reteEvaluator) {
         return finisher.apply((ResultContainer_) context);
     }
 }
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
@@ -20,7 +20,6 @@ import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.function.Supplier;
-import java.util.function.UnaryOperator;
 
 import org.drools.core.common.InternalFactHandle;
 import org.drools.core.common.ReteEvaluator;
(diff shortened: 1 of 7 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
```diff
@@ -85,6 +93,37 @@ public final ResultContainer_ init(Object workingMemoryContext, Object context,
...
+         * Therefore, initialization of our accumulator can only happen the first time the method is called,
+         * and therefore the initialization needs to be properly synchronized.
+         *
+         * We do not care if the initialization happens multiple times,
+         * as each of the attempts will fill the variables with the same data.
+         * We just need to ensure that the variables are initialized before the actual accumulation happens.
+         *
+         * Technically running this more than once is wasteful,
+         * but the initialized() check will make sure this only happens during the cold start of the session.
+         */
```
triceo, line 114: This explains the problem and the solution.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
```diff
@@ -85,6 +60,37 @@ public final ResultContainer_ init(Object workingMemoryContext, Object context,
...
+        if (!initialized.get()) {
+            initialize(leftTuple, innerDeclarations);
+            initialized.set(true);
+        }
```
ge0ffrey, line 85: I fear this might be broken.
Read the entire section "Usage in Java" of https://en.wikipedia.org/wiki/Double-checked_locking

Race condition proof:
- Thread 1 calls accumulate() line 64
- Thread 1 checks if (!initialized.get()) { line 82 => goes to line 83
- Thread 2 calls accumulate() line 64
- Thread 2 checks if (!initialized.get()) { line 82 => goes to line 83
- Thread 1 calls initialize(...) line 83
- Thread 2 calls initialize(...) line 83

Solution: this entire if block should be wrapped by `synchronized(initialized) {`

ge0ffrey: Not that uncontested locks are cheap.

Also it's too back that AtomicBoolean doesn't support something like setIfFalse(Function<Boolean, Boolean>), similar to Maps.putIfAbsent() etc.

triceo: I am aware of this issue, and I am explaining in the comment why this is not an issue.
Yes, it will cause initialize() to be called multiple times - but we do not care.
Each call to initialize() will set *the same* values to all variables.

The code does what it is supposed to - each thread will see the initialized accumulator.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/BiAccumulator.java
```diff
@@ -46,28 +44,22 @@ public BiAccumulator(Variable<A> varA, Variable<B> varB,
...
+    protected Runnable accumulate(ResultContainer_ context, Tuple leftTuple, InternalFactHandle handle,
+            Declaration[] innerDeclarations) {
+        A a = valueExtractorA.apply(leftTuple);
+        B b = valueExtractorB.apply(leftTuple);
+        return accumulator.apply(context, a, b);
     }
 
-    private void init(Tuple leftTuple, Declaration[] innerDeclarations) {
+    @Override
+    protected void initialize(Tuple leftTuple, Declaration[] innerDeclarations) {
```
ge0ffrey, line 55: Why can't we do this code in the Accumulator.init(...) method that Drools guarantees to call once for us?
Can we do an RFE to @mariofusco for whatever guarantees is missing?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
triceo: If we want to prevent the multiple initialization, I agree we need to run this in a synchronized block. Is it worth it to force this code to always synchronize, so that we don't have a harmless double call to the method?

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
```diff
@@ -85,6 +60,37 @@ public final ResultContainer_ init(Object workingMemoryContext, Object context,
...
+        return accumulate((ResultContainer_) context, leftTuple, handle, innerDeclarations);
+    }
+
+    protected abstract Runnable accumulate(ResultContainer_ context, Tuple leftTuple, InternalFactHandle handle,
+            Declaration[] innerDeclarations);
+
+    protected abstract void initialize(Tuple leftTuple, Declaration[] innerDeclarations);
```
ge0ffrey, line 92: This is similar to different to Accumulator.init(). Can we add a comment here to explain why init() doesn't furfill our needs? Something like "// WORKAROUND because init() doesn't do X"

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/BiAccumulator.java
triceo: I also explain that in the long comment.
I discussed this with Mario, and he considers `Accumulator` to be internal API - we are only using it to squeeze max performance out of Drools. As such, he doesn't want to do any changes to this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
triceo: The comment is already there, and it is very long.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/AbstractAccumulator.java
ge0ffrey: Non contested synchronization is extremely cheap. It's better to be correct than fast. IIRC Brian Goetz's talk correctly, we shouldn't try to be a smartass and just synchronize it.

But even better would be if we can get Drools to allow us to just use the init() method for this work.

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/stream/drools/common/BiAccumulator.java
ge0ffrey: Let's create a DROOLS jira to spec out exactly what we want, and more importantly why.

If it's an internal API, then it doesn't need to be Backwards Compatible :)

ge0ffrey (approved):
LGTM, Lukas's call on how to deal with my comments :)

PR #1665, opened 2021-11-11
Reduce mutability in Phase related code

yurloc:
The goal of this PR is to make it easier to understand and reason about the Phases code. Declaring fields `final` and eliminating the corresponding setters reduces the effort of searching for the places from where the field can be changed. With `final`, the intention is clear: the value is determined during object construction and then never changes.

The Builder pattern is taken from Item 2, Effective Java. In this case it solves the problem with optional assertion flags. The Builder has a named setter method for each optional argument, so instead of something like
```java
XyPhase phase;
if (environment.isLevelA() && environment.isLevelB()) {
    phase = new XyPhase(requiredArgs, true, true);
}
if (environment.isLevelA()) {
    phase = new XyPhase(requiredArgs, true);
}
// etc.
```
and multiple telescoping constructors for `XyPhase`, we have something where the client code is almost the same as before but immutability is achieved once the object is built:
```java
// Initilialize builder with mandatory arguments.
Builder builder = new XyPhase.Builder(requiredArgs);
// Conditionally enable any combination of optional assertion flags.
if (environment.isLevelA()) {
    builder.setAssertStepScoreFromScratch(true);
}
if (environment.isLevelB()) {
    builder.setAssertExpectedStepScore(true);
}
// Once the object is built, even the optional arguments become final.
XyPhase phase = builder.build();
```

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
@@ -46,47 +46,21 @@ import org.optaplanner.core.impl.solver.termination.Termination;
 public class DefaultExhaustiveSearchPhase<Solution_> extends AbstractPhase<Solution_>
         implements ExhaustiveSearchPhase<Solution_> {
 
-    protected Comparator<ExhaustiveSearchNode> nodeComparator;
-    protected EntitySelector<Solution_> entitySelector;
-    protected ExhaustiveSearchDecider<Solution_> decider;
+    protected final Comparator<ExhaustiveSearchNode> nodeComparator;
+    protected final EntitySelector<Solution_> entitySelector;
+    protected final ExhaustiveSearchDecider<Solution_> decider;
 
-    protected boolean assertWorkingSolutionScoreFromScratch = false;
-    protected boolean assertExpectedWorkingSolutionScore = false;
+    protected final boolean assertWorkingSolutionScoreFromScratch;
+    protected final boolean assertExpectedWorkingSolutionScore;
 
-    public DefaultExhaustiveSearchPhase(int phaseIndex, String logIndentation, Termination<Solution_> termination) {
-        super(phaseIndex, logIndentation, termination);
-    }
+    private DefaultExhaustiveSearchPhase(Builder<Solution_> builder) {
+        super(builder);
+        nodeComparator = builder.nodeComparator;
+        entitySelector = builder.entitySelector;
+        decider = builder.decider;
 
-    public Comparator<ExhaustiveSearchNode> getNodeComparator() {
-        return nodeComparator;
-    }
-
-    public void setNodeComparator(Comparator<ExhaustiveSearchNode> nodeComparator) {
-        this.nodeComparator = nodeComparator;
-    }
-
-    public EntitySelector<Solution_> getEntitySelector() {
-        return entitySelector;
-    }
-
-    public void setEntitySelector(EntitySelector<Solution_> entitySelector) {
-        this.entitySelector = entitySelector;
-    }
-
-    public ExhaustiveSearchDecider<Solution_> getDecider() {
-        return decider;
-    }
-
-    public void setDecider(ExhaustiveSearchDecider<Solution_> decider) {
-        this.decider = decider;
-    }
-
-    public void setAssertWorkingSolutionScoreFromScratch(boolean assertWorkingSolutionScoreFromScratch) {
-        this.assertWorkingSolutionScoreFromScratch = assertWorkingSolutionScoreFromScratch;
-    }
-
-    public void setAssertExpectedWorkingSolutionScore(boolean assertExpectedWorkingSolutionScore) {
-        this.assertExpectedWorkingSolutionScore = assertExpectedWorkingSolutionScore;
+        assertWorkingSolutionScoreFromScratch = builder.assertWorkingSolutionScoreFromScratch;
+        assertExpectedWorkingSolutionScore = builder.assertExpectedWorkingSolutionScore;
     }
 
     @Override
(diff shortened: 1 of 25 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/HeuristicConfigPolicy.java
```diff
@@ -144,23 +113,19 @@ public boolean isInitializedChainedValueFilterEnabled() {
...
+                .withLogIndentation(logIndentation);
     }
 
-    public HeuristicConfigPolicy<Solution_> createFilteredPhaseConfigPolicy() {
-        HeuristicConfigPolicy<Solution_> heuristicConfigPolicy = createPhaseConfigPolicy();
-        heuristicConfigPolicy.reinitializeVariableFilterEnabled = true;
-        heuristicConfigPolicy.initializedChainedValueFilterEnabled = true;
-        return heuristicConfigPolicy;
+    public HeuristicConfigPolicy<Solution_> createPhaseConfigPolicy() {
+        return buildClone().build();
```
yurloc, line 122: Or maybe `clonerBuilder().build()`?

optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
```diff
@@ -269,4 +243,35 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
         decider.solvingEnded(solverScope);
     }
 
+    public static class Builder<Solution_> extends AbstractPhase.Builder<Solution_> {
```
ge0ffrey, line 246: Do we need a builder pattern?

optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
```diff
@@ -269,4 +243,35 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
...
+            this.assertWorkingSolutionScoreFromScratch = assertWorkingSolutionScoreFromScratch;
+        }
+
+        public void setAssertExpectedWorkingSolutionScore(boolean assertExpectedWorkingSolutionScore) {
+            this.assertExpectedWorkingSolutionScore = assertExpectedWorkingSolutionScore;
+        }
+
+        @Override
+        public DefaultExhaustiveSearchPhase<Solution_> build() {
+            return new DefaultExhaustiveSearchPhase<>(this);
```
ge0ffrey, line 274: Typically, in a builder pattern, the builder itself is not given to the thing it builds IIRC.

ge0ffrey:
> ```java
> XyPhase phase;
> if (environment.isLevelA() && environment.isLevelB()) {
>     phase = new XyPhase(requiredArgs, true, true);
> }
> if (environment.isLevelA()) {
>     phase = new XyPhase(requiredArgs, true);
> }
> // etc.
> ```
Personally, I prefer this before code over the after code. What do the other think?

ge0ffrey:
+1 to try to make more fields final of the phases (not the scopes), where it doesn't increase complexity.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/HeuristicConfigPolicy.java
rsynek: Sounds much better, as the `buildClone` method is not creating the clone itself.

optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
rsynek: An alternative to passing a builder instance is creating a constructor that accepts all possible options as parameters. In that case, this `build` method becomes yet another place where the options need to be passed to the constructor as arguments. For that reason, I prefer the way @yurloc proposes.

optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
rsynek: It depends on if all the properties are mandatory. If they are not, removing setters leads to a combinatorial explosion of constructors.

rsynek (approved):

triceo (approved):
LGTM. The endless mutability of core has been bugging me for a while.

optaplanner-core/src/main/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
yurloc: Exactly as Radek says. Passing the builder instance avoids repetition. I've seen this approach in multiple sources, including the Effective Java book.

yurloc:
> Do we need a builder pattern? Why not just use a simple constructor? KISS principle.

I agree the Builder code is a boilerplate but I think it does pull its weight:
- It's pretty straightforward and it's nicely separated at the end of its enclosing class so it doesn't interrupt reading it.
- It fits the current state where all the affected classes have some optional parameters that are being conditionally set based on the `Environment`. It preserves the existing API that makes the client code read nicely. Optional parameters are being set using named methods (`builder.setAssertExpectedStepScore(true)`) vs. being passed as an anonymous value in the constructor (`new DefaultLocalSearchPhase(..., true)`) while allowing them to stay immutable once the object is constructed.

triceo:
2 approvals, no rejections, merging.

PR #1698, opened 2021-11-30
Simplify SolutionDescriptor.getMovableEntityCount()

yurloc:
It's unnecessary to count all movable entities when it's enough to find any and quit early.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -931,12 +931,9 @@ public class SolutionDescriptor<Solution_> {
      * @param scoreDirector never null
      * @return {@code >= 0}
      */
-    public int getMovableEntityCount(ScoreDirector<Solution_> scoreDirector) {
+    public boolean hasMovableEntities(ScoreDirector<Solution_> scoreDirector) {
         return extractAllEntitiesStream(scoreDirector.getWorkingSolution())
-                .mapToInt(entity -> findEntityDescriptorOrFail(entity.getClass()).isMovable(scoreDirector, entity)
-                        ? 1
-                        : 0)
-                .sum();
+                .anyMatch(entity -> findEntityDescriptorOrFail(entity.getClass()).isMovable(scoreDirector, entity));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -931,12 +931,9 @@ public int getEntityCount(Solution_ solution) {
...
      * @return {@code >= 0}
      */
-    public int getMovableEntityCount(ScoreDirector<Solution_> scoreDirector) {
+    public boolean hasMovableEntities(ScoreDirector<Solution_> scoreDirector) {
         return extractAllEntitiesStream(scoreDirector.getWorkingSolution())
-                .mapToInt(entity -> findEntityDescriptorOrFail(entity.getClass()).isMovable(scoreDirector, entity)
-                        ? 1
-                        : 0)
-                .sum();
+                .anyMatch(entity -> findEntityDescriptorOrFail(entity.getClass()).isMovable(scoreDirector, entity));
```
triceo, line 936: This is a beneficial change on its own. But looking at `findEntityDescriptorOrFail(...)`, it has a performance issue of its own.

It does try to cache the results of the lookup. Unfortunately, when nothing is found, it doesn't store anything in the cache - so in that case (= no entity descriptor found) the lookups will be happening over and over again.

If we're touching this code, I'd vote to address this issue too.

triceo (approved):
I like this, but I made a suggestion inside.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
yurloc: @triceo Done. Please take a look.

yurloc:
> I will leave it to him to have the final word. That said, I'd definitely keep the `null` return type for `findEntityDescriptor(...)`, as it would be inconsistent for the API to use `Optional` in this one place and not anywhere else.

I will gladly make the rest of the API consistent by using Optional anywhere where `null` could be returned :slightly_smiling_face:

yurloc:
If there is not a clear consensus on how to address the memoized entity descriptor map optimization, I propose to merge the PR in its original form (the first commit only) and handle the rest separately.

triceo:
> I will gladly make the rest of the API consistent by using Optional anywhere where `null` could be returned slightly_smiling_face

Preaching to the choir here. :-) In this case, I am not the one who needs to be convinced.

> If there is not a clear consensus on how to address the memoized entity descriptor map optimization, I propose to merge the PR in its original form (the first commit only) and handle the rest separately.

Yeah, let's kick the can further down the road.

PR #1699, opened 2021-12-01
Optimize SolutionDescriptor.findEntityDescriptor()

yurloc:
This PR offers one approach to address a trivial performance issue brought up in https://github.com/kiegroup/optaplanner/pull/1698#discussion_r759236750.

Use `Optional.empty()` instead of `null` as a value in the memoization map
to designate that the search for a descriptor has been made and none has
been found. As opposed to `null` meaning the search hasn't been made yet.

This proposal doesn't include the controversial internal API change that exposes `Optional` as the `findEntityDescriptor()` return type. This is simply achieved by returning `null` if the `Optional` representing the result of the lookup is empty.

### Referenced pull requests
Split from #1698.

### Checklist

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -39,7 +39,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Stream;
@@ -169,7 +168,7 @@ public class SolutionDescriptor<Solution_> {
     private ConstraintConfigurationDescriptor<Solution_> constraintConfigurationDescriptor;
     private final Map<Class<?>, EntityDescriptor<Solution_>> entityDescriptorMap;
     private final List<Class<?>> reversedEntityClassList;
-    private final ConcurrentMap<Class<?>, Optional<EntityDescriptor<Solution_>>> lowestEntityDescriptorMemoization =
+    private final ConcurrentMap<Class<?>, EntityDescriptor<Solution_>> lowestEntityDescriptorMemoization =
             new ConcurrentMemoization<>();
 
     private SolutionCloner<Solution_> solutionCloner;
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -741,11 +742,11 @@ public boolean hasEntityDescriptor(Class<?> entitySubclass) {
             // Reverse order to find the nearest ancestor
             for (Class<?> entityClass : reversedEntityClassList) {
                 if (entityClass.isAssignableFrom(entitySubclass)) {
-                    return entityDescriptorMap.get(entityClass);
+                    return Optional.of(entityDescriptorMap.get(entityClass));
                 }
             }
-            return null;
-        });
+            return Optional.empty();
```
yurloc, line 748: Insert a log statement before this line and try with the Nurse Rostering example to see the effect.

triceo (approved):
I like this. That said, I will leave the final approval to @ge0ffrey, as he has his own opinions on `Optional`.

ge0ffrey:
Yea... I don't like the use of Optional outside of streams, where it was created for. IIRC Brian Goetz had the same advise on that.

I don't want to see this merged... Proliferating Optional will not make the code easier to read.

ge0ffrey (changes requested):
I'don't want to introduce Optional, except for code that uses it to deal with stream operations that return zero or one results.
It doesn't make our code, as a whole, easier to read or to maintenance if this is introduced in some places. It doesn't make it faster.
And AFAIK, it's not its intended use.

triceo:
I still think that this change has value. Can we create a dummy entity descriptor which we would use in the map? The method should always return `null` though.

yurloc:
There is one thing that makes me hesitant to take the `Optional` approach.

In "Item 55: Return optionals judiciously", Joshua Bloch says:
> It is almost never appropriate to use an optional as a key, value, or element in a collection or array.

But I still think that in this case it's a correct and elegant usage. Btw, the fact that `Optional` is often used in `Stream` operation doesn't mean that it's limited to streams. It represents a value that is either present or missing and forces the user to check that before using the value. It's a good compromise between returning `null` and throwing a checked exception (to force the user to handle the case when the value is not present).

Nevertheless, the performance impact of the inefficient computation of the memoization map is unknown and probably negligible. So I consider this a premature optimization.

PR #1712, opened 2021-12-07
PLANNER-2586 Don't create undo moves of undo moves

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -94,8 +94,7 @@ public void solve(SolverScope<Solution_> solverScope) {
 
     private void doStep(ConstructionHeuristicStepScope<Solution_> stepScope) {
         Move<Solution_> step = stepScope.getStep();
-        Move<Solution_> undoStep = step.doMove(stepScope.getScoreDirector());
-        stepScope.setUndoStep(undoStep);
+        step.doMoveOnly(stepScope.getScoreDirector());
```
triceo, line 97: Note: undo was not used anywhere, so I deleted it.

optaplanner-core/src/test/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseTest.java
```diff
@@ -100,19 +100,19 @@ public void restoreWorkingSolution() {
...
-        verify(node3B.getUndoMove(), times(0)).doMove(any(ScoreDirector.class));
-        verify(node4B.getMove(), times(1)).doMove(scoreDirector);
-        verify(node4B.getUndoMove(), times(0)).doMove(any(ScoreDirector.class));
+        verify(node3A.getUndoMove(), times(1)).doMoveOnly(scoreDirector);
+        verify(node2B.getMove(), times(1)).doMoveOnly(scoreDirector);
+        verify(node2B.getUndoMove(), times(0)).doMoveOnly(any(ScoreDirector.class));
+        verify(node3B.getMove(), times(1)).doMoveOnly(scoreDirector);
+        verify(node3B.getUndoMove(), times(0)).doMoveOnly(any(ScoreDirector.class));
+        verify(node4B.getMove(), times(1)).doMoveOnly(scoreDirector);
+        verify(node4B.getUndoMove(), times(0)).doMoveOnly(any(ScoreDirector.class));
```
triceo, line 115: The test passes like this, but I do not really understand why it needs to be like that.
The test is somewhat cryptic.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/AbstractMove.java
```diff
@@ -33,9 +33,14 @@
     @Override
     public final AbstractMove<Solution_> doMove(ScoreDirector<Solution_> scoreDirector) {
         AbstractMove<Solution_> undoMove = createUndoMove(scoreDirector);
+        doMoveOnly(scoreDirector);
+        return undoMove;
+    }
+
+    @Override
+    public void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
```
ge0ffrey, line 41: This method should be final here, like doMove is:
```suggestion
    public final void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
```

We don't want to allow users to overwrite this or think they need to implement this.

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
```diff
@@ -87,6 +87,18 @@
      */
     Move<Solution_> doMove(ScoreDirector<Solution_> scoreDirector);
 
+    /**
+     * As defined by {@link #doMove(ScoreDirector)}, but does not return an undo move.
+     * It is recommended that {@link #doMove(ScoreDirector)} is implemented by wrapping {@link #doMoveOnly(ScoreDirector)}.
```
ge0ffrey, line 92: This sentence confuses me, especially with the implementation.

In general, we recommend users to implement AbstractMove and only implement doGenuineMove there. Let's keep it simple and recommend that and only that.

CompositiveMove cannot extend AbstractMove because it needs to sequence undo move creation and do move execution correctly. But for that one exception I don't think we need to confuse the javadocs, Wdyt?

ge0ffrey:
```suggestion
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java
```diff
@@ -87,6 +87,18 @@
...
+     *
+     * @param scoreDirector never null, the {@link ScoreDirector} that needs to get notified of the changes
+     */
+    default void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
+        // For backwards compatibility, this method is default and calls doMove(...).
+        // In reality, the relationship would be inversed, as implemented in AbstractMove.
```
ge0ffrey, line 98: (nitpick) pressgang/ibm/etc docs style guidelines: avoid "would", etc as they are ambiguous signals

```suggestion
        // Normally, the relationship is typically inversed, as implemented in AbstractMove.
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/partitionedsearch/scope/PartitionChangeMove.java
```diff
@@ -98,9 +98,7 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
 
     @Override
     protected PartitionChangeMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
-        // HACK Creating an undoMove is a waste of time because it is never used,
-        // but this method is called anyway so it returns null to fail fast if it is used
-        return null;
+        throw new UnsupportedOperationException("Programming error: undo move should not have been called.");
```
ge0ffrey, line 101: (nitpick) we typically start such errors with "Impossible state:" to indicate that it's defenitly an optaplanner bug.
```suggestion
        throw new UnsupportedOperationException("Impossible state: undo move should not be called.");
```

optaplanner-core/src/test/java/org/optaplanner/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseTest.java
ge0ffrey: IIRC, it's the delta set of moves to go from one solution state to a totally different one, somewhere in another node in the big exponential search tree.

ge0ffrey (approved):

PR #1720, opened 2021-12-12
With methods for union move selector and cartesian product move selector

ge0ffrey:


optaplanner-core/src/main/java/org/optaplanner/core/config/heuristic/selector/move/composite/CartesianProductMoveSelectorConfig.java
```diff
@@ -65,18 +66,40 @@
...
     public CartesianProductMoveSelectorConfig(List<MoveSelectorConfig> moveSelectorConfigList) {
         this.moveSelectorConfigList = moveSelectorConfigList;
     }
 
+    /**
+     * @deprecated in favor of {@link #getMoveSelectorList()}.
+     * @return sometimes null
+     */
+    @Deprecated
     public List<MoveSelectorConfig> getMoveSelectorConfigList() {
```
ge0ffrey, line 85: See `SolverConfig.setPhaseList(List<PhaseConfigList)`
The naming convention is to drop the "Config" in the name (but not in the typing).

rsynek (approved):

PR #1732, opened 2021-12-16
PLANNER-2611: Fix rare race condition in DefaultSolverJob terminateEarly

Christopher-Chianelli:
There was a race condition where a SolverJob can cancel
the completition of a different SolverJob with the same
problem id. In particular, the old code did not have any locking
on SolverStatus changes, which allowed the following sequence
of operations on three threads (T1, T2, T3):

Preconditions:
solverManager.getSolverStatus(1L) == SOLVING_SCHEDULED

```
---
T1: solverManager.terminateEarly(1L)
T2: solverManager.terminateEarly(1L)
T3: solverJob = solverManager.solve(1L, problem)

---

T1: solvingTerminated() (in SolverJob.terminateEarly, under SOLVING_SCHEDULED)
T2: solvingTerminated() (in SolverJob.terminateEarly, under SOLVING_SCHEDULED)
T3: solverJob = solverManager.solve(1L, problem) (No Action Yet)

---

T1: Thread Ended (the old SolverJob was removed from SolverManager)
T2: solvingTerminated() (No Action Yet)
T3: solverJob = solverManager.solve(1L, problem) (No Action Yet)

---

T1: Thread Ended
T2: solvingTerminated() (No Action Yet)
T3: assert solverJob.getStatus() == solverManager.getSolverStatus(1L)
    // (A new SolverJob was successfully created)

---

T1: Thread Ended
T2: Thread Ended (The new SolverJob was removed)
T3: assert solverJob.getStatus() == solverManager.getSolverStatus(1L)
    // No Action Yet

---

T1: Thread Ended
T2: Thread Ended
T3: AssertionError("SOLVING_SCHEDULED != NOT_SOLVING")

---
```

This create an inconsistency in the SolverManager, and allows
two SolverJobs with the same ProblemId to be active at the
same time (which can have strange side effects, like one
SolverJob cancelling the other).

The solution I came up with was to lock whenever SolverStatus
was going to be potentially written to. A new
PhaseLifecycleEventListener was added that unlock said lock when
Solving started to prevent a race condition where terminateEarly
is called on the SolverJob but the Solver did not start yet,
which will cause the Solver to continue solving and not cancel
the Job.

### JIRA

https://issues.redhat.com/browse/PLANNER-2611

### Checklist

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
@@ -23,6 +23,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -31,6 +32,9 @@ import org.optaplanner.core.api.domain.solution.PlanningSolution;
 import org.optaplanner.core.api.solver.Solver;
 import org.optaplanner.core.api.solver.SolverJob;
 import org.optaplanner.core.api.solver.SolverStatus;
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListener;
+import org.optaplanner.core.impl.phase.scope.AbstractPhaseScope;
+import org.optaplanner.core.impl.phase.scope.AbstractStepScope;
 import org.optaplanner.core.impl.solver.scope.SolverScope;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -51,7 +55,8 @@ public final class DefaultSolverJob<Solution_, ProblemId_> implements SolverJob<
     private final BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler;
 
     private final AtomicReference<SolverStatus> solverStatusReference;
-    private CountDownLatch terminatedLatch;
+    private final CountDownLatch terminatedLatch;
+    private final ReentrantLock solverStatusModifyingLock;
 
     private Future<Solution_> future;
 
@@ -73,6 +78,7 @@ public final class DefaultSolverJob<Solution_, ProblemId_> implements SolverJob<
         this.exceptionHandler = exceptionHandler;
         solverStatusReference = new AtomicReference<>(SolverStatus.SOLVING_SCHEDULED);
         terminatedLatch = new CountDownLatch(1);
+        solverStatusModifyingLock = new ReentrantLock();
     }
 
     public void setFuture(Future<Solution_> future) {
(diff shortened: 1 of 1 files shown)


Christopher-Chianelli:
In the extremely rare case that terminateEarly is called when trying to reacquire the lock (in the `!solverStatusModifyingLock.isHeldByCurrentThread()` guard),  it only calls `solver.terminateEarly()`, which does nothing and releases the lock, so it should not cause an issue.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -51,7 +55,8 @@
     private final BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler;
 
     private final AtomicReference<SolverStatus> solverStatusReference;
```
triceo, line 57: The `AtomicReference` is here to deal with concurrency issues around `SolverStatus`.
If you say that is insufficient, and have implemented locking to work around that, I'd argue that the `AtomicReference` is no longer necessary and confusing.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -108,7 +118,12 @@ public Solution_ call() {
             exceptionHandler.accept(problemId, e);
             throw new IllegalStateException("Solving failed for problemId (" + problemId + ").", e);
         } finally {
+            if (!solverStatusModifyingLock.isHeldByCurrentThread()) {
+                // reacquire the lock if we don't have it
+                solverStatusModifyingLock.lock();
+            }
             solvingTerminated();
+            solverStatusModifyingLock.unlock();
```
triceo, line 126: This is not particularly safe, I'm afraid. What if `lock()` throws?  (I'd expect it can throw at least `InterruptedException`.)
WHat if  `solverTerminated()` throws?

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -132,20 +147,25 @@ private void solvingTerminated() {
 
     @Override
     public void terminateEarly() {
```
triceo, line 149: We should really use the lock-try-finally-unlock pattern here, too.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -177,4 +197,38 @@ public Duration getSolvingDuration() {
         }
         return Duration.ofMillis(endingSystemTimeMillis - startingSystemTimeMillis);
     }
+
+    private class UnlockLockPhaseLifecycleListener implements PhaseLifecycleListener {
+
+        @Override
+        public void solvingStarted(SolverScope solverScope) {
+            solverStatusModifyingLock.unlock();
```
triceo, line 205: Seeing a listener like this makes me thing that this whole locking mechanism needs very good Javadocs so that the future us understand what is going on and when.

triceo (commented):
I'm proposing some changes to make this safer and, more importantly, easier to understand.

triceo:
Also please file a JIRA for this, we really shouldn't be letting untracked commits in more than we already have.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
Christopher-Chianelli: It is insufficient; in particular, two threads in `terminateEarly` can enter `case SOLVING_SCHEDULED: solvingTerminated();` at the same time (which leads to a double free, which can lead to a use after free). An easier way to work around it is to make `terminateEarly` synchronized, but a `call()` could be done in a way where `solverManager.getSolverStatus()` returns NOT_SOLVING yet the Solver is solving (call get the first set, stop before solver.solve, solvingTerminated perform the second set and unregistered the job, then solver.solve is called, which continues normally since it resets the isTerminated flag), which also can lead to a double free). Concurrency is fun! It would very hard to reproduce these conditions, but it is possible.

Christopher-Chianelli: +1 for converting to `private volatile SolverStatus`

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverJob.java
```diff
@@ -177,4 +200,59 @@ public Duration getSolvingDuration() {
...
+     *
+     * The solver.solve() call resets the terminateEarly flag, and thus the solver will not be terminated
+     * by the call, which means terminatedLatch will not be decremented, causing Thread 2 to wait forever
+     * (at least until another Thread calls terminateEarly again).
+     *
+     * To prevent Thread 2 from potentially waiting forever, we only unlock the lock after the
+     * solvingStarted phase lifecycle event is fired, meaning the terminateEarly flag will not be
+     * reset and thus the solver will actually terminate.
+     */
+    private final class UnlockLockPhaseLifecycleListener implements PhaseLifecycleListener<Solution_> {
```
yurloc, line 225: We have the `PhaseLifecycleListenerAdapter`. You may want to extend that and only override `solvingStarted()` and skip the "Do nothing" overrides.

triceo (approved):

PR #1735, opened 2021-12-17
[DROOLS-6707] move tms in its own module

mariofusco:
Related PRs:
https://github.com/kiegroup/drools/pull/4075
https://github.com/kiegroup/kogito-runtimes/pull/1805
https://github.com/kiegroup/optaplanner/pull/1735

### DIFF
--- a/optaplanner-core/pom.xml
+++ b/optaplanner-core/pom.xml
@@ -55,6 +55,11 @@
     </dependency>
     <dependency>
       <groupId>org.drools</groupId>
+      <artifactId>drools-tms</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
       <artifactId>drools-wiring-dynamic</artifactId>
       <scope>runtime</scope>
     </dependency>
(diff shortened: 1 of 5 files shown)


danielezonca (approved):

optaplanner-core/pom.xml
```diff
@@ -53,6 +53,11 @@
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-tms</artifactId>
+      <scope>runtime</scope>
```
danielezonca, line 59: For the record: I'm not sure this works when used with Quarkus + OptaPlanner + DRL file that uses logical insert.
There is a drools check that fails if you are compiling a DRL with logical insert but TMS is not available and I guess this check is performed during deployment (aka compile time) in Quarkus and that that classpath does not contain runtime dependencies (like this one).
We should have a quarkus integration test with DRL with a logical insert to verify 🤔

triceo: @danielezonca Good point. We do have a Quarkus DRL integration test, but not one that uses logical inserts. Let's fix that after the new year, since CI is dead this year.

PR #1739, opened 2021-12-19
Reduce GC pressure in VariableListenerNotificationSupport

triceo:
Creates much less objects.
Also while I was there, cleans up some code.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/support/VariableListenerSupport.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/support/VariableListenerSupport.java
@@ -40,17 +40,17 @@ import org.optaplanner.core.impl.score.director.InnerScoreDirector;
 /**
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
-public class VariableListenerSupport<Solution_> implements SupplyManager<Solution_> {
+public final class VariableListenerSupport<Solution_> implements SupplyManager<Solution_> {
 
-    protected final InnerScoreDirector<Solution_, ?> scoreDirector;
+    private final InnerScoreDirector<Solution_, ?> scoreDirector;
 
-    protected final List<VariableListenerNotifiable> notifiableList;
-    protected final Map<VariableDescriptor<Solution_>, List<VariableListenerNotifiable>> sourceVariableToNotifiableMap;
-    protected final Map<EntityDescriptor<Solution_>, List<VariableListenerNotifiable>> sourceEntityToNotifiableMap;
-    protected final Map<Demand<Solution_, ?>, Supply> supplyMap;
-    protected int nextGlobalOrder = 0;
+    private final List<VariableListenerNotifiable<Solution_>> notifiableList;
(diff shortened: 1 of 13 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -82,11 +82,13 @@ public void linkVariableListeners() {
...
-                    VariableListenerNotifiable notifiable = new VariableListenerNotifiable(variableListener, globalOrder);
+                    VariableListenerNotifiable<Solution_> notifiable =
+                            new VariableListenerNotifiable<>(variableListener, globalOrder);
                     for (VariableDescriptor<Solution_> source : shadowVariableDescriptor.getSourceVariableDescriptorList()) {
-                        List<VariableListenerNotifiable> variableNotifiableList = sourceVariableToNotifiableMap.get(source);
+                        List<VariableListenerNotifiable<Solution_>> variableNotifiableList =
+                                sourceVariableToNotifiableMap.get(source);
                         variableNotifiableList.add(notifiable);
-                        List<VariableListenerNotifiable> entityNotifiableList = sourceEntityToNotifiableMap
+                        List<VariableListenerNotifiable<Solution_>> entityNotifiableList = sourceEntityToNotifiableMap
```
triceo, line 91: This is the meat. The class is now created out of the loop and reused; it is immutable, it can be safely shared.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -134,27 +137,30 @@ public void linkVariableListeners() {
...
+        if (!notifiableList.isEmpty()) {
+            VariableListenerNotification notification =
+                    new VariableListenerNotification(entity, VariableListenerNotificationType.ENTITY_ADDED);
+            for (VariableListenerNotifiable<Solution_> notifiable : notifiableList) {
+                Collection<VariableListenerNotification> notificationQueue = notifiable.getNotificationQueue();
+                boolean added = notificationQueue.add(notification);
+                if (added) {
+                    notifiable.getVariableListener().beforeEntityAdded(scoreDirector, entity);
+                }
             }
```
triceo, line 164: This is the meat. The class is now created out of the loop and reused; it is immutable, it can be safely shared.

rsynek (approved):
LGTM, thanks.

yurloc:
@triceo Did you measure the performance impact? Sometimes less code is better even though a few extra objects are created. So I wonder whether this change brings any observable performance improvement that would outweigh the extra if-s.

triceo:
@yurloc I did measure it as part of the DIMACS competition. This happens on the hot path, and so there is *a lot* of these instances being created and thrown away. And before they are thrown away, they are stored in multiple collections, which further increases memory use.

I'm not saying that the performance benefits of this are miraculous, but I am consistently seeing marginally higher score calc counts.

yurloc:
There is one problem with the unfinished Jenkins job - no Sonar analysis.

PR #1744, opened 2021-12-20
Modernize and unit-test Benchmarker statistic code

triceo:
Went looking for perf improvements, in the end I just cleaned up the code, using some modern Java.

### DIFF
--- a/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticPoint.java
+++ b/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticPoint.java
@@ -17,66 +17,62 @@
 package org.optaplanner.benchmark.impl.statistic;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
 
 /**
  * Implementations must be immutable.
  */
 public abstract class StatisticPoint {
 
+    private static final Pattern DOUBLE_QUOTE = Pattern.compile("\"\"");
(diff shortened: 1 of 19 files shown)


optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/SubSingleStatistic.java
```diff
@@ -152,40 +152,30 @@ private void readCsvStatisticFile() {
...
-                    String originalToken = stringDuplicationRemovalMap.get(token);
-                    if (originalToken == null) {
-                        stringDuplicationRemovalMap.put(token, token);
-                    } else {
-                        it.set(originalToken);
-                    }
-                }
+                List<String> csvLine = StatisticPoint.parseCsvLine(line)
+                        .stream()
+                        .map(String::intern)
```
rsynek, line 177: Very nice trick.

optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/statistic/StatisticPoint.java
```diff
@@ -17,66 +17,62 @@
...
 import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
 
 /**
  * Implementations must be immutable.
  */
 public abstract class StatisticPoint {
```
rsynek, line 30: I haven't seen the coverage data (we will have to wait for them), but I suspect this is one of the less tested classes. Maybe this is a good opportunity to not only modernize the code but also improve its coverage.

rsynek (approved):
LGTM, but let's wait for the CI and SonarCloud to see if we need to increase test coverage.

PR #1751, opened 2022-01-03
PLANNER-2508 Add externalized index supply

yurloc:
I paid a lot of attention to keep high level of code style consistency between external supplies and listeners and with the existing code. Some changes on the older code (before this PR) reflect that.

### Question 1: uniqueness requirement

Can we afford to require unique entity events in the inverse/index listeners/supplies? It has a performance cost but is necessary to make the strict checks work.

Alternatively, I can eliminate<sup>:monkey:</sup> the redundant before/after variable changed notifications in list moves if they happen on the same entity. I could then lift the requirement for unique events but that would open the door to breaking the checks for users that implement custom moves on a list variable.

Finally, I can do :monkey: even if the requirement for unique events is kept. Simply because it makes sense.
- Pros: A few (very few) before/after notifications won't be emitted.
- Cons: A bit of extra code and a check to detect if the move happens on the same entity.
- Neutral: `SmallScalingOrderedSet` will still be used due to the uniqueness requirement.

### JIRA
https://issues.redhat.com/browse/PLANNER-2508

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/ExternalizedIndexVariableSupply.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/ExternalizedIndexVariableSupply.java
@@ -19,6 +19,7 @@ package org.optaplanner.core.impl.domain.variable.index;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.optaplanner.core.api.score.director.ScoreDirector;
 import org.optaplanner.core.impl.domain.entity.descriptor.EntityDescriptor;
@@ -61,6 +62,13 @@ public class ExternalizedIndexVariableSupply<Solution_>
     }
 
     @Override
+    public boolean requiresUniqueEntityEvents() {
+        // A move on a single entity produces multiple before/after variable changed events for the given entity
+        // but the corrupted supply checks in insert/retract methods require a unique pair of before/after events.
+        return true;
+    }
+
+    @Override
     public void beforeEntityAdded(ScoreDirector<Solution_> scoreDirector, Object entity) {
         // Do nothing
     }
@@ -92,16 +100,32 @@ public class ExternalizedIndexVariableSupply<Solution_>
     }
 
     private void insert(ScoreDirector<Solution_> scoreDirector, Object entity) {
-        int i = 0;
-        for (Object shadowEntity : sourceVariableDescriptor.getListVariable(entity)) {
-            indexMap.put(shadowEntity, i);
-            i++;
+        int index = 0;
+        for (Object value : sourceVariableDescriptor.getListVariable(entity)) {
+            Integer oldIndex = indexMap.put(value, index);
+            if (oldIndex != null) {
+                throw new IllegalStateException("The supply (" + this + ") is corrupted,"
+                        + " because the entity (" + entity
+                        + ") for sourceVariable (" + sourceVariableDescriptor.getVariableName()
+                        + ") cannot be inserted: one of its values (" + value
+                        + ") at index (" + index + ") already has a non-null oldIndex (" + oldIndex + ").");
+            }
+            index++;
         }
     }
 
     private void retract(ScoreDirector<Solution_> scoreDirector, Object entity) {
-        for (Object shadowEntity : sourceVariableDescriptor.getListVariable(entity)) {
-            indexMap.remove(shadowEntity);
+        int index = 0;
+        for (Object value : sourceVariableDescriptor.getListVariable(entity)) {
+            Integer oldIndex = indexMap.remove(value);
+            if (!Objects.equals(oldIndex, index)) {
+                throw new IllegalStateException("The supply (" + this + ") is corrupted,"
+                        + " because the entity (" + entity
+                        + ") for sourceVariable (" + sourceVariableDescriptor.getVariableName()
+                        + ") cannot be retracted: one of its values (" + value
+                        + ") at index (" + index + ") already has an unexpected oldIndex (" + oldIndex + ").");
+            }
+            index++;
         }
     }
 
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/ExternalizedIndexVariableSupply.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/ExternalizedIndexVariableSupply.java
@@ -100,8 +100,12 @@ public class ExternalizedIndexVariableSupply<Solution_>
     }
 
     private void insert(ScoreDirector<Solution_> scoreDirector, Object entity) {
+        List<Object> listVariable = sourceVariableDescriptor.getListVariable(entity);
+        if (listVariable == null) {
+            return;
+        }
         int index = 0;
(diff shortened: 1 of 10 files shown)


optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/DefaultLocalSearchPhaseTest.java
```diff
@@ -213,4 +215,15 @@ public void solveListVariable() {
...
+    @Test
+    public void solveListVariableWithExternalizedInverseAndIndexSupplies() {
+        SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(
+                TestdataListSolutionExternalized.class, TestdataListEntityExternalized.class);
+
+        TestdataListSolutionExternalized solution = TestdataListSolutionExternalized.generateUninitializedSolution(6, 2);
+
+        solution = PlannerTestUtils.solve(solverConfig, solution);
+        assertThat(solution).isNotNull();
+    }
```
yurloc, line 227: Maybe this test and the one above deserve a better place (a separate test class) but where? I can't find a suitable existing package for that.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/ExternalizedIndexVariableSupply.java
```diff
@@ -0,0 +1,149 @@
...
+    @Override
+    public VariableDescriptor<Solution_> getSourceVariableDescriptor() {
+        return sourceVariableDescriptor;
+    }
+
+    @Override
+    public void resetWorkingSolution(ScoreDirector<Solution_> scoreDirector) {
+        EntityDescriptor<Solution_> entityDescriptor = sourceVariableDescriptor.getEntityDescriptor();
+        List<Object> entityList = entityDescriptor.extractEntities(scoreDirector.getWorkingSolution());
+        indexMap = new HashMap<>();
```
ge0ffrey, line 53: (important) Use an new IdentityHashMap() instead of a HashMap to avoid being affected by how the user decides to define equals/hashcode on their domain objects.

yurloc: Absolutely! I overlooked the fact that although the map holds _Integers_, user domain objects are the keys!

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/IndexVariableListener.java
```diff
@@ -66,38 +73,56 @@ public void afterEntityRemoved(ScoreDirector<Solution_> scoreDirector, Object en
...
-                indexShadowVariableDescriptor.setValue(shadowEntity, index);
-                scoreDirector.afterVariableChanged(indexShadowVariableDescriptor, shadowEntity);
+            Object oldIndex = shadowVariableDescriptor.getValue(value);
+            if (oldIndex != null) {
+                throw new IllegalStateException("The entity (" + entity
+                        + ") has a list variable (" + sourceVariableDescriptor.getVariableName()
+                        + ") and one of its values (" + value
+                        + ") which has a shadow variable (" + shadowVariableDescriptor.getVariableName()
+                        + ") has an oldIndex (" + oldIndex + ") which is not null.\n"
+                        + "Verify the consistency of your input problem for that shadow variable.");
```
ge0ffrey, line 92: nitpick: we typically start the advise line after the \n with "Maybe", so "Maybe verify ...".
Why? Because the first line is 100% sure: that is the problem, that is the error. The second line, the advise, is never 100% sure. It could be something else too, in theory at least. The "maybe" signals to users that we're not guaranteeing that this advise always works, we just think it works most of the time.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/ExternalizedSingletonListInverseVariableSupply.java
```diff
@@ -27,7 +27,7 @@
 import org.optaplanner.core.impl.domain.variable.listener.SourcedVariableListener;
 
 /**
- * Alternative to {@link SingletonInverseVariableListener}.
+ * Alternative to {@link SingletonListInverseVariableListener}.
  */
 public class ExternalizedSingletonListInverseVariableSupply<Solution_>
```
ge0ffrey, line 32: Are we sure that "...SingletonList..." is a good name?

optaplanner-core/src/test/java/org/optaplanner/core/impl/localsearch/DefaultLocalSearchPhaseTest.java
ge0ffrey: Adding it here is consistent with what we've done so far.

But if this is the best place for all those methods here, that is indeed not sure. Maybe there's a better structure that would be easier to find them, good for another PR maybe. I am not too worried about this now.

ge0ffrey (approved):

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/index/IndexVariableListener.java
yurloc: Correct but this is currently consistent with rest of the built-in listeners. I've filed https://issues.redhat.com/browse/PLANNER-2619 to review all of them.

optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/ExternalizedSingletonListInverseVariableSupply.java
yurloc: The existing classes for the basic variable are `SingletonInverseVariableListener` and `ExternalizedSingletonInverseVariableSupply`. I assume that adding `List` somewhere is a good way to differentiate the corresponding pair of classes designed for the list variable. But I agree that gluing five nouns together leaves a space for many interpretations.

In both cases they are a supply/listener that provide an "inverse relation" "(shadow) variable" that is a "singleton". So probably we should keep "SingletonInverseVariable" together and just prefix it with "List", so:

- `SingletonInverseVariableListener`
- `ExternalizedSingletonInverseVariableSupply`
- `ListSingletonInverseVariableListener`
- `ExternalizedListSingletonInverseVariableSupply`

Wdyt?

We could also prefix the existing classes with "Basic" but we'd also have to apply that to the "Collection" variant, so:

- `BasicCollectionInverseVariableListener`
- `ExternalizedBasicCollectionInverseVariableSupply`
- `BasicSingletonInverseVariableListener`
- `ExternalizedBasicSingletonInverseVariableSupply`
- `ListSingletonInverseVariableListener`
- `ExternalizedListSingletonInverseVariableSupply`

yurloc: Not a blocker for this PR. Let's continue the naming discussion in the design document.

PR #1765, opened 2022-01-10
PLANNER-2621 Cache score director factory

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -65,24 +65,27 @@
     private static final long DEFAULT_RANDOM_SEED = 0L;
 
     private final SolverConfig solverConfig;
+    private volatile InnerScoreDirectorFactory<Solution_, ?> scoreDirectorFactory;
 
     public DefaultSolverFactory(SolverConfig solverConfig) {
         this.solverConfig = Objects.requireNonNull(solverConfig, "The solverConfig (" + solverConfig + ") cannot be null.");
     }
 
     public InnerScoreDirectorFactory<Solution_, ?> getScoreDirectorFactory() {
```
triceo, line 74: I considered whether or not to make this lazy and I decided not to do the work in the constructor. It seems the user should only be paying the penalty if they actually ask for the solver. That said, I am open to reconsidering, if you have a better argument.

optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -65,24 +66,37 @@
...
     private final SolverConfig solverConfig;
+    private volatile InnerScoreDirectorFactory<Solution_, ?> scoreDirectorFactory;
 
     public DefaultSolverFactory(SolverConfig solverConfig) {
         this.solverConfig = Objects.requireNonNull(solverConfig, "The solverConfig (" + solverConfig + ") cannot be null.");
     }
 
-    public InnerScoreDirectorFactory<Solution_, ?> getScoreDirectorFactory() {
-        return buildScoreDirectorFactory(solverConfig.determineEnvironmentMode());
+    public <Score_ extends Score<Score_>> InnerScoreDirectorFactory<Solution_, Score_> getScoreDirectorFactory() {
```
ge0ffrey, line 75: This builds the score directory factory lazily. Which means the first solving requests pays the cost initialization cost.
I don't think it should be build lazily - it should be build when the SolverFactory is created.

ge0ffrey: Nitpick: naming wise calling this get* when it actually also builds it, doesn't feel entirely right to me.

ge0ffrey (commented):
LGTM except for the lazy initialization part

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/test/java/org/optaplanner/quarkus/constraints/OptaPlannerProcessorConstraintProviderAndDrlTest.java
```diff
@@ -1,70 +0,0 @@
...
-import org.optaplanner.quarkus.KieRuntimeBuilderMock;
-import org.optaplanner.quarkus.deployment.config.OptaPlannerBuildTimeConfig;
-import org.optaplanner.quarkus.testdata.normal.constraints.TestdataQuarkusConstraintProvider;
-import org.optaplanner.quarkus.testdata.normal.domain.TestdataQuarkusEntity;
-import org.optaplanner.quarkus.testdata.normal.domain.TestdataQuarkusSolution;
-
-import io.quarkus.deployment.builditem.CapabilityBuildItem;
-import io.quarkus.test.QuarkusUnitTest;
-
-public class OptaPlannerProcessorConstraintProviderAndDrlTest {
```
triceo, line 40: @Christopher-Chianelli This is a side-effect of my change. Specifically:

    The kieBaseSupplier (org.optaplanner.quarkus.gizmo.OptaPlannerDroolsInitializer$Implementation$$function$$1@ba17be6) is not a KieBaseDescriptor. Maybe remove calls to ScoreDirectorFactoryConfig.setKieBaseSupplier(Supplier)?
    	at org.optaplanner.core.impl.score.director.stream.DroolsConstraintStreamScoreDirectorFactory.assertIsKieBaseDescriptor(DroolsConstraintStreamScoreDirectorFactory.java:83)
	    at org.optaplanner.core.impl.score.director.stream.DroolsConstraintStreamScoreDirectorFactory.<init>(DroolsConstraintStreamScoreDirectorFactory.java:89)
	    
I think this happens because:
- `SolverFactory` now creates `ScoreDirectorFactory` in the constructor,
- and therefore if the constructor gets a wrong `SolverConfig`, it throws this exception immediately.

The situation that you were testing therefore can not happen, this exception will be thrown earlier.
That said, I am not too excited about deleting this test - can you suggest a better approach, with your knowledge of the OP-Quarkus codebase?

triceo:
Jenkins run native tests.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/test/java/org/optaplanner/quarkus/constraints/OptaPlannerProcessorConstraintProviderAndDrlTest.java
Christopher-Chianelli: This happens because DRL need to get the generated KieBase in Quarkus, so a custom Supplier is written and passed to GizmoKieBaseSupplier (which get said generated KieBase). However, the KieBase Supplier required by DroolsConstraintStream is required to be a particular type, and since the same KieBase Supplier is used for both Drools score calculation method, this exception is thrown.

The situation being tested is a helpful error message is thrown if both DRL and Constraint Streams are used (in particular, both DRL on classpath + ConstraintProvider.

We might be able to throw this exception at Quarkus build time. What happens if you set a constraint provider + score drl in base OptaPlanner now?

Christopher-Chianelli: Look like this happens because all the factories are built first, and then the "only one" factory constraint is validated.

Christopher-Chianelli: The fix would be to either validate the "only one" score directory factory constraint before the factories are built, or to throw the exception at Quarkus build time (which will do the "only one" score directory factory constraint check).

triceo: I was thinking the same. I may have to do the SPI in two stages - first check if configured, then instantiate if selected. That way, I will be able to avoid touchich Quarkus altogether.

PR #1768, opened 2022-01-10
PLANNER-2626, PLANNER-2624: Fix inherited benchmark config

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-2624
https://issues.redhat.com/browse/PLANNER-2626

### Checklist

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
@@ -520,7 +520,6 @@ public class TerminationConfig extends AbstractConfig<TerminationConfig> {
     @XmlTransient
     public boolean isConfigured() {
         return terminationClass != null ||
-                terminationCompositionStyle != null ||
                 spentLimit != null ||
                 millisecondsSpentLimit != null ||
                 secondsSpentLimit != null ||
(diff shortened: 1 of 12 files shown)


Christopher-Chianelli:
Note: currently does not inherit the score calculation method, since different benchmarks might use different score calculation methods. Debating how it should be inherited.

triceo:
I'm thinking that we should:

- Inherit everything, it's inheritance after all.
- But then remove some things that do not make sense to be inherited. (Such as if I decided to use different phases; currently, they are just added at the end of the existing list.)

We can discuss tomorrow.

triceo:
(Question: what does the normal benchmarker do? We should do the same, or if it has the same bug, fix it there too.)

triceo (changes requested):
We should completely solve the inheritance issue before we merge.

optaplanner-core/src/main/java/org/optaplanner/core/config/solver/termination/TerminationConfig.java
```diff
@@ -520,21 +520,27 @@ public Long calculateUnimprovedTimeMillisSpentLimit() {
     @XmlTransient
     public boolean isConfigured() {
         return terminationClass != null ||
+                terminationCompositionStyle != null ||
```
triceo, line 523: Does this mean that termination is configured? Having just the composition style and nothing else still means there is no actual termination, right?

Christopher-Chianelli: It was missing scoreCalculationCount before, so I copied and pasted all fields in prop order. I think in most cases, if at least one property is set, then there a high probability that an actual termination property is set. `isConfigured` returning true does not guarantee it will terminate. So either, we pick a subset of properties, increasing maintenance cost (since we need to know if a missing property should be checked or not), or we include everything, reducing maintenance cost but allowing some unconfigured termination to return true.

triceo: @Christopher-Chianelli That is an interesting trade-off. In my opinion, we should prefer correctness over maintenance costs. In order words: we should only improve maintenance costs if the code is already correct.

Assuming we can centralize this logic in a single place, I vote that we reduce the set of null checks to just the actual terminations. After that, I think this can be merged.

triceo (approved):

PR #1769, opened 2022-01-10
PLANNER-2303 Fix problem scale for list variable

yurloc:
### JIRA

https://issues.redhat.com/browse/PLANNER-2303

The formula is `v^2 / e + v`. It's designed so that it gives the same problem scale with the list variable as with the chained basic variable. I can't see any good meaning behind it. Nevertheless, I can live with this being merged provided that we replace the "problem size" metric with solution space as proposed in https://issues.redhat.com/browse/PLANNER-2623.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
@@ -567,7 +567,16 @@ public class EntityDescriptor<Solution_> {
     public long getProblemScale(Solution_ solution, Object entity) {
         long problemScale = 1L;
         for (GenuineVariableDescriptor<Solution_> variableDescriptor : effectiveGenuineVariableDescriptorList) {
-            problemScale *= variableDescriptor.getValueCount(solution, entity);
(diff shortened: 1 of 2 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -567,7 +567,15 @@ public long getMaximumValueCount(Solution_ solution, Object entity) {
     public long getProblemScale(Solution_ solution, Object entity) {
         long problemScale = 1L;
         for (GenuineVariableDescriptor<Solution_> variableDescriptor : effectiveGenuineVariableDescriptorList) {
-            problemScale *= variableDescriptor.getValueCount(solution, entity);
+            long valueCount = variableDescriptor.getValueCount(solution, entity);
+            problemScale *= valueCount;
+            if (variableDescriptor.isListVariable()) {
+                // This formula probably makes no sense other than that it results in the same problem scale for both
+                // chained and list variable models.
```
ge0ffrey, line 574: optional: add a // TODO with a url to the jira about changing "problem scale" to "search space size"

ge0ffrey (approved):

PR #1770, opened 2022-01-10
Fail fast if the inverse shadow variable type is a Collection

yurloc:
A list variable's inverse shadow variable must a "singleton" inverse because each value can only be assigned to a single entity's list variable. Therefore, a Collection type is not supported and it should fail fast.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/InverseRelationShadowVariableDescriptor.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/InverseRelationShadowVariableDescriptor.java
@@ -62,6 +62,22 @@ public class InverseRelationShadowVariableDescriptor<Solution_> extends ShadowVa
         linkShadowSources(descriptorPolicy);
     }
 
+    /**
+     * The following table describes which combinations of source and shadow variable types are supported.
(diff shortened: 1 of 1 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/inverserelation/InverseRelationShadowVariableDescriptor.java
```diff
@@ -62,6 +62,22 @@ public void linkVariableDescriptors(DescriptorPolicy descriptorPolicy) {
...
+     *
+     * <pre>
+     * {@code
+     *          | singleton | collection
+     * -----------------------------------
+     *   basic  |     ✕     |     ✓
+     * chained  |     ✓     |     ✕
+     *    list  |     ✓     |     ✕
+     *}
+     * </pre>
```
ge0ffrey, line 77: (nitpick)
In the table its not clear to me if the rows or columns are the "source" side.

It also seems elaborate. How about this instead:
```suggestion
     * Sourced on a basic genuine planning variable, the shadow type is a Collection (such as List or Set).
     * Sourced on a list or chained planning variable, the shadow variable type is a single instance.
```

ge0ffrey (approved):
Good to merge, with or without nitpick suggestions

PR #1771, opened 2022-01-10
PLANNER-2627: Move step score map to solve-lifetime scope

Christopher-Chianelli:
Before, the step score map was configured to a phase-lifetime object.
This is fine when there is only one phase, but there is typically
many phases. Thus, when a new phase is started, the step score map
that the gauge tracks is never updated (since it linked to the old,
forgotten, phase). The step score map now lives in SolverScope, so
it should work across the entire solver run.

Maybe should be backported.

### JIRA

https://issues.redhat.com/browse/PLANNER-2627

### Checklist

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
@@ -17,10 +17,6 @@
 package org.optaplanner.core.impl.phase;
 
 import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicReference;
 
 import org.optaplanner.core.api.domain.solution.PlanningSolution;
 import org.optaplanner.core.api.score.Score;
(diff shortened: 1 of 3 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
```diff
@@ -53,7 +47,6 @@
 
     protected final int phaseIndex;
     protected final String logIndentation;
-    protected final Map<Tags, List<AtomicReference<Number>>> stepScoreMap = new ConcurrentHashMap<>();
```
triceo, line 56: I checked the code and I'm not really sure why we need the `AtomicReference` there.
The `Number` is there because we do not know what the `Score` will return - fair enough. But the `AtomicReference`?

If you need concurrent modifications to the list, there is the `CopyOnWriteArrayList`.
Is there any particular reason why the change of the list element should be atomic? `List.set(...)` should be enough here.

triceo (approved):
LGTM, but check the comment inside.

optaplanner-core/src/main/java/org/optaplanner/core/impl/phase/AbstractPhase.java
Christopher-Chianelli: It more complex to do it with a list, since the Gauge function is given whatever is passed to it, meaning we would need to also store the index in the list to fetch versus just the atomic reference (currently, the atomic reference is passed to the gauge, and the gauge read directly from the atomic reference, bypassing the List). See Section 9 in https://micrometer.io/docs/concepts . In theory, it does not need to be atomic, just settable. However, the concept page only show atomic types.

PR #1778, opened 2022-01-12
PLANNER-2614 Remove easy forward references from the API package

triceo:


optaplanner-core/src/main/java/org/optaplanner/core/api/domain/valuerange/ValueRange.java
```diff
@@ -70,7 +58,7 @@
      * Scales well because it does not require caching.
      *
      * @param workingRandom never null, the {@link Random} to use when any random number is needed,
-     *        so {@link EnvironmentMode#REPRODUCIBLE} works correctly. {@link RandomUtils} can be useful too.
+     *        so we can achieve proper reproducible environments.
```
ge0ffrey, line 61: Avoid "we" and "I" in javadocs and docs

```suggestion
     *        so runs are reproducible when desired.
```

ge0ffrey (approved):

optaplanner-core/src/main/java/org/optaplanner/core/api/domain/valuerange/CountableValueRange.java
```diff
@@ -19,17 +19,12 @@
 import java.util.Iterator;
 
 import org.optaplanner.core.api.domain.variable.PlanningVariable;
-import org.optaplanner.core.impl.domain.valuerange.AbstractCountableValueRange;
 
 /**
  * A {@link ValueRange} that is ending. Therefore, it has a discrete (as in non-continuous) range.
- * <p>
- * An implementation must extend {@link AbstractCountableValueRange}
- * to ensure backwards compatibility in future versions.
```
yurloc, line 28: Is it safe to remove this? It is part of the API's *contract* that the implementer must follow in order to be guaranteed backward compatibility.

triceo: We do the same thing everywhere else (Score directors have their abstract counterparts, Scores do etc.), and we do not warn about it. In fact, this Javadoc is a bit of a lie - `AbstractCountableValueRange` is not public API, therefore we are making no guarantees on it either. :-)

yurloc: It depends whether users are encouraged to implement this or not. They're definitely not encouraged to implement Score or or ScoreDirector so there it's OK to have internal abstract base classes and be silent about them.

On the other hand, _if_ (and I might be wrong) we tell users it's a good idea to implement their own value range then we must provide them with complete information on how to do that correctly.

Yes, the contract does require them to use an impl code that might break in compile time after major upgrade. But at least it's visible as opposed to breaking the behavior of their code more or less silently if they accidentally don't extend the abstract base.

yurloc: > we are making no guarantees on it either

I don't think so. If I understand it correctly. We reserve the right to cause compilation errors in a major update but we guarantee that if they extend the abstract base, their code will work correctly.

triceo: > > we are making no guarantees on it either
> 
> I don't think so. If I understand it correctly. We reserve the right to cause compilation errors in a major update but we guarantee that if they extend the abstract base, their code will work correctly.

I have to say I am aware of no such requirement. When I need to, I break abstract/inner classes left and right.

> It depends whether users are encouraged to implement this or not. They're definitely not encouraged to implement Score or or ScoreDirector so there it's OK to have internal abstract base classes and be silent about them.

We currently do support custom scores, yet we do not mention abstract score. They need to implement it using abstract score, because otherwise OP will never work - but we do not actually mention it. As far as I know.
 
> On the other hand, _if_ (and I might be wrong) we tell users it's a good idea to implement their own value range then we must provide them with complete information on how to do that correctly.

That is a good question. I do not really see value ranges being used at all. Still, I think what I said above applies.

yurloc: > > It depends whether users are encouraged to implement this or not. They're definitely not encouraged to implement Score or or ScoreDirector so there it's OK to have internal abstract base classes and be silent about them.
> 
> We currently do support custom scores, yet we do not mention abstract score. They need to implement it using abstract score, because otherwise OP will never work - but we do not actually mention it. As far as I know.

I've checked the docs and there are no instructions to implement a custom value range. There are even no implementations in the examples. Only the built-in ones. So it's safe to remove that. It's an implementation detail that doesn't belong to the public API Javadoc.

A good example of where this type of information is necessary is the `Move` interface. Although it's an internal interface, users are encouraged to implement custom moves. And so, they need to know this:
https://github.com/kiegroup/optaplanner/blob/61c6f24469566fd7cf273cfdd43a6c72398fee59/optaplanner-core/src/main/java/org/optaplanner/core/impl/heuristic/move/Move.java#L46

rsynek: Both are valid arguments. If we really want to support `CountableValueRange`, the `AbstractCountableValueRange` should be a part of the API too. If nobody is using it, can we afford to deprecate the value ranges without any replacement?

triceo: At the moment, I see no reason for deprecating anything. I've just removed a pointer to a non-public API - I'd argue I did not practically change anything for anyone.

PR #1779, opened 2022-01-12
PLANNER-2635 Switch Task Assigning with list variable to Constraint Streams

yurloc:
Replicates 9ab908311b22acf121fcc340cedbc0c4c4f2b382 onto the list variable version of the Task Assigning example.

### DIFF
--- a/optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning_listvariable/score/TaskAssigningConstraintProvider.java
+++ b/optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning_listvariable/score/TaskAssigningConstraintProvider.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.optaplanner.examples.taskassigning_listvariable.optional.score;
+package org.optaplanner.examples.taskassigning_listvariable.score;
 
 import org.optaplanner.core.api.score.buildin.bendable.BendableScore;
 import org.optaplanner.core.api.score.stream.Constraint;
@@ -47,7 +47,7 @@ public final class TaskAssigningConstraintProvider implements ConstraintProvider
 
     private UniConstraintStream<Task> getTaskWithPriority(ConstraintFactory constraintFactory, Priority priority) {
         return constraintFactory.forEach(Task.class)
-                .filter(task -> task.getPriority() == priority);
+                .filter(task -> task.getEmployee() != null && task.getPriority() == priority);
     }
 
     private Constraint noMissingSkills(ConstraintFactory constraintFactory) {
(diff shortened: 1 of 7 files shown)


optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning_listvariable/score/TaskAssigningConstraintProvider.java
```diff
@@ -47,7 +47,7 @@
 
     private UniConstraintStream<Task> getTaskWithPriority(ConstraintFactory constraintFactory, Priority priority) {
         return constraintFactory.forEach(Task.class)
-                .filter(task -> task.getPriority() == priority);
+                .filter(task -> task.getEmployee() != null && task.getPriority() == priority);
```
yurloc, line 50: This is the only difference from 9ab908311b22acf121fcc340cedbc0c4c4f2b382.

I have to do this to avoid NPE in CH and because `forEach()` doesn't filter out shadow planning entities with uninitialized shadow variables.

@triceo Is this the most effective way to do that?

triceo: Yeah, I do not see any other way.
That said, maybe it should be discussed? The point of `forEach(...)` is to eliminate null checks, and this seems like a corner case that escaped.

triceo (approved):
Good on its own.

Two questions, however:
- forEach() and shadow entities. That looks like something that totally escaped us when we were designing the API.
- Why have this example? Or rather, why keep the old one? I'd suggest to keep one implementation of the old chained model. (Maybe in TSP which is trivial?) And to move everything else forward. List variable is so much better than I really think we should push it hard.

optaplanner-examples/src/main/java/org/optaplanner/examples/taskassigning_listvariable/score/TaskAssigningConstraintProvider.java
yurloc: According to Proposal I, this behavior was chosen on purpose. I have to say I don't fully understand the reasoning that explains Proposal I rejection but I assume it was understood during the discussion.

In other words, this corner case hasn't escaped our attention. Whether it is a design flaw or not, I'm not sure though.

yurloc:
> forEach() and shadow entities. That looks like something that totally escaped us when we were designing the API.

I don't think it has escaped. See above.
 
> Why have this example? Or rather, why keep the old one? I'd suggest to keep one implementation of the old chained model. (Maybe in TSP which is trivial?) And to move everything else forward. List variable is so much better that I really think we should push it hard.

This is going to `main` soon, overriding the original chained Task Assigning. PR by the end of this week.

PR #1785, opened 2022-01-13
PLANNER-2614 Remove forward dependency on ClassAndPlanningIdComparator

triceo:
Coincidentally also significantly speeds up CM as comparisons no longer require `MemberAccessor`s.

### DIFF
--- a/optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
+++ b/optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
@@ -22,25 +22,39 @@ import java.util.Comparator;
 import java.util.List;
 
 import org.optaplanner.core.api.score.Score;
-import org.optaplanner.core.impl.domain.lookup.ClassAndPlanningIdComparator;
 
 /**
  * Retrievable from {@link ConstraintMatchTotal#getConstraintMatchSet()}
  * and {@link Indictment#getConstraintMatchSet()}.
  *
- * <p>
- * This class has a {@link #compareTo(ConstraintMatch)} method which is inconsistent with equals.
- * (See {@link Comparable}.)
- * Two different {@link ConstraintMatch} instances with the same justification list aren't
- * {@link Object#equals(Object) equal} because some ConstraintStream API methods can result in duplicate facts,
- * which are treated as independent matches.
- * Yet two instances may {@link #compareTo(ConstraintMatch)} equal in case they come from the same constraint and their
- * justifications are equal.
- * This is for consistent ordering of constraint matches in visualizations.
- * 
  * @param <Score_> the actual score type
  */
 public final class ConstraintMatch<Score_ extends Score<Score_>> implements Comparable<ConstraintMatch<Score_>> {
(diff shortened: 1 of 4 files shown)


optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -90,34 +104,38 @@ public String getIdentificationString() {
         return getConstraintId() + "/" + justificationList;
     }
 
+    /**
+     * As defined by {@link Comparable#compareTo(Object)}.
```
ge0ffrey, line 108: I would remove this line, as everyone knows compareTo

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -90,34 +104,38 @@ public String getIdentificationString() {
...
+     *
+     * <p>
+     * The point of this method is to provide consistency when a collection of matches is visualized.
```
ge0ffrey, line 111:
```suggestion
     * This sole purpose of this compareTo method to allow reproducible and consistent visualization of a collection of matches.
```

optaplanner-core/src/main/java/org/optaplanner/core/api/score/constraint/ConstraintMatch.java
```diff
@@ -90,34 +104,38 @@ public String getIdentificationString() {
...
+     * Instances of this class should never be compared to one another for any other reason.
```
ge0ffrey, line 112:
```suggestion
     * Instances of this class should never be compared to each another for any other reason.
```

optaplanner-core/src/main/java/org/optaplanner/core/impl/score/director/ClassAndPlanningIdComparator.java
```diff
@@ -14,41 +14,32 @@
...
 import java.util.HashMap;
 import java.util.Map;
 
 import org.optaplanner.core.api.domain.common.DomainAccessType;
-import org.optaplanner.core.api.domain.lookup.PlanningId;
 import org.optaplanner.core.config.util.ConfigUtils;
 import org.optaplanner.core.impl.domain.common.accessor.MemberAccessor;
 
-public class ClassAndPlanningIdComparator implements Comparator<Object> {
+final class ClassAndPlanningIdComparator implements Comparator<Object> {
```
ge0ffrey, line 28: I'd still make it "final public", in case anyone is using it. It's in impl, so they aren't allowed to, but no gain in breaking them if they do.

```suggestion
final public class ClassAndPlanningIdComparator implements Comparator<Object> {
```

ge0ffrey (changes requested):
PlanningId changed might affect demos, but this change might be worth it nonetheless.

On hold until we prove we can (or until have to plan to) complete the separate api jar story on the other fronts, so this change is worth it.

triceo:
Closing as unnecessary.

PR #1787, opened 2022-01-14
Reorganize new variable descriptor code to improve consistency

yurloc:


optaplanner-core/src/main/java/org/optaplanner/core/impl/domain/variable/descriptor/BasicVariableDescriptor.java
```diff
@@ -27,10 +27,18 @@
     private boolean chained;
     private boolean nullable;
 
+    // ************************************************************************
+    // Constructors and simple getters/setters
```
yurloc, line 31: Remove getters/setters part.

yurloc: No, it's truthful. What should be done instead is to move getters to the bottom but that's out of scope of this PR.

PR #1794, opened 2022-01-17
PLANNER-2641 Rely on Drools 8 only and remove code dependencies on Kogito snapshots

triceo:
Simplify the build by not pulling in a kogito snapshot and just use Drools 8.


ge0ffrey (approved):

Christopher-Chianelli (approved):
LGTM; although it does break BC for anyone using drools in Quarkus Native

PR #1870, opened 2022-03-06
Enable FilteringEntitySelector to support list iterator

artysidorenko:
Hello 👋 

I ran into difficulties benchmarking with a dataset that included pinned entities, and came to learn the cause was I couldn't run `VARIABLE_NEIGHBORHOOD_DESCENT` with pinned entities, because [FilteringEntitySelector#listIterator](https://github.com/kiegroup/optaplanner/blob/32b4fb0ad0ccc13f72ac940a9709c15db1bdd9e0/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/entity/decorator/FilteringEntitySelector.java#L123-L126) had not yet been implemented.

I was interested in the internals so had a stab at doing this by extending the existing [UpcomingSelectionIterator](https://github.com/kiegroup/optaplanner/blob/32b4fb0ad0ccc13f72ac940a9709c15db1bdd9e0/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java) into a new class that could also iterate backwards, and using that in the implementation.

Updated unit tests to include this, though wasn't sure if you guys would want a more comprehensive integration test as well? e.g. could add some pinned entities into the tennis example [here for instance](https://github.com/kiegroup/optaplanner/blob/32b4fb0ad0ccc13f72ac940a9709c15db1bdd9e0/optaplanner-examples/data/tennis/unsolved/munich-7teams.xml#L174), and run one of the tennis tests with variable descent to check it all works end-to-end, if that's advisable.

Let me know if there's any questions or feedback, or if there's other reasons I missed for why this can't/shouldn't be supported.

### JIRA

Was not explicitly dealing with a JIRA ticket, though I came across this one that's related: https://issues.redhat.com/browse/PLANNER-196 (status "won't fix")

### DIFF
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/iterator/UpcomingSelectionListIterator.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/iterator/UpcomingSelectionListIterator.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2012 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


triceo:
Thank you for the PR! Someone will hopefully review it soon.
Ok to test.

PR #1886, opened 2022-03-22
Align meetingscheduling constraint provider with DRL

triceo:
- Every time `calculateOverlap(...)` is used, a non-null startingTimeGrain is required.
- Using a filtered nested stream is preferrable to using a filtering joiner.


rsynek (approved):
LGTM; please consider enhancing the constraint provider tests as these changes went unnoticed.

PR #1934, opened 2022-04-11
PLANNER-2674 Override logback-test in modules that would otherwise inherit them

triceo:
In a Maven build, we exclude logback-test.xml from JARs.
Therefore, each module needs to bring its own.
This may result in IDE warnings about duplicate configs, because IDEs typically do not use this Maven JAR plugin config.
We now warn about that directly in the inherited configs.


core/optaplanner-constraint-streams/pom.xml
```diff
@@ -22,7 +22,7 @@
   <url>https://www.optaplanner.org</url>
 
   <properties>
-    <java.module.name>org.optaplanner.constraint.drl</java.module.name>
+    <java.module.name>org.optaplanner.constraint.streams</java.module.name>
```
triceo, line 25: Unrelated, but I noticed it at the same time.

rsynek (approved):

PR #1950, opened 2022-04-23
CS-B: Clean up the ComparisonIndexer

triceo:
This PR has no measurable effect on performance.

### DIFF
--- a/core/optaplanner-constraint-streams/src/main/java/org/optaplanner/constraint/streams/bavet/common/index/ComparisonIndexer.java
+++ b/core/optaplanner-constraint-streams/src/main/java/org/optaplanner/constraint/streams/bavet/common/index/ComparisonIndexer.java
@@ -16,6 +16,7 @@
 
 package org.optaplanner.constraint.streams.bavet.common.index;
 
+import java.util.Comparator;
 import java.util.Map;
 import java.util.NavigableMap;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


core/optaplanner-constraint-streams/src/main/java/org/optaplanner/constraint/streams/bavet/common/index/ComparisonIndexer.java
```diff
@@ -28,33 +29,34 @@
 import org.optaplanner.constraint.streams.bavet.common.Tuple;
 import org.optaplanner.core.impl.score.stream.JoinerType;
 
-final class ComparisonIndexer<Tuple_ extends Tuple, Value_> implements Indexer<Tuple_, Value_> {
+final class ComparisonIndexer<Tuple_ extends Tuple, Value_, Key_ extends Comparable<Key_>>
+        implements Indexer<Tuple_, Value_> {
 
-    private final BiFunction<NavigableMap<Object, Indexer<Tuple_, Value_>>, Comparable, Map<Object, Indexer<Tuple_, Value_>>> submapFunction;
-    private final Function<IndexProperties, Comparable> comparisonIndexPropertyFunction;
+    private final SubmapBiFunction<Tuple_, Value_, Key_> submapExtractor;
```
triceo, line 35: This shows that the new interface introduced significantly simplifies the code.

core/optaplanner-constraint-streams/src/main/java/org/optaplanner/constraint/streams/bavet/common/index/ComparisonIndexer.java
```diff
@@ -103,4 +104,18 @@ public boolean isEmpty() {
...
     }
 
+    private static final class KeyComparator<Key_ extends Comparable<Key_>> implements Comparator<Key_> {
+
+        @Override
+        public int compare(Key_ o1, Key_ o2) { // Exists so that the comparison operations can be more easily debugged.
+            return o1.compareTo(o2);
+        }
+
+    }
```
triceo, line 114: This comparator is technically unnecessary, but it has two nice benefits:

- First, it allows us to easily debug the comparison operations by putting a breakpoint here, and not somewhere deep in the `TreeMap`.
- Second, this gets rid of the weird async profiler comparator behavior.

I did not find any performance impact of this.

ge0ffrey: I don't mind this, but this is actually a natural comperator and I suspect over time users will want to have `Joiners.lessThan(X::getAge, Y::getAge, new MyOwnAgeComperator())`

PR #1960, opened 2022-04-27
No value hack (do not merge as is)

ge0ffrey:
An experiment, to see if we need to redesign Indexer to avoid keeping the value set.
Clearly faster on course scheduling 07


triceo:
Interesting. Good idea, if it works.

ge0ffrey:
We don't want to merge this. But as it shows perf gain, there are 2 ways to investigate:
A) adjust Indexer to no longer contain a Value type (which is that Set or Map of the AB downstream tuples), but it still contains a Set<Tuple<A>> for each indexProperties
B) Use a linked list instead of a Set/Map of Tuple<A> in the indexer. Track those linked list entries such that they can be removed without being looked up by index. Basically replace indexer.remove(indexProperties, tupleA) by tupleA.indexerEntry.remove().

triceo:
I like A better. Sounds simpler. As long as it performs well enough, good for me.

triceo:
Cleaned up and resubmitted as https://github.com/kiegroup/optaplanner/pull/1986.

PR #1979, opened 2022-05-16
PLANNER-2716 Simplified RandomUnionMoveIterator without probabilities

triceo:


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/composite/UnionMoveSelectorFactory.java
```diff
@@ -52,20 +48,6 @@ public MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<Solut
             }
             selectorProbabilityWeightFactory = ConfigUtils.newInstance(config,
                     "selectorProbabilityWeightFactoryClass", config.getSelectorProbabilityWeightFactoryClass());
-        } else if (randomSelection) {
```
ge0ffrey, line 55: Removing this seems wierd

ge0ffrey: because this is the code that calls innerMoveSelectorConfig.getFixedProbabilityWeight()

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/composite/BiasedRandomUnionMoveIterator.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public Move<Solution_> next() {
+        if (stale) {
+            refreshMoveIteratorMap();
+        }
+        double randomOffset = RandomUtils.nextDouble(workingRandom, probabilityWeightTotal);
+        Map.Entry<Double, Iterator<Move<Solution_>>> entry = moveIteratorMap.floorEntry(randomOffset);
+        // entry is never null because randomOffset < probabilityWeightTotal
```
ge0ffrey, line 73: (nitpick)
```suggestion
        // The entry is never null because randomOffset < probabilityWeightTotal
```

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/composite/UnionMoveSelectorFactory.java
```diff
@@ -52,20 +48,6 @@ public MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<Solut
...
-            Map<MoveSelector<Solution_>, Double> fixedProbabilityWeightMap =
-                    new HashMap<>(config.getMoveSelectorList().size());
-            for (int i = 0; i < config.getMoveSelectorList().size(); i++) {
-                MoveSelectorConfig<?> innerMoveSelectorConfig = config.getMoveSelectorList().get(i);
-                MoveSelector<Solution_> moveSelector = moveSelectorList.get(i);
-                Double fixedProbabilityWeight = innerMoveSelectorConfig.getFixedProbabilityWeight();
```
ge0ffrey, line 61: (hard) This is a regression.

There are two ways to create a selectorProbabilityWeightFactory, as a user:
- with a custom one, with config.getSelectorProbabilityWeightFactoryClass()
- with a syntactic sugar one, with moveSelectorConfig.getFixedProbabilityWeight()

The second one is being removed in this PR and we have users/customers using that. They will silently fall back on uniform distribution (which is probably why the integration tests don't fail).

ge0ffrey (commented):
Good to merge if regression is fixed.

PR #2024, opened 2022-06-07
CS-B: GroupBy update

ge0ffrey:
I need to clean AbstractGroupNode up a bit further. This might already be faster because inserts and retracts use update on the out tuple if possible. That is more important than implementing the input update() method, which I'll do now.

### DIFF
--- a/core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
+++ b/core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
@@ -75,38 +75,40 @@ public abstract class AbstractGroupNode<InTuple_ extends Tuple, OutTuple_ extend
                     + ") was already added in the tupleStore.");
         }
         GroupKey_ groupKey = createGroupKey(tuple);
-        Group<OutTuple_, GroupKey_, ResultContainer_> group = groupMap.computeIfAbsent(groupKey,
-                k -> new Group<>(k, hasCollector ? supplier.get() : null));
+        Group<OutTuple_, GroupKey_, ResultContainer_> group = groupMap.get(groupKey);
+        if (group == null) {
+            ResultContainer_ resultContainer = hasCollector ? supplier.get() : null;
+            OutTuple_ outTuple = createOutTuple(groupKey);
+            outTuple.setState(BavetTupleState.CREATING);
+            group = new Group<>(groupKey, resultContainer, outTuple);
+            groupMap.put(groupKey, group);
+            // Don't add it if (state == CREATING), but (group != null), which is a 2th insert of the same groupKey.
+            dirtyGroupQueue.add(group);
+        }
         group.parentCount++;
-
         Runnable undoAccumulator = hasCollector ? accumulate(group.resultContainer, tuple) : null;
         GroupPart<Group<OutTuple_, GroupKey_, ResultContainer_>> groupPart = new GroupPart<>(group, undoAccumulator);
         tupleStore[groupStoreIndex] = groupPart;
 
-        if (group.state == null) {
-            group.state = BavetTupleState.CREATING;
-            dirtyGroupQueue.add(group);
-        } else {
-            switch (group.state) {
-                case CREATING:
-                    break;
-                case UPDATING:
-                    break;
-                case OK:
-                    group.state = BavetTupleState.UPDATING;
-                    dirtyGroupQueue.add(group);
-                    break;
-                case DYING:
-                    group.state = BavetTupleState.UPDATING;
-                    break;
-                case ABORTING:
-                    group.state = BavetTupleState.CREATING;
-                    break;
-                case DEAD:
-                default:
-                    throw new IllegalStateException("Impossible state: The group (" + group + ") in node (" +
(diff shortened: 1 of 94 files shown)


core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
```diff
@@ -22,35 +22,47 @@
...
     private final Queue<Group<OutTuple_, GroupKey_, ResultContainer_>> dirtyGroupQueue;
 
-    protected AbstractGroupNode(int groupStoreIndex, Supplier<ResultContainer_> supplier,
-            Consumer<OutTuple_> nextNodesInsert, Consumer<OutTuple_> nextNodesRetract) {
+    protected AbstractGroupNode(int groupStoreIndex,
+            Supplier<ResultContainer_> supplier,
+            Function<ResultContainer_, Result_> finisher,
+            Consumer<OutTuple_> nextNodesInsert,
+            Consumer<OutTuple_> nextNodesUpdate,
+            Consumer<OutTuple_> nextNodesRetract) {
```
triceo, line 59: Considering how much these three consumers appear *everywhere* and how many repetitious code lines they actually bring, may I suggest creating a new type to encapsulate them?

    interface TupleLifecycle<Tuple_ extends Tuple> {

        void insert(Tuple_ tuple);
        void update(Tuple_ tuple);
        void retract(Tuple_ tuple);

    }

Doesn't need to be done in this PR, but IMO needs to be done. This PR could have been half the diff if we had this already.

core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/uni/UniTuple.java
```diff
@@ -21,7 +21,8 @@
 
 public final class UniTuple<A> implements Tuple {
 
-    public final A factA;
+    // Only a tuple's origin node may modify a fact.
+    public A factA;
```
triceo, line 25: I understand why this exists, and I know there is no other way.
Still, I have to say that I absolutely hate that anyone can just come in and modify the tuples.

triceo (commented):
Some comments, no dealbreakers.

core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
ge0ffrey: Agreed, but different issue indeed.

core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/uni/UniTuple.java
ge0ffrey: Yea, but performance begs for it I hope. Let's see how the benchmarks turn out.

core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
```diff
@@ -118,29 +170,58 @@ public void retract(InTuple_ tuple) {
...
-                // Delay calculating right until it propagates
-                OutTuple_ tuple = createOutTuple(group);
-                group.tuple = tuple;
-                nextNodesInsert.accept(tuple);
-                tuple.setState(BavetTupleState.OK);
+            // Delay calculating finisher right until the tuple propagates
+            switch (group.state) {
+                case CREATING:
+                    outTuple = createOutTuple(group.groupKey);
+                    updateOutTupleToFinisher(outTuple, group.resultContainer);
```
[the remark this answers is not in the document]
ge0ffrey: Refactor done, should make sense now

core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
ge0ffrey: See zulip

core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/AbstractGroupNode.java
```diff
@@ -63,23 +75,149 @@ public void insert(InTuple_ tuple) {
...
+                group.outTuple.setState(BavetTupleState.CREATING);
+                break;
+            case DEAD:
+            default:
+                throw new IllegalStateException("Impossible state: The group (" + group + ") in node (" +
+                        this + ") is in an unexpected state (" + group.outTuple.getState() + ").");
         }
     }
 
     public void update(InTuple_ tuple) {
```
triceo, line 115: This method is very long, and its control flow confusing. Ifs, switches, breaks, ...
Maybe split it into some submethods for functional blocks, so that it becomes easier to reason about what happens when?

triceo (approved):
I think that the update code needs more structure. But overall, good PR.

PR #2056, opened 2022-07-04
Fix void with-method

chrfin-4:
Trivial two-line bug fix.

CustomPhaseConfig.withCustomProperties(Map) was void. Now returns the
instance after setting the property.

### JIRA

There is no issue for this, as far as I know.


triceo (approved):

triceo:
Thanks for the contribution!
OK to test.

PR #2075, opened 2022-07-12
Desugared use of APIs incompatible with Android

Namnodorel:
Android unfortunately supports only a subset of the current JDK API. This has so far not been a problem for OptaPlanner, but after some recent changes OptaPlanner apparently started using some of the unsupported APIs, causing runtime crashes.

[ServiceLoader.findFirst()](https://developer.android.com/reference/java/util/ServiceLoader) is not currently supported by Android. `Collection.toArray(IntFunction<T[]>)` [is available as of Android API Level 33](https://developer.android.com/reference/java/util/Collection#toArray(java.util.function.IntFunction%3CT[]%3E)) but [is not available as a "desugared" API yet](https://developer.android.com/studio/write/java8-support-table), which means it can not be used in apps that want to maintain backwards compatibility.

Fortunately, this is easily fixed. Both APIs are convenience methods, not adding any crucial new features. This PR replaces their use with the slightly-less-nice-looking-but-Android-compatible counterparts. I didn't touch any usage for example in the Drools module, as that isn't compatible with Android in the first place.

### DIFF
--- a/core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/BavetConstraintSessionFactory.java
+++ b/core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/BavetConstraintSessionFactory.java
@@ -74,7 +74,7 @@ public final class BavetConstraintSessionFactory<Solution_, Score_ extends Score
                 }
             }
         }
-        return new BavetConstraintSession<>(scoreInliner, declaredClassToNodeMap, nodeList.toArray(AbstractNode[]::new));
+        return new BavetConstraintSession<>(scoreInliner, declaredClassToNodeMap, nodeList.toArray(new AbstractNode[0]));
     }
 
 }
--- a/core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/BavetConstraintStreamScoreDirectorFactory.java
+++ b/core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/BavetConstraintStreamScoreDirectorFactory.java
@@ -49,7 +49,7 @@ public final class BavetConstraintStreamScoreDirectorFactory<Solution_, Score_ e
 
     @Override
     public Constraint[] getConstraints() {
-        return constraintList.toArray(Constraint[]::new);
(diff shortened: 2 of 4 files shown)


triceo:
Thanks for the contribution, OK to test.

I am curious though - we have never invested any effort into supporting Android, nor are we testing it or even considering the impact of our work on Android. May I ask about your use case? What have we broken? How can we find out before we do it next time?

Namnodorel:
Yeah, I doubt there are many people using OptaPlanner on Android :D There is a small (useful) [section about using OptaPlanner on Android](https://www.optaplanner.org/docs/optaplanner/latest/integration/integration.html#integrationWithAndroid) in the documentation, which makes me think that at least at some point someone has done this kind of thing before.

If you wanted to avoid such problems in the future, I imagine that you could configure code scanning tools to warn about Android-incompatible APIs. I have never needed that so far though, so I don't know if there are any ready-made configs for such a thing. 

Ultimately, this is a problem that _should_ be addressed by Google, but unfortunately it's just easier to work around it than it is to get Google to provide a full API.

Regarding the Use Case: The idea of the project I'm using OptaPlanner in is basically to utilize the power of constraint optimization for problems on an individual-scale. The app is effectively a very smart ToDo-App - you specify tasks, duration estimates, dependency relationships, etc. and after computing an appropriate plan the app tells you what tasks to complete on each day to meet deadlines while also distributing the workload equally over all days, and meeting various other goals.

Because it only needs to plan for one person, the search space stays relatively reasonable in size, and the high performance of Bavet make it feasible to perform this computation on mobile devices (though it's still more computationally expensive than most regular apps, leaving out games).

I'd like to provide a link for you to somewhere with more details, but the app unfortunately isn't public yet. I have used it daily for over a year now, though, and have found it quite useful. If you'd like, I could tell you about it when it eventually does go public, and you can have a look yourself.

rsynek:
@Namnodorel thanks for bringing this up. As you can see, OptaPlanner uses SonarCloud for static code analysis. Do you happen to know if some Sonar rules cover this situation? Unfortunately, I haven't found any.

triceo (approved):
LGTM. I took the liberty of updating the Android section of the docs, which I honestly didn't know existed.

When you're ready to share your project, reach out to us on Zulip. It does sound interesting.

Thanks for the contribution!

PR #2085, opened 2022-07-19
Fix key in documentation for log separation

simontiffert:
Change the key for the sifting appender to subSingleBenchmark.name in the documentation as referenced in the code at https://github.com/kiegroup/optaplanner/blob/5e36aecfa6b18b6f36e4e019fed5b70449176fea/optaplanner-benchmark/src/main/java/org/optaplanner/benchmark/impl/SubSingleBenchmarkRunner.java#L28


triceo:
Thanks for the contribution!
OK to test.

triceo (approved):

PR #2122, opened 2022-08-09
Clean up selector factories' build methods

yurloc:
I made some improvements to the code that builds selectors from configs. It should be now easier to understand because it's slightly less code and its structure is more consistent across all factories.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

- for <b>pull request checks</b>  
  Please add comment: <b>Jenkins retest this</b>

- for a <b>specific pull request check</b>  
  please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] tests</b>
  
- for a <b>full downstream build</b> 
  - for <b>jenkins</b> job: 
    please add comment: <b>Jenkins run fdb</b>
  - for <b>github actions</b> job: 
    add the label `run_fdb`

- for a <b>compile downstream build</b>  
  please add comment: <b>Jenkins run cdb</b>

- for a <b>full production downstream build</b>  
  please add comment: <b>Jenkins execute product fdb</b>

- for an <b>upstream build</b>  
  please add comment: <b>Jenkins run upstream</b>

- for <b>quarkus branch checks</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins run quarkus-branch</b>

- for a <b>quarkus branch specific check</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] quarkus-branch</b>

- for <b>quarkus main checks</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins run quarkus-main</b>

- for a <b>specific quarkus main check</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] quarkus-branch</b>

- for <b>native checks</b>  
  Run native checks  
  Please add comment: <b>Jenkins run native</b>

- for a <b>specific native check</b>  
  Run native checks 
  Please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] native</b>

- for <b>mandrel checks</b>  
  Run native checks against Mandrel image
  Please add comment: <b>Jenkins run mandrel</b>

- for a <b>specific mandrel check</b>  
  Run native checks against Mandrel image  
  Please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] mandrel</b>
</details>

### CI Status

 You can check OptaPlanner repositories CI status from [Chain Status webpage](https://kiegroup.github.io/optaplanner/).

### DIFF
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
@@ -37,6 +37,14 @@ public abstract class AbstractFromConfigFactory<Solution_, Config_ extends Abstr
         return entitySelectorConfig;
     }
 
+    protected EntityDescriptor<Solution_> deduceEntityDescriptor(HeuristicConfigPolicy<Solution_> configPolicy,
+            Class<?> entityClass) {
+        SolutionDescriptor<Solution_> solutionDescriptor = configPolicy.getSolutionDescriptor();
+        return entityClass == null
+                ? deduceEntityDescriptor(solutionDescriptor)
+                : deduceEntityDescriptor(solutionDescriptor, entityClass);
+    }
+
     protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor,
             Class<?> entityClass) {
         EntityDescriptor<Solution_> entityDescriptor =
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
@@ -41,14 +41,13 @@ public abstract class AbstractFromConfigFactory<Solution_, Config_ extends Abstr
             Class<?> entityClass) {
         SolutionDescriptor<Solution_> solutionDescriptor = configPolicy.getSolutionDescriptor();
         return entityClass == null
-                ? deduceEntityDescriptor(solutionDescriptor)
-                : deduceEntityDescriptor(solutionDescriptor, entityClass);
+                ? getTheOnlyEntityDescriptor(solutionDescriptor)
+                : getEntityDescriptorForClass(solutionDescriptor, entityClass);
     }
 
-    protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor,
+    private EntityDescriptor<Solution_> getEntityDescriptorForClass(SolutionDescriptor<Solution_> solutionDescriptor,
             Class<?> entityClass) {
-        EntityDescriptor<Solution_> entityDescriptor =
-                solutionDescriptor.getEntityDescriptorStrict(Objects.requireNonNull(entityClass));
+        EntityDescriptor<Solution_> entityDescriptor = solutionDescriptor.getEntityDescriptorStrict(entityClass);
         if (entityDescriptor == null) {
             throw new IllegalArgumentException("The config (" + config
                     + ") has an entityClass (" + entityClass + ") that is not a known planning entity.\n"
@@ -60,7 +59,7 @@ public abstract class AbstractFromConfigFactory<Solution_, Config_ extends Abstr
         return entityDescriptor;
     }
 
-    protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor) {
+    protected EntityDescriptor<Solution_> getTheOnlyEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor) {
         Collection<EntityDescriptor<Solution_>> entityDescriptors = solutionDescriptor.getGenuineEntityDescriptors();
         if (entityDescriptors.size() != 1) {
             throw new IllegalArgumentException("The config (" + config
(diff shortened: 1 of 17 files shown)


yurloc:
Please squash.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/constructionheuristic/placer/QueuedValuePlacerFactory.java
```diff
@@ -31,9 +31,7 @@ public QueuedValuePlacerFactory(QueuedValuePlacerConfig placerConfig) {
 
     @Override
     public QueuedValuePlacer<Solution_> buildEntityPlacer(HeuristicConfigPolicy<Solution_> configPolicy) {
-        EntityDescriptor<Solution_> entityDescriptor =
-                config.getEntityClass() == null ? deduceEntityDescriptor(configPolicy.getSolutionDescriptor())
-                        : deduceEntityDescriptor(configPolicy.getSolutionDescriptor(), config.getEntityClass());
+        EntityDescriptor<Solution_> entityDescriptor = deduceEntityDescriptor(configPolicy, config.getEntityClass());
```
triceo, line 34: I'm thinking out loud... Is it worth having the `deduceEntityDescriptor` method?
If we keep the null check here, we can call the two new methods directly.
Not sure if that is better - probably if the null check would proliferate, then keeping it inside the extra method is better.

triceo, line 34: I'm thinking out loud... Is it worth having the deduceEntityDescriptor method?
If we keep the null check here, we can call the two new methods directly.
Not sure if that is better - probably if the null check would proliferate, then keeping it inside the extra method is better.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/PillarSwapMoveSelectorFactory.java
```diff
@@ -26,17 +26,16 @@ protected MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<So
             SelectionCacheType minimumCacheType, boolean randomSelection) {
         PillarSelectorConfig leftPillarSelectorConfig =
                 Objects.requireNonNullElseGet(config.getPillarSelectorConfig(), PillarSelectorConfig::new);
-        PillarSelector<Solution_> leftPillarSelector =
-                buildPillarSelector(leftPillarSelectorConfig, configPolicy, minimumCacheType, randomSelection);
         PillarSelectorConfig rightPillarSelectorConfig =
                 Objects.requireNonNullElse(config.getSecondaryPillarSelectorConfig(), leftPillarSelectorConfig);
+        PillarSelector<Solution_> leftPillarSelector =
+                buildPillarSelector(leftPillarSelectorConfig, configPolicy, minimumCacheType, randomSelection);
```
triceo, line 32: Seems like a change for change's sake. I can guess that the idea was to improve the locality of `leftPillarSelector`, but that in turns worsens the locality of `leftPillarSelectorConfig`. 

Six of one, half a dozen of other; I'd revert the change.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
```diff
@@ -52,7 +59,7 @@ protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<
         return entityDescriptor;
     }
 
-    protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor) {
+    protected EntityDescriptor<Solution_> getTheOnlyEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor) {
```
triceo, line 62: Is this called often, perhaps on the hot path? Because it is expensive. Collection size, creation of a new iterator... If it is on the hot path, may I suggest caching the result?

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
```diff
@@ -77,8 +90,7 @@ protected GenuineVariableDescriptor<Solution_> deduceVariableDescriptor(
         return variableDescriptor;
     }
 
-    protected GenuineVariableDescriptor<Solution_> deduceVariableDescriptor(
-            EntityDescriptor<Solution_> entityDescriptor) {
+    protected GenuineVariableDescriptor<Solution_> getTheOnlyVariableDescriptor(EntityDescriptor<Solution_> entityDescriptor) {
```
triceo, line 93: Dtto.

triceo (approved):
Minor comments, no showstoppers.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
yurloc: No, `getTheOnlyEntityDescriptor()` is called only once or twice when the CH phase is being built. Flame Graph surface negligible, confirming it's not on the hot path.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
```diff
@@ -52,7 +59,7 @@ protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<
         return entityDescriptor;
     }
 
-    protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor) {
+    protected EntityDescriptor<Solution_> getTheOnlyEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor) {
         Collection<EntityDescriptor<Solution_>> entityDescriptors = solutionDescriptor.getGenuineEntityDescriptors();
```
yurloc, line 63: `getGenuineEntityDescriptors()` is the expensive part (I think you were referring to this in the comment above). IntelliJ Profiler's Method List doesn't show it at all (which means 0 samples) so I don't think it's on the hot path (in a basic scenario).

On the other hand, it's also being called from other places, on of them being `BestSolutionMutationCountStatistic`, which calls it whenever the best solution changes (if the statistic is enabled). I've filed https://issues.redhat.com/browse/PLANNER-2773 for that.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/AbstractFromConfigFactory.java
yurloc: This is being called once or twice when the CH phase is built. Unlike the case above, `entityDescriptor.getGenuineVariableDescriptorList()` is cheap.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/PillarSwapMoveSelectorFactory.java
yurloc: Without any other context, I'd agree. But this is not about the locality of the variables. It's about the procedure.

The intention is to first focus on the config resolution and then on building the selectors.

The reason for keeping config resolution together is that it makes it easier to notice and understand that the left pillar config falls back to empty config but the right pillar config _does not_. It falls back to the left pillar config. This is significant and it's important to notice the intention. I think I should add a comment about that.

Other selector factories follow the same structure.

PR #2129, opened 2022-08-12
CS-B: Do not index simple joins

triceo:
Joins that do not use Joiners do not need to be sent through the IndexProperties+NoneIndexer machinery.
Some examples improve by ~10 %.


triceo:
Will open a different PR shortly.

PR #2142, opened 2022-08-23
CS-B: Fix cheaptime score corruption

triceo:
Also introduces new test coverage to catch these problems earlier.
SmokeTest scores have been adjusted across the board so that all the tests run in under 5 seconds.
Finally, updated methodology overview SVG, which was largely outdated at this point.


core/optaplanner-constraint-streams-bavet/src/main/java/org/optaplanner/constraint/streams/bavet/common/index/ThreeIndexProperties.java
```diff
@@ -34,7 +34,7 @@ public <Type_> Type_ toKey(int index) {
     public <Type_> Type_ toKey(int from, int to) {
         switch (to - from) {
             case 1:
-                return toKey(to);
+                return toKey(from);
```
triceo, line 37: This is the fix.

optaplanner-examples/src/test/java/org/optaplanner/examples/common/app/SolverSmokeTest.java
```diff
@@ -0,0 +1,241 @@
...
+        SolverFactory<Solution_> solverFactory =
+                buildSolverFactory(constraintStreamImplType, bestScoreLimit, environmentMode, moveThreadCount);
+        Solution_ problem = solutionFileIO.read(unsolvedDataFile);
+        logger.info("Opened: {}", unsolvedDataFile);
+        Solver<Solution_> solver = solverFactory.buildSolver();
+        Solution_ bestSolution = solver.solve(problem);
+        assertScoreAndConstraintMatches(solverFactory, bestSolution, bestScoreLimit);
+    }
+
+    private SolverFactory<Solution_> buildSolverFactory(ConstraintStreamImplType constraintStreamImplType,
```
rsynek, line 137: (does not block merging) Building the `SolverFactory` is a relatively expensive operation, maybe we could cache it ` @BeforeAll` to avoid rebuilding it for every test data combination? This applies to `buildMutualCorrectnessSolverFactory` as well.

rsynek (approved):
One soft suggestion inline, plus let's make the tests run 5 seconds to save some time.

PR #2177, opened 2022-09-06
PLANNER-1709 Improved penalize/reward/impact overloads

triceo:


rsynek (commented):
The API changes look good, but I think we need better wording in Javadoc.

rsynek (approved):

PR #2147, opened 2022-08-25
Skip certain useless moves

triceo:
### Checklist
- [ ] Documentation updated if applicable.
- [ ] Upgrade recipe provided if applicable.

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

- for <b>pull request checks</b>  
  Please add comment: <b>Jenkins retest this</b>

- for a <b>specific pull request check</b>  
  please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] tests</b>
  
- for a <b>full downstream build</b> 
  - for <b>jenkins</b> job: 
    please add comment: <b>Jenkins run fdb</b>
  - for <b>github actions</b> job: 
    add the label `run_fdb`

- for a <b>compile downstream build</b>  
  please add comment: <b>Jenkins run cdb</b>

- for a <b>full production downstream build</b>  
  please add comment: <b>Jenkins execute product fdb</b>

- for an <b>upstream build</b>  
  please add comment: <b>Jenkins run upstream</b>

- for <b>quarkus branch checks</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins run quarkus-branch</b>

- for a <b>quarkus branch specific check</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins (re)run [optaplanner|optaplanner-quickstarts|optaweb-vehicle-routing] quarkus-branch</b>

- for <b>quarkus main checks</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins run quarkus-main</b>

- for a <b>specific quarkus main check</b>  
  Run checks against Quarkus main branch  
[...]

PR #2286, opened 2022-11-07
Prevent useless change moves

triceo:
Prevents moves where an entity's variable value would be changed to itself.

### DIFF
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/iterator/AbstractRandomChangeIterator.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/iterator/AbstractRandomChangeIterator.java
@@ -1,7 +1,9 @@
 package org.optaplanner.core.impl.heuristic.selector.common.iterator;
 
 import java.util.Iterator;
+import java.util.Objects;
 
+import org.optaplanner.core.impl.domain.variable.descriptor.VariableDescriptor;
 import org.optaplanner.core.impl.heuristic.move.Move;
 import org.optaplanner.core.impl.heuristic.selector.entity.EntitySelector;
 import org.optaplanner.core.impl.heuristic.selector.value.ValueSelector;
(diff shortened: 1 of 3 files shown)


rsynek (approved):

rsynek:
@triceo Unfortunately, the `OutOfMemoryError: Java heap space` coming from `org.apache.maven.plugins:maven-assembly-plugin:3.4.2:single (package-generated-docs) on project optaplanner-doc` didn't go away.

yurloc:
> @triceo Unfortunately, the `OutOfMemoryError: Java heap space` coming from `org.apache.maven.plugins:maven-assembly-plugin:3.4.2:single (package-generated-docs) on project optaplanner-doc` didn't go away.

@rsynek Where do you see an OOME? The CI is green.

yurloc:
@triceo What is the supposed benefit of this change?

triceo:
> @triceo What is the supposed benefit of this change?

I repeatedly see change moves being generated and executed where the change is from value X to value X (= noop). 

You could argue that these values should be filtered out by `isMoveDoable(...)` on the change move. Unfortunately, I can [prove that they are not](https://stackoverflow.com/questions/74325614/optaplanner-score-corruption-using-bavet), even though I do not exactly understand why.

And even if it worked, which it doesn't, doing it in `isMoveDoable(...)` is suboptimal - there is no point in generating a move to be ignored, if I can cheaply prevent it from being generated altogether.

yurloc:
> if I can cheaply prevent it from being generated altogether.

Do you have any indications that the prevention proposed in this PR is cheaper than filtering it out later? The `isMoveDoable()` code seems really simple at first sight.

yurloc:
> And even if it worked, which it doesn't

I really think it does.

> I can [prove that they are not](https://stackoverflow.com/questions/74325614/optaplanner-score-corruption-using-bavet) (just one case out of many) , even though I do not exactly understand why.

I'm _afraid_ you've been misled by the log message. I need some time to confirm this statement but based on my previous experience, I believe that the log message is _wrong_ because the move's `toString()` reads the variable's old value each time it's called. So, the message is not showing the variable's value as it was before the move is done, which is unexpected, confusing, and wrong, but it doesn't prove the `isMoveDoable()` doesn't work.

triceo:
Fair point. If you can prove that the message is wrong, and not the behavior, we do not have a problem to solve. Even then, I would argue that the change here is beneficial, for the reason I already explained.

yurloc:
I appreciate the intention behind this change but unfortunately, it's incompatible with the caching system. If move selector caching is enabled, move doability must be evaluated just in time (when the move is being done) because it depends on the state of the solution.

For example, if `PHASE` selection cache type is used, the move iterator is used once at the beginning of the phase, and all the moves it creates are stored in a cache. The iterator is never used again during the phase. However, moves that are not doable at the beginning of the phase might very well become doable after some steps. If the iterator didn't return such moves they wouldn't be added to the cache and the local search algorithm could never make those moves during that phase. Its ability to explore the solution space would be reduced.

Complementary to that, we can't remove the `isMoveDoable()` method because moves that are doable at the beginning of the phase become undoable each time they are done and remain undoable until some other move is done on the same entity.

I'm closing this now because I can't see any way how this could be merged without either breaking the behavior of cached move selectors or removing the caching system completely.

yurloc:
For the record, a similar attempt was made in #2147. The PR was closed for a different reason before I could elaborate on the caching problem.

I'm worried that this is the second time we almost changed the behavior of cached move selection without breaking a unit test. I think we should really have one to catch this in the future.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/iterator/AbstractRandomChangeIterator.java
```diff
@@ -37,24 +39,31 @@ protected Move_ createUpcomingSelection() {
...
-        // This loop is mostly only relevant when the entityIterator or valueIterator is non-random or shuffled
-        while (!valueIterator.hasNext()) {
-            // Try the next entity
-            if (!entityIterator.hasNext()) {
-                entityIterator = entitySelector.iterator();
-                entityIteratorCreationCount++;
-                if (entityIteratorCreationCount >= 2) {
-                    // All entity-value combinations have been tried (some even more than once)
-                    return noUpcomingSelection();
+        while (true) {
```
ge0ffrey, line 45: There could be no doable moves. If that happens, it's very bad.
Some situations of no doable moves can be detected, for example with ChangeMove or SwapMove selectors that see there is only 1 planning value. But for more complex moves, not all situations of no doable moves can be detected without false positives.

If there are no doable moves, this **loops forever**. Calling solver.terminateEarly() will NOT end it. Time spend up will NOT end it. Asking the JVM to shutdown, which will interrupte its thread, will NOT end it. It will need to be killed with fire (kill -9). That kills OptaPlanner's reputation.

The move evaluation loop does check termination (by terminateEarly, by interruption or by configured termination such as timeSpent), so users think optaplanner stops immediately if the time is up, which is important for them. In fact, optaplanner stops at the first next move (or next step or next phase) iteration, which is as good as immediately normally.

@triceo @yurloc I found it good to be paranoid about avoiding "forever loops".

PR #2330, opened 2022-11-24
PLANNER-2843 Convert nqueens to Jackson

triceo:


optaplanner-examples/src/test/java/org/optaplanner/examples/nqueens/score/NQueensConstraintProviderTest.java
```diff
@@ -115,8 +116,8 @@ private static NQueens readSolution(String resource) throws IOException {
     void givenSolutionMultipleConstraints(ConstraintVerifier<NQueensConstraintProvider, NQueens> constraintVerifier)
             throws IOException {
         constraintVerifier.verifyThat()
-                .givenSolution(readSolution("256queensScore-30.xml"))
-                .scores(SimpleScore.of(-30));
+                .givenSolution(readSolution("256queensScore-29.json"))
```
rsynek, line 119: Could you please explain how has the score changed?

triceo: It hasn't. I just didn't catch the solver in time for 30, and I stopped it at 29. This doesn't change anything - it's just a different point in the solver run.

triceo: There is no reason why the score should be 30. It is an arbitrary point in time. So I just picked a different one.

PR #2333, opened 2022-11-25
PLANNER-2840 Convert curriculumcourse to Jackson

triceo:
This is the final one. After that, cleanups and leftovers.


rsynek (approved):
LGTM

PR #2364, opened 2022-11-25
PLANNER-2836 Finish conversion of examples to Jackson

triceo:
All references to XStream are now gone from the examples.
The remaining XLSX examples have been converted to use the new `AbstractPersistable`.
All generators and importers have been re-run to ensure that they function properly.
This also required stabilizing some collections, so that the data sets are generated reproducibly.


optaplanner-examples/src/main/java/org/optaplanner/examples/flightcrewscheduling/persistence/FlightCrewSchedulingGenerator.java
```diff
@@ -190,16 +180,12 @@ private void createFlightList(FlightCrewSolution solution, int flightRoundTripsP
...
-                    flight.setId(flightId);
-                    flightId++;
-                    flight.setFlightNumber(flightNumber);
-                    flight.setDepartureAirport(departureAirport);
-                    flight.setDepartureUTCDateTime(date.atTime(departureMinute / 60, departureMinute % 60));
-                    flight.setArrivalAirport(arrivalAirport);
-                    flight.setArrivalUTCDateTime(date.atTime(arrivalMinute / 60, arrivalMinute % 60));
+                    Flight flight = new Flight(flightId, flightNumber, departureAirport,
+                            date.atTime(departureMinute / 60, departureMinute % 60), arrivalAirport,
+                            date.atTime(arrivalMinute / 60, arrivalMinute % 60));
```
rsynek, line 185: Local variables with the results of these calculations would improve readability.

rsynek (approved):
Could you please elaborate on the `.xlsx` files cleanup? For some examples, we are removing them, while adding new ones for others.

triceo:
What specifically am I removing or deleting?
I don't see anything in the list of files that would be green or red - only yellow for modifications.

rsynek:
> What specifically am I removing or deleting? I don't see anything in the list of files that would be green or red - only yellow for modifications.

Correct, these are only additions/removal from the same files. Please ignore my previous comment.

PR #2404, opened 2022-12-01
CS: Reduce the use of capturing lambdas

triceo:
Besides code that is easier to understand, it also performs ~3 % better in the score director benchmarks.


rsynek (approved):
One suggestion about reducing minor duplication.

PR #2487, opened 2023-01-06
Benchmark favor CP over DRL + replace use of non-existing testdataConstraints.drl

ge0ffrey:


optaplanner-benchmark/src/test/resources/org/optaplanner/benchmark/config/testBenchmarkConfigWithNamespace.xml
```diff
@@ -32,7 +32,7 @@
       <solutionClass>org.optaplanner.core.impl.testdata.domain.TestdataSolution</solutionClass>
       <entityClass>org.optaplanner.core.impl.testdata.domain.TestdataEntity</entityClass>
       <scoreDirectorFactory>
-        <scoreDrl>org/optaplanner/core/api/solver/testdataConstraints.drl</scoreDrl>
```
ge0ffrey, line 35: As far as I can tell, this file has been removed ages ago, but the tests somehow still used it.
Clearly, these tests should be using a ConstraintProvider too.

optaplanner-benchmark/src/test/resources/org/optaplanner/benchmark/api/classloaderTestdataBenchmarkConfig.xml
```diff
@@ -16,7 +16,9 @@
 
       <!-- Score configuration -->
       <scoreDirectorFactory>
-        <scoreDrl>divertThroughClassLoader/org/optaplanner/benchmark/api/classloaderTestdataConstraints.drl</scoreDrl>
+        <!-- Using these classnames doesn't work because the className still differs from class.getName()-->
+        <!--<constraintProviderClass>divertThroughClassLoader.org.optaplanner.core.impl.testdata.domain.TestdataConstraintProvider</constraintProviderClass>-->
+        <constraintProviderClass>org.optaplanner.core.impl.testdata.domain.TestdataConstraintProvider</constraintProviderClass>
```
ge0ffrey, line 21: One can argue wheter the PlannerBenchmarkFactoryTest shouldn't check the DRL redirect too, but I'd argue it's more important to at least test the ConstraintProvider.

As for the comment: same limitations as for the entityClass etc a few lines above it.

triceo:
OK to test

optaplanner-benchmark/src/test/resources/org/optaplanner/benchmark/api/classloaderTestdataBenchmarkConfig.xml
ge0ffrey: This test is probably too convoluted anyway. It tests if a classloader works correctly, by checking if a redirect in the custom class loader succeeds, without going directly to the correct resource in the parent class...

triceo (changes requested):
It appears as if there's whitespace issues in the XML files.

ge0ffrey:
> It appears as if there's whitespace issues in the XML files.

I don't understand. Where exactly?

Things I tried:
- Check for any tab chars (\t in all files that changed)
- Check the CI results, all green. I presume that also confirms that formatting is ok.
- Manually check the changes here on github (diff), seems ok as far as I can tell.

ge0ffrey:
Note: my first push had a mistake with commenting out, but I forced pushed a minute or so after that, with the fix.

triceo (approved):
Nevermind, it was a problem with my browser. LGTM.

PR #2517, opened 2023-01-23
PLANNER-2889 Allow to set Drools version in release (#2516)

radtriste:
https://issues.redhat.com/browse/PLANNER-2889

This will help to release OptaPlanner at the same time as Kogito and Drools in case we don't have the time to release one after the other.
All artifacts will so be staged to JBoss Nexus and then release altogether.

NOTE: This remains an optional parameter


rsynek (approved):

radtriste:
worked like a charm on 8.33.x :)

PR #2527, opened 2023-01-25
Reduce memory churn in Demand hashCode

triceo:
`Objects.hashCode(...)` is a varargs method, which means it creates an array of arguments every time it is called. 
This is typically not a problem, but in the case of `Demand` implementations, this actually shows up on the hot path.
This change significantly reduces the number of `Object[]` arrays getting allocated.

### DIFF
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/variable/supply/AbstractVariableDescriptorBasedDemand.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/variable/supply/AbstractVariableDescriptorBasedDemand.java
@@ -6,10 +6,8 @@ import org.optaplanner.core.impl.domain.variable.descriptor.VariableDescriptor;
 
 /**
  * Some {@link Demand} implementation classes are defined by their {@link VariableDescriptor} and nothing else.
- * However, they still must not have the same {@link #hashCode()} as other {@link Demand} implementation classes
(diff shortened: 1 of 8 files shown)


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/variable/supply/AbstractVariableDescriptorBasedDemand.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Objects;
+
+import org.optaplanner.core.impl.domain.variable.descriptor.VariableDescriptor;
+
+/**
+ * Some {@link Demand} implementation classes are defined by their {@link VariableDescriptor} and nothing else.
+ * However, they still must not have the same {@link #hashCode()} as other {@link Demand} implementation classes
+ * defined by the same {@link VariableDescriptor}.
+ * Otherwise these different {@link Demand} implementations would be considered identical in hash maps
+ * and the same supply would be used for them.
```
yurloc, line 12: This is slightly inaccurate and may be confusing for beginners or non-Java programmers. HashCode determines the bucket. If non-equal keys have the same hashCode the lookup will be less efficient but it's not "illegal". I would rephrase this to state that different Demand impls must not be equal.

yurloc (approved):
Minor Javadoc suggestion, otherwise LGTM.

PR #2554, opened 2023-02-02
commons-io is no longer necessary

triceo:
A previous change moved commons-io to test scope.
Unfortunately, this transitively broke Apache POI, which requires that dependency at runtime.
The real fix is to stop using commons-io on our end.


rsynek (approved):
LGTM

PR #2608, opened 2023-02-15
KOGITO-8086 quarkus-lts PR check as default

radtriste:
https://issues.redhat.com/browse/KOGITO-8086

- https://github.com/kiegroup/kogito-pipelines/pull/852
- https://github.com/kiegroup/drools/pull/5006
- https://github.com/kiegroup/optaplanner/pull/2608


rsynek (commented):
Could you please explain the motivation for this change? Is it to ensure that we never ever break the LTS (so far only optional)?

radtriste:
@rsynek not something you would want on 9.x-dsl-config I guess as LTS is still 2.13

radtriste:
> Could you please explain the motivation for this change? Is it to ensure that we never ever break the LTS (so far only optional)?

Exactly. On Kogito and Drools, we realized that it breaks quite easily, thus having the quarkus-lts change makes sure we are still compatible with RHBQ version (or at least we provide a patch/workaround to be compatible)

radtriste:
the `quarkus-lts` will run the repository with `productized` profile and against RHBQ 2.13

radtriste:
This is not mandatory. I just felt it would be great to have it as well on OP, like we have on Kogito and Drools.

rsynek (approved):

PR #2681, opened 2023-03-08
PLANNER-2906 Group key hashCodes must remain constant

triceo:
Changes the docs and Javadoc to say that.
Introduces assertions to fail fast early.

### DIFF
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/solver/EnvironmentMode.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/solver/EnvironmentMode.java
@@ -31,7 +31,7 @@ public enum EnvironmentMode {
      * <p>
      * This mode is horribly slow.
      */
-    FULL_ASSERT,
+    FULL_ASSERT(true),
     /**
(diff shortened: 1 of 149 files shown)


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/solver/EnvironmentMode.java
```diff
@@ -81,62 +81,34 @@ public enum EnvironmentMode {
...
-            case FAST_ASSERT:
-                return true;
-            case REPRODUCIBLE:
-            case NON_REPRODUCIBLE:
-                return false;
-            default:
-                throw new IllegalStateException("The environmentMode (" + this + ") is not implemented.");
-        }
+        return asserted;
     }
```
triceo, line 94: This is another potentially interesting piece.
Since `isAsserted()` is now called on the hot path in groupBy, I need to make sure it is as fast as possible.

rsynek (approved):
Looks good; it's a bit worrying how many files have to change to pass the `EnvironmentMode` there.

PR #2690, opened 2023-03-14
Clean up Gizmo accessors

triceo:
This PR cleans up the logic in Gizmo accessors,
removing instanceof checks during read and write operations.
This should make the code read better.

### DIFF
--- a/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
+++ b/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
@@ -1,7 +1,6 @@
 package org.optaplanner.core.impl.domain.common.accessor.gizmo;
 
 import java.lang.reflect.Field;
-import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.util.function.Consumer;
 
@@ -12,10 +11,16 @@ import io.quarkus.gizmo.ResultHandle;
 
 final class GizmoFieldHandler implements GizmoMemberHandler {
 
+    private final Class<?> declaringClass;
+    private final Field field;
     private final FieldDescriptor fieldDescriptor;
+    private final boolean canBeWritten;
 
-    GizmoFieldHandler(FieldDescriptor fieldDescriptor) {
+    GizmoFieldHandler(Class<?> declaringClass, Field field, FieldDescriptor fieldDescriptor, boolean canBeWritten) {
+        this.declaringClass = declaringClass;
+        this.field = field;
         this.fieldDescriptor = fieldDescriptor;
+        this.canBeWritten = canBeWritten;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            bytecodeCreator.writeInstanceField(fieldDescriptor, thisObj, newValue);
+            return true;
+        } catch (Exception ex) {
+            if (canBeWritten) {
+                throw new IllegalStateException("Cannot write field (" + field + ") of class (" + declaringClass + ").", ex);
+            } else {
+                return false;
+            }
+        }
```
triceo, line 53: This is kinda strange. In order to make all tests pass, I had to attempt to write in all cases, and then only throw the exception if the write should have succeeded.

Instead, I'd expect not to have to do the write if I know it should fail.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoMemberHandler.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
+import java.util.function.Consumer;
+
+import io.quarkus.gizmo.BytecodeCreator;
+import io.quarkus.gizmo.FieldDescriptor;
+import io.quarkus.gizmo.MethodDescriptor;
+import io.quarkus.gizmo.ResultHandle;
+
+interface GizmoMemberHandler {
```
triceo, line 13: A new interface that handles the differences between member and field access.

optaplanner-quarkus-integration/optaplanner-quarkus/deployment/src/test/java/org/optaplanner/quarkus/testdata/gizmo/PrivateNoArgsConstructorSolution.java
```diff
@@ -15,15 +15,17 @@ public class PrivateNoArgsConstructorSolution {
     List<PrivateNoArgsConstructorEntity> planningEntityList;
 
     @PlanningScore
-    public final SimpleScore score;
+    public SimpleScore score;
```
triceo, line 18: Score must not be `final`, ever. I am kinda surprised there is no fail-fast for this.

Christopher-Chianelli (commented):
The code reads better, but I think the `try...catch` can be removed if `canBeWritten` is forced to `true` in the `public GizmoMemberDescriptor(String name, Object memberDescriptor, Class<?> declaringClass)` constructor (which is only called from Quarkus, and tells the code to ignore any final checks since it will remove final modifiers)

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
triceo: Hmm. I explicitly check if the field is final. Are you saying that Quarkus makes the field non-final, while the field is still reporting itself to be final?

Christopher-Chianelli: Yes; this code is called BEFORE the transformation happens, hence it will report as final.

triceo: Ah, ok, so that explain the issue very well. Gap in my understanding, as usual.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/common/accessor/gizmo/GizmoMemberHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Creates handler for a {@link Field}.
+     *
+     * @param declaringClass never null, class that declares the {@link Field} in question
+     * @param name never null, name of the field
+     * @param memberDescriptor never null, descriptor of the {@link Field} in question
+     * @param ignoreFinalChecks true if Quarkus will make the field non-final for us
+     * @return never null
+     */
+    static GizmoMemberHandler of(Class<?> declaringClass, String name, FieldDescriptor memberDescriptor,
```
rsynek, line 25: Soft suggestion: keep the parameter name consistent with its type (`methodDescriptor` / `fieldDescriptor`).

rsynek (approved):

PR #2696, opened 2023-03-15
Enable nearby selection in VRP benchmark config

yurloc:
This is still not identical to the benchmark config before the list variable because we don't have a replacement for the TailChainSwapMoveSelector but it's getting close.

I think we can either:
- A) introduce ListTailSwapMoveSelector
- B) add nearby selection support to KOptListMoveSelector and replace TailChainSwapMoveSelector with a combination of sublist swap and k-opt moves (with nearby selection).


triceo (approved):
LGTM. I'd vote for B.

PR #2698, opened 2023-03-16
PLANNER-1506 Further improve performance of solution cloning

triceo:
Reduces some overhead and fixes a bug that would still clone enums and primitives.


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/solution/cloner/DeepCloningUtils.java
```diff
@@ -134,7 +134,7 @@ public boolean isFieldDeepCloned(Field field, Class<?> owningClass) {
 
     static boolean isImmutable(Class<?> clz) {
         if (clz.isPrimitive() || clz.isEnum()) {
-            return false;
+            return true;
```
triceo, line 137: This was the bug.

rsynek (approved):

PR #2727, opened 2023-03-27
Update documentation with nearby selection for list variable

yurloc:


optaplanner-docs/src/modules/ROOT/pages/move-and-neighborhood-selection/move-and-neighborhood-selection.adoc
```diff
@@ -126,6 +126,14 @@ The root of this tree is a `MoveSelector` which is injected into the optimizatio
 |Swap 2 sets of entities with the same values
 |`+[Process-A, Process-B, Process-C] {Computer-1} <-> [Process-E, Process-F] {Computer-2}+`
 
+|<<listChangeMoveSelector,List change move>>
+|Move a list element to a different index or another entity's list variable
+|`+Customer-3 {Vehicle-4[3] -> Vehicle-4[2]}+`
+
+|<<listSwapMoveSelector,List swap move>>
+|Swap 2 list variable elements
+|`+Customer-3 {Vehicle-3[2]} <-> Customer-10 {Vehicle-0[2]}+`
```
triceo, line 135: Does it makes sense to mention that this applies specifically to a planning list variable, and not just to any list?

optaplanner-docs/src/modules/ROOT/pages/release-notes/release-notes-8.adoc
```diff
@@ -1,6 +1,13 @@
 [[releaseNotes-8.x]]
 == OptaPlanner 8.x Release Notes
 
+[[releaseNotes-8.37.0.Final]]
+=== OptaPlanner 8.37.0.Final
+
+==== `PlanningListVariable` supports nearby selection
+
+xref:move-and-neighborhood-selection/move-and-neighborhood-selection.adoc#nearbySelection[Nearby selection] is now available for planning domains using a planning list variable.
```
triceo, line 9: I think that this already conflicts with `main`.

triceo (approved):
LGTM, minor comments.

optaplanner-docs/src/modules/ROOT/pages/move-and-neighborhood-selection/move-and-neighborhood-selection.adoc
yurloc: Thanks for this question. I realize that "list variable" is redundant here. It is more than obvious that moves work with planning variables. So I think it is safe to assume that "list" implies "planning list variable", and not just _any_ list.

While it is technically possible to have a `@PlaningVariable List<E>`, which is a "basic" planning variable, and not a planning list variable, it is impossible for any kind of list move to modify elements in such a list because an attempt to configure any `*List*MoveSelector` for a `@PlaningVariable` will fail fast (even if its type is a `List`).

```java
@PlanningEntity class MyEntity {
    @PlanningVariable(...) List<MyValue> valueList;
}
```

> java.lang.IllegalArgumentException: The listSwapMoveSelector (ListSwapMoveSelectorConfig(null)) cannot unfold because there are no planning list variables for the only entity (EntityDescriptor(com.example.MyEntity)) or no planning list variables at all.

Therefore, in this context, it is safe to simply use "list" instead of "list variable" or "planning list variable".

PR #1028, opened 2020-11-11
PLANNER-2250 Disable flaky tests

michaltomco:
Tests identified as flaky:

DefaultPartitionedSearchPhaseTest.shutdownMainThreadAbruptly
CloudBalancingDaemonTest.daemon

These tests have failed sporadically in the past and need to be reviewed.

PR #2752, opened 2023-04-08
PLANNER-2445 Ruin phase

MichalDudkiewicz:
### JIRA

https://issues.redhat.com/browse/PLANNER-2445

### Checklist
- [ ] Release notes updated if applicable.

<details>
<summary>
Additional info
</summary>
I decided to split the issue into 2 separate PRs. This one is about ruin phase and moves and the following one will be about the loop phase.
</details>


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/move/CompositeMove.java
```diff
@@ -102,6 +102,11 @@ public CompositeMove<Solution_> doMove(ScoreDirector<Solution_> scoreDirector) {
         return new CompositeMove<>(undoMoves);
     }
 
+    @Override
+    public void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
+        Arrays.stream(moves).filter(move -> move.isMoveDoable(scoreDirector)).forEach(move -> move.doMoveOnly(scoreDirector));
```
triceo, line 107: I recommend not using the streams here, rather switch to a normal for loop.
`doMoveOnly(...)` is on the solver hot path, and therefore we need the fastest possible way of iterating.
(Streams have a lot of overhead. Good for one-time work, bad on the hot path.)

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/RandomNearbyIterator.java
```diff
@@ -33,7 +35,7 @@ public RandomNearbyIterator(MemoizingSupply<NearbyDistanceMatrix<Object, Object>
 
     @Override
     public boolean hasNext() {
-        return replayingIterator.hasNext() && nearbySize > 0;
+        return (origin != null || replayingIterator.hasNext()) && nearbySize > 0;
```
triceo, line 38: Would you mind explaining why you needed to touch nearby selection?
I'd expect ruin moves to use nearby as is, without changes - did you perhaps fix a bug?
If so, please file the bug and submit the fix in a different PR, with its own tests.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/ruin/RuinPhase.java
```diff
@@ -0,0 +1,129 @@
...
+import org.optaplanner.core.impl.score.director.InnerScoreDirector;
+import org.optaplanner.core.impl.solver.scope.SolverScope;
+import org.optaplanner.core.impl.solver.termination.Termination;
+
+/**
+ * Default implementation of {@link RuinPhase}.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ */
+public class RuinPhase<Solution_> extends AbstractPhase<Solution_> implements CustomPhase<Solution_> {
```
triceo, line 19: I don't think the entire phase is necessary. Here's what I propose:

- The ruin config creates a custom phase.
- The custom phase gets the ruin command.

As a result, the logic of running a phase stays in custom phase, and ruination is only a command for it. That way, code duplication is reduced to a minimum.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/heuristic/selector/move/generic/RuinMoveSelectorConfig.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public void setPercentageToRuin(Integer percentageToRuin) {
+        this.percentageToRuin = percentageToRuin;
+    }
+
+    @Override
+    public SelectionOrder getSelectionOrder() {
+        SelectionOrder selectionOrder = super.getSelectionOrder();
+        return Objects.requireNonNullElse(selectionOrder, ORIGINAL);
+    }
```
triceo, line 77: Why is this here? Why not just keep the parent method?

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/ruin/RuinPhaseConfig.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public TerminationConfig getTerminationConfig() {
+        TerminationConfig terminationConfig = super.getTerminationConfig();
+        if (terminationConfig != null) {
+            return terminationConfig;
+        }
+        TerminationConfig defaultTerminationConfig = new TerminationConfig();
+        defaultTerminationConfig.setStepCountLimit(1);
+        return defaultTerminationConfig;
+    }
```
triceo, line 44: The ruin phase needs no termination. 
It does one move, and the move needs to finish.
I suggest overriding `setTermination(...)` to throw `UnsupportedOperationException`.
`getTerminationConfig()` should probably return `null` or whatever is convenient.

triceo (commented):
Thank you for the PR. Looking much better already, @MichalDudkiewicz !

I've done a quick review of the configs, which is where we'll need some changes first.
I'll get to reviewing the actual ruin impl, the tests and the examples early next week.

Also, please make sure you run `mvn clean install` before committing, as that fixes your code formatting and gives you CI results here on Github.

optaplanner-examples/src/main/resources/org/optaplanner/examples/vehiclerouting/vehicleRoutingSolverConfig.xml
```diff
@@ -80,4 +80,44 @@
       <acceptedCountLimit>1</acceptedCountLimit>
     </forager>
   </localSearch>
+
+  <ruin>
+    <ruinMoveSelector>
```
triceo, line 85: This nesting appears unnecessary. There will never be any selector here, other than a ruin selector, right?

I'd rather see the config without the `ruinMoveSelector`, going from `<ruin>` straight to `<percentage...>`, `<entitySelector>` etc.

optaplanner-examples/src/main/resources/org/optaplanner/examples/machinereassignment/machineReassignmentSolverConfig.xml
```diff
@@ -27,6 +24,33 @@
...
+      <entityTabuSize>7</entityTabuSize>
+      <!--<lateAcceptanceSize>2000</lateAcceptanceSize>-->
+    </acceptor>
+    <forager>
+      <acceptedCountLimit>2000</acceptedCountLimit>
+      <!--<acceptedCountLimit>500</acceptedCountLimit>-->
+    </forager>
+  </localSearch>
+
+  <ruin/>
```
triceo, line 44: Instead of putting the ruin configs in the default solver configs, I'd remove it from here and create a new general benchmark config [here](https://github.com/kiegroup/optaplanner/tree/main/optaplanner-examples/src/main/resources/org/optaplanner/examples/app/benchmark), allowing to compare:

- Multiple examples
- on multiple datasets
- with ruin+nearby enabled, simple ruin enabled, and ruin disabled.

Then you can run the benchmarker once, let it run for hours, and have a comprehensive report.
I'm assuming this report will help you in your thesis just as well.

I also have an additional reason for doing it like this - I don't want to expose the ruin phase by default. I want to give users the opportunity to play with it, before we make it "officially official". So we first introduce it in a less prominent space, and only put it in the primary solver configs a few releases later.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/common/nearby/RandomNearbyIterator.java
MichalDudkiewicz: _MimicReplayingEntitySelector_ doesn't allow us to call next more than once, because after first one _recordingAlreadyReturned_ is set to true. It looks intentional, but in the same time prevent us from using nearby selection for a bigger group of entities than 2. The documentation of both _nearbyEntityIterators_ however states that "origin here will be constant unless _next()_ on the original recording iterator is called first". Then there is really no need to call next on the _replayingEntitySelector_ as long as it doesn't change in the meantime. Therefore the workaround not to call it in this case, which also overcomes the limitation of _MimicReplayingEntitySelector_ . Does it look like a bug for you? Or maybe I am using the iterators in a wrong way?

MichalDudkiewicz:
> @MichalDudkiewicz please add the `package-info.java` (see any other package under `config`) to the `ruin` package to ensure that the new element belongs to the same namespace as the rest of the `SolverConfig`. Also, don't forget to add the `RunPhaseConfig` here: https://github.com/kiegroup/optaplanner/blob/main/core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/phase/PhaseConfig.java#L22 so that the polymorphism of `PhaseConfig` keeps working with (de)serialization to XML.

thanks, I missed that package-info change

core/optaplanner-core-impl/src/build/revapi-config.json
```diff
@@ -1045,6 +1045,28 @@
...
+          "annotationType": "javax.xml.bind.annotation.XmlSeeAlso",
+          "attribute": "value",
+          "oldValue": "{org.optaplanner.core.config.constructionheuristic.ConstructionHeuristicPhaseConfig.class, org.optaplanner.core.config.phase.custom.CustomPhaseConfig.class, org.optaplanner.core.config.exhaustivesearch.ExhaustiveSearchPhaseConfig.class, org.optaplanner.core.config.localsearch.LocalSearchPhaseConfig.class, org.optaplanner.core.config.phase.NoChangePhaseConfig.class, org.optaplanner.core.config.partitionedsearch.PartitionedSearchPhaseConfig.class}",
+          "newValue": "{org.optaplanner.core.config.constructionheuristic.ConstructionHeuristicPhaseConfig.class, org.optaplanner.core.config.phase.custom.CustomPhaseConfig.class, org.optaplanner.core.config.exhaustivesearch.ExhaustiveSearchPhaseConfig.class, org.optaplanner.core.config.localsearch.LocalSearchPhaseConfig.class, org.optaplanner.core.config.ruin.RuinPhaseConfig.class, org.optaplanner.core.config.phase.NoChangePhaseConfig.class, org.optaplanner.core.config.partitionedsearch.PartitionedSearchPhaseConfig.class}",
```
triceo, line 1057: AFAIK this can be removed.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/ruin/RuinPhaseConfig.java
```diff
@@ -0,0 +1,109 @@
...
+import org.optaplanner.core.config.solver.termination.TerminationConfig;
+import org.optaplanner.core.config.util.ConfigUtils;
+
+@XmlType(propOrder = {
+        "percentageToRuin",
+        "entitySelectorConfig",
+        "secondaryEntitySelectorConfig",
+        "variableNameIncludeList"
+})
+public class RuinPhaseConfig extends PhaseConfig<RuinPhaseConfig> {
```
triceo, line 23: Please add the `with...` methods in this class.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/heuristic/selector/move/generic/RuinMoveSelectorConfig.java
```diff
@@ -0,0 +1,82 @@
+package org.optaplanner.core.config.heuristic.selector.move.generic;
+
+import java.util.List;
+import java.util.function.Consumer;
+
+import org.optaplanner.core.config.heuristic.selector.entity.EntitySelectorConfig;
+import org.optaplanner.core.config.heuristic.selector.move.MoveSelectorConfig;
+import org.optaplanner.core.config.util.ConfigUtils;
+
+public class RuinMoveSelectorConfig extends MoveSelectorConfig<RuinMoveSelectorConfig> {
```
triceo, line 10: At first, I was going to ask why does this exist, if there already is a `RuinPhaseConfig` with the same properties. Then I realized this is because `MoveSelectorFactory` actually requires a `MoveSelectorConfig`, so the phase config needs to create one.

Although this is completely logical, it poses a problem. Everything in the `config` package is considered API, and it will be recommended to people in their IDE. So people will be using this class, not knowing or understanding that they should be using `RuinPhaseConfig` instead. Ideally, this class would be package-private, but then that would make it inaccessible to the `MoveSelectorFactory`.

I'm not exactly sure of the best way to solve it, but I know I don't want this type exposed publicly.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/move/CompositeMove.java
```diff
@@ -102,6 +102,15 @@ public CompositeMove<Solution_> doMove(ScoreDirector<Solution_> scoreDirector) {
         return new CompositeMove<>(undoMoves);
     }
 
+    @Override
+    public void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
```
triceo, line 106: Well spotted, this will improve the performance of the `CompositeMove`.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/CompositeRuinMoveSelector.java
```diff
@@ -0,0 +1,266 @@
...
+    private final HashMap<GenuineVariableDescriptor<Solution_>, SingletonInverseVariableSupply> variableDescriptorToInverseVariableSupply =
+            new HashMap<>();
+    private final boolean chained;
+
+    public CompositeRuinMoveSelector(EntitySelector<Solution_> leftEntitySelector,
+            EntitySelector<Solution_> rightEntitySelector,
+            List<GenuineVariableDescriptor<Solution_>> variableDescriptorList,
+            Collection<ShadowVariableDescriptor<Solution_>> shadowVariableDescriptors, Integer percentageToBeRuined) {
+        assert 0 < percentageToBeRuined && percentageToBeRuined <= 100
+                : "CompositeRuinMoveSelector: percentage to ruin the solution must be greater than 0 and smaller or equal to 100%";
```
triceo, line 40: We generally don't use these JDK assertions. They need to be specifically enabled to work, and nobody ever enables them in their JDK.

Please make this an `IllegalArgumentException` instead.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/CompositeRuinMoveSelector.java
```diff
@@ -0,0 +1,266 @@
+package org.optaplanner.core.impl.heuristic.selector.move.generic;
+
+import static org.optaplanner.core.impl.heuristic.move.CompositeMove.buildMove;
+
+import java.util.*;
```
triceo, line 5: Asterisk imports are disallowed by our code style. Please configure your IDE not to generate them and look throughout your PR for all instances of this anti-pattern.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/CompositeRuinMoveSelector.java
```diff
@@ -0,0 +1,266 @@
...
+            variableDescriptorToInverseVariableSupply.clear();
+        }
+    }
+
+    @Override
+    public Iterator<Move<Solution_>> iterator() {
+        return new RuinMoveIterator();
+    }
+
+    private class RuinMoveIterator extends UpcomingSelectionIterator<Move<Solution_>> {
```
triceo, line 100: This iterator is extremely complex. 150 lines of code. Please break it down into smaller logical units, such as methods, with descriptive names and comments. Also, I'd expect a lot of test coverage for such a complex piece of code.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/CompositeRuinMoveSelector.java
```diff
@@ -0,0 +1,266 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            Comparator<ListUnassignMove<Solution_>> listUnassignMoveComparator =
+                    Comparator.comparing(ListUnassignMove::getSourceIndex);
+            listUnassignMoves.sort(listUnassignMoveComparator.reversed());
+            moves.addAll(listUnassignMoves);
```
triceo, line 254: Please add a comment describing why it's necessary to reverse the order of unassign moves. It's a ruin move - why do we care in which order is the solution ruined?

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/ruin/RuinPhaseConfig.java
```diff
@@ -0,0 +1,109 @@
...
+    private EntitySelectorConfig entitySelectorConfig = null;
+    @XmlElement(name = "secondaryEntitySelector")
+    private EntitySelectorConfig secondaryEntitySelectorConfig = null;
+
+    @XmlElementWrapper(name = "variableNameIncludes")
+    @XmlElement(name = "variableNameInclude")
+    private List<String> variableNameIncludeList = null;
+
+    @XmlElement(name = "percentageToRuin")
+    private Integer percentageToRuin = 20;
```
triceo, line 40: Usually when a magic number appears, it's good to add a comment explaining how you arrived at the number. Providing some context is always a good thing, for the people in the future who have to understand your intentions.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/RuinMoveSelectorFactory.java
```diff
@@ -0,0 +1,50 @@
...
+        super(moveSelectorConfig);
+    }
+
+    @Override
+    protected MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<Solution_> configPolicy,
+            SelectionCacheType minimumCacheType, boolean randomSelection) {
+        EntitySelectorConfig entitySelectorConfig =
+                Objects.requireNonNullElseGet(config.getEntitySelectorConfig(), EntitySelectorConfig::new);
+        EntitySelectorConfig secondaryEntitySelectorConfig =
+                Objects.requireNonNullElseGet(config.getSecondaryEntitySelectorConfig(), EntitySelectorConfig::new);
```
triceo, line 31: I'm wondering... the two selectors exist for nearby only.
If the user decides not to use nearby, why do we have this complicated code handling two selectors? Why not introduce a much simpler variant with just the one selector, simply picking the entities to ruin?

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/phase/PhaseFactory.java
```diff
@@ -35,6 +37,9 @@ static <Solution_> PhaseFactory<Solution_> create(PhaseConfig<?> phaseConfig) {
             return new DefaultExhaustiveSearchPhaseFactory<>((ExhaustiveSearchPhaseConfig) phaseConfig);
         } else if (NoChangePhaseConfig.class.isAssignableFrom(phaseConfig.getClass())) {
             return new NoChangePhaseFactory<>((NoChangePhaseConfig) phaseConfig);
+        }
+        if (RuinPhaseConfig.class.isAssignableFrom(phaseConfig.getClass())) {
```
triceo, line 41: `else if`?

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/ruin/RuinPhaseCommand.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void changeWorkingSolution(ScoreDirector<Solution_> scoreDirector) {
+        Move<Solution_> nextRuinMove = moveSelector.iterator().next();
+        if (nextRuinMove != null) {
+            nextRuinMove.doMoveOnly(scoreDirector);
+            scoreDirector.triggerVariableListeners();
+        } else {
+            NoChangeMove<Solution_> noChangeMove = new NoChangeMove<>();
+            noChangeMove.doMoveOnly(scoreDirector);
```
triceo, line 32: Should this even be possible? A ruin phase that does nothing?
Shouldn't we rather throw an exception here?

core/optaplanner-core-impl/src/test/java/org/optaplanner/core/impl/heuristic/move/CompositeMoveTest.java
```diff
@@ -80,6 +80,23 @@ void doMove() {
         verify(c, times(1)).doMove(scoreDirector);
     }
 
+    @Test
+    void doMoveOnly() {
```
triceo, line 84: I appreciate this new test - however, it is identical to the `doMove` test.
In that case, the other test should make sure that some undo move is actually returned.

optaplanner-examples/src/main/java/org/optaplanner/examples/nqueens/domain/solver/nearby/QueenNearbyDistance.java
```diff
@@ -0,0 +1,16 @@
+package org.optaplanner.examples.nqueens.domain.solver.nearby;
```
triceo, line 1: Personally, I wouldn't touch nqueens at all.
The example is so trivial that introducing ruin doesn't help it much, but it does introduce extra complexity, which makes the "teaching value" of the example smaller.

optaplanner-examples/src/main/resources/org/optaplanner/examples/machinereassignment/machineReassignmentSolverConfig.xml
```diff
@@ -16,9 +16,6 @@
       <!--<incrementalScoreCalculatorClass>org.optaplanner.examples.machinereassignment.optional.score.MachineReassignmentIncrementalScoreCalculator</incrementalScoreCalculatorClass>-->
     <!--</assertionScoreDirectorFactory>-->
   </scoreDirectorFactory>
-  <termination>
-    <minutesSpentLimit>5</minutesSpentLimit>
-  </termination>
```
triceo, line 21: As agreed, please do not change the default solver configs.
Ruin should be introduced in a separate benchmark config first.

optaplanner-examples/src/main/resources/org/optaplanner/examples/tsp/tspSolverConfig.xml
```diff
@@ -16,13 +16,13 @@
...
   </scoreDirectorFactory>
 
-  <termination>
-    <minutesSpentLimit>2</minutesSpentLimit>
-  </termination>
   <constructionHeuristic>
     <constructionHeuristicType>FIRST_FIT_DECREASING</constructionHeuristicType>
   </constructionHeuristic>
   <localSearch>
+    <termination>
```
triceo, line 23: Dtto.

triceo (changes requested):
I finished my first round of review. Good work so far!
Once the comments are resolved and we see the benchmark results, I will do a final round of review - this time mostly formal.

triceo:
Also:

I'd prefer

triceo:
My apologies, I mistakenly closed this PR.

rsynek:
Jenkins run optaplanner tests

triceo:
Jenkins run optaplanner-quickstarts

kie-ci3:
**(tests) - optaplanner job** `#1028` was: **UNSTABLE**
Possible explanation: This should be test failures

Please look here: https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/optaplanner/job/main/job/pullrequest/job/optaplanner.tests.optaplanner/1028/display/redirect

**Test results:**
- PASSED: 4995
- FAILED: 5

Those are the test failures: 
<details>
<summary><a href="https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/optaplanner/job/main/job/pullrequest/job/optaplanner.tests.optaplanner/1028/testReport/org.optaplanner.examples.tsp.app/TspSmokeTest/Build_projects___runSpeedTest/">Build projects / org.optaplanner.examples.tsp.app.TspSmokeTest.runSpeedTest</a></summary>
[The bestScore (-226928940) must be at least the bestScoreLimit (-217364246).] <br/>Expecting actual:<br/>  -226928940<br/>to be greater than or equal to:<br/>  -217364246<br/>
</details>
<details>
<summary><a href="https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/optaplanner/job/main/job/pullrequest/job/optaplanner.tests.optaplanner/1028/testReport/org.optaplanner.examples.vehiclerouting.app/VehicleRoutingSmokeTest/Build_projects___runSpeedTest/">Build projects / org.optaplanner.examples.vehiclerouting.app.VehicleRoutingSmokeTest.runSpeedTest</a></summary>
[The bestScore (0hard/-866917soft) must be at least the bestScoreLimit (0hard/-788369soft).] <br/>Expecting actual:<br/>  0hard/-866917soft<br/>to be greater than or equal to:<br/>  0hard/-788369soft<br/>
</details>
<details>
<summary><a href="https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/optaplanner/job/main/job/pullrequest/job/optaplanner.tests.optaplanner/1028/testReport/org.optaplanner.examples.vehiclerouting.app/VehicleRoutingSmokeTest/Build_projects___runSpeedTest/">Build projects / org.optaplanner.examples.vehiclerouting.app.VehicleRoutingSmokeTest.runSpeedTest</a></summary>
[The bestScore (0hard/-830666soft) must be at least the bestScoreLimit (0hard/-788369soft).] <br/>Expecting actual:<br/>  0hard/-830666soft<br/>to be greater than or equal to:<br/>  0hard/-788369soft<br/>
</details>
<details>
<summary><a href="https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/optaplanner/job/main/job/pullrequest/job/optaplanner.tests.optaplanner/1028/testReport/org.optaplanner.examples.vehiclerouting.app/VehicleRoutingSmokeTest/Build_projects___runSpeedTest/">Build projects / org.optaplanner.examples.vehiclerouting.app.VehicleRoutingSmokeTest.runSpeedTest</a></summary>
[The bestScore (0hard/-2215387soft) must be at least the bestScoreLimit (0hard/-2113238soft).] <br/>Expecting actual:<br/>  0hard/-2215387soft<br/>to be greater than or equal to:<br/>  0hard/-2113238soft<br/>
</details>
<details>
<summary><a href="https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/optaplanner/job/main/job/pullrequest/job/optaplanner.tests.optaplanner/1028/testReport/org.optaplanner.examples.vehiclerouting.app/VehicleRoutingSmokeTest/Build_projects___runSpeedTest/">Build projects / org.optaplanner.examples.vehiclerouting.app.VehicleRoutingSmokeTest.runSpeedTest</a></summary>
[The bestScore (0hard/-2211056soft) must be at least the bestScoreLimit (0hard/-2156060soft).] <br/>Expecting actual:<br/>  0hard/-2211056soft<br/>to be greater than or equal to:<br/>  0hard/-2156060soft<br/>
</details>

triceo:
@MichalDudkiewicz Now that the PR can actually run CI, you will have noticed two things:

- Test failures, please fix the tests. They will be fixed by simply removing the ruin phase config.
- Sonar analysis, pointing out potential problems in your code. You don't need to fix all of them, as many are not issues. But go over the report, think about what you see there, and see which of them you think you should fix.

optaplanner-examples/src/main/resources/org/optaplanner/examples/machinereassignment/machineReassignmentSolverConfig.xml
MichalDudkiewicz: the benchmark config is there. I will run it as soon as we finish with the refactoring

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/ruin/RuinPhaseCommand.java
MichalDudkiewicz: when you run ruin phase 5 times in a row with lets say 20% of ruin size and then you try to apply ruin once more then the iterator will search for entities that are not yet ruined. Since 5*20%~100% is already ruined then it will find no more entities left to be ruined. In this case the iterator returns null and no action is taken.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/ruin/RuinPhaseConfig.java
MichalDudkiewicz: For now I left both a comment and todo. I will come back to it when I have the benchmarks

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/ruin/RuinPhaseCommand.java
MichalDudkiewicz: I left a comment as well

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/config/heuristic/selector/move/generic/RuinMoveSelectorConfig.java
MichalDudkiewicz: [4e2f380](https://github.com/kiegroup/optaplanner/pull/2752/commits/4e2f38037b8ad3720a0295ca0921ae720d04d5cd), let me know what you think

MichalDudkiewicz: Hence we loose this AbstractMoveSelectorFactory.buildMoveSelector logic (caching, shuffling, sorting, filtering...), but I wonder whether we even need it for ruin move

MichalDudkiewicz:
Please don't review the last commit yet. I will amend it later. I needed to push it, cause I want to pull it on another machine for benchmarking

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/RuinMoveSelectorFactory.java
MichalDudkiewicz: I separated these two, but I will refactor and better test them in an upcoming commit

MichalDudkiewicz:
**Benchmark results:**

200computers-600processesBestScoreStatisticLevel1
800lectures-32periods-50roomsBestScoreStatisticLevel1
83-a3f7-866dadb87e11.png)
comp07BestScoreStatisticLevel1
6-152538266dd
800lectures-32periods-50roomsBestScoreStatisticLevel1
d.png)
exam_comp_set2BestScoreStatisticLevel1
-95d62f6
exam_comp_set7BestScoreStatisticLevel1
0-3b13-4f65-8
medium01BestScoreStatisticLevel1
047-43e0-ac53-1712ccceafd9.png)
3a9-8a36107e0ca3.png)
long01BestScoreStatisticLevel1
0-03255ef7-6ab
model_b_10BestScoreStatisticLevel1
4-4f1e-b093-d7766ec2bdfd.png)

**Some initial considerations:**
1. Almost all nearby random ruins benchmark failed - I need to figure out why.
2. The way we calculate percentage to ruin must be changed to take into account nullable entities (sometime we ruin too much)
3. Some kind of caching the best global score would be useful (so that we don't lose it after solving is done)
4. Quite obvious, but ruining is not good for small size problems

triceo:
Good to see some data!

> 1. Almost all nearby random ruins benchmark failed - I need to figure out why.

+1. Once you've figured it out, please add test coverage for the problem.

> 2. The way we calculate percentage to ruin must be changed to take into account nullable entities (sometime we ruin too much)

+1.

> 3. Some kind of caching the best global score would be useful (so that we don't lose it after solving is done)

The solver/score director has API to give you the best solution found, that's the best solution overall. It includes the score. Isn't that enough?

> 4. Quite obvious, but ruining is not good for small size problems

Yes, that makes sense. In fact, it is part of the reason why I don't want to put it in all the configs, but rather leave it to the users to enable it if they want.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/CompositeRuinMoveSelector.java
MichalDudkiewicz: done, but requires some more testing

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -602,6 +602,15 @@ public boolean hasNoNullVariables(Object entity) {
...
     }
 
+    public boolean hasAnyNoNullGenuineAndShadowVariables(Object entity) {
+        for (VariableDescriptor<Solution_> variableDescriptor : effectiveVariableDescriptorMap.values()) {
+            if (variableDescriptor.getValue(entity) != null) {
+                return true;
+            }
+        }
+        return false;
+    }
```
triceo, line 612: `EntityDescriptor` already has `hasNoNullVariables(...)`.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -602,6 +602,15 @@ public boolean hasNoNullVariables(Object entity) {
         return true;
     }
 
+    public boolean hasAnyNoNullGenuineAndShadowVariables(Object entity) {
```
triceo, line 605: Consistently with `hasNoNullVariables()`, let's call this `hasNullVariables()`.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/NearbyCompositeRuinMoveSelector.java
```diff
@@ -1,16 +1,18 @@
...
 
 import org.optaplanner.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
 import org.optaplanner.core.impl.domain.variable.descriptor.ShadowVariableDescriptor;
 import org.optaplanner.core.impl.heuristic.move.Move;
 import org.optaplanner.core.impl.heuristic.selector.entity.EntitySelector;
+import org.optaplanner.core.impl.heuristic.selector.entity.decorator.FilteringEntitySelector;
 
 public class NearbyCompositeRuinMoveSelector<Solution_> extends SimpleCompositeRuinMoveSelector<Solution_> {
-    private final EntitySelector<Solution_> rightEntitySelector;
+    private EntitySelector<Solution_> rightEntitySelector;
```
triceo, line 15: I'm confused - why do you need to overwrite the selector?
If you only need the filtering selector locally inside `iterator`, create a local variable there.
If you need the filtering selector globally, why not store it in this field?

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/SimpleCompositeRuinMoveSelector.java
```diff
@@ -12,18 +13,21 @@
 import org.optaplanner.core.impl.domain.variable.supply.SupplyManager;
 import org.optaplanner.core.impl.heuristic.move.Move;
 import org.optaplanner.core.impl.heuristic.selector.entity.EntitySelector;
+import org.optaplanner.core.impl.heuristic.selector.entity.decorator.FilteringEntitySelector;
 import org.optaplanner.core.impl.solver.scope.SolverScope;
 
 public class SimpleCompositeRuinMoveSelector<Solution_> extends GenericMoveSelector<Solution_> {
 
-    protected final EntitySelector<Solution_> originEntitySelector;
+    protected EntitySelector<Solution_> originEntitySelector;
```
triceo, line 21: Dtto.

triceo (commented):
Some comments on the latest commit.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/domain/entity/descriptor/EntityDescriptor.java
MichalDudkiewicz: It is not only about null variables or about genuine variables at all. It needs to take into account also empty list variables and null shadow variables as well. We don't won't to count them in. At first glance word "initialized" would sound good to me, but then I realized list variables are always considered initialized, so are the nullable entities.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/NearbyCompositeRuinMoveSelector.java
MichalDudkiewicz: The filtering selector seems to be a kind of wrapper over the basic selector. I cannot create it locally without having the base selector, that was set within the constructor.

Speaking of the second option: it's too early to do it in the constructor. The actual creation of the ruin move selector by the factory is before any solving is started. It is too early to figure out which entities are initialized and which not. We can do this only after we start iterating over the move selector (e.g. after some local search phase, before ruin). This is why it is done in the iterator()

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/SimpleCompositeRuinMoveSelector.java
MichalDudkiewicz: answered above

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/ruin/RuinCustomPhase.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public void solvingEnded(SolverScope<Solution_> solverScope) {
+        if (bestScoreBeforeRuin.compareTo((Score_) solverScope.getBestScore()) > 0) {
+            solverScope.setBestSolution(bestSolutionBeforeRuin);
+            solverScope.setBestScore(bestScoreBeforeRuin);
+            solverScope.setBestSolutionTimeMillis(System.currentTimeMillis());
+            solverScope.getScoreDirector().setWorkingSolution(bestSolutionBeforeRuin);
+            solver.getBestSolutionRecaller().updateBestSolutionAndFire(solverScope);
+        }
```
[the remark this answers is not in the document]
triceo: Ok, I understand the use case now, thank you for that. 
Still, it seems that we are solving a symptom, and not the root cause.

> If we allow destructive moves that worsen solution and set that as current best solution

Why would we allow to set something as best, if it's not in fact better than the best we already have? That destroys the concept of the best solution - which is always the best solution that the solver reached so far, regardless of what happened in the meantime.

If there is code that sets the best solution to a *worse* solution than we had before, I'd argue that this is the problem we have to fix. New best solution must only ever be better than the old best solution.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/heuristic/selector/move/generic/NearbyCompositeRuinMoveSelector.java
MichalDudkiewicz: edit: I was wrong. For some reason I thought the filtering is done in FilteringEntitySelector constructor, which is not true. Applied

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/ruin/RuinCustomPhase.java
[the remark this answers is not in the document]
MichalDudkiewicz: > If there is code that sets the best solution to a _worse_ solution than we had before, I'd argue that this is the problem we have to fix. 

Actually this is exactly what DefaultCustomPhase do in stepEnded. It updates the best solution only when score is not improved. Is this a bug?

But even if we invert this condition and also disable assigning working solution from best solution in AbstractSolver runPhases for ruin phase only we still cannot "save" the best solution. It will be overwritten by following construction heuristic phase. It by design updates the best solution with working solution in BestSolutionRecaller::processWorkingSolutionDuringConstructionHeuristicsStep. It looks like construction heuristic here is a real problem. We know the order of phases in the AbstractSolver and we could possibly store this information, whether ruin was used before particular construction heuristic and update the best solution accordingly. However I am not sure whether it will not cause some problems in the following local search phase. I quickly commented out this updating best solution part and the local phase claims the solution is uninitialized.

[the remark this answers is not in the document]
Christopher-Chianelli: The commit that caused this is 258b1c5a0836d7fb684b0ac904860f484015ea7f (which from the commit log, is related to PLANNER-2008).
In particular, the `forceUpdateBestSolution` option was removed, and the condition changed from `if (forceUpdateBestSolution && !bestScoreImproved) {` to `!bestScoreImproved`. Is `forceUpdateBestSolution` changed to always be true or false @triceo ?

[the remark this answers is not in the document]
triceo: The current CH is built upon the assumption that it is *the only* CH. Therefore it makes sense for this single CH to set the best solution - after all, it is the first thing that runs, therefore the score *has* to be better.

Obviously, this assumption is no longer valid with the ruin phase. CH can now be run at any time, and therefore it can not assume that it will always provide the best score. I think the CH code needs to be updated, to deal with this new situation.

I don't think LS needs to be updated; LS was always capable of running after another LS, for example. So I think the problem is exclusively in the CH.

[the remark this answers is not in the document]
triceo: @Christopher-Chianelli I don't really remember this commit anymore; after all, it's been 3 years already. I'll take a look next week, rebuild my understanding of what goes on in there.

core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/solver/AbstractSolver.java
```diff
@@ -81,9 +82,10 @@ protected void runPhases(SolverScope<Solution_> solverScope) {
         while (!solverTermination.isSolverTerminated(solverScope) && it.hasNext()) {
             Phase<Solution_> phase = it.next();
             phase.solve(solverScope);
-            // If there is a next phase, it starts from the best solution, which might differ from the working solution.
+            // If current phase was not ruin and there is a next phase, it starts from the best solution,
+            // which might differ from the working solution.
             // If there isn't, no need to planning clone the best solution to the working solution.
-            if (it.hasNext()) {
+            if (it.hasNext() && !(phase instanceof RuinCustomPhase)) {
```
MichalDudkiewicz, line 88: this condition is still not enough. It may happen that CH and any other phase should also not set working solution from best one

triceo: Consider two local search configurations, one after another (LS1 and LS2). When LS1 exits, it is very likely that working solution != best solution. Yet LS2 needs to start from the best solution, as starting from the working solution would be a waste. This is in contrast to CH after ruin, which needs to start from the ruined solution, and that - by definition - is not the best solution. 

Practically, it should be the decision of the phase what solution it starts from. I think that the best solution to the problem is to introduce a new method on the phase, something like `pickStartingSolution(...)`. Each phase impl will override this method to implement its own logic for picking the starting solution.

Most phases will start from the best solution, but some may choose to start from the working solution, and that could even be the default impl of the pick method. But if a CH follows a ruin phase, it has to make a choice of not starting from the best solution, but rather from the working solution, overriding the pick method with its own. 

That said, it will be difficult to figure out if a CH follows a ruin, so maybe the switch should instead be "if working solution not initialized, then use working solution, else use best solution".

MichalDudkiewicz: (edited before I saw the answer)

This is not only about CH. This is about every phase. LS should also start from working solution in this case. Lets say CH after ruin ends with a worse score than best. Then we should still start LS from working solution not the best one

triceo: Good point, didn't think of that. This is a relatively tough nut to crack.

First, we have the backward compatibility consideration - we can not change what LS starts with, we would break every current user. (Are we sure it currently starts from the best solution?) So there would have to be a check that the ruin phase was included, in which case starting from the working solution would be permitted.

Second, what if ruin follows a ruin (ruin-CH-LS-ruin)? Does it start from the best, or the working? Arguably, it starts from the working. So, only the *first* ruin starts from the best solution? 

This is becoming very complicated. We need to figure out a system that *makes sense* and then make it work. As opposed to figuring out what works, and ignoring if it makes sense. :-)

triceo: It's starting to sound like ruin should be *a move*. Which means it would have to *include*  its own CH. which opens another can of worms. (But eliminates the need for loop phase.)

triceo: > This is becoming very complicated. We need to figure out a system that _makes sense_ and then make it work. As opposed to figuring out what works, and ignoring if it makes sense. :-)
> It's starting to sound like ruin should be a move inside LS, not its own phase. Which means it would have to include its own CH, which opens another can of worms. (But eliminates the need for loop phase.)

@Christopher-Chianelli , WDYT? Any opinions?

MichalDudkiewicz: I can also think of 3 other approaches:
1. Starting from after ruin phase start using working solution for every phase instead of the best one - relatively easy to do and should do the work for simple cases. On the other hand when the solver config is more complicated and has couple of LS in a row it will make the LS start from worse solution and probably result in a worse final score. 
2. Revert the caching of preruin solution at phase start by ruin phase and comparing at solver end. Also another revert required to make the CH update best solutions again - also easy and has no downsides of the first approach, but it not elegant way I guess
3. Introducing something like bestGlobalScore and solution (something which indeed is the best score) in solver scope? across all the phases used - ? no idea how much work needed, but doesn't seem to be big

triceo: I'll have to think about this a bit more. For now, I'm thinking something similar to 3, but it needs more time to brew.

Christopher-Chianelli: > > This is becoming very complicated. We need to figure out a system that _makes sense_ and then make it work. As opposed to figuring out what works, and ignoring if it makes sense. :-)
> > It's starting to sound like ruin should be a move inside LS, not its own phase. Which means it would have to include its own CH, which opens another can of worms. (But eliminates the need for loop phase.)
> 
> @Christopher-Chianelli , WDYT? Any opinions?

I was honestly thinking about something similar yesterday, a "paranoid" move that tries to move to a radically different solution by composing many different move. For such a "paranoid" move, it would be helpful if the move itself can evaluate the score in the middle of the move (so it can decide if it need to compose an additional move or not). Alternatively, it might be nice to have access to the score calculator inside a move selector (which allows the selector to see how "good" a particular move is and modify it selection accordingly).

MichalDudkiewicz: Thank you for the suggestion about the "paranoid" move. I may be a good alternative indeed. It sounds to me more like a "Levy flight" than "Ruin&Recreate" principle, which may also do the job.

However I wouldn't like to do this mostly because I am doing it in the scope of my master thesis, which already has "R&R" in its topic, so it is not that flexible and also because my time is limited. We don't know what obstacles we encounter with this new move approach, while "Ruin phase" is pretty much well known at this point and mostly implemented. I still believe we can overcome these issues with assigning best solution to working solution at every phase start with one of the approaches we discussed above.

My suggestion would be to finish what we started with the ruin phase and maybe create another ticket for this paranoid move inside LS. Then at the end of the day when we have both we can compare them and see which approach is better.

MichalDudkiewicz:
I also discovered why the benchmarker config I did is failing for some examples. This is not related with ruin phase. It happens also on the main branch when we copy-paste the local search config from either vehicle routing example or tsp example, so that there are 2 LS one after another (with some termination) and run either of the examples:

For the TSP example:
```
Cannot invoke "org.optaplanner.core.api.domain.valuerange.CountableValueRange.getSize()" because "this.cachedValueRange" is null
```

For VRP example:

```
java.lang.NullPointerException: Cannot invoke "java.util.List.size()" because "this.cachedEntityList" is null
	at org.optaplanner.core.impl.heuristic.selector.entity.FromSolutionEntitySelector.getSize(FromSolutionEntitySelector.java:106)
```

Any ideas why we cannot have these 2 LS config one after another for these examples? Because of that I am not able to simulate looping (also real loop phase would not work I guess).

triceo:
It is not possible for me to easily find the code you mention in a 22k line file.
That said, if you already read my e-mail, you know we're not going to support the looping anyway - so if you remove this from the solver config, I'm guessing the issue will go away too.

(This is why I didn't want to get into the loop phase just yet - configs with so many phases are stretching the solver to a place where it's likely never been. That way lie bugs, most likely.)

triceo:
Also, now that we know we're not doing the loop phase, I suggest we slightly update the benchmark.
If we want to get any information about large data sets, they need to run unimproved for longer, minutes. Small datasets, we can probably keep within the minute.
After the ruin, there should be enough unimproved time as well. Off the cuff, let's use half of what was used in that case before ruin.

MichalDudkiewicz:
> at least we'll have a file where the problem will be easier to point out

Here are the configs from optaplannes-examples (TSP and VRP) that also can reproduce the crash on main branch:
[LS after LS config.zip](https://github.com/kiegroup/optaplanner/files/11249789/LS.after.LS.config.zip)

This is a blocker for my benchmarks

triceo:
@MichalDudkiewicz Fix here: https://github.com/kiegroup/optaplanner/pull/2775

Will be merged later today.

triceo:
@MichalDudkiewicz FYI https://github.com/kiegroup/optaplanner/pull/2785

optaplanner-examples/src/main/resources/org/optaplanner/examples/app/benchmark/ruinWithConstantTimeOptaPlannerBenchmarkConfigTemplate.xml.ftl
```diff
@@ -5,9 +5,9 @@
 	<inheritedSolverBenchmark>
 		<solver>
 			<termination>
-				<minutesSpentLimit>15</minutesSpentLimit>
+				<minutesSpentLimit>10</minutesSpentLimit>
 			</termination>
-			<subSingleCount>10</subSingleCount>
+			<subSingleCount>5</subSingleCount>
```
triceo: I respect that you want to be thorough, but considering how many benchmarks you need to run, are you sure you want to multiply that number by 5?

triceo: Don't get me wrong - 5 is good, for results that are more statistically relevant. :-) But unless you have a very powerful machine, you will be waiting for a long time.

MichalDudkiewicz:
declined since it is not going to be merged. The work proves ruin move cannot be implemented as a custom phase, but rather a separate move within LS. 

However benchmarks results are going to be presented in my master thesis and linked also in here or [Testimonials and Case Studies](https://www.optaplanner.org/learn/testimonialsAndCaseStudies.html)

MichalDudkiewicz:
benchmark results:
https://drive.google.com/file/d/1B7ceFuHdopYxubEPU9XJVjGwlJjGKdmp/view?usp=share_link

Unfortunately I cannot aggregate them now, because of [PLANNER-2925](https://issues.redhat.com/browse/PLANNER-2925). I will choose the most interesting ones and summarize it in my master thesis. In general ruining helps.

PR #2757, opened 2023-04-12
PLANNER-2919 Correct Quarkus checks

radtriste:
https://issues.redhat.com/browse/PLANNER-2919

Related:
- https://github.com/kiegroup/optaplanner/pull/2757
- https://github.com/kiegroup/optaplanner-quickstarts/pull/545

### DIFF
--- a/.ci/environments/update.sh
+++ b/.ci/environments/update.sh
@@ -43,6 +43,3 @@ if [ -f "${env_path}/after.sh" ]; then
     echo "Run after script"
     ${env_path}/after.sh $@
 fi
-
-# Download `setup_integration_branch` script and execute
-curl -s https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/dsl/seed/scripts/setup_integration_branch.sh | bash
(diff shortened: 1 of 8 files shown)


.ci/environments/update.sh
```diff
@@ -0,0 +1,48 @@
...
+fi
+
+# If update script is present, apply it
+if [ -f "${env_path}/after.sh" ]; then
+    echo "Run after script"
+    ${env_path}/after.sh $@
+fi
+
+# Download `setup_integration_branch` script and execute
+curl -s https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/dsl/seed/scripts/setup_integration_branch.sh | bash
```
rsynek, line 48: Why do we create any integration branch to update the quarkus version?

radtriste: optional->script is executed only if `INTEGRATION_BRANCH` env var is present, else it ignores and go on
see also https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/dsl/seed/scripts/setup_integration_branch.sh

rsynek: If we don't need it, I prefer simplicity.

radtriste: removed from both PRs

PR #2785, opened 2023-04-20
Clean up around scopes

triceo:
Some best score improvement checks were unnecessary.
Some code was unused.


core/optaplanner-core-impl/src/main/java/org/optaplanner/core/impl/phase/scope/AbstractPhaseScope.java
```diff
@@ -64,8 +64,8 @@ public void setBestSolutionStepIndex(int bestSolutionStepIndex) {
 
     public void reset() {
         bestSolutionStepIndex = -1;
-        // TODO Usage of solverScope.getBestScore() would be better performance wise but is null with a uninitialized score
-        startingScore = solverScope.calculateScore();
+        // solverScope.getBestScore() is null with a uninitialized score
```
rsynek, line 67:
```suggestion
        // solverScope.getBestScore() is null with an uninitialized score
```

rsynek (approved):

PR #2871, opened 2023-06-10
fix NPE for Optional lookup in MockProblemChangeDirector

RobJohansen:
### Issue
- NPE when using lookUpWorkingObject for non-registered object.
- Believe intention is that this should return Optional.empty(), which isn't working without ofNullable.

### Resolution
- Change to Optional.ofNullable since lookUpTable shouldn't contain object.


mareknovotny:
ok to test

lucamolteni (approved):

PR #3103, opened 2024-06-20
NO-ISSUE: Fix weekly Jenkins job build params

rodrigonull:


jstastny-cz:
I think the optaplanner:main now should map to quickstarts: development, no other option valid from now on.

jstastny-cz (approved):

PR #3121, opened 2024-08-23
[10.0.x] NO-ISSUE: Disable generate-code-tests goal and fixes for the release job

rodrigonull:


baldimir (approved):

porcelli (approved):
this has been problematic for the release, root cause should be investigated for upcoming releases... 

however we believe this shouldn't be a blocker for the release

tiagobento:
Can we please file an issue on `kie-issues` so we don't forget that we had to do this? We can continue the conversations there on what's the proper solution.

tiagobento:
@rodrigonull ^

rodrigonull:
Issue: https://github.com/apache/incubator-kie-issues/issues/1452

PR #3215, opened 2026-03-27
Add `forEachUniquePairIncludingNullVars`

eranl:
Add a `forEachUniquePair` vararg overload that takes two additional parameters:
- `boolean includingNullVars` - if true, facts with null variables are returned as well. This is useful in cases where you want to e.g. further penalize multiple unscheduled appointments for the same patient.
- `boolean noIdJoiner` - if true, returned facts are not joined based on `id`. This is useful in cases where you want to join facts by a different property.

These two parameters can be replaced by some kind of flags.
I chose `false` values to result in current behavior.
Would make sense to add non-vararg overloads too.

You can work around not having `noIdJoiner` by using `constraintFactory.forEach(sourceClass).join(constraintFactory.forEach(sourceClass), ...)`. 
To work around not having `includingNullVars`, you could use `constraintFactory.forEachIncludingNullVars(sourceClass).join(constraintFactory.forEachIncludingNullVars(sourceClass), ...)`, but you'd also have to expose the id in order to join by it.

Please let me know if I'm missing something and there's a better way to do this.

### Checklist
- [ ] Release notes updated if applicable.

### DIFF
--- a/core/optaplanner-constraint-streams-common/src/main/java/org/optaplanner/constraint/streams/common/InnerConstraintFactory.java
+++ b/core/optaplanner-constraint-streams-common/src/main/java/org/optaplanner/constraint/streams/common/InnerConstraintFactory.java
@@ -73,17 +73,11 @@ public abstract class InnerConstraintFactory<Solution_, Constraint_ extends Cons
     }
 
     @Override
-    public <A> BiConstraintStream<A, A> forEachUniquePair(Class<A> sourceClass, boolean includingNullVars, boolean noIdJoiner,
-            BiJoiner<A, A>... joiners) {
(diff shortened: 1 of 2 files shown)


Rikkola:
Hello, an unit test or tests would be good to cover the change.

eranl:
I made two changes:
1. I've realized that not adding the id joiner would result in non-unique pairs, so given that a simple workaround exists, I dropped the `noIdJoiner` parameter
2. To be consistent with `forEachIncludingNullVars`, I converted the method from an overload to a new `forEachUniquePairIncludingNullVars`

Updated the title. Not sure if I should update the description or not.

> Hello, an unit test or tests would be good to cover the change.

Hi, I couldn't find any existing `ConstraintFactory` tests, so I didn't know where to put the tests and how to test it.

PR #3221, opened 2026-04-28
[incubator-kie-issues#2288] - Springboot 4.0.x upgrade 

### LINKED ISSUE #2288: Springboot 4.0.x upgrade
This issue is to upgrade Spring Boot to a newer supported version, as the current Spring Boot version(3.5.x) used in incubator KIE repositories is nearing its end of support (EoS/EOL) by 30 Jun 2026. The upgrade ensures continued security fixes, stability, and long‑term maintainability.

nrknithin:
Closes: https://github.com/apache/incubator-kie-issues/issues/2288

Related PRs
- Drools: https://github.com/apache/incubator-kie-drools/pull/6686
- Kogito Runtimes: https://github.com/apache/incubator-kie-kogito-runtimes/pull/4270
- Kogito Apps: https://github.com/apache/incubator-kie-kogito-apps/pull/2331
- Kogito Examples: https://github.com/apache/incubator-kie-kogito-examples/pull/2210

### Why
Optaplanner's Spring integration (`optaplanner-spring-boot-autoconfigure` and its starter) needs to be on the same Spring Boot version that the rest of the KIE stack is moving to. This PR moves it from Spring Boot `3.5.12` to `4.0.5` and Spring Framework `6.2.17` to `7.0.6`. Two changes in Spring Boot 4 affect optaplanner directly:

1. **Persistence auto-configuration moved into its own opt-in module.** `@EntityScan`, `EntityScanPackages`, and `EntityScanner` are no longer part of `spring-boot-autoconfigure`. They moved from `org.springframework.boot.autoconfigure.domain.*` to `org.springframework.boot.persistence.autoconfigure.*`, and the new `spring-boot-persistence` artifact must be on the classpath. Optaplanner uses these classes in its auto-configuration and a few test configurations.
2. **Spring Framework 7 depends on `commons-logging` directly.** Spring 6 used the `spring-jcl` bridge module instead. Optaplanner's Maven Enforcer config has a rule that bans `commons-logging:commons-logging` to keep the slf4j bridge in place. With Spring 7 that ban now blocks the build, so the rule needs an exception for `commons-logging` when it comes from Spring Framework.

### Things to call out

- **`commons-logging` exception in the Enforcer rule.** This is the part most likely to look wrong at first reading. The original intent of the ban (force the slf4j bridge) is preserved — the exception only allows the dependency in the `compile` scope, and only because Spring Framework 7 needs it as a direct dependency. Code that pulls in `commons-logging` transitively from non-Spring sources still has to exclude it.
- **`spring-boot-persistence` is a runtime dependency.** Added in two places: the build-parent's `<dependencyManagement>` (to pin the version) and the autoconfigure module's `<dependencies>` (so the artifact actually lands on the classpath). Without it, the import rewrite compiles but tests fail at runtime with `ClassNotFoundException` on `EntityScanPackages`.

### How to review

- `build/optaplanner-build-parent/pom.xml`: three version bumps plus the new `spring-boot-persistence` entry and the Enforcer exception. The Enforcer exception is the line that needs careful reading.
- `optaplanner-spring-integration/optaplanner-spring-boot-autoconfigure/pom.xml`: one new runtime dependency on `spring-boot-persistence`.
- The five Java files (`OptaPlannerAutoConfiguration` + 4 test configurations): pure import path rewrites — `org.springframework.boot.autoconfigure.domain.*` → `org.springframework.boot.persistence.autoconfigure.*`. No behavior change.
- A second commit removes a temporary `spring-boot-properties-migrator` runtime dependency that was used during the upgrade to detect deprecated property keys. None were found, so the dependency is no longer needed.

### Verification
`mvn clean install` on this branch passes locally, including the five Spring auto-configure test configurations and the Quarkus integration tests.


yesamer (approved):
Thank you @nrknithin

gitgabrio (approved):

